PR #7, opened 2015-02-23
Karma Nodeunit

mattheworiordan:
@paddybyers and @kouno, I ended up going down a rabbit hole trying to get Jasmine to work with Chai in the way our existing test suites ran.  I really disliked the huge amount of hacking that was needed to get Chai to report into Jasmine correctly because Jasmine does not expose the running tests in anyway.  And when I looked back at why I was using Jasmine + Chai in the first place, it was to simplify things but keep similar assert like syntax that is currently used in the tests.

So yesterday I did U-turn and tried to get Karma using NodeUnit so that we can run the same NodeUnit tests in the browser, with Karma using headless browsers, and in Node.  This is all working now, with one small caveat - to run the NodeUnit tests you must run `grunt nodeunit` as it ensures necessary set up is done before NodeUnit starts running the tests.

If you open the repo, you can do any of the following:

`grunt karma`
`grunt nodeunit`
In your browser, open spec/nodeunit.html

Note I will move all the tests back into the `/test` folder once I have migrated them all across to the new format.

Shout if you have any comments.


Gruntfile.js
```diff
@@ -181,9 +181,33 @@ module.exports = function (grunt) {
...
 	grunt.initConfig(gruntConfig);
-
 	grunt.registerTask('compiler', ['curl:compiler', 'unzip:compiler']);
-	grunt.registerTask('all', ['copy', 'concat', 'closure-compiler']);
+	grunt.registerTask('all', ['copy', 'concat', 'requirejs', 'closure-compiler']);
+
+	grunt.loadTasks('spec/tasks');
+
+	var browsers = grunt.option('browsers') || 'default';
+  grunt.registerTask('test',
```
kouno: Yep, you really have to fix your indent.

mattheworiordan: I really don't care that much! 

Regards,
Matthew

kouno: Mixing tabs and spaces? You are kidding, right :smile:? You are pretty good to keep a consistent style usually. I didn't expect you wouldn't care about this.

mattheworiordan: I do, but right now I just have bigger things to worry about than to fix an issue that is only visible in Github because it assumes tabs are 4 spaces instead of 2. The rest of the library uses spaces, so this file is the exception. I'll do a fix up, but I am more concerned with the task in hand to be honest which is to get the Javascript library running tests automaticalky in a range of browsers. 

Regards,
Matthew

kouno: > I do, but right now I just have bigger things to worry about than to fix an issue that is only visible in Github because it assumes tabs are 4 spaces instead of 2.

Fair enough.

> The rest of the library uses spaces, so this file is the exception.

Actually I think most of the library contains tabs. You are actually bringing the 2 spaces indent in the repo. Check the other files you modified/deleted.

mattheworiordan:
@paddybyers, I have got a lot of this work done now but think I now need your help.

I have migrated all the Realtime & Rest tests across to the new structure, and have also migrated the `/browser` tests excluding the Crypto tests that are in that folder.  I have some questions and problems:

* Note if you want to run the tests yourself it's easier to run them against a localhost Realtime server which responds on port 80 & 443 correctly.  Once you have that running, simply change https://github.com/ably/ably-js/blob/ff1b348d546ecd685a0fca450df5282764784982/spec/common/globals/environment.js#L7 to 'local' instead of 'sandbox'.
* When Running the Nodeunit webserver tests, all tests currently pass in Chrome.  You can runt he tests locally easily by running `grunt nodeunit:webserver` and opening http://localhost:3000.  We run all the Realtime & REST tests as well as the browser specific tests, see https://github.com/ably/ably-js/blob/ff1b348d546ecd685a0fca450df5282764784982/karma.conf.js#L62-L65.
* Running the Nodeunit Node.js tests pass every time too, run `grunt test:nodeunit`.  As you can see at https://github.com/ably/ably-js/blob/ff1b348d546ecd685a0fca450df5282764784982/spec/tasks/grunt-nodeunit.js#L37, I don't include the browser tests in the folder `/spec/browser`
* Running the Karma tests against PhantomJS fails unfortunately with "Executed 74 of 151 ERROR (2 mins 40.933 secs / 2 mins 40.792 secs)".  You can see the problem when running `grunt test:karma`. It seems to specifically be a PhantomJS issue though.  When I run the tests against Chrome using Karma it doesn't pass either although more passes with "Executed 94 of 151 ERROR (2 mins 37.179 secs / 2 mins 36.949 secs)".  You can see this too with `grunt test:karma --browsers Chrome`.  I fear we have some PhantomJS issues that may be quite hard to diagnose unfortunately.  FYI, one way to test the Karma test suite more easily is to run the Karma web server and then open a browser against the web server, see https://github.com/ably/ably-js/blob/ff1b348d546ecd685a0fca450df5282764784982/README.md#debugging-the-tests-in-a-remote-browser-with-nodeunit-test-runner for more info on how to do this.  If you want to test using SauceLabs browsers, you can start a local Nodeunit web server with `grunt nodeunit:webserver`, then [create a SauceLabs tunnel](https://docs.saucelabs.com/reference/sauce-connect/).  The login details are in the private repo.
* I have not changed the stats tests to use the the stats set up POST end point yet.  If you could do that at some point in the future that would be great.

### Couple of additional thoughts:

* Once the final tests are migrated across, I plan on deleting the old `/test` folder and renaming `/spec` to `/test`
* I plan on migrating the PubNub and Pusher tests later on once everythign else is working.  PubNub & Pusher are lower priority.
* You can run the SauceLabs tests yourself if you want by doing the following `grunt karma --browsers sl_firefox`
* When running any of the tests or tasks in Grunt, I find that Grunt uses 70% of the CPU for some reason.  Oddly I've tried removing all dependencies, and it stays like that.  I have no idea why this is and we'll need to fix this at some point.
* I still need to integrate Travis correctly so that it runs the Nodeunit and Karma tests against a base set of browsers.
* The Nodeunit test suite passess, but never actually completes because of issue https://github.com/ably/ably-js/issues/9.  We need to figure out what in the Realtime library is causing the problem.

We should probably chat about this all tomorrow.

mattheworiordan:
I am closing this PR in favour of another PR that I will create that includes fixes to make Karma run the tests using Firefox which is supported by Travis CI

PR #49, opened 2015-05-11
Allow data parameter to be optional for presence#enter and channell#subscribe

### LINKED ISSUE #47: Presence#enterClient errors if not supplied with a data param
`channel.presence.enterClient('bob', 'foobar');` works fine, but `channel.presence.enterClient('bob');` fails with `"Ably: ConnectionManager.sendQueuedMessages(): Unexpected exception in transport.send(): TypeError: ob is undefined"`

(Docs say data parameter is optional, but might just be that the docs are out of date. Ruby client works without it.)

SimonWoolf:
Fixes #47 

@paddybyers The thing I'm confused by is that, though running the nodeunit tests (with spec/nodeunit.html) works as I'd expect, if running them in the console (with `grunt test:nodeunit`) I get tests incorrectly passing. E.g. checkout commit 233aa09 (after adding new tests but before I fix the problem) and run presence.test.js in the browser and the console -- the `enter5` test fails in the browser (correctly, as the behaviour it tests doesn't work), but somehow passes in the console.


SimonWoolf:
Abandoning in favour of https://github.com/ably/ably-js/pull/50, so we don't lose the comment thread linked to commit 0b09c66.

PR #50, opened 2015-05-11
Allow data parameter to be optional for presence#enter and channell#subscribe (2)

SimonWoolf:
Fixes #47

@paddybyers

PR #51, opened 2015-05-12
Make sure boolean data is encoded correctly

SimonWoolf:
Fixes ci breakage of https://github.com/ably/ably-js/pull/50

@paddybyers

### DIFF
--- a/browser/static/ably.js
+++ b/browser/static/ably.js
@@ -4478,7 +4478,7 @@ var Message = (function() {
 
 	Message.encode = function(msg, options) {
 		var data = msg.data, encoding;
-		if(data && typeof(data) != 'string' && !BufferUtils.isBuffer(data)) {
+		if(data !== null && data !== undefined && typeof(data) != 'string' && !BufferUtils.isBuffer(data)) {
 			msg.data = JSON.stringify(data);
 			msg.encoding = (encoding = msg.encoding) ? (encoding + '/json') : 'json';
 		}
--- a/browser/static/ably.noencryption.js
+++ b/browser/static/ably.noencryption.js
@@ -3130,7 +3130,7 @@ var Message = (function() {
 
 	Message.encode = function(msg, options) {
 		var data = msg.data, encoding;
-		if(data && typeof(data) != 'string' && !BufferUtils.isBuffer(data)) {
+		if(data !== null && data !== undefined && typeof(data) != 'string' && !BufferUtils.isBuffer(data)) {
 			msg.data = JSON.stringify(data);
 			msg.encoding = (encoding = msg.encoding) ? (encoding + '/json') : 'json';
 		}
--- a/common/lib/types/message.js
+++ b/common/lib/types/message.js
@@ -79,7 +79,7 @@ var Message = (function() {
 
 	Message.encode = function(msg, options) {
 		var data = msg.data, encoding;
-		if(data && typeof(data) != 'string' && !BufferUtils.isBuffer(data)) {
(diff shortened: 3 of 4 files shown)


mattheworiordan:
This looks good, but out of interest, why are you supporting the `Boolean` type?  We agreed some time back that we would support Strings, Byte Arrays and JSON objects only, see message encoding under http://docs.ably.io/client-lib-development-guide/features/#rest-channel

paddybyers:
It says "objects capable of JSON representation" which includes numbers and booleans.

mattheworiordan:
True, but not seen tests for this before so we can't expect this will work as expected in every language.  I will modify the spec then to explicitly state that our expectation includes boolean, numbers, arrays and dictionary/hash/collection-like objects (specific to the language) so that in every language there is a suitable test to ensure portability.  Currently if you [look at PHP](https://github.com/bladeSk/ably-php/blob/86b157adb9f69b21764d781f3955687996234b14/tests/ChannelMessagesTest.php#L38-L44) this has not been tested, neither is [Java](https://github.com/ably/ably-java/blob/master/test/io/ably/test/rest/RestChannelPublish.java) or [Ruby](https://github.com/ably/ably-ruby/blob/master/spec/unit/models/message_encoders/json_spec.rb#L7-L10) it's not covered.

paddybyers:
> I will modify the spec then to explicitly state that our expectation includes boolean, numbers, arrays and dictionary/hash/collection-like objects (specific to the language) so that in every language there is a suitable test to ensure portability.

I think the documentation for each language we should document the mapping between types; ie for each of the supported types in JSON (Number, Boolean, String, Object, Array):
- how it is passed to the client in the relevant language, if received from Ably;
- how the client is expected to pass it if they want to pass that value.

It is not always true that the mapping will be to the "natural" representation of that type in the target language; only that there is a defined type that makes it possible. For example in Java, the JSON-encodable types are JSONObject and JSONArray, not general Java collections.

mattheworiordan:
The problem is that we are not actually following the JSON spec if we do this.  Supporting the value "123" as a JSON representation is in fact incorrect, from [json.org](http://json.org/):

> JSON is built on two structures:
> A collection of name/value pairs. In various languages, this is realized as an object, record, struct, dictionary, hash table, keyed list, or associative array.
> An ordered list of values. In most languages, this is realized as an array, vector, list, or sequence.

So for example, using Ruby's native JSON library, parsing a non JSON like object (i.e. not starting with a `{` or `[`) raises an error, see below:

```
[4] pry(main)> JSON.parse('true')
JSON::ParserError: 757: unexpected token at 'true'
```

There is a serious risk that in any language that strictly supports the JSON spec we're going to have to build our own pre-parser to try and identify value types as opposed to actual JSON objects.

We spent a lot of time removing code some time back responsible for type support, and we agreed it was not a wise idea.  However we backtracked in favour or JSON, however I do not think a Numeric is really a JSON type, it's simply one of the supported value types.

mattheworiordan:
FYI, I've [implemented JSON ruby type support in Ruby](https://github.com/ably/ably-ruby/commit/49df0ddfe57e5ca45ffb9f51e91a7bd1c5971fd7) however it fails because the JSON decoder does not accept an invalid JSON string

paddybyers:
OK, you're right, the RFC is explicit: "JSON text is a serialized object or array"

http://www.ietf.org/rfc/rfc4627.txt

By contrast, the ECMA spec permits any text that satisfies the grammar for a JSON value (which does include primitives): 

http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf

PR #53, opened 2015-05-14
Throw 411 error for unsupported data types

### LINKED ISSUE #52: Agree on handling of JSON value types
Creating this issue to cover the discussion at https://github.com/ably/ably-js/pull/51

Probably better if we continue the discussion here as an issue

SimonWoolf:
Fixes #52


SimonWoolf:
> I would like the success case first (ie string or buffer).

Done

> I'm also mildly irked that the isObject() and isArray() checks each call toString() but I'll live with it.

I tried `data.constructor == Object`, but that kept evaluating false when run in nodeunit for, even though was true when in-browser, not sure why

paddybyers:
> but that kept evaluating false

Different contexts (eg iframes) each have a copy of these globals. The check is the correct one.

SimonWoolf:
forgot about this pr :)

> check string, then buffer, then others

nb needed to tweak `isArrayBuffer` to stop it erroring when fed null/undefined, if you want that check to come before the null check - https://github.com/ably/ably-js/commit/1a038eeca27b8c87249791c18f74ca3e93df853a

PR #80, opened 2015-07-30
Add tests for authCallback and update its docstring

### LINKED ISSUE #72: authCallback bug, test coverage & additional type support
authCallback currently has no test coverage in the library, we need to add some.

Additionally, we need to add support for the callback to return:

* a token string
* a TokenDetails object
* a TokenRequest object

Based on the docs at https://github.com/ably/ably-js/blob/1c8dd8f35d50eb488d8cde4fa4b3510d21ceb470/common/lib/client/auth.js#L155-L156, I assume only a TokenRequest is supported.

See http://www.rubydoc.info/gems/ably/0.8.2/Ably%2FAuth%3Ainitialize as an example (under auth_callback),.

In our spec as well, we specify `When authCallback option is set, it will execute the callback and will expect either a token string, a TokenDetails object or a TokenRequest object which will in turn be used to request a token from Ably.`

SimonWoolf:
Fixes #72

### DIFF
--- a/common/lib/client/auth.js
+++ b/common/lib/client/auth.js
@@ -160,7 +160,8 @@ var Auth = (function() {
 	 * - key:           the key to use.
 	 *
 	 * - authCallback:  (optional) a javascript callback to be used, passing a set of token
-	 *                  request params, to get a signed token request.
+	 *                  request params, to get a signed token request, a token details object,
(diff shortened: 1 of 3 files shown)


paddybyers:
Thanks

Issue #84, opened 2015-08-03
presence#history only available for rest channels, not realtime ones


SimonWoolf:

SimonWoolf:
Fixed by #85 / #87

PR #85, opened 2015-08-03
Make presence#history available to both realtime and rest channels

SimonWoolf:
Fixes https://github.com/ably/ably-js/issues/84

### DIFF
--- a/common/lib/client/channel.js
+++ b/common/lib/client/channel.js
@@ -10,7 +10,7 @@ var Channel = (function() {
 		this.rest = rest;
 		this.name = name;
 		this.basePath = '/channels/' + encodeURIComponent(name);
-		this.presence = new Presence(this);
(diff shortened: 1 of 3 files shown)


paddybyers:
Would it not work simply to have Realtime Presence extend Rest Presence?

mattheworiordan:
Looks good, :+1:

SimonWoolf:
@paddybyers Utils.inherits seems to break if inheriting from more than one object, but I only want history() so I guess it's not too messy to do that manually, and I it is a bit neater extracted. I pushed that approach up to [presence-history-2](https://github.com/ably/ably-js/compare/presence-history-2?expand=1) branch

SimonWoolf:
Closing in favour of https://github.com/ably/ably-js/pull/87

Issue #82, opened 2015-07-31
Saucelabs websockets support apparently sucks..?


SimonWoolf:
So a lot of the ably-js-browser tests are failing, across all browsers, with the following:

```Chrome 42.0.2311 (Windows 8.1) LOG 'Ably: WebSocketTransport.onError(): Unexpected error from WebSocket: undefined'```

```IE 11.0.0 (Windows 8.1 0.0.0) LOG: 'Ably: WebSocketTransport.onError(): Unexpected error from WebSocket: undefined'```

Ie the websocket is just suddently randomly failing. not closing, the onclose event isn't getting fired - apparently just vanishing into thin air. (The 'undefined' is because the 'error' is just a naked event, with no `message` property - the full object [here](https://gist.githubusercontent.com/SimonWoolf/45b8722e73b881e3ac41/raw/381cde85ca3e6835e85780207c6530cd1080c01d/error.json))

This only happens on saucelabs, never locally.

It's not that saucelabs doesn't support websockets; some of the websocket tests pass fine. In [Nov 2014](http://sauceio.com/index.php/2014/11/sauce-connect-gets-a-speed-boost-websocket-support/) they said websocket support was arriving "over the coming weeks", maybe they're still having teething troubles?

As an experiment I pushed [a branch](https://github.com/ably/ably-js/tree/no-ws-in-test) through ci that forces tests to use the comet transport instead of websockets. Results [here](https://ci.ably.io/job/ably-js-browsers-min/83/consoleFull): fewer errors, but still a bunch due to the comet transport not behaving the same as the ws transport, which I'll file separate bugs about (eg https://github.com/ably/ably-js/issues/81 ).

cc @mattheworiordan

mattheworiordan:
@SimonWoolf thanks for the update. One of the very frustrating things about WebSockets is that up until the point the web socket is established (i.e. following a successful HTTP upgrade), no error information is really available.  So it's quite plausible that for example the server has returned an error when trying to do the upgrade, but unfortunately the error will never be received by the client in these cases.  There is perhaps an argument to always upgrade and then send an error protocol message and then close the connection in these cases so that we can capture there error.  However, I am not sure if your errors above are triggered with an already established connection or not.  @paddybyers one for you to chime on I think.

paddybyers:
In the cases we can, we already accept a connection and send an error protocol message, instead of just refusing the connection. We then do a normal close, which is seen by the client as a close, not as an error as in this case.

[...]

PR #90, opened 2015-08-05
Force saucelabs to use varnish rather than squid as a proxy

SimonWoolf:
Fixes https://github.com/ably/ably-js/issues/82.

Before: https://ci.ably.io/job/ably-js-browsers-min/99/consoleFull
After: https://ci.ably.io/job/ably-js-browsers-min/100/console

The tests still fail, but at least now they fail because IE and mobile safari suck, not because some websocket connections randomly fail.

@mattheworiordan @paddybyers 

untitled


paddybyers:
LGTM. @mattheworiordan ?

mattheworiordan:
Good work, :+1:

Issue #27, opened 2015-04-09
Comet transport error


mattheworiordan:
Discussion from Slack

----

> Hi guys, I've been having some trouble with the following error: ` {"statusCode":404,"code":80009,"message":"Connection not established (no transport handle): EEEE-EEEEEE"} `

> I'm pretty sure I am connected to Ably before receiving the error, as I do receive a "connected" callback before the error. An example URL it 404's on is: [click](https://rest.ably.io/comet/EEEE!EEEEEEEE/send?access_token=EEEEE.EEEE-E6XCfSiT-bGd7amOKPSg1jRnTBugb4sexad7rSbfqCs07M7H8jCjXHzCgj9BxjBAcSL-7W6h63dNZun-o2eSwbW1kB2trrcVGDnMct0&rnd=5599157747346908)

> Here's a part of my log when I receive the error:

```
Ably: RealtimeChannel.setAttached: activating channel; name = LQMLC; message flags = undefined
ably.js?_=1428610526373:4067 Ably: WebSocketTransport.onWsData(): data received; length = 186; type = string
ably.js?_=1428610526373:4067 Ably: Presence.setPresence(): received presence for 1 participants; syncChannelSerial = undefined
ably.js?_=1428610526373:4067 Ably: PresenceMap.endSync(); channel = LQMLC; syncInProgress = true: undefined
ably.js?_=1428610526373:4067 Ably: WebSocketTransport.onWsData(): data received; length = 186; type = string
ably.js?_=1428610526373:4067 Ably: Presence.setPresence(): received presence for 1 participants; syncChannelSerial = undefined
ably.js?_=1428610526373:4067 Ably: PresenceMap.endSync(); channel = LQMLC; syncInProgress = false: undefined
ably.js?_=1428610526373:4067 Ably: WebSocketTransport.onWsData(): data received; length = 22; type = string
ably.js?_=1428610526373:4067 Ably: ConnectionManager on(ack): serial = undefined; count = 1
ably.js?_=1428610526373:4067 Ably: ConnectionManager.ackMessage(): serial = undefined; count = 1
[...]

PR #91, opened 2015-08-11
Partial attack on IE11 ci tests

SimonWoolf:
So this does *not* make the saucelabs integration test pass in IE11. What it does do it make it consistently pass locally, in a VM. It makes sure that tests close their connections at the end of each test to avoid running into IE's limit of six simultaneous websocket connections.

The saucelabs test still fails - several tests time out. This then causes other failures, since when a test times out, its websocket connections stay open (as the `withTimeout` helper doesn't close them). So after six timeouts, all slots are used and any test which requires a websocket transport will fail (with a `SecurityError`, which are what IE throws if you try exceed the six-websocket limit).  I tried a quick fix for this of putting something in the `AblyRealtime` helper that registered the connection (on the test app object), and then had the timeout close all open connections -- or just overrid test.done() to force all connections to close at the end of every test -- which worked well but broke all the tests which set up a connection in one test and then went on to use it in others.

To demonstrate: increasing the timeout to 120s reduces the number of saucelabs timeouts to just two, see https://ci.ably.io/job/ably-js-browsers-min/113/consoleText . (Obv not included in the PR).

Also tried the obvious fix of [removing web_socket requirements](https://github.com/ably/ably-js/commit/cd844df) - but that broke everything that was forcing websockets to get around issue https://github.com/ably/ably-js/issues/27. (Had a quick go at fixing that issue but didn't get very far).

So, something is causing IE11 on saucelabs to just be incredibly slow, and combined with the six connections limit make it very annoying to get working on saucelabs. A temporary solution could be to remove ie11 from ably-js-browsers-min and just test it manually in virtualbox (or use one of the karma virtualbox IE runners), which with this pr seems to work ok; I've included that commit here.

Most of this is just test code so should be uncontroversial; the exception is https://github.com/ably/ably-js/commit/077b8679791efc8c094feed71264ce763e8fa5a0 (adding a transport `disconnect` event to comettransport no matter how the connection was closed). [The websocket transport does something similar already](https://github.com/ably/ably-js/blob/master/common/lib/transport/websockettransport.js#L107) so I figured it could make sense, but my understanding of the details of the different transport lifecycles is shaky, so wanted to flag it up to @paddybyers . (The normal connection `closed` event would work for comet, but for websockets it comes too early - it's emitted before the websocket object is actually disposed of, and on IE that seems to sometimes take a while).

cc @mattheworiordan @paddybyers

### DIFF
--- a/spec/realtime/channel.test.js
+++ b/spec/realtime/channel.test.js
@@ -2,7 +2,9 @@
 
 define(['ably', 'shared_helper'], function(Ably, helper) {
 	var exports = {},
-		displayError = helper.displayError;
+		displayError = helper.displayError,
+		closeAndFinish = helper.closeAndFinish,
+		monitorConnection = helper.monitorConnection;
 
 	exports.setupauth = function(test) {
 		test.expect(1);
@@ -32,22 +34,13 @@ define(['ably', 'shared_helper'], function(Ably, helper) {
 						test.ok(false, 'Attach failed with error: ' + err);
 					else
 						test.ok(true, 'Attach to channel 0 with no options');
-					test.done();
-					realtime.close();
+					closeAndFinish(test, realtime);
 				});
 			});
-			var exitOnState = function(state) {
-				realtime.connection.on(state, function () {
-					test.ok(false, transport + ' connection to server failed');
-					test.done();
-					realtime.close();
-				});
-			};
-			exitOnState('failed');
-			exitOnState('suspended');
+			monitorConnection(test, realtime);
 		} catch(e) {
 			test.ok(false, 'Channel attach failed with exception: ' + e.stack);
-			test.done();
+			closeAndFinish(test, realtime);
 		}
 	};
 
@@ -67,26 +60,13 @@ define(['ably', 'shared_helper'], function(Ably, helper) {
 						test.ok(false, 'Attach failed with error: ' + err);
 					else
 						test.ok(true, 'Attach to channel1 with no options');
-
-					setTimeout(function() {
-						test.done();
-
-					}, 3000);
-					realtime.close();
+					closeAndFinish(test, realtime);
 				});
 			});
-			var exitOnState = function(state) {
-				realtime.connection.on(state, function () {
-					test.ok(false, transport + ' connection to server failed');
-					test.done();
-					realtime.close();
-				});
-			};
-			exitOnState('failed');
-			exitOnState('suspended');
+			monitorConnection(test, realtime);
 		} catch(e) {
 			test.ok(false, 'Channel attach failed with exception: ' + e.stack);
-			test.done();
+			closeAndFinish(test, realtime);
 		}
 	};
 
@@ -105,21 +85,12 @@ define(['ably', 'shared_helper'], function(Ably, helper) {
 					test.ok(false, 'Attach failed with error: ' + err);
 				else
 					test.ok(true, 'Attach to channel 0 with no options');
-				test.done();
-				realtime.close();
+				closeAndFinish(test, realtime);
 			});
-			var exitOnState = function(state) {
-				realtime.connection.on(state, function () {
-					test.ok(false, transport + ' connection to server failed');
-					test.done();
-					realtime.close();
-				});
-			};
-			exitOnState('failed');
-			exitOnState('suspended');
+			monitorConnection(test, realtime);
 		} catch(e) {
 			test.ok(false, 'Channel attach failed with exception: ' + e.stack);
-			test.done();
+			closeAndFinish(test, realtime);
 		}
 	};
 
@@ -137,36 +108,25 @@ define(['ably', 'shared_helper'], function(Ably, helper) {
 				channel0.attach(function(err) {
 					if(err) {
 						test.ok(false, 'Attach failed with error: ' + err);
-						test.done();
-						realtime.close();
+						closeAndFinish(test, realtime);
 					}
 					channel0.detach(function(err) {
 						if(err) {
 							test.ok(false, 'Detach failed with error: ' + err);
-							test.done();
-							realtime.close();
+							closeAndFinish(test, realtime);
 						}
 						if(channel0.state == 'detached')
 							test.ok(true, 'Attach then detach to channel 0 with no options');
 						else
 							test.ok(false, 'Detach failed: State is '+channel0.state);
-						test.done();
-						realtime.close();
+						closeAndFinish(test, realtime);
 					});
 				});
 			});
-			var exitOnState = function(state) {
-				realtime.connection.on(state, function () {
-					test.ok(false, transport + ' connection to server failed');
-					test.done();
-					realtime.close();
-				});
-			};
-			exitOnState('failed');
-			exitOnState('suspended');
+			monitorConnection(test, realtime);
 		} catch(e) {
 			test.ok(false, 'Channel attach failed with exception: ' + e.stack);
-			test.done();
+			closeAndFinish(test, realtime);
 		}
 	};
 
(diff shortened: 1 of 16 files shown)


spec/realtime/channel.test.js
```diff
@@ -2,7 +2,9 @@
 
 define(['ably', 'shared_helper'], function(Ably, helper) {
 	var exports = {},
-		displayError = helper.displayError;
+		displayError = helper.displayError,
```
paddybyers: Whitespace problems?

paddybyers:
> force all connections to close at the end of every test -- which worked well but broke all the tests which set up a connection in one test and then went on to use it in others.

Can we have a look at these and see if we can reduce the interdependency between the tests?

> but that broke everything that was forcing websockets to get around issue #27. (Had a quick go at fixing that issue but didn't get very far).

Please don't spend time on this issue - I am working on changes to the upgrade mechanism atm.

> the exception is 077b867 (adding a transport disconnect event to comettransport no matter how the connection was closed). The websocket transport does something similar already so I figured it could make sense, but my understanding of the details of the different transport lifecycles is shaky, so wanted to flag it up

The ws code you linked is indicating disconnection of the transport in the case of a remotely-initiated disconnection, whereas the change you made for comet is different. I'd like to understand more  about how the comet transport is failing to shutdown in the case of the tests (which should all be client-initiated disconnections).

> (The normal connection closed event would work for comet, but for websockets it comes too early - it's emitted before the websocket object is actually disposed of, and on IE that seems to sometimes take a while).

I'd also like to understand more about this point. The ws object indicates a change in connection state, yet the underlying resource isn't available for allocation to a new ws object?

The test changes look good so perhaps while we look at the above issues we could land those changes?

mattheworiordan:
@SimonWoolf, good work on this, I have seen the pain you have gone through.  If this is now passing in CI, I think we should simply use one of the VirtualBox Karma launchers to run the IE tests locally on CI, and rely on SauceLabs for the other browsers.  I am not sure how difficult that would be, but I expect not too difficult.

@paddybyers Simon is now away on holiday in the US so won't be doing any more on this unfortunately for a little while.  In terms of the issues with IE11, I did sit down briefly with Simon to see the issues in Saucelabs.  What we would see is a few things:

* Calling close on a websocket connection would not necessarily free the resource.  As such, subsequent tests would start and could not obtain a WS connection and would eventually fail.
* Comet connections were similar, especially because we did not wait until those connections were closed.  So requests to open new XHR connections would simply pause, and eventually the tests would fail.

@paddybyers what is the work you are doing to upgrade the mechanism?  Will that effect the events emitted on the connection lifecycle channel i.e. will it state it's an upgrade instead of a new connection?

SimonWoolf:
> Can we have a look at these and see if we can reduce the interdependency between the tests?

Sure, I'll do that. Looks like presence.test.js is the only really tangled one.

> The ws code you linked is indicating disconnection of the transport in the case of a remotely-initiated disconnection, whereas the change you made for comet is different.

I don't follow -- in the ws code `Transport.onDisconnect` is called (in the line I linked) by `WebSocketTransport.onWsClose`, which is called when the ws is closed no matter how that was initiated. Should it not be? (note that the 'disconnected' event on the transport object is distinct from the 'disconnected' event on the actual main connection object, the one I'm using is the former.)

>  I'd like to understand more about how the comet transport is failing to shutdown in the case of the tests (which should all be client-initiated disconnections).

To be clear, my reason for adding that commit wasn't to fix a comet problem (xhr connections having to wait a bit doesn't seem to cause much of an issue), but rather because, as I'm listening for the transport `disconnected` event to solve the websocket issue (as that's the only even emitted after the websocket is actually closed), I wanted the comet transport to emit a similar one. (If I listen for connection `close`, then the websocket connections is called too early). There's no direct analogue as xhr doesn't have an onclose - I could add one but not sure it's necessary, so I figured the closed existing analogue was in the `dispose()` method, as that gets called no matter how the connection was initiated.

Note: if I [jack up the timeout to something ridiculous](https://github.com/ably/ably-js/commit/f2de0cb5ee34537f3e533712ce6f67f886ae8b74), the only tests that still fail in saucelabs (but pass in a VM) are the crypto multisend tests. They do so in an interesting way -- have a look at https://ci.ably.io/job/ably-js-browsers-min/116/consoleText (after `started; name = multiple_send_text_2_200'`). You get the following 17 lines, repeated six times, after which you start to get securityerrors as the websocket resources are all used up:
```
IE 11.0.0 (Windows 8.1 0.0.0) LOG: 'Ably: ConnectionManager.requestState(): requested state: connecting'
IE 11.0.0 (Windows 8.1 0.0.0) LOG: 'Ably: ConnectionManager.cancelTransitionTimer(): '
IE 11.0.0 (Windows 8.1 0.0.0) LOG: 'Ably: ConnectionManager.enactStateChange: setting new state: connecting; reason = undefined'
IE 11.0.0 (Windows 8.1 0.0.0) LOG: 'Ably: ConnectionManager.connectImpl(): starting connection'
IE 11.0.0 (Windows 8.1 0.0.0) LOG: 'Ably: ConnectionManager.startTransitionTimer(): transitionState: connecting'
IE 11.0.0 (Windows 8.1 0.0.0) LOG: 'Ably: ConnectionManager.chooseTransport(): '
IE 11.0.0 (Windows 8.1 0.0.0) LOG: 'Ably: ConnectionManager.chooseTransport(): Transport recovery mode = clean'
IE 11.0.0 (Windows 8.1 0.0.0) LOG: 'Ably: ConnectionManager.chooseTransport(): No http transports available; ignoring fallback hosts'
IE 11.0.0 (Windows 8.1 0.0.0) LOG: 'Ably: ConnectionManager.chooseTransportForHost(): trying web_socket'
IE 11.0.0 (Windows 8.1 0.0.0) LOG: 'Ably: WebSocketTransport.connect(): starting'
IE 11.0.0 (Windows 8.1 0.0.0) LOG: 'Ably: WebSocketTransport.connect(): uri: wss://sandbox-realtime.ably.io:443/'
IE 11.0.0 (Windows 8.1 0.0.0) LOG: 'Ably: WebSocketTransport.connect(): authParams: key: lupTqQ.VLSVjw:hnBKETAyo5VLcqsl;'
IE 11.0.0 (Windows 8.1 0.0.0) LOG: 'Ably: ConnectionManager connect timer expired: requesting new state: disconnected'
IE 11.0.0 (Windows 8.1 0.0.0) LOG: 'Ably: ConnectionManager.notifyState(): new state: disconnected'
IE 11.0.0 (Windows 8.1 0.0.0) LOG: 'Ably: ConnectionManager.cancelTransitionTimer(): '
IE 11.0.0 (Windows 8.1 0.0.0) LOG: 'Ably: ConnectionManager.enactStateChange: setting new state: disconnected; reason = Connection to server temporarily unavailable'
IE 11.0.0 (Windows 8.1 0.0.0) LOG: 'Ably: ConnectionManager.queuePendingMessages(): queueing 0 pending messages'
```

Fixed some bugs in that test (b91b7d2, 04de328), but no difference. So, two problems -- one, whatever strange IE/saucelabs issue is causing the websocket connection attempt to timeout, and two, ably-js not disposing of the resource after the TransitionTimer times out. @paddybyers - for the second one, shall I file a new issue or will you be looking at it with your connection upgrade rewrite?

Maybe the best thing at this point is to wait for your upgrade work? Assuming this fixes #27, we can remove forcing-web_socket from the tests, which with this PR has a good chance of making IE11 work in saucelabs (just because it'll fall back to comet), without needing to spin up vms in ci.

Incidentally, I mentioned the iphone failure -- looks like that wasn't a real failure (https://ci.ably.io/job/ably-js-browsers-min/114/consoleText), just that the iphone emulator takes around 60s to start, and we have the karma launch timeout set to 60s -  c46d1d1 increases that to 120.

mattheworiordan:
Good progress @SimonWoolf, the end is near, :fireworks:

SimonWoolf:
Conclusion of discussion with paddy:
- Added https://github.com/ably/ably-js/issues/98
- should emit a new transport dispose event on dispose (for comet: need to be called when this.recvRequest.abort(); completes if there's a recvrequest), not override the semantics of disconnected
- IE: work out why crypto tests failing. the;'re doing so before the transport even connects (?!?!). running out of entropy? shouldn't be, getrandombytes is nonblocking.

SimonWoolf:
@paddybyers ably-js-browsers-min now passes with chrome 42, mobile safari, and IE11 - albeit on win10 rather than win8.1 . https://ci.ably.io/job/ably-js-browsers-min/129/changes

Commits that are new since my previous comment are
- e50af01 ensure websocket object is disposed if connect timeout expires
- adcaa40 emit an event on transport dispose
- 9ff69c4 IE11 uses window.msCrypto rather than window.crypto
- 6fbdd44 increase test timeout (saucelabs is slooooow)
- d58a2ec saucelabs change ie11 tests from win8.1 to win10 -- saucelabs ie11 on win8.1 seems to exhibit strange pathologies not present on saucelabs ie11 on win10 or running ie11 win8.1 in virtualbox. rather than spending more days debugging saucelabs I figure switching to win10 is acceptable for now.

mattheworiordan:
Nice work @SimonWoolf

spec/realtime/channel.test.js
SimonWoolf: Don't think so..? Was two tabs before, is too tabs after. The diff was changing `;` to `,`.

paddybyers:
Other than the comments on specific commits, LGTM.

paddybyers:
> at this point, ConnectionManager doesn't have a reference to the transport

This issue is different after the upgrade changes, so pls remove this commit from the PR and we'll merge the PR and address this specific issue afterwards.

SimonWoolf:
Merged to master as 3d057a6 (without e50af01 or e5b9b54)

Issue #99, opened 2015-08-22
Detection of connection state change is too slow


mattheworiordan:
Whilst testing the [mobile Cordova demo](https://github.com/ably/demo-mobile-phonegap-cordova) in my browser, I see that the connection state change detection is far too slow to be useful.

I have compared this to the web versions of WhatsApp and Slack, and the results vary quite a lot.  WhatsApp literally detects the Wifi connection being turned off immediately, Slack takes a good 10s or so, and it seems like our Cordova demo almost never detects that the connection is gone.  See https://dl.dropboxusercontent.com/u/1575409/Screenshots/video-capture-connection-state.mov.

The connection state does eventually drop, but it takes far too long.  I have not tested this in Cordova yet, but I expect we will need to add code specific to Cordova to detect network state, fortunately Cordova provides this already, see http://cordova.apache.org/docs/en/2.4.0/cordova_events_events.md.html#offline.

I expect we need to approach this in a number of ways:

* Catch online / offline events from the device if available
  * Browsers http://html5-demos.appspot.com/static/navigator.onLine.html (very well supported in browsers it seems, https://developer.mozilla.org/en-US/docs/Web/API/NavigatorOnLine/onLine)
  * Cordova http://cordova.apache.org/docs/en/2.4.0/cordova_events_events.md.html#offline
* Detect a lack of heartbeat (perhaps increase the frequency for browser connections, can we specify a param in the connection?) and then check if the internet up check passes, if not, assume we are disconnected.
* See if we can tweak the transports we use to be more sensitive to changing connection state?

@paddybyers thoughts?

paddybyers:
I agree, but lets also look at what pusher and pubnub do.

With a websocket connection we're generally waiting for the tcp connection to drop, and the idea of the heartbeat is that it should prompt the connection itself to notice that it's broken - it shouldn't need to be the job of the client to have timers running for the heartbeat. However, this strategy is flawed, now I think about it, because only the server (which is attempting to send the heartbeat) then knows there's a problem. So really we should have the client reply to the heartbeat, or initiate sending a heartbeat itself, to trigger a disconnection event from the tcp layer.

I agree that where a platform provides it we should register for network status events, although then using them in place of the internet-up check is perhaps putting a bit too much trust in them.

mattheworiordan:
[...]

PR #100, opened 2015-08-22
Tests (WIP) for subscribing to device connection state change events

mattheworiordan:
See https://github.com/ably/ably-js/issues/99

This does not address the heartbeat discussion in any way, this only tests for subscription to online / offline events from the browser or Cordova device

@paddybyers can you plug in the changes to subscribe document `online` and `offline` events to make the tests pass, see http://caniuse.com/#feat=online-status?  Also, I was not sure how to force the suspended state as I don't see any easy way to change the reconnect defaults.  In Ruby I change them using stubs, see https://github.com/ably/ably-ruby/blob/master/spec/acceptance/realtime/connection_failures_spec.rb#L256-L260, however I think we should simply make this an option in the client options anyway, or perhaps expose a new method on the connection to simply change the defaults.

### DIFF
--- a/spec/browser/connection.test.js
+++ b/spec/browser/connection.test.js
@@ -1,0 +1,120 @@
+"use strict";
+
+define(['ably', 'shared_helper'], function(Ably, helper) {
+  var exports = {};
+
+  exports.setup_realtime_history = function(test) {
+    test.expect(1);
+    helper.setupApp(function(err) {
+      if(err) {
+        test.ok(false, helper.displayError(err));
+      } else {
+        test.ok(true, 'app set up');
+      }
+      test.done();
+    });
+  };
+
+  exports.device_going_offline_causes_disconnected_state = function(test) {
+    var realtime = helper.AblyRealtime(),
+        connection = realtime.connection,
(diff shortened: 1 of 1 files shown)


paddybyers:
> I was not sure how to force the suspended state

As per the other discussion, I think we should simply be disconnecting the transport explicitly if the network is indicated as having gone down. This will transition to `disconnected`, and will (by the usual mechanism) then immediately trigger a reconnection attempt. The eventual transition to `suspended` will follow the usual sequence of retries.

mattheworiordan:
> I was not sure how to force the suspended state

I know that, but that is not relevant to the PR where I want to add test coverage for an `online` event when in a `suspended` state, waiting 2 minutes in the tests is clearly not practical, hence the question and https://github.com/ably/ably-js/commit/f175a8c985cd3a2358ea83de367e2c3cf074987b#diff-f6677636d5f88be4eb6bffd537887c3aR83.

SimonWoolf:
I raised a new PR against new-upgrade branch, so closing this in favour of https://github.com/ably/ably-js/pull/126

PR #102, opened 2015-08-25
Allow connection to inherit clientId from connectionDetails

SimonWoolf:
Changes needed for https://github.com/ably/realtime/pulls/274.

Test currently fails just because sandbox isn't running a branch that supports connectionDetails. Run with local realtime on the `connection-details-tentative` branch.

### DIFF
--- a/common/lib/transport/connectionmanager.js
+++ b/common/lib/transport/connectionmanager.js
@@ -650,8 +650,9 @@ var ConnectionManager = (function() {
 				});
 				return;
 			}
-			/* FIXME: decide if fatal */
-			var fatal = false;
+			/* all codes in the 40-50k zone indicate a fault with the request, so
+			 * should not be retried unmodified */
+			var fatal = err.code >= 40000 && err.code <= 50000;
 			if(fatal)
 				self.notifyState({state: 'failed', error: err});
 			else
--- a/common/lib/transport/connectionmanager.js
+++ b/common/lib/transport/connectionmanager.js
@@ -331,7 +331,7 @@ var ConnectionManager = (function() {
 
 		var self = this;
 		var handleTransportEvent = function(state) {
-			return function(error, connectionKey, connectionSerial, connectionId) {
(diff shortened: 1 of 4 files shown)


common/lib/transport/connectionmanager.js
```diff
@@ -331,7 +331,7 @@ var ConnectionManager = (function() {
 
 		var self = this;
 		var handleTransportEvent = function(state) {
-			return function(error, connectionKey, connectionSerial, connectionId) {
+			return function(error, connectionKey, connectionSerial, connectionId, clientId) {
```
paddybyers: This change needs to be reworked when rebasing to `new-upgrade`.

common/lib/transport/connectionmanager.js
```diff
@@ -406,6 +409,14 @@ var ConnectionManager = (function() {
 				this.persistConnection();
 			this.msgSerial = 0;
 		}
+		if(clientId) {
+			if(this.realtime.clientId && this.realtime.clientId != clientId) {
+				/* Should never happen in normal circumstances as realtime should
+				 * return an error code in case of mismatch */
+				throw new Error('Unexpected mismatch between expected and received clientId');
```
paddybyers: This doesn't make sense because you're just throwing an exception back into the handler for the event that triggered this transport activation. Just log the error and take no further action.

common/lib/transport/connectionmanager.js
```diff
@@ -639,8 +650,9 @@ var ConnectionManager = (function() {
 				});
 				return;
 			}
-			/* FIXME: decide if fatal */
-			var fatal = false;
+			/* all codes in the 40-50k zone indicate a fault with the request, so
```
paddybyers: Are you sure this is what's specified? A 502 for example would suggest that a fallback host should be attempted.

paddybyers:
This LGTM except for the comments above, but please also rebase to `new-upgrade`.

SimonWoolf:
Reraised against new-upgrade: https://github.com/ably/ably-js/pull/110

PR #110, opened 2015-08-28
Allow connection to inherit clientId from connectionDetails

SimonWoolf:
Changes needed for https://github.com/ably/realtime/pulls/274.

Test currently fails just because sandbox isn't running a branch that supports connectionDetails. Run with local realtime on the connection-details-tentative branch.

Re-raised against new-upgrade.

### DIFF
--- a/common/lib/client/auth.js
+++ b/common/lib/client/auth.js
@@ -131,6 +131,10 @@ var Auth = (function() {
 	Auth.prototype.authorise = function(authOptions, tokenParams, callback) {
 		var token = this.tokenDetails;
 		if(token) {
+			if(this.rest.clientId && token.clientId && this.rest.clientId !== token.clientId) {
+				callback(new ErrorInfo('ClientId in token was ' + token.clientId + ', but library was instantiated with clientId ' + this.rest.clientId, 40102, 401));
+				return;
+			}
 			if(token.expires === undefined || (token.expires > this.getTimestamp())) {
 				if(!(authOptions && authOptions.force)) {
 					Logger.logAction(Logger.LOG_MINOR, 'Auth.getToken()', 'using cached token; expires = ' + token.expires);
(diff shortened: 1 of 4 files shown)


common/lib/transport/connectionmanager.js
```diff
@@ -408,6 +410,14 @@ var ConnectionManager = (function() {
 		if(connectionKey && this.connectionKey != connectionKey)  {
 			this.setConnection(connectionId, connectionKey, connectionSerial);
 		}
+		if(clientId) {
+			if(this.realtime.clientId && this.realtime.clientId != clientId) {
+				/* Should never happen in normal circumstances as realtime should
+				 * recognise mismatch and return an error */
+				Logger.logAction(Logger.LOG_ERROR, 'ConnectionManager.activateTransport()', 'Unexpected mismatch between expected and received clientId');
```
paddybyers: Sorry, I think I confused/misled you with my last comment on this. I agree this should never happen but if it does :) it's a fatal error I think so we should fail the entire connection.

paddybyers:
Other than the above, LGTM.

SimonWoolf:
@paddybyers to fail the connection, https://github.com/ably/ably-js/commit/50e10654b15cbf0618682afd5357dd0d52c0a2cd (aborting the transport with a fatal err code) is sufficient, right? I don't also need to do a `this.notifyState({state: 'failed'});`?

paddybyers:
> I don't also need to do a this.notifyState({state: 'failed'});?

Correct, because that happens in the ensuing `deactivateTransport` if the transport that generated the event was also active.

common/lib/client/auth.js
```diff
@@ -131,6 +131,9 @@ var Auth = (function() {
 	Auth.prototype.authorise = function(authOptions, tokenParams, callback) {
 		var token = this.tokenDetails;
 		if(token) {
+			if(this.rest.clientId && token.clientId && this.rest.clientId !== token.clientId) {
+				callback(new ErrorInfo('ClientId in token was ' + token.clientId + ', but library was instantiated with clientId ' + this.rest.clientId, 40102, 401));
+			}
```
paddybyers: Need to `return;` here

SimonWoolf: oops, thanks

paddybyers:
Thanks

PR #111, opened 2015-08-28
Explicitly disallow empty string clientIds

SimonWoolf:
Since realtime treats that as equivalent to no clientid

### DIFF
--- a/common/lib/client/auth.js
+++ b/common/lib/client/auth.js
@@ -363,6 +363,11 @@ var Auth = (function() {
 			return;
 		}
 
+		if(tokenParams.clientId === '') {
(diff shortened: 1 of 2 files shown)


paddybyers:
We either need a specific error code in `errors.json` in the 400 series, and then we should return the appropriate `ErrorInfo`. Then the test can check for the error code instead of specific message string.

SimonWoolf:
Added; ably-common pr https://github.com/ably/ably-common/pull/5

paddybyers:
Oops, see comment on the other PR; it needs now to be `40012`.

SimonWoolf:
Updated

paddybyers:
Thanks

PR #112, opened 2015-08-28
Define error codes that result in failed rather than disconnected

SimonWoolf:


SimonWoolf:
@paddybyers refactored the errfatal helper to be a bit less horrible, made the discussed changes, and added some tests.

But got stuck in a jsonp problem: the 'should fail on auth-type errors' tests fail for the jsonp transport, since AFAICT, when the browser gets a response with a 404 statuscode (with the details in the body), it throws away the body and just calls the `script.onerror` callback with a generic event error (which is indistinguishable from what's called for network failure), and I can't see a way of getting the body. Not sure how big an issue you think this as, so gave up and commented out the jsonp test for the moment.

paddybyers:
> tests fail for the jsonp transport

JSONP should set `envelope=json` on the request, which means that the server responds with a 200l but the headers and actual status code are in the body. Is this not happening?

paddybyers:
Looking better. I still think we should move the `80015, 80017, 80030` codes so that we can decide with a single (or at least simpler test).

WRT the "break transport" test: how do you think we should arrange for only a subset of these tests to run on browsers that don't support all transports?

SimonWoolf:
> JSONP should set envelope=json on the request ... Is this not happening?

Nope. Only url params are key, callback, and body. 

> I still think we should move the 80015, 80017, 80030 codes so that we can decide with a single (or at least simpler test).

Makes sense, I'll file an issue in ably-common for discussion.

> WRT the "break transport" test: how do you think we should arrange for only a subset of these tests to run on browsers that don't support all transports?

Good point, fixed: https://github.com/ably/ably-js/commit/154f5f5bd1cf38837f03b2b9303894e25ff2e880

paddybyers:
> SONP should set envelope=json on the request ... Is this not happening?
> Not AFAICS. Only url params are key, callback, and body.

Open an issue for this then :(

This param needs to be added to the request, and then the envelope needs to be unwrapped in `complete()`.

> I still think we should move the 80015, 80017, 80030 codes so that we can decide with a single (or at least simpler test).
> Makes sense, I'll file an issue in ably-common for discussion.

Thanks

SimonWoolf:
rebased on new-upgrade post-`jsonp-envelope`-merge

paddybyers:
Thanks

Issue #108, opened 2015-08-27
Token Params are not being sent through to authUrl


mattheworiordan:
See https://github.com/ably/demo-mobile-phonegap-cordova/blob/master/www/js/app.js#L100, the authUrl requests do not have the token params appended.

We need a test for this.

SimonWoolf:
@mattheworiordan I'm not clear on what the desired behaviour is here. there's an `authParams` option which says "Query params to be included in any request made by the library to the authURL" (which works correctly) - is the proposal that the contents of that should be merged with tokenparams?

SimonWoolf:
(Eg currently the clientid that the library is instantiated with isn't even passed to the builtin token request, never mind an authCallback or authUrl -- if you do `new Ably.Realtime({key: <key>, clientId: "foo", useTokenAuth: true})`, the resulting token doesn't include the clientId. Should that also be changed?)

mattheworiordan:
@SimonWoolf I was not previously aware that this was the case, but @paddybyers told me that he understood that the authParams are passed through to the authUrl as query params when a GET request is being made, and as part of the body when using POST.  I understood the authParams should also be passed to the authCallback. See the definition in Java https://github.com/ably/ably-java/blob/83b365feb722e24093ee969c5aee85af48e4bd36/src/io/ably/rest/Auth.java#L321-L322

SimonWoolf:
@mattheworiordan 

> the authParams are passed through to the authUrl as query params 

Yeah, and they are. But that phonegap demo you linked doesn't specify an `authParams`. so was wondering if you meant the clientId (that the lib was initialized with) should be passed as a param as well

paddybyers:
> clientId (that the lib was initialized with) should be passed as a param as well

yes, because that's implicitly an `authParam` if not explicitly

SimonWoolf:
Fixed by https://github.com/ably/ably-js/pull/115/files

PR #115, opened 2015-09-02
If client lib initialised with clientid, use that for auth if no new tokenParams object is supplied

SimonWoolf:
Fixes https://github.com/ably/ably-js/issues/108


paddybyers:
Thanks

Issue #44, opened 2015-05-06
ToJson bug


mattheworiordan:
See https://github.com/ably/ably-js/blob/master/common/lib/types/presencemessage.js#L28, shouldn't this be action?

mattheworiordan:
In fact, it looks like a number of supported fields are missing from the toJson and toString functions

paddybyers:
You'll see that `action` is already there, so the bug is that `name` is included.

Other fields are intentionally omitted because `toJSON()` is only used when sending messages, and this is the subset of fields that are sent by the client.

paddybyers:
https://github.com/ably/ably-js/commit/b1632c3

mattheworiordan:
Thanks, but I think `id` is missing too and `connectionId` from the string representation

paddybyers:
`toString()` is intended to provide a useful human-readable summary containing the most relevant fields.

mattheworiordan:
Seems a bit arbitrary that an `action` enum integer  value and `time` in milliseconds is considered useful human-readable value, but `connectionId` is not IMHO.

mattheworiordan:
Matt to remove unneeded fields such as timestamp, connectionId

SimonWoolf:
Is the action here to remove the timestamp & connectionId fields from the presencemessage object entirely?

mattheworiordan:
> Is the action here to remove the timestamp & connectionId fields from the presencemessage object entirely?

Yes, plus if `clientId` matches the authenticated `clientId`, or if `clientId` is empty, then we don't need to send that field.

SimonWoolf:
I don't see how that can be right. The lib uses the timestamp and connectionId in the presence message to decide whether new presence messages should supersede items with the same clientId + connectionId in the presenceMap: https://github.com/ably/ably-js/blob/master/common/lib/client/realtimepresence.js#L256:L261. If we're removing those two from the object entirely (not just from the fromJSON() and toString() calls), then that problem'll be unsolved

SimonWoolf:
PR that only changes the json/string rep: https://github.com/ably/ably-js/pull/117

paddybyers:
`timestamp` is server-assigned, not client-assigned, so doesn't need to be sent by the client.

`connectionId`is implicit on a realtime connection, so also doesn't need to be sent by the client. Since the client never sends `PresenceMessage`s via REST, then there is no need to include it. `Message`s are of course different in that respect.

mattheworiordan:
[...]

PR #117, opened 2015-09-02
Don’t include timestamp/connectionid in string/json representations of messages

SimonWoolf:
And only include clientId in json rep if present

Fixes https://github.com/ably/ably-js/issues/44

### DIFF
--- a/common/lib/types/message.js
+++ b/common/lib/types/message.js
@@ -20,7 +20,6 @@ var Message = (function() {
 			name: this.name,
 			clientId: this.clientId,
 			connectionId: this.connectionId,
-			timestamp: this.timestamp,
 			encoding: this.encoding
 		};
 
--- a/common/lib/types/presencemessage.js
+++ b/common/lib/types/presencemessage.js
@@ -26,8 +26,6 @@ var PresenceMessage = (function() {
 	PresenceMessage.prototype.toJSON = function() {
 		var result = {
 			clientId: this.clientId,
-			connectionId: this.connectionId,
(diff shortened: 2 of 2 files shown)


common/lib/types/message.js
```diff
@@ -18,12 +18,12 @@ var Message = (function() {
 	Message.prototype.toJSON = function() {
 		var result = {
 			name: this.name,
-			clientId: this.clientId,
-			connectionId: this.connectionId,
```
paddybyers: It is valid to include a `connectionId` in `Message` when sent over REST so I don't think it's valid to remove it here.

common/lib/types/presencemessage.js
```diff
@@ -50,8 +50,6 @@ var PresenceMessage = (function() {
 		result += '; action=' + this.action;
 		if(this.id)
 			result += '; id=' + this.id;
-		if(this.timestamp)
```
paddybyers: Why is this removed?

common/lib/types/message.js
```diff
@@ -43,12 +43,8 @@ var Message = (function() {
 			result += '; name=' + this.name;
 		if(this.id)
 			result += '; id=' + this.id;
-		if(this.timestamp)
```
paddybyers: Why is this removed (also `connectionId` below) ?

SimonWoolf:
@paddybyers sorry, I misread the issue asking for timestamp/connectionId to be removed from toString as well as toJSON

paddybyers:
Thanks

PR #118, opened 2015-09-03
'utf-8' encoding currently assumes a wordArray?

SimonWoolf:
So currently, if ably-js is fed a message like [this one](https://github.com/ably/ably-common/blob/master/test-resources/test-app-setup.json#L21) (not encrypted, but with a `utf-8` encoding), it feeds it to `BufferUtils.utf8Decode` because of the `utf-8` in the encoding, which in a browser, passes it to `CryptoJS.enc.Utf8.stringify`, which [expects](https://code.google.com/p/crypto-js/#Encoders) a wordArray so outputs an empty string. (in node it just calls `toString('utf-8')`, which works fine).

This PR fixes that immediate problem, but I'm not sure if there's a deeper issue here - not clear to me if there's a significant difference in how the different libs interpret `utf-8` (ably-js [treating it](https://github.com/ably/ably-js/blob/master/browser/lib/util/bufferutils.js) as meaning wordArray-encoded, ably-ruby [treating it](https://github.com/ably/ably-ruby/blob/master/lib/ably/models/message_encoders/utf8.rb) as meaning just a utf-8-encoded string), or if I'm misunderstanding the code

@mattheworiordan @paddybyers

### DIFF
--- a/browser/lib/util/bufferutils.js
+++ b/browser/lib/util/bufferutils.js
@@ -101,8 +101,10 @@ var BufferUtils = (function() {
 	};
 
 	BufferUtils.utf8Decode = function(buf) {
-		if(CryptoJS)
+		if(isWordArray(buf) && CryptoJS)
 			return CryptoJS.enc.Utf8.stringify(buf);
+		else
+			return buf.toString();
 	};
 
 	BufferUtils.bufferCompare = function(buf1, buf2) {


paddybyers:
The test case is wrong.

`json` is a transformation that converts and object to a string. Then that resulting string becomes the `data` in the `Message`. There is no `utf-8` transformation, and it should not be present in the encoding string.

If Ruby is doing something different, then it's broken.

paddybyers:
Your commit is just skipping the `utf-8` decode in the problematic case, which is a clue to the fact that it shouldn't be there :)

SimonWoolf:
I *think* it's fine since with ruby, I think the openssl library must be doing the wordArray conversion implicitly as part of encryption/decryption; so if it is never valid to have `utf-8` in the encoding string unless it's encrypted, then it doesn't matter if one lib puts the wordarray conversion in the utf-8 step and the other in the encryption step, as the end result's the same.

(Though in that case there might still be differences between the libs if a string is encrypted but there's no `utf-8` in the encoding string -- except that's a crypto-data fixture, so that's tested to work... I'll try & understand that more tomorrow).

So, OK, I'll close this PR in favour of [this ably-common one](https://github.com/ably/ably-common/pull/7)

paddybyers:
The `utf-8` transformation converts a string to a byte array. So if `utf-8` is present as the final encoding step, or is present with all later steps being binary->binary (eg `cipher`), then the `data` payload on the wire **must** be binary instead of a string. If Ruby emits messages whose `data` is encoded as a string, but whose encoding contains `utf-8`, then it won't be interoperable with valid clients.

It is valid to have `utf-8` without `cipher`, provided that `data` is binary. But it isn't required in any circumstances (currently), because you can omit `utf-8` and send a string instead.

So I agree if the library only has this transient anomaly where there is notionally an as-yet-unencrypted-and-unencoded string with an encoding containing `utf-8`, but that state is never exposed, then I suppose that's ok, but it's at least going to be confusing to anyone trying to understand the code.

And just to be clear, we're using `WordArray` as the binary representation in the browser case (at least for browsers that don't have `ArrayBuffer`), and that (in principle) is the binary format irrespective of whether or not that data is to be encrypted. It just happens that that format came from the crypto library.

SimonWoolf:
Thanks for clarifying :)

> If Ruby emits messages whose data is encoded as a string, but whose encoding contains utf-8, then it won't be interoperable with valid clients.

Double-checked, and no, the only time ruby client adds 'utf-8' to the encoding string is by the cipher encoding, which results in binary. So just a bad ably-common fixture. (But will still file a bug in the ably-ruby library because [`MessageEncoders::Utf8`](https://github.com/ably/ably-ruby/blob/master/lib/ably/models/message_encoders/utf8.rb) is a bit misleading right now)

paddybyers:
ok, thanks

mattheworiordan:
> If Ruby emits messages whose data is encoded as a string, but whose encoding contains utf-8, then it won't be interoperable with valid clients.

We've been here before, but Ruby is different in that a string is not effectively word array, but instead is  a char array with an encoding.  So passing a UTF8 string to OpenSSL works because it understands the UTF-8 encoded char array, and will thus ensure it is indeed interoperable with other libraries, I believe that is specifically the reason we introduced the 128 & 256 fixtures to ensure that.  We can add more tests cases if you are concerned about this, in fact I recall requesting this some time back for this reason.

> So I agree if the library only has this transient anomaly where there is notionally an as-yet-unencrypted-and-unencoded string with an encoding containing utf-8, but that state is never exposed, then I suppose that's ok, but it's at least going to be confusing to anyone trying to understand the code.

I agree it may be confusing, but I would prefer that this PR lands with a comment about why it is there, as opposed to failing as it does now.  

Either way, we clearly need documentation about the encoders as this is unclear, I will add that to the spec in a PR for review.

@paddybyers any objection to the `ably-js` library handling this incorrect encoding when it clearly is able to do so?  We can certainly add a comment about why.  Given we have had shared fixture data that the other client libraries have been built against, I would prefer to make sure it works as opposed to changing all the client libraries.

paddybyers:
> We've been here before, but Ruby is different

That can affect how things are implemented but it can't affect what goes out on the wire.

> any objection to the ably-js library handling this incorrect encoding when it clearly is able to do so?

Yes, because there is no valid reason for the library ever to handle such malformed messages.

mattheworiordan:
> Yes, because there is no valid reason for the library ever to handle such malformed messages.

How about the fact that the shared (incorrect) fixture data has been around since April and everyone has been using that, see https://github.com/ably/ably-common/commit/34b3937475b037e5827cafd1b6b1f3a57b7f06b2?  In the interests of keeping things moving forwards, I don't see any downside in a workaround to allow this for backwards support.  We do that currently in the realtime system in various places, so don't see how this is different.

paddybyers:
I'm sorry, I disagree. The fixture data was wrong, and so there is a risk that every implementation depending on it is broken. We should fix it.

mattheworiordan:
> I'm sorry, I disagree. The fixture data was wrong, and so there is a risk that every implementation depending on it is broken. We should fix it.

It is fixed, https://github.com/ably/ably-common/commit/5d4c83eebcafc96e6f188ee9a898e10fd6d8a88a

I still don't see any downside in having this fallback.  You have fallbacks for the params in camelCase versus snake_case, how is that any different?

paddybyers:
We have a valid reason for admitting both camel case and snake case - because of implementation convenience in different languages.

This case is different because there is no valid reason for any library to send an `encoding` string that doesn't correspond to the type of `data`.

According to our spec, when decoding a received message, if any component of the `encoding` string cannot be reversed, the `data` is left unchanged at that point, and that `data` and the residual `encoding` string are returned to the client. We don't say that the encoding step is simply discarded because of a mismatch between the expected and actual types.

In the case of a message whose data is a String, the `utf-8` component of the encoding can't be "unapplied" because that only applies to binary data. So according to our spec we should return the String and the encoding with the `utf-8` component still present, and not apply any further decoding steps.

SimonWoolf:
> So according to our spec we should return the String and the encoding with the utf-8 component still present, and not apply any further decoding steps.

FWIW that's not the current behaviour. As it stands `CryptoJS.enc.Utf8.stringify`, if fed a string, returns an empty string. So the lib continues on happily with the data vanished, until JSON.parse throws because an empty string isn't valid JSON.

I'll do another PR later to fix that.

PR #119, opened 2015-09-04
Use ably-common submodule instead of inline json for test app setup

SimonWoolf:
- move file loading code from testdata module to testapp-manager, as need it to load test-app-setup.json (and testdata module already depended on testapp-manager, so would be a circular dependency)
- Change capabilities tests to use keys in ably-common app setup


paddybyers:
Thanks

PR #122, opened 2015-09-07
Don't process duplicate messages or messages on inactive transports

SimonWoolf:
So, in new-upgrade, sometimes you get duplicate messages from realtime due to races. (Eg: a new transport is made active, it sends a sync message with connectionSerial -1, it receives a new action:15 message with connectionSerial 0, realtime processes the sync and sends that message again. Most of the tests in spec/realtime/message.test.js (run against local realtime for less lag to make it more likely) can show this behaviour.)

But that shouldn't be a problem: the messages have the same connectionSerial, so the client should just discard the second one. Isn't that part of the point of a connectionSerial? But it currently doesn't: messages are processed even if their connectionSerial is <= the client's connectionSerial.

In addition, messages are processed even if they come through on a no-longer-active transport (although in that case the client's connectionSerial is not updated). So just adding a check for connectionSerial isn't enough, as then you have a race condition where, say, closing the connection clears the connection state (which sets the connectionSerial to `undefined`) - and then a message comes in on that transport and gets processed, even though the connection's nominally closed, as the connectionManager can't tell it's a duplicate due to lack of a connectionSerial. Why do we want messages coming in on inactive transports to be processed at all?

### DIFF
--- a/common/lib/transport/connectionmanager.js
+++ b/common/lib/transport/connectionmanager.js
@@ -852,14 +852,26 @@ var ConnectionManager = (function() {
 	};
 
 	ConnectionManager.prototype.onChannelMessage = function(message, transport) {
-		/* do not update connectionSerial for messages received
-		 * on transports that are no longer the current transport */
 		if(this.activeProtocol && transport === this.activeProtocol.getTransport()) {
 			var connectionSerial = message.connectionSerial;
-			if(connectionSerial !== undefined)
+			if(connectionSerial <= this.connectionSerial) {
+				Logger.logAction(Logger.LOG_MICRO, 'ConnectionManager.onChannelMessage() received message with connectionSerial ' + connectionSerial + ', but current connectionSerial is ' + this.connectionSerial + '; assuming message is a duplicate and discarding it');
+				return;
+			}
+			if(connectionSerial !== undefined) {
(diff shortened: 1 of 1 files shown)


paddybyers:
> Why do we want messages coming in on inactive transports to be processed at all?

It is still legitimate to continue to pubish messages on an outgoing transport while the SYNC on the new transport is in progress, and even after the new transport is fully active (ie sync is complete) the outgoing transport remains alive to get the `ACK` messages that are outstanding on messages sent on it.

Apart from that, all messages received on an outgoing transport should be discarded from the moment that the `SYNC` is sent on the new transport.

SimonWoolf:
@paddybyers makes sense, updated PR. (My problems were happening on close rather than upgrade, but I guess the same logic (only allowing acks/nacks/errors) applies there too.)

> Apart from that, all messages received on an outgoing transport should be discarded from the moment that the SYNC is sent on the new transport.

I think we can get away with just using whether transport is the active transport rather than creating any new logic, since scheduleTransportActivation calls `sync()` right after `activateTransport()`

Everything ok with the connectionSerial logic? (was surprised the lib didn't do it already, wasn't sure if there was some reason for that I hadn't worked out)

paddybyers:
Thanks

Issue #121, opened 2015-09-04
Invalid encoding should not result in the data vanishing


SimonWoolf:
https://github.com/ably/ably-js/pull/118#issuecomment-137819459

paddybyers:
You had me confused with the title there ....

SimonWoolf:
Fixed by https://github.com/ably/ably-js/pull/123

Issue #39, opened 2015-04-22
Emit errors


mattheworiordan:
See https://github.com/ably/ably-java/issues/16

mattheworiordan:
See https://github.com/ably/ably-js/issues/43

mattheworiordan:
This will be covered in the spec

PR #123, opened 2015-09-07
Invalid encodings

SimonWoolf:
Fixes https://github.com/ably/ably-js/issues/121

Throw errors if input to utf8Decode is not a wordArray or an arrayBuffer, and properly handle when it's an arrayBuffer.

Also remove unnecessary if(CryptoJS) calls. if CryptoJS is undefined (though it shouldn't be, we bundle it in), we want it to throw an error (which it will do when it tries to call a method on undefined) to be caught by the Message.decode try/catch block, not silently return null. (Though it'll error because of [this](https://github.com/ably/ably-js/blob/master/browser/lib/util/bufferutils.js#L2) anyway)

Also add tests for the changed behaviour.

The first commit (86b86ef) is a separate PR: https://github.com/ably/ably-js/pull/122. included here as it's needed to make the test work; so that one should be merged first.

### DIFF
--- a/browser/lib/util/bufferutils.js
+++ b/browser/lib/util/bufferutils.js
@@ -90,19 +90,19 @@ var BufferUtils = (function() {
 	BufferUtils.base64Decode = function(str) {
 		if(ArrayBuffer)
 			return base64ToArrayBuffer(str);
-
-		if(CryptoJS)
-			return CryptoJS.enc.Base64.parse(str);
+		return CryptoJS.enc.Base64.parse(str);
 	};
 
 	BufferUtils.utf8Encode = function(string) {
-		if(CryptoJS)
-			return CryptoJS.enc.Utf8.parse(string);
+		return CryptoJS.enc.Utf8.parse(string);
 	};
 
 	BufferUtils.utf8Decode = function(buf) {
-		if(CryptoJS)
+		if(isArrayBuffer(buf))
+			buf = BufferUtils.toWordArray(buf) // CryptoJS only works with WordArrays
+		if(isWordArray(buf))
 			return CryptoJS.enc.Utf8.stringify(buf);
+		throw new Error("Expected input of utf8Decode to be a buffer or CryptoJS WordArray");
 	};
 
 	BufferUtils.bufferCompare = function(buf1, buf2) {
--- a/common/lib/client/realtimechannel.js
+++ b/common/lib/client/realtimechannel.js
@@ -274,11 +274,8 @@ var RealtimeChannel = (function() {
 					var presenceMsg = presence[i];
 					PresenceMessage.decode(presenceMsg, options);
 				} catch (e) {
-					/* decrypt failed .. the most likely cause is that we have the wrong key */
-					var errmsg = 'Unexpected error decoding message; err = ' + e;
-					Logger.logAction(Logger.LOG_ERROR, 'RealtimeChannel.onMessage()', errmsg);
-					var err = new Error(errmsg);
-					this.emit('error', err);
+					Logger.logAction(Logger.LOG_ERROR, 'RealtimeChannel.onMessage()', e.toString());
+					this.emit('error', e);
 				}
 				if(!presenceMsg.connectionId) presenceMsg.connectionId = connectionId;
 				if(!presenceMsg.timestamp) presenceMsg.timestamp = timestamp;
@@ -300,10 +297,8 @@ var RealtimeChannel = (function() {
 					Message.decode(msg, options);
 				} catch (e) {
 					/* decrypt failed .. the most likely cause is that we have the wrong key */
-					var errmsg = 'Unexpected error decoding message; err = ' + e;
-					Logger.logAction(Logger.LOG_ERROR, 'RealtimeChannel.onMessage()', errmsg);
-					var err = new Error(errmsg);
-					this.emit('error', err);
+					Logger.logAction(Logger.LOG_ERROR, 'RealtimeChannel.onMessage()', e.toString());
+					this.emit('error', e);
 				}
 				if(!msg.connectionId) msg.connectionId = connectionId;
 				if(!msg.timestamp) msg.timestamp = timestamp;
--- a/common/lib/types/message.js
+++ b/common/lib/types/message.js
@@ -128,16 +128,19 @@ var Message = (function() {
 								var xformAlgorithm = match[3], cipher = options.cipher;
 								/* don't attempt to decrypt unless the cipher params are compatible */
 								if(xformAlgorithm != cipher.algorithm) {
-									Logger.logAction(Logger.LOG_ERROR, 'Message.decode()', 'Unable to decrypt message with given cipher; incompatible cipher params');
+									throw new Error('Unable to decrypt message with given cipher; incompatible cipher params');
 									break;
 								}
 								data = cipher.decrypt(data);
 								continue;
 							}
 						default:
+							throw new Error("Unknown encoding");
 					}
 					break;
 				}
+			} catch(e) {
+				throw new ErrorInfo('Error processing the ' + xform + ' encoding, decoder returned ‘' + e.message + '’', 40013, 400);
 			} finally {
 				message.encoding = (i <= 0) ? null : xforms.slice(0, i).join('/');
 				message.data = data;
@@ -145,13 +148,18 @@ var Message = (function() {
 		}
 	};
 
-	Message.fromResponseBody = function(body, options, format) {
+	Message.fromResponseBody = function(body, options, format, channel) {
 		if(format)
 			body = (format == 'msgpack') ? msgpack.decode(body) : JSON.parse(String(body));
 
 		for(var i = 0; i < body.length; i++) {
 			var msg = body[i] = Message.fromDecoded(body[i]);
-			Message.decode(msg, options);
+			try {
+				Message.decode(msg, options);
+			} catch (e) {
+				Logger.logAction(Logger.LOG_ERROR, 'Message.fromResponseBody()', e.toString());
+				channel.emit('error', e);
+			}
 		}
 		return body;
 	};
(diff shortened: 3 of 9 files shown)


browser/lib/util/bufferutils.js
```diff
@@ -90,19 +90,19 @@ var BufferUtils = (function() {
...
+		return CryptoJS.enc.Utf8.parse(string);
 	};
 
 	BufferUtils.utf8Decode = function(buf) {
-		if(CryptoJS)
+		if(isArrayBuffer(buf))
+			buf = BufferUtils.toWordArray(buf) // CryptoJS only works with WordArrays
+		if(isWordArray(buf))
 			return CryptoJS.enc.Utf8.stringify(buf);
+		throw new Error("Expected input of utf8Decode to be a buffer");
```
paddybyers: ErrorInfo please ....

SimonWoolf: They see `Ably: RealtimeChannel.onMessage(): Unexpected error decoding message; err = [Error: Expected input of utf8Decode to be a buffer]`, or with errorInfo:  `Ably: RealtimeChannel.onMessage(): Unexpected error decoding message; err = [ErrorInfo: Expected input of utf8Decode to be a buffer; statusCode=123; code=12345]` (or whatever)

(so the statusCode is not present in the final emitted error either way - it's emitted by https://github.com/ably/ably-js/blob/master/common/lib/client/realtimechannel.js#L267)

Can change the error message to `'Message had encoding component utf-8, which ably-js expect to be represented as a binary data buffer or CryptoJS WordArray; but the given data was a ' + typeof buf` or something. What statusCode were you thinking? None seem to really fit afaics.

paddybyers: Ideally when the exception is caught in decode(), the resultant error will identity the encoding component that failed (ie `utf-8`), and will include the error message from the root error. I don't think it needs to be quite as verbose as what you suggest there.

I think 400 is the most appropriate `statusCode`, and probably a new `code` `40013` `Invalid message data or encoding`

I understand that the `ErrorInfo` fields don't make it into the final error given to the client but I'd still prefer that it's an ErrorInfo. We might in the future have a `cause` field, for example, where outer errors wrap the underlying error.

SimonWoolf: done

paddybyers:
Sorry for not being clearer, but this isn't quite what I meant.

I think the error thrown by the generic `utf8Decode()` function needs to have a generic error message; ie 'Unable to decode non-binary data' or whatever I suggested in my earlier (now deleted) comment.

Then that exception is caught, and a different error is thrown that indicates that it was a message decoding error. That's the operation that needs to form the error message indicating the encoding component that was the culprit.

However, I see now that `Message.decode()` doesn't intercept the thrown exception at all, and it's caught in `RealtimeChannel.onMessage()`. This is a problem because at this point you don't know what the encoding component was.

So I think `Message.decode() needs a `catch` (lexically before the `finally`) which catches the error, forms a new error with the composite message, and the 400013 code, and throws it.

Then we can remove the code that wraps the error with a new error in `RealtimeChannel.onMessage()`, because we simply want the (now-wrapped-in-`Message.decode()`) error logged and emitted.

Also check what happens with decode errors in the Rest history case: it looks to me like `Message.fromResponseBody()` needs to do some catching as well. And it needs a test ....

SimonWoolf:
@paddybyers is that more what you were thinking?

(Couldn't find a nice way to test the Logger output for the rest history request, where there's no channel event emitted (and we definitely don't want to callback(err) on the history request), so it's just a console log. In the end I just mocked out the console, which is a horrendously bad idea and you should :-1: me for it.

Should it emit an event on the channel? Rest channels do inherit from eventEmitter, but AFAICT right now only realtime channels ever emit any events

common/lib/client/realtimechannel.js
```diff
@@ -274,11 +274,8 @@ var RealtimeChannel = (function() {
 					var presenceMsg = presence[i];
 					PresenceMessage.decode(presenceMsg, options);
 				} catch (e) {
-					/* decrypt failed .. the most likely cause is that we have the wrong key */
-					var errmsg = 'Unexpected error decoding message; err = ' + e;
-					Logger.logAction(Logger.LOG_ERROR, 'RealtimeChannel.onMessage()', errmsg);
-					var err = new Error(errmsg);
-					this.emit('error', err);
+					Logger.logAction(Logger.LOG_ERROR, 'RealtimeChannel.onMessage()', e.toString());
+					this.emit('error', e.toString());
```
paddybyers: we should emit an `ErrorInfo` instance, not just a string

common/lib/client/realtimechannel.js
```diff
@@ -300,10 +297,8 @@ var RealtimeChannel = (function() {
 					Message.decode(msg, options);
 				} catch (e) {
 					/* decrypt failed .. the most likely cause is that we have the wrong key */
-					var errmsg = 'Unexpected error decoding message; err = ' + e;
-					Logger.logAction(Logger.LOG_ERROR, 'RealtimeChannel.onMessage()', errmsg);
-					var err = new Error(errmsg);
-					this.emit('error', err);
+					Logger.logAction(Logger.LOG_ERROR, 'RealtimeChannel.onMessage()', e.toString());
+					this.emit('error', e.toString());
```
paddybyers: ditto

common/lib/types/message.js
```diff
@@ -151,7 +154,11 @@ var Message = (function() {
 
 		for(var i = 0; i < body.length; i++) {
 			var msg = body[i] = Message.fromDecoded(body[i]);
-			Message.decode(msg, options);
+			try {
+				Message.decode(msg, options);
+			} catch (e) {
+				Logger.logAction(Logger.LOG_ERROR, 'Message.fromResponseBody()', e.toString());
+			}
```
paddybyers: emit here

common/lib/types/presencemessage.js
```diff
@@ -76,7 +76,11 @@ var PresenceMessage = (function() {
 
 		for(var i = 0; i < body.length; i++) {
 			var msg = body[i] = PresenceMessage.fromDecoded(body[i]);
-			PresenceMessage.decode(msg, options);
+			try {
+				PresenceMessage.decode(msg, options);
+			} catch (e) {
+				Logger.logAction(Logger.LOG_ERROR, 'PresenceMessage.fromResponseBody()', e.toString());
+			}
```
paddybyers: ditto

paddybyers:
> So I think Message.decode() needs acatch(lexically before thefinally`) which catches the error, forms a new error with the composite message, and the 400013 code, and throws it.

I think you missed this bit?

common/lib/types/message.js
SimonWoolf: emit on what? Messages aren't EventEmitters

SimonWoolf:
> I think you missed this bit?

it's here: https://github.com/ably/ably-js/pull/123/files#diff-da763f0daf85654e40970bc4867e316cR143 . (If I'd missed that, the tests that check for 40013 would've failed)

paddybyers:
> don't think so

sorry, my mistake

browser/lib/util/bufferutils.js
```diff
@@ -90,19 +90,19 @@ var BufferUtils = (function() {
...
+		return CryptoJS.enc.Utf8.parse(string);
 	};
 
 	BufferUtils.utf8Decode = function(buf) {
-		if(CryptoJS)
+		if(isArrayBuffer(buf))
+			buf = BufferUtils.toWordArray(buf) // CryptoJS only works with WordArrays
+		if(isWordArray(buf))
 			return CryptoJS.enc.Utf8.stringify(buf);
+		throw new ErrorInfo("Expected encoding component utf-8 to be a buffer or CryptoJS WordArray, but was " + typeof buf, 40013, 400);
```
paddybyers: This should now be a generic message relating to utf8Decode, not one implying it was called from `Message.decode()`.

common/lib/types/presencemessage.js
```diff
@@ -70,13 +70,18 @@ var PresenceMessage = (function() {
...
 			body = (format == 'msgpack') ? msgpack.decode(body) : JSON.parse(String(body));
 
 		for(var i = 0; i < body.length; i++) {
 			var msg = body[i] = PresenceMessage.fromDecoded(body[i]);
-			PresenceMessage.decode(msg, options);
+			try {
+				PresenceMessage.decode(msg, options);
+			} catch (e) {
+				Logger.logAction(Logger.LOG_ERROR, 'PresenceMessage.fromResponseBody()', e.toString());
+				channel.emit('error', e);
```
paddybyers: can you just double-check that the spec doesn't say something more specific here? Or maybe you did already/

paddybyers:
Just a couple of comments, otherwise LGTM

common/lib/types/presencemessage.js
SimonWoolf: Only relevant one I can see is (RSL6b), which is for generic decoding in rest lib - "*If, for example, incompatible encryption details are provided or invalid Base64 is detected in the message payload, an error message will be sent to the logger, but the message will still be delivered with last successful decoding and the encoding field. For example, if a message had a decoding of “utf-8/cipher+aes-128-cbc/base64”, and the payload was successfully Base64 decoded but the payload could not be encrypted because the CipherParam details were not configured, the message would be delivered with a binary payload and an encoding with the value “utf-8/cipher+aes-128-cbc”*"

..which doesn't even mention events. (hence the question earlier about whether we should be emitting events on a rest channel at all. RTL7e (for realtime channels) mentions emitting an error event but doesn't say anything specific, just "*and an error should be emitted on the channel*")

paddybyers: See here also: https://github.com/ably/ably-js/issues/39

I think we've agreed to emit an error event on the channel object but I think Ruby is the only lib that does that presently.

PR #124, opened 2015-09-08
Add some rest presence tests

SimonWoolf:
much simpler now there's the ably-common fixtures and the normalised cipher interface, so based off the `CipherParams` branch


paddybyers:
Thanks

Issue #98, opened 2015-08-21
ably-js doesn't dispose of the websocket resource if the connection attempt times out


SimonWoolf:
See https://github.com/ably/ably-js/pull/91#issuecomment-132729832

SimonWoolf:
needs to be reexamined with new-upgrade

SimonWoolf:
@me to resurrect https://github.com/ably/ably-js/commit/e50af01f04678ef0b6051493b717f3dfa2fa4180#diff-d62974e471aa1550b450c96d34575c78R134 with renamed timer

SimonWoolf:
Fixed by https://github.com/ably/ably-js/pull/125

PR #125, opened 2015-09-09
Ensure websocket object is disposed if it fails to connect

SimonWoolf:
Per conversation yesterday. Fixes https://github.com/ably/ably-js/issues/98


paddybyers:
Thanks

PR #126, opened 2015-09-10
Device connection state events 2

SimonWoolf:
New PR as wanted to raise against `new-upgrade` rather than `master`.

Addresses https://github.com/ably/ably-js/issues/99 (apart from the heartbeat thing)

@paddybyers

### DIFF
--- a/common/lib/transport/connectionmanager.js
+++ b/common/lib/transport/connectionmanager.js
@@ -800,6 +800,31 @@ var ConnectionManager = (function() {
 		this.notifyState({state: 'closed'});
 	};
 
+	ConnectionManager.prototype.disconnectAllTransports = function() {
+		Logger.logAction(Logger.LOG_MINOR, 'ConnectionManager.disconnectAllTransports()', 'disconnecting all transports');
+
+		function disconnectTransport(transport) {
+			if(transport) {
+				try {
+					Logger.logAction(Logger.LOG_MINOR, 'ConnectionManager.disconnectAllTransports()', 'disconnecting transport: ' + transport);
+					transport.disconnect();
+				} catch(e) {
+					var msg = 'Unexpected exception attempting to disconnect transport; e = ' + e;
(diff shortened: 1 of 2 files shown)


common/lib/transport/connectionmanager.js
```diff
@@ -123,6 +123,29 @@ var ConnectionManager = (function() {
...
+				}
+			});
+			window.addEventListener('offline', function() {
+				if(self.state == states.connected) {
+					Logger.logAction(Logger.LOG_MINOR, 'ConnectionManager caught browser ‘offline’ event', 'disconnecting active transport');
+					// Not sufficient to just go to the 'disconnected' state, want to
+					// force all transports to reattempt the connection
+					self.disconnectAllTransports();
+					// Try a reconnect immediately rather than wait for the disconnect timeout
+					Utils.nextTick(function() {
```
paddybyers: Is this necessary? I thought we would always go immediately into `connecting` when a disconnection occurs from `connected`. I thought we only had the disconnected timeout if entering `disconnected` from `connecting`.

SimonWoolf: > I thought we would always go immediately into connecting when a disconnection occurs from connected

Doesn't seem to do that, no. Might be missing something, but I can't see any mechanism for that to happen - the retry timer in notifyState is set to whatever the retryDelay for the new state is, doesn't look at the previous state (except to check if it's already there or is a terminal state).

paddybyers: Well it should :) I was thinking of the java library which checks for this explicitly. Maybe in `deactivateTransport` we need to explicitly reattempt connection instead of just notifying the `disconnected` state from the transport.

SimonWoolf: How about something like https://github.com/ably/ably-js/commit/482d277d589182f8b27807cb181c88a6ad46431a? (on PR #130)

SimonWoolf:
Closing in favour of https://github.com/ably/ably-js/pull/130 (raised against `master` as `new-upgrade`'s been merged)

PR #127, opened 2015-09-12
Tests no force ws; fix failure to callback on presence event causing implicit attach

SimonWoolf:
Started out as a PR to just stop forcing web_sockets when not necessary (fixes https://github.com/ably/ably-js/pull/127), ended up also fixing a couple bugs that that exposed:

- If there's an error in an implicit attach caused by sending a presence message, callback with that err. (was only a problem with presence not with normal messages since normal messages get their callback called in `RealtimeChannel#failPendingMessages`). + test (and another test for normal messages, though that one was already passing)

- Fixed an onNack callback typo in protocol.js. (indicator that codepath for NACKs is not well-tested?)

### DIFF
--- a/common/lib/client/realtimepresence.js
+++ b/common/lib/client/realtimepresence.js
@@ -56,7 +56,14 @@ var RealtimePresence = (function() {
 				break;
 			case 'initialized':
 			case 'detached':
-				channel.attach();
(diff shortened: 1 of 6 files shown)


paddybyers:
Thanks

PR #130, opened 2015-09-15
Connection state events 3

SimonWoolf:
Reraising https://github.com/ably/ably-js/pull/126 against master as new-upgrade has merged.

Added additional commit https://github.com/ably/ably-js/commit/482d277d589182f8b27807cb181c88a6ad46431a per discussion in old PR, and redid the second test to stub out `chooseTransport` rather than calling `disconnectAllTransports` on('connected') to fix the immediate reconnection making that test fail.

### DIFF
--- a/common/lib/transport/connectionmanager.js
+++ b/common/lib/transport/connectionmanager.js
@@ -800,6 +800,31 @@ var ConnectionManager = (function() {
 		this.notifyState({state: 'closed'});
 	};
 
+	ConnectionManager.prototype.disconnectAllTransports = function() {
+		Logger.logAction(Logger.LOG_MINOR, 'ConnectionManager.disconnectAllTransports()', 'disconnecting all transports');
+
+		function disconnectTransport(transport) {
+			if(transport) {
+				try {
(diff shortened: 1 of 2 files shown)


common/lib/transport/connectionmanager.js
```diff
@@ -90,6 +79,17 @@ var ConnectionManager = (function() {
 		EventEmitter.call(this);
 		this.realtime = realtime;
 		this.options = options;
+		states = {
```
paddybyers: What's the purpose of reinitialising this each time a `ConnectionManager` is initialised, when it's reinitialising something with global scope? Either the structure is immutable, in which case it can be initialised once, or it's mutable (with per-instance state), in which case it should have per-instance scope.

SimonWoolf: Good point.  d6aef9e

paddybyers:
Thanks

mattheworiordan:
FYI, it would have been better to change the text on my commit to remove the (WIP) bit...

mattheworiordan:
Nice work @SimonWoolf

PR #132, opened 2015-09-24
Avoid ambiguity in received transport protocol message

mattheworiordan:
@paddybyers @SimonWoolf 

Whilst looking through some code, I found the name `onChannelMessage` incredibly misleading, especially when `ChannelManager.onChannelMessage` expects a ChannelMessage, whereas `Transport.onChannelMessage` expects any type of ProtocolMessage.


paddybyers:
Thanks

Issue #134, opened 2015-09-25
XHR requests timing out leads to connection going into failed state


SimonWoolf:
An xhr request timing out (the sendTimeout) results in this chain of events:
- [xhr.abort() called](https://github.com/ably/ably-js/blob/7560241/browser/lib/transport/xhrrequest.js#L69)
- [xhr request completes with a 400/80000 error](https://github.com/ably/ably-js/blob/7560241/browser/lib/transport/xhrrequest.js#L100-L108)
- [That error is processed as a protocol error](https://github.com/ably/ably-js/blob/756024196ca404a7e9a42cccbb1297ef377533d2/common/lib/transport/comettransport.js#L195)
- [Transport is aborted with that error](https://github.com/ably/ably-js/blob/756024196ca404a7e9a42cccbb1297ef377533d2/common/lib/transport/transport.js#L99)
- [Transport#abort emits 'failed' with that error](https://github.com/ably/ably-js/blob/756024196ca404a7e9a42cccbb1297ef377533d2/common/lib/transport/transport.js#L52)
- [ConnectionManager catches the event and calls deactivateTransport with a 'failed' argument](https://github.com/ably/ably-js/blob/756024196ca404a7e9a42cccbb1297ef377533d2/common/lib/transport/connectionmanager.js#L359-L368)
- [ConnectionManage#deactivateTransport calls notifyState with {state: 'failed'}](https://github.com/ably/ably-js/blob/756024196ca404a7e9a42cccbb1297ef377533d2/common/lib/transport/connectionmanager.js#L502)

Net effect: a single xhr request timing out results in the connection going into the 'failed' state (so it won't be retried), which is suboptimal.

None of the individual steps seem obviously wrong. My stab at the least correct step is the last one: I'm not sure a failed transport should necessarily imply a failed connection. PR based on that: #135

SimonWoolf:
Fixed by https://github.com/ably/ably-js/pull/135

PR #135, opened 2015-09-25
Transport failures don’t necessarily imply connection failures

SimonWoolf:
Proposed fix for #134

### DIFF
--- a/common/lib/transport/comettransport.js
+++ b/common/lib/transport/comettransport.js
@@ -92,10 +92,10 @@ var CometTransport = (function() {
 		this.dispose();
 	};
 
-	CometTransport.prototype.abort = function() {
-		Logger.logAction(Logger.LOG_MINOR, 'CometTransport.abort()', '');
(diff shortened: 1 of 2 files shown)


common/lib/transport/comettransport.js
```diff
@@ -92,10 +92,10 @@ var CometTransport = (function() {
 		this.dispose();
 	};
 
-	CometTransport.prototype.abort = function() {
-		Logger.logAction(Logger.LOG_MINOR, 'CometTransport.abort()', '');
+	CometTransport.prototype.abort = function(err) {
+		Logger.logAction(Logger.LOG_MINOR, 'CometTransport.abort()', 'err: ' + JSON.stringify(err));
```
paddybyers: We have `Utils.inspect()` for this kind of thing. Or, if we know that `err` is certain to be an `ErrorInfo`, then just `err` will do because `toString()` does what we want.

paddybyers:
LGTM apart from the comment above.

PR #139, opened 2015-09-30
Log protocol messages at micro level

mattheworiordan:
@paddybyers and @SimonWoolf please review this PR as I find it quite difficult to understand where failures are occurring over Realtime and REST when using the `ably-js` library.

I would like to merge this in if possible so that I can add the necessary logging easily from the realtime QOS tests that are failing.

### DIFF
--- a/browser/lib/transport/xhrtransport.js
+++ b/browser/lib/transport/xhrtransport.js
@@ -9,8 +9,12 @@ var XHRTransport = (function() {
 	XHRTransport.isAvailable = XHRRequest.isAvailable;
 
 	XHRTransport.checkConnectivity = function(callback) {
-		Http.Request('http://internet-up.ably.io.s3-website-us-east-1.amazonaws.com/is-the-internet-up.txt', null, null, null, function(err, responseText) {
-			callback(null, (!err && responseText == 'yes'));
+		var upUrl = 'http://internet-up.ably.io.s3-website-us-east-1.amazonaws.com/is-the-internet-up.txt';
+		Logger.logAction(Logger.LOG_MICRO, 'XHRTransport.checkConnectivity()', 'Sending; ' + upUrl);
+		Http.Request(upUrl, null, null, null, function(err, responseText) {
+			var result = (!err && responseText == 'yes');
+			Logger.logAction(Logger.LOG_MICRO, 'XHRTransport.checkConnectivity()', 'Result: ' + result);
+			callback(null, result);
 		});
 	};
 
--- a/common/lib/client/auth.js
+++ b/common/lib/client/auth.js
@@ -239,8 +239,15 @@ var Auth = (function() {
 				}
 			}
 			tokenRequestCallback = function(params, cb) {
-				var authHeaders = Utils.mixin({accept: 'application/json'}, authOptions.authHeaders);
-				Http.getUri(rest, authOptions.authUrl, authHeaders || {}, Utils.mixin(params, authOptions.authParams), function(err, body, headers, unpacked) {
+				var authHeaders = Utils.mixin({accept: 'application/json'}, authOptions.authHeaders),
+						authParams = Utils.mixin(params, authOptions.authParams);
+				Logger.logAction(Logger.LOG_MICRO, 'Auth.requestToken().tokenRequestCallback', 'Sending; ' + authOptions.authUrl + '; Params: ' + JSON.stringify(authParams));
+				Http.getUri(rest, authOptions.authUrl, authHeaders || {}, authParams, function(err, body, headers, unpacked) {
+					if (err) {
+						Logger.logAction(Logger.LOG_MICRO, 'Auth.requestToken().tokenRequestCallback', 'Received Error; ' + JSON.stringify(err));
+					} else {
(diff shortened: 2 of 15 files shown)


paddybyers:
A few comments above but overall lgtm.

mattheworiordan:
@paddybyers & @SimonWoolf I think I have addressed all your comments + I refactored the way we do the REST logging because it was not decoding MsgPack payloads.  Mind reviewing the recent commits again?

Also, can you please advise how you would like me to handle 4724f17.  Should I remove the unrelated code for now?

common/lib/types/protocolmessage.js
```diff
@@ -66,5 +66,47 @@ var ProtocolMessage = (function() {
...
+		var result = [];
+		if (array) {
+			for (var i = 0; i < array.length; i++) {
+				result.push(array[i].toString());
+			}
+		}
+		return '[ ' + result.join(', ') + ' ]';
+	}
+
+	ProtocolMessage.stringify = function(msg) {
```
paddybyers: Why not `.prototype.toString` ?

common/lib/client/auth.js
```diff
@@ -239,8 +239,15 @@ var Auth = (function() {
...
-				var authHeaders = Utils.mixin({accept: 'application/json'}, authOptions.authHeaders);
-				Http.getUri(rest, authOptions.authUrl, authHeaders || {}, Utils.mixin(params, authOptions.authParams), function(err, body, headers, unpacked) {
+				var authHeaders = Utils.mixin({accept: 'application/json'}, authOptions.authHeaders),
+						authParams = Utils.mixin(params, authOptions.authParams);
+				Logger.logAction(Logger.LOG_MICRO, 'Auth.requestToken().tokenRequestCallback', 'Sending; ' + authOptions.authUrl + '; Params: ' + JSON.stringify(authParams));
+				Http.getUri(rest, authOptions.authUrl, authHeaders || {}, authParams, function(err, body, headers, unpacked) {
+					if (err) {
+						Logger.logAction(Logger.LOG_MICRO, 'Auth.requestToken().tokenRequestCallback', 'Received Error; ' + JSON.stringify(err));
+					} else {
+						Logger.logAction(Logger.LOG_MICRO, 'Auth.requestToken().tokenRequestCallback', 'Received; body: ' + JSON.stringify(BufferUtils.isBuffer(body) ? body.toString() : body));
```
paddybyers: Why `stringify()` this?

browser/lib/transport/xhrtransport.js
```diff
@@ -9,8 +9,12 @@ var XHRTransport = (function() {
 	XHRTransport.isAvailable = XHRRequest.isAvailable;
 
 	XHRTransport.checkConnectivity = function(callback) {
-		Http.Request('http://internet-up.ably.io.s3-website-us-east-1.amazonaws.com/is-the-internet-up.txt', null, null, null, function(err, responseText) {
-			callback(null, (!err && responseText == 'yes'));
+		var upUrl = 'http://internet-up.ably.io.s3-website-us-east-1.amazonaws.com/is-the-internet-up.txt';
```
paddybyers: I thought the URL was being moved to `Defaults` ?

paddybyers:
> Also, can you please advise how you would like me to handle 4724f17. Should I remove the unrelated code for now?

Do your other changes depend on it? If not, then we should handle it as a separate PR.

common/lib/types/protocolmessage.js
mattheworiordan: Because this a class method, not an instance method.  We are using the `ProtocolMessage` class method to convert a JSON object into a string representation.

browser/lib/transport/xhrtransport.js
mattheworiordan: That's a different PR

common/lib/client/auth.js
mattheworiordan: Because I am not sure what object the `body` will aways be, so I felt it's safer to do this as opposed to ending up with `[Object object]` which is all too common when not using stringify.  What is the downside?

mattheworiordan:
> Do your other changes depend on it? If not, then we should handle it as a separate PR.

@SimonWoolf can you confirm that the logging will break when using MsgPack encrypted payloads in a browser if we don't merge this in?  If so, I would like to merge and then add a separate issue to review.

browser/lib/transport/xhrtransport.js
paddybyers: ah ok

common/lib/client/auth.js
paddybyers: > I am not sure what object the body will aways be

Either a buffer or string at this point. If it's neither then it's a bug elsewhere.

> What is the downside?

If we know it's not necessary it should be removed.

SimonWoolf:
> can you confirm that the logging will break when using MsgPack encrypted payloads in a browser if we don't merge this in?

Yes, the binary crypto tests will all break.

mattheworiordan:
> Yes, the binary crypto tests will all break.

Ok, I will merge in as is for now and raise a separate issue that @paddybyers can look at it if he wishes.

Issue #42, opened 2015-04-30
Use the correct internet-up CDN


mattheworiordan:
Should be https://internet-up.ably-realtime.com/is-the-internet-up.txt

mattheworiordan:
Matt to correct

mattheworiordan:
https://github.com/ably/ably-js/issues/4

PR #140, opened 2015-09-30
Use the correct internet up URL

mattheworiordan:
@paddybyers and @SimonWoolf, this fixes https://github.com/ably/ably-js/issues/42

### DIFF
--- a/browser/lib/transport/xhrtransport.js
+++ b/browser/lib/transport/xhrtransport.js
@@ -9,7 +9,7 @@ var XHRTransport = (function() {
 	XHRTransport.isAvailable = XHRRequest.isAvailable;
 
 	XHRTransport.checkConnectivity = function(callback) {
-		var upUrl = 'http://internet-up.ably.io.s3-website-us-east-1.amazonaws.com/is-the-internet-up.txt';
+		var upUrl = Defaults.internetUpUrlWithoutExtension + '.txt';
 		Logger.logAction(Logger.LOG_MICRO, 'XHRTransport.checkConnectivity()', 'Sending; ' + upUrl);
 		Http.Request(upUrl, null, null, null, function(err, responseText) {
 			var result = (!err && responseText == 'yes');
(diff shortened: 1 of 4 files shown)


browser/lib/transport/xhrtransport.js
```diff
@@ -9,7 +9,7 @@ var XHRTransport = (function() {
 	XHRTransport.isAvailable = XHRRequest.isAvailable;
 
 	XHRTransport.checkConnectivity = function(callback) {
-		Http.Request('http://internet-up.ably.io.s3-website-us-east-1.amazonaws.com/is-the-internet-up.txt', null, null, null, function(err, responseText) {
+		Http.Request(Defaults.internetUpUrlWithoutExtension + '.txt', null, null, null, function(err, responseText) {
```
mattheworiordan: @paddybyers can you confirm if this check is only made when in a browser?  If not, I will need to append the current protocol to this request when in a browser environment I assume.  Also, is internet connectivity only checked when falling back to XHR & JSONP? I assume that is intentional.

paddybyers: > can you confirm if this check is only made when in a browser?

Also in the node comet transport.

> If not, I will need to append the current protocol to this request when in a browser environment I assume.

I don't know what this means.

> Also, is internet connectivity only checked when falling back to XHR & JSONP? I assume that is intentional.

Since we don't fall back but we fall forward, upgrade to web socket never occurs until one of the HTTP transports connects. So we only use the HTTP transports to check connectivity for the purposes of deciding whether or not to use a fallback host.

mattheworiordan: >> If not, I will need to append the current protocol to this request when in a browser environment I assume.

> I don't know what this means.

If you look in the change below in defaults.js, I no longer include `http` in the URL.  This is intentional so that the internet up check will succeed if the current connection is over HTTPS when in a browser.  However, this will therefore now fail under Node.

mattheworiordan: Ah, I see there are 2 different `Defaults.js` files anyway, so it's a very easy fix

paddybyers: > this will therefore now fail under Node.

Are you going to address that?

mattheworiordan: See https://github.com/ably/ably-js/commit/b102c8747e2dfe972f82fef7de04100c6c4ba68e

PR #141, opened 2015-09-30
NodeCometTransport: clear request timeout on request error

SimonWoolf:
Fixes https://github.com/ably/ably-js/commit/d5fc19a327bde198126238bbb824a9b019cae99a.

(The connectionmanager doesn't know about the transport till it calls the preconnect callback, which it never does if the initial connect fails - which is fine, it's just the transport wasn't clearing the timeout when it got a 404)

@paddybyers not sure if the transport should also dispose() itself if the connect request errors (ie put a `self.dispose()` in [here](https://github.com/ably/ably-js/blob/master/common/lib/transport/comettransport.js#L72-L75)), wdyt?

### DIFF
--- a/common/lib/transport/comettransport.js
+++ b/common/lib/transport/comettransport.js
@@ -71,6 +71,9 @@ var CometTransport = (function() {
 				self.recvRequest = null;
 				if(err) {
 					self.emit('error', err);
+					/* If connect errors before the preconnect, connectionManager is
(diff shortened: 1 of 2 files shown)


paddybyers:
> put a self.dispose() in here), wdyt?

yes.

otherwise lgtm

mattheworiordan:
Nice, good spot @SimonWoolf

Issue #142, opened 2015-09-30
MsgPack cipher workaround


mattheworiordan:
See https://github.com/ably/ably-js/pull/139#commitcomment-13520216 for a complete explanation.  Raising this issue so that we keep track of this if we wish to resolve in future.

SimonWoolf:
I've have thought this was pretty high-priority, surely msgpack + encrypted channel isn't *that* rare a combination?

PR #143, opened 2015-09-30
Convert WordArrays to ArrayBuffers when msgpack-encoding

SimonWoolf:
Fixes https://github.com/ably/ably-js/issues/142, and makes msgpack more efficient when used with encryption, as encrypted binary data is encoded as binary data rather than as arrays of ints.

The actual bit-packing code [already existed](  https://code.google.com/p/crypto-js/source/browse/branches/3.x/src/lib-typedarrays.js?r=661) in the 3.x development branch of the google code project, but the version on github (which is what's packaged on npm) was forked from the 3.1.2 release, which doesn't have it. For simplicity I just copied & pasted the relevant code into bufferutils.

### DIFF
--- a/browser/lib/util/bufferutils.js
+++ b/browser/lib/util/bufferutils.js
@@ -81,6 +81,29 @@ var BufferUtils = (function() {
 
 	BufferUtils.isBuffer = function(buf) { return isArrayBuffer(buf) || isWordArray(buf); };
 
+	BufferUtils.toArrayBuffer = function(buf) {
+		if(!ArrayBuffer)
+			throw new Error("Can't convert to ArrayBuffer: ArrayBuffer not supported");
+
+		if(isArrayBuffer(buf))
+			return buf;
+
+		if(isWordArray(buf)) {
+			/* Backported from unreleased CryptoJS
+			* https://code.google.com/p/crypto-js/source/browse/branches/3.x/src/lib-typedarrays.js?r=661 */
+			var arrayBuffer = new ArrayBuffer(buf.sigBytes);
+			var uint8View = new Uint8Array(arrayBuffer);
+
+			for (var i = 0; i < buf.sigBytes; i++) {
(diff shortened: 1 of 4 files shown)


paddybyers:
> makes msgpack more efficient

err .. makes it interoperable, because sending a map would never work except for other browser clients.

browser/lib/util/bufferutils.js
```diff
@@ -81,6 +76,29 @@ var BufferUtils = (function() {
...
+			return buf;
+
+		if(isWordArray(buf)) {
+			/* Backported from unreleased CryptoJS
+			* https://code.google.com/p/crypto-js/source/browse/branches/3.x/src/lib-typedarrays.js?r=661 */
+			var arrayBuffer = new ArrayBuffer(buf.sigBytes);
+			var uint8View = new Uint8Array(arrayBuffer);
+
+			for (var i = 0; i < buf.sigBytes; i++) {
+				uint8View[i] = (buf.words[i >>> 2] >>> (24 - (i % 4) * 8)) & 0xff;
```
paddybyers: This still isn't great efficiency-wise, because we should be writing words into a `unit32View`, but we can leave this and file a lower-priority issue to improve it unless you want to do it now anyway.

nodejs/lib/util/bufferutils.js
```diff
@@ -7,6 +7,8 @@ this.BufferUtils = (function() {
 
 	BufferUtils.isBuffer = Buffer.isBuffer;
 
+	BufferUtils.toArrayBuffer = function(buf) { return buf; };
```
paddybyers: Not sure I like the name of this, but it's not better than `mightBeWordArrayOrBufferToBuffer`.

browser/lib/util/bufferutils.js
mattheworiordan: It is definitely not worth optimising now, let's please land this as it works and performance of this obscure requirement for now is not a high priority

mattheworiordan:
Nice work

Issue #133, opened 2015-09-24
connection#ping throws exception if you don't pass it a callback


SimonWoolf:
`Ably: EventEmitter.emit(): Unexpected listener exception: TypeError: callback is not a function`

PR #146, opened 2015-10-08
Connection#ping: allow no callback, return a responseTime, add tests

SimonWoolf:
Fixes https://github.com/ably/ably-js/issues/133


common/lib/client/connection.js
```diff
@@ -35,6 +35,8 @@ var Connection = (function() {
 
 	Connection.prototype.ping = function(callback) {
 		Logger.logAction(Logger.LOG_MINOR, 'Connection.ping()', '');
+		if(typeof(callback) !== 'function')
```
paddybyers: why not just `callback = callback || function() {}` ?

paddybyers:
One comment otherwise LGTM

Issue #55, opened 2015-06-16
IFrameTransport errors when connecting with tls: false


SimonWoolf:
Context:
```
ably.js:4067 Ably: IframeTransport.connect(): starting
ably.js:4067 Ably: Auth.getToken(): using cached token; expires = 1434461878631
ably.js:4067 Ably: IframeTransport.createIframe(): destUri: https://sandbox-realtime.ably.io:443/static/iframe-0.8.0.html?origin=file%3…YQ3H5DJHvk3InWqSy63gj0gtcLqRsevTjT7644LeOQK7vRw7Jm6iujC153Y7d8ObMRFbTWew1k
ably.js:4067 Ably: IframeTransport.tryConnect(): viable transport IframeTransport; uri=undefined
ably.js:4067 Ably: ConnectionManager.chooseTransport(): transport iframe connecting
ably.js:4067 Ably: ConnectionManager.setTransportPending(): transport = IframeTransport; uri=undefined
ably.js:4067 Ably: ConnectionManager.setTransportPending: on state = disconnected
ably.js:4067 Ably: ConnectionManager.setTransportPending: reason =  Connection to server temporarily unavailable
ably.js:4067 Ably: ConnectionManager.deactivateTransport(): transport = IframeTransport; uri=undefined
ably.js:4067 Ably: IframeTransport.dispose(): 
ably.js:4067 Ably: ConnectionManager.chooseTransport(): Establishing http transport: IframeTransport; uri=undefined
```
Error:
`Failed to execute 'postMessage' on 'DOMWindow': The target origin provided ('file://') does not match the recipient window's origin ('null').`

Stack:
```
IframeAgent.postMessageEvent @ iframe-0.8.0.html?origin=file%3A%2F%2F&access_token=rkPwKA.CBL7U5DiBnAAO66Z0qtvkbiMxoL08awwA1YQ3H5D…:1294
IframeAgent.onData @ iframe-0.8.0.html?origin=file%3A%2F%2F&access_token=rkPwKA.CBL7U5DiBnAAO66Z0qtvkbiMxoL08awwA1YQ3H5D…:1290
(anonymous function) @ iframe-0.8.0.html?origin=file%3A%2F%2F&access_token=rkPwKA.CBL7U5DiBnAAO66Z0qtvkbiMxoL08awwA1YQ3H5D…:1147
[...]

PR #148, opened 2015-10-08
Allow iframe transport to work with local (file://) pages

SimonWoolf:
Fixes https://github.com/ably/ably-js/issues/55. (Turns out was only broken because I was testing from a local file).

But -- **possibly introduces security hole**? In theory should only happen with local files, which means only for people testing / playing about with ably. But I don't know enough about what's going on to know if this could be an security hole when not a local file, eg through some malicious script manipulating `window.location`.

There doesn't seem to be any way around it at the moment (MDN says "posting a message to a page at a file: URL currently requires that the targetOrigin argument be `*`. file:// cannot be used as a security restriction; this restriction may be modified in the future"), so if it is a security hole, then I'll do an another PR that just makes `IframeTransport.isAvailable` return false when using local files.

### DIFF
--- a/browser/lib/transport/iframetransport.js
+++ b/browser/lib/transport/iframetransport.js
@@ -1,6 +1,11 @@
 var IframeTransport = (function() {
 	var origin = location.origin || location.protocol + "//" + location.hostname + (location.port ? ':' + location.port: '');
 
+	/* Allows iframe to work from local (file://) files - firefox sets to origin
+	* to 'null', chrome to 'file://', neither of which are valid security
+	* restrictions. */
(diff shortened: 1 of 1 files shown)


paddybyers:
> I don't know enough about what's going on to know if this could be an security hole when not a local file, eg through some malicious script manipulating window.location.

If there is a malicious script running in the scope of the document that loaded the Ably library, then that script has the ability to do anything that the library can do, and more. I don't think this change enables the malicious script to do anything it couldn't do otherwise, since you've not modified anything running in the Iframe's origin.

paddybyers:
Thanks

PR #151, opened 2015-10-15
ably-js-browsers-all

SimonWoolf:
https://ci.ably.io/job/ably-js-browsers-all/100/ passing ably-js-browsers-all test run! \o/

Fixes an xhr issue on IE10 by using an envelope, to get around [this bug](http://stackoverflow.com/a/16320339).

I only did it for true XHR (which IE10 can use), not XDR, because I don't understand the current xdr code -- it [seems to be assuming](https://github.com/ably/ably-js/blob/master/browser/lib/transport/xhrrequest.js#L292-L295) that all responses from realtime are chunked, even when streaming is off, which they aren't -- I didn't want to assume it was completely broken and rewrite from scratch if I might be missing something. (XDR's only used by IE9 and below anyway, which I explicitly wasn't targeting with this piece of work (it's not in ably-js-browsers-all), so can be sorted later when we do IE9).

This makes IE10 pass. But -- not on saucelabs. I've completely failed at getting the IE10 saucelabs VM to work properly. (It just seems to lose all external connectivity (xhr, jsonp, whatever) about 30s after the test starts, after which every test fails). Having banged by head against that for a few days, I gave up and added a VM karma test. Run with `grunt test:karma --browsers "IE10\ -\ Win7"` after installing the IE10 vm with `curl -s https://raw.githubusercontent.com/xdissent/ievms/master/ievms.sh | env IEVMS_VERSIONS="10" bash`. Will talk to Lewis about getting it working in ci. For now can be run manually.

The way envelopes are used at the moment is a bit messy: they're sometimes added at the restclient level, and sometimes at the transport level (eg jsonp, and now xhr on IE10), but the restclient assumes that if it added one it can remove it (which it can't if the transport already has). 6ab1c31 is a quick fix to stop the resource trying to unwrap an already-unwrapped envelope, but still, a bit inelegant.

Also a few more test fixes and an update to browser selection used in browsers-all.

9624f6f is a temp fix to stop everything failing due to CORS issues, just until https://github.com/ably/realtime/pull/337 is deployed to sandbox. Remove before merging if that's deployed.

### DIFF
--- a/browser/lib/transport/xhrrequest.js
+++ b/browser/lib/transport/xhrrequest.js
@@ -32,6 +32,10 @@ var XHRRequest = (function() {
 		return xhr.getResponseHeader && xhr.getResponseHeader('content-type');
 	}
 
+	function isEncodingChunked(xhr) {
+		return xhr.getResponseHeader && xhr.getResponseHeader('transfer-encoding') === 'chunked';
+	}
+
 	function XHRRequest(uri, headers, params, body, requestMode) {
 		EventEmitter.call(this);
 		params = params || {};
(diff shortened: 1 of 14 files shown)


paddybyers:
Thanks

PR #152, opened 2015-10-19
Presence map: normalise enter/update actions to present in #put

SimonWoolf:
ITSM should be the presence map's responsibility to normalise its actions - neater than juggling different broadcast and map-member presence messages in #setPresence

Fixes presence tests broken by 9c4f8d7

### DIFF
--- a/common/lib/client/realtimepresence.js
+++ b/common/lib/client/realtimepresence.js
@@ -175,7 +175,6 @@ var RealtimePresence = (function() {
 					break;
 				case presenceAction.UPDATE:
 				case presenceAction.ENTER:
-					presence.action = presenceAction.PRESENT;
 				case presenceAction.PRESENT:
 					if(members.put(presence)) {
 						broadcastMessages.push(presence);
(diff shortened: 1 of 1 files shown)


common/lib/client/realtimepresence.js
```diff
@@ -259,6 +258,10 @@ var RealtimePresence = (function() {
 	};
 
 	PresenceMap.prototype.put = function(item) {
+		if(Utils.arrIn([presenceAction.ENTER, presenceAction.UPDATE], item.action)) {
```
paddybyers: This might look elegant but it's hopelessly inefficient compared to just doing (x==a || x==b).

common/lib/client/realtimepresence.js
```diff
@@ -259,6 +258,10 @@ var RealtimePresence = (function() {
...
+			item = Utils.copy(item);
```
paddybyers: We want a clone of the object with the correct prototype, so I think this should be `PresenceMessage.fromValues()`.

paddybyers:
Thanks

Issue #161, opened 2015-11-13
Presence callback error following an immediate disconnect


mattheworiordan:
Not sure how easy this is to replicate, but when running the demo chat app, if the connection fails immediately (due to known realtime issue 370), strangely the `RealtimePresence` class throws an odd error about a callback not exist.  The callback is not user provided, so this is an internal error.

This is happening when running the demo at https://github.com/ably/demo-mobile-phonegap-cordova, the initialisation of the lib is https://github.com/ably/demo-mobile-phonegap-cordova/blob/master/www/js/app.js#L112-L116 (with log level 5), and the channel attachment and presence is pretty standard https://github.com/ably/demo-mobile-phonegap-cordova/blob/master/www/js/app.js#L150-L172

Complete log is:

```
ably-0.8.7.js:4103Ably: Rest(): started
2015-11-13 10:46:46.131 ably-0.8.7.js:4103Ably: Auth(): using token auth with authUrl
2015-11-13 10:46:46.132 ably-0.8.7.js:4103Ably: Realtime.ConnectionManager(): started
2015-11-13 10:46:46.132 ably-0.8.7.js:4103Ably: Realtime.ConnectionManager(): requested transports = [web_socket,xhr,iframe,jsonp]
2015-11-13 10:46:46.132 ably-0.8.7.js:4103Ably: Realtime.ConnectionManager(): available http transports = [xhr,iframe,jsonp]
2015-11-13 10:46:46.132 ably-0.8.7.js:4103Ably: Realtime.ConnectionManager(): available transports = [web_socket,xhr,iframe,jsonp]
2015-11-13 10:46:46.133 ably-0.8.7.js:4103Ably: Realtime.ConnectionManager(): http hosts = [rest.ably.io,A.ably-realtime.com,B.ably-realtime.com,C.ably-realtime.com,D.ably-realtime.com,E.ably-realtime.com]
2015-11-13 10:46:46.133 ably-0.8.7.js:4103Ably: Connection.connect(): 
2015-11-13 10:46:46.133 ably-0.8.7.js:4103 Ably: ConnectionManager.requestState(): requested state: connecting
2015-11-13 10:46:46.133 ably-0.8.7.js:4103 Ably: ConnectionManager.cancelTransitionTimer(): 
2015-11-13 10:46:46.134 ably-0.8.7.js:4103 Ably: ConnectionManager.enactStateChange: setting new state: connecting; reason = undefined
[...]

PR #162, opened 2015-11-13
Don’t assume a callback exists if error occurs during presence enter/update

SimonWoolf:
Quick fix to https://github.com/ably/ably-js/issues/161


paddybyers:
Thanks

Issue #166, opened 2015-11-19
HTTP requests ignore `tls: false` attribute


mattheworiordan:
Initialising a Realtime library with `tls: false` connects to a realtime end point correctly not using TLS, however when performing the following I found the client library attempted to connect over TLS:

* Implicit token request for auth
* History requests on a channel

We need to ensure that `tls: false` is respected for REST operations performed by a Realtime client and add suitable test coverage.

@SimonWoolf mind taking a look?

SimonWoolf:
TBH I'd always assumed that was deliberate - `rest.baseUri` has always ignored the tls option. I've done a PR to change it https://github.com/ably/ably-js/pull/167, but will wait for @paddybyers to weigh in on whether there's a reason it was done that way before merging

PR #167, opened 2015-11-19
REST client: respect tls clientOption

SimonWoolf:
Fixes https://github.com/ably/ably-js/issues/166. See discussion there

### DIFF
--- a/common/lib/util/defaults.js
+++ b/common/lib/util/defaults.js
@@ -33,6 +33,10 @@ Defaults.getPort = function(options, tls) {
 	return (tls || options.tls) ? options.tlsPort : options.port;
 };
 
+Defaults.getHttpScheme = function(options) {
+	return options.tls ? 'https://' : 'http://';
+};
+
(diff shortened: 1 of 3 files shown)


mattheworiordan:
LGTM.

common/lib/util/defaults.js
```diff
@@ -33,6 +33,10 @@ Defaults.getPort = function(options, tls) {
 	return (tls || options.tls) ? options.tlsPort : options.port;
 };
 
+Defaults.getHttpScheme = function(options, tls) {
```
paddybyers: The second argument is never used. Did you have a plan for that?

SimonWoolf: planned to use it for stuff like the iframetransport which needs to force tls, but didn't bother in the end (simpler to leave it as hardcoded "https://") and forgot to remove the param. oops

paddybyers:
Thanks

PR #168, opened 2015-11-20
Fix exception when doing jsonp rest requests

SimonWoolf:


browser/lib/transport/jsonptransport.js
```diff
@@ -52,7 +52,9 @@ var JSONPTransport = (function() {
 	};
 
 	var createRequest = JSONPTransport.prototype.createRequest = function(uri, headers, params, body, requestMode) {
-		return new Request(undefined, uri, headers, params, body, requestMode, this.timeouts);
+		/* JSONP requests are used outside the context of a realtime transport, in which case use the default timeouts */
+		var timeouts = (this && this.timeouts) ? this.timeouts : Defaults.TIMEOUTS;
```
paddybyers: (this && this.timeouts) || Defaults.TIMEOUTS;

SimonWoolf: good point

SimonWoolf:
Merged as 150fda9

PR #176, opened 2015-12-09
EventEmitter fixes

mattheworiordan:
@SimonWoolf and @paddybyers please see my PR which fixes an inconsistency we had with the EventEmitter.  Firstly, it could fire twice when adding an `on` event handler with the Connection object, but this would not fire for `once` or for the Channel object.  I don't see the logic in this exception.

Also, I discovered that the EventEmitter did not correctly handle exceptions by continuing with other handlers when there was an exception, and it would also cause an exception to be thrown in the caller of the `emit`

I added a convenience (undocumented) method `onceOrIf` that is the equivalent of https://github.com/ably/ably-ruby/blob/master/lib/ably/modules/state_emitter.rb#L74

### DIFF
--- a/common/lib/util/eventemitter.js
+++ b/common/lib/util/eventemitter.js
@@ -8,6 +8,13 @@ var EventEmitter = (function() {
 		this.eventsOnce = {};
 	}
 
+	/* Call the listener, catch any exceptions and log, but continue operation*/
+	function callListener(eventThis, listener, args) {
+		try { listener.apply(eventThis, args); } catch(e) {
+			Logger.logAction(Logger.LOG_ERROR, 'EventEmitter.emit()', 'Unexpected listener exception: ' + e + '; stack = ' + e.stack);
+		}
+	}
+
 	/**
 	 * Add an event listener
 	 * @param event (optional) the name of the event to listen to
@@ -104,18 +111,11 @@ var EventEmitter = (function() {
 		var args = Array.prototype.slice.call(arguments, 1);
 		var eventThis = {event:event};
 
-		/* wrap the try/catch in a function improves performance by 30% */
-		function callListener(listener) {
-			try { listener.apply(eventThis, args); } catch(e) {
-				Logger.logAction(Logger.LOG_ERROR, 'EventEmitter.emit()', 'Unexpected listener exception: ' + e + '; stack = ' + e.stack);
-				throw e;
-			}
-		}
 		if(this.anyOnce.length) {
 			var listeners = this.anyOnce;
 			this.anyOnce = [];
 			for(var i = 0; i < listeners.length; i++)
-				callListener((listeners[i]));
+				callListener(eventThis, listeners[i], args);
 		}
 		for(var i = 0; i < this.any.length; i++)
 			this.any[i].apply(eventThis, args);
@@ -123,12 +123,12 @@ var EventEmitter = (function() {
 		if(listeners) {
 			delete this.eventsOnce[event];
 			for(var i = 0; i < listeners.length; i++)
-				callListener((listeners[i]));
+				callListener(eventThis, listeners[i], args);
 		}
 		var listeners = this.events[event];
 		if(listeners)
 			for(var i = 0; i < listeners.length; i++)
-				callListener((listeners[i]));
+				callListener(eventThis, listeners[i], args);
 	};
 
 	/**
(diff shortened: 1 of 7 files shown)


paddybyers:
> I added a convenience (undocumented) method onceOrIf that is the equivalent of

Actually I think this functionality is more likely to be the thing you want to call than `once()` in most situations. So if we're going to add it we should try to think of a less awkward name and make it part of the public API.

If this method only dealt with the connected state then it would be natural to call it `onceConnected()`. Dealing with all possible states, perhaps `onceAt('connected', ...)`, `onceState('connected', ...)` ?

mattheworiordan:
> Actually I think this functionality is more likely to be the thing you want to call than once() in most situations.

Perhaps, but I am not sure we should change the spec for this method, and also I am not sure we really know what "people expect" either.

> So if we're going to add it we should try to think of a less awkward name and make it part of the public API.

Fine, but that needs to be moved into the 0.9 spec.  This is not needed, and is only a convenience.

> If this method only dealt with the connected state then it would be natural to call it onceConnected(). Dealing with all possible states, perhaps onceAt('connected', ...), onceState('connected', ...) ?

I fear that that naming is confusing because `once('connected')` would then mean once the state changes to `connected`, yet for example `onceConnected` would mean if the state is or changes to `connected`.  I am not convinced that any of the permutations with `once` work, and as such would prefer to not include in the spec.

As this is an undocumented feature and quite likely to not use the same naming in future anyway, can we add this now and add a new issue to address this problem in the 0.9 spec?  As a side note, I consciously  differentiated between an `EventEmitter` and `StateEmitter` because of these issues, and thus introduced methods in the `StateEmitter` mixin.

common/lib/util/eventemitter.js
```diff
@@ -147,5 +147,30 @@ var EventEmitter = (function() {
...
+			throw("onceOrIf requires a valid event String argument");
+		} else if (typeof(listener) !== 'function') {
+			throw("onceOrIf requires a valid listener argument");
+		} else {
+			if(targetState === currentState) {
+				callListener(eventThis, listener, listenerArgs);
+			} else {
+				this.once(targetState, listener);
+			}
+		}
```
SimonWoolf: Might be more readable as `if {} else if {} else if {} else {}`, which is equivalent and avoids nesting `if`s

mattheworiordan: Sure, sensible, I'll amend. Not quite sure how I ended up with that mess ;)

SimonWoolf:
Agree with Paddy that onceOrIf is likely to be what people will generally want (with connection and channel state, anyway) -- looking at its use in the tests it usually is. Re naming, perhaps `when`, `whenIn`, or `whenInState`?

mattheworiordan:
Agreed but I am reluctant now to change anything in the spec unless it's
wrong. We're struggling to complete the libraries to the spec as it is, so
I'd really prefer we release undocumented features now, and formalise it in
0.9. I don't care what the method name is now or in 0.9, but I do care if
we start changing the 0.8 spec.

mattheworiordan:
Sorry last comment. The problem is that we use on as an interface of the EventEmitter, which makes sense when emitting events. Once was introduced as a slight variation that simply does the same thing but only fires once, and in that regard the naming feels right. When or whenInState really does not clearly convey whether the callback should be called once or many times, and as such, just adds to the confusion. Making once fire once the event is emitted is fine but an EventEmitter does not have state, so in the case of an object that has mixed in an EventEmitter, using the same method but with different behaviour feels very wrong. We need a new name that clearly conveys the fact that it is run only once, and is called immediately if the state is met. I don't have the answer, but I don't think we need the answer either, let's just defer to 0.9

mattheworiordan:
Excuse the multitude of typos, on phone and can't be arsed correcting posts in tiny text boxes ;)

SimonWoolf:
Fair enough. OK, :+1: to this for now then.

mattheworiordan:
> Fair enough. OK, :+1: to this for now then.

@paddybyers you OK with this too?

common/lib/util/eventemitter.js
paddybyers: You don't even need the first two `else`s because the first alternative never continues.

````
if() {
  throw
}

if() {
  throw
}

if() {
} else {
}
````

paddybyers:
> @paddybyers you OK with this too?

Well I think `onceOrIf` is an abomination but other than that I'm fine.

common/lib/util/eventemitter.js
mattheworiordan: Ok, fixed with https://github.com/ably/ably-js/commit/d7844232a46e489ccaad1f90029df853260cd7ea

mattheworiordan:
I could change it to `whenState` if you like?  I don't really care that much, so long as we don't document it.

mattheworiordan:
@SimonWoolf were you going to merge this in with your other changes?

SimonWoolf:
@mattheworiordan now other changes are in, I need to rebase this on master (& make paddy's requested changes), then will merge.

mattheworiordan:
Thanks, sounds good if you're happy to do it.

SimonWoolf:
Merged into master as https://github.com/ably/ably-js/compare/150fda9...193a25f

Went with whenState per Paddy:
paddy

Issue #183, opened 2015-12-17
No internet up test coverage


mattheworiordan:
See https://github.com/ably/infrastructure/issues/476 and the error in the browser:
```
?logLevel=5:1 XMLHttpRequest cannot load https://rest.ably.io/comet/connect?resume=2a523!Onbyx4yRXVLI1LQn-a772a523&c…tXjwqu1tkPPqxbLS2wanjdXJRhUufkwOOSGvrL0azuR8IlldFyA7Fqqh716y0r8YbyDCG6x8Dw. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://ably-phonegap-cordova-demo.herokuapp.com' is therefore not allowed access. The response had HTTP status code 504.
```

We didn't realise this would happen because we don't run tests in the browser to check the internet up test works.  We need to add test coverage for this ASAP.  See https://github.com/ably/ably-ruby/blob/master/spec/acceptance/realtime/connection_spec.rb#L1052-L1140 for the equivalent test coverage in Ruby

@SimonWoolf mind looking at this?

PR #184, opened 2015-12-17
Connectivity check fixes & tests

SimonWoolf:
Closes https://github.com/ably/ably-js/issues/183 ; currently fails due to https://github.com/ably/infrastructure/issues/476

### DIFF
--- a/browser/lib/transport/jsonptransport.js
+++ b/browser/lib/transport/jsonptransport.js
@@ -27,12 +27,17 @@ var JSONPTransport = (function() {
 		}
 		checksInProgress = [callback];
 		Logger.logAction(Logger.LOG_MICRO, 'JSONPTransport.checkConnectivity()', 'Sending; ' + upUrl);
-		request(upUrl, null, null, null, false, function(err, response) {
+
+		var req = new Request('_isTheInternetUp', upUrl, null, null, null, CometTransport.REQ_SEND, Defaults.TIMEOUTS);
+		req.once('complete', function(err, response) {
 			var result = !err && response;
 			Logger.logAction(Logger.LOG_MICRO, 'JSONPTransport.checkConnectivity()', 'Result: ' + result);
 			for(var i = 0; i < checksInProgress.length; i++) checksInProgress[i](null, result);
 			checksInProgress = null;
 		});
+		Utils.nextTick(function() {
+			req.exec();
+		})
 	};
 
 	JSONPTransport.tryConnect = function(connectionManager, auth, params, callback) {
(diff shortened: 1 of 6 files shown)


mattheworiordan:
Looks great @SimonWoolf :+1:

common/lib/transport/connectionmanager.js
```diff
@@ -66,6 +68,8 @@ var ConnectionManager = (function() {
 			params.echo = 'false';
 		if(this.format !== undefined)
 			params.format = this.format;
+		if(this.stream !== undefined)
```
paddybyers: All of this special-case crap for one of the transport params was what I was trying to get away from with the previous commit. The idea is that the stream param is added to params along with all other given transport params. The only special-casing needed is to take notice of whether or not it's a stream when creating the request.

browser/lib/transport/jsonptransport.js
```diff
@@ -112,6 +113,10 @@ var JSONPTransport = (function() {
 			self.complete(err);
 		};
 
+		_["_isTheInternetUp"] = function() {
```
paddybyers: Please can we have single quotes wherever possible.

common/lib/transport/connectionmanager.js
SimonWoolf: I'm not quite clear what you want. It needs to be set as a connect params somewhere. If not in `getConnectParams`, we could do it in `CometTransport#connect` after the `getConnectParams` call [here](https://github.com/ably/ably-js/blob/master/common/lib/transport/comettransport.js#L47) if you prefer.

Or do you mean you want it added directly into the jsonp Request object, ie [here](https://github.com/ably/ably-js/blob/master/browser/lib/transport/jsonptransport.js#L65)? If so, not sure I agree - that will let us remove it from getConnectParams, but it means specifying it in two places (it still needs to be an attribute on the Transport as well as that's what the requestMode looks at), which isn't very DRY. Or am I misunderstanding?

SimonWoolf:
Uploaded a new internet-up.js file with `Ably._._isTheInternetUp({"response":true,"statusCode": 200});`; it'll presumably change whenever S3's caching gets around to it. Changed the jsonp connectivity check accordingly.

SimonWoolf:
(Turns out Cloudflare was caching it more or less permanently despite setting a 4 hour cache-control header, needed to log on to cloudflare and explicitly purge the cache for that individual file...)

common/lib/transport/connectionmanager.js
paddybyers: > It needs to be set as a connect params somewhere.

I was intending that it gets into `params` only because it's in `transportParams` and then, if it is still unspecified, the default for that transport is applied. But that disperses the logic into different places so in retrospect I'm fine with this.

paddybyers:
@SimonWoolf so if you're happy this can merge and be released.

PR #189, opened 2015-12-18
Fix iframe connections not starting a recv after the connect req closes

SimonWoolf:
Exactly the same as https://github.com/ably/ably-js/pull/181, except for iframes (I didn't notice that iframetransport doesn't use the shared comettransport#connect)


paddybyers:
LGTM, merge if you're satisfied.

SimonWoolf:
Merged in b013bcc

PR #193, opened 2016-01-08
Fix unsubscribe for all events & listeners

mattheworiordan:
@paddybyers @SimonWoolf can you review please?  This is a live issue for a client so I'd like to release a fix ASAP

### DIFF
--- a/common/lib/util/eventemitter.js
+++ b/common/lib/util/eventemitter.js
@@ -24,7 +24,7 @@ var EventEmitter = (function() {
 	EventEmitter.prototype.on = function(event, listener) {
 		if(arguments.length == 1 && typeof(event) == 'function') {
 			this.any.push(event);
-		} else if(event === null) {
+		} else if(Utils.isEmptyArg(event)) {
 			this.any.push(listener);
 		} else {
 			var listeners = (this.events[event] || (this.events[event] = []));
@@ -41,7 +41,7 @@ var EventEmitter = (function() {
 	 *        supplied, all listeners are removed.
 	 */
 	EventEmitter.prototype.off = function(event, listener) {
-		if(arguments.length == 0 || (!event && !listener)) {
+		if(arguments.length == 0 || (Utils.isEmptyArg(event) && Utils.isEmptyArg(listener))) {
 			this.any = [];
 			this.events = {};
 			this.anyOnce = [];
@@ -57,7 +57,7 @@ var EventEmitter = (function() {
 			/* ... or we take event to be the actual event name and listener to be all */
 		}
 		var listeners, idx = -1;
-		if((event === null) || (event === undefined)) {
+		if(Utils.isEmptyArg(event)) {
 			/* "any" case */
 			if(listener) {
 				if(!(listeners = this.any) || (idx = Utils.arrIndexOf(listeners, listener)) == -1) {
@@ -72,7 +72,7 @@ var EventEmitter = (function() {
 			}
 			return;
 		}
-		/* "normal* case where event is an actual event */
+		/* "normal" case where event is an actual event */
 		if(listener) {
 			var listeners, idx = -1;
 			if(!(listeners = this.events[event]) || (idx = Utils.arrIndexOf(listeners, listener)) == -1) {
@@ -139,7 +139,7 @@ var EventEmitter = (function() {
 	EventEmitter.prototype.once = function(event, listener) {
 		if(arguments.length == 1 && typeof(event) == 'function') {
 			this.anyOnce.push(event);
-		} else if(event === null) {
+		} else if(Utils.isEmptyArg(event)) {
 			this.anyOnce.push(listener);
 		} else {
 			var listeners = (this.eventsOnce[event] || (this.eventsOnce[event] = []));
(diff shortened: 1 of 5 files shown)


common/lib/util/eventemitter.js
```diff
@@ -41,7 +41,7 @@ var EventEmitter = (function() {
 	 *        supplied, all listeners are removed.
 	 */
 	EventEmitter.prototype.off = function(event, listener) {
-		if(arguments.length == 0) {
+		if(arguments.length == 0 || (!event && !listener)) {
```
SimonWoolf: I think this change means https://github.com/ably/ably-js/blob/e4c8cf7/common/lib/util/eventemitter.js#L69-L72 should now be unreachable, so can be removed

common/lib/util/eventemitter.js
```diff
@@ -57,7 +57,7 @@ var EventEmitter = (function() {
 			/* ... or we take event to be the actual event name and listener to be all */
 		}
 		var listeners, idx = -1;
-		if(event === null) {
+		if((event === null) || (event === undefined)) {
```
SimonWoolf: If we're accepting `undefined` as well as `null` in `off` should we make the equivalent change in `on` (L27) and `once` (L142)?

SimonWoolf:
:+1:

common/lib/util/eventemitter.js
mattheworiordan: Yes, true.
I am a little uncomfortable with the explicit checks for `null` in some cases, and `undefined` in others.  @paddybyers do you have any reason to not simply evaluate the arguments as truthy like I have done in this line throughout?

common/lib/util/eventemitter.js
mattheworiordan: Yes, probably, @paddybyers can you share your review as I would like to really just use `if (!event)`

paddybyers: Well we also have to worry about `false`, `0`, `''`. Either we have an API that detects the types of all arguments and rejects invalid arguments, or, it behaves on the assumption that it is given a string, or we have something in-between as now.

I think the logic should really be:

- decide which overloaded signature is being used: for this check if `event === undefined`
- assume that the `event` argument, if provided, is a string (or coerce to a string) and then reject any invalid string values.

I don't think `event === null` has a right to special handling any more than `typeof(event) == 'number'`.

mattheworiordan: The reason I implemented this is because `null` was being passed to this method, see https://github.com/ably/ably-js/blob/master/common/lib/client/realtimechannel.js#L226-L227.

If we are going to insist that it's indeed `undefined`, then I am happy to change all methods that call it to assume `undefined` too, and I would also like to put a guard on there to reject in valid values to avoid confusion i.e. if a user assumes calling `unsubscribe(null, listener)` will unsubscribe listeners for all events. 

Are you happy for me to try and track down all uses of `null` then where we expected `undefined` where they use the `EventEmitter`?

paddybyers: OK, I partially take back what I said because now we have a new can of worms because this method, when called from `RealtimeChannel.unsubscribe()` can no longer rely on its own logic to know which overloaded signature applies, because that was done before it was called. So the overloading logic is re-performed here which is a bit crap; ideally we'd have an `_off()` or something which knows it doesn't need to do that. It's not such an issue for `on()`/`off()` though.

That aside, we now have one or more values for `event` which signify that no `event` argument was supplied, and the intended meaning is that the overloaded variant with no `event` argument - ie unsubscribe from all events - was called. The special value could be `null` or `undefined` but while this is only an argument passed to an internal method it is moot..

However, the question this raises is whether or not we intend that this behaviour also forms part of the public API: is it intended that a client can pass `null`, or `undefined`, or any falsy value, and thereby request to unsubscribe from all events?

The standard that defines best practice for these situations is WebIDL: http://www.w3.org/TR/WebIDL/

I think we should in fact expose an API publically which is in WebIDL terminology 

    off([TreatUndefinedAs=Null] DOMString? event, Listener? listener)

where the `event` argument, if `null` (IDL null, not JS `null`) means unsubscribe from all events.

Now WebIDL tells us which JS argument values map to IDL `null`: they are `undefined` and `null`.

The other falsy values map `false` -> `'false'`, `0` -> `'0'`, `''` -> `''`; ie they are naturally coerced to a string if not already a string.

The upshot is that we should be checking for `(x === null || x === undefined)` instead of `!x`, and it's fine (and arguably more WebIDL-esque) for the overload-resolution functionality in the derived class to pass `null`.

mattheworiordan: Ok, I will add a commit that works with `(x === null || x === undefined)`.  Is there a helper method that exists anywhere to test this?  If not, it would be useful to have one in the Ably scope.

mattheworiordan: I've looked and not seen one, I will propose a new method in my revised PR

paddybyers: function isIDLNull(x) { return x === null || x === undefined; }

mattheworiordan: I wasn't asking for the code, I was wondering if a similar method existed elsewhere to reduce code duplication!

mattheworiordan:
@paddybyers @SimonWoolf thoughts on the recent 3 commits?

paddybyers:
LGTM

PR #197, opened 2016-01-13
Tweak xhr error handling

SimonWoolf:
Not sure whether bundling the errorEvent and the current statusText in there will actually be useful, but, shrug

### DIFF
--- a/browser/lib/transport/xhrrequest.js
+++ b/browser/lib/transport/xhrrequest.js
@@ -122,19 +122,19 @@ var XHRRequest = (function() {
 		for(var h in headers)
 			xhr.setRequestHeader(h, headers[h]);
 
-		var onerror = xhr.onerror = function(err) {
-			err.code = 80000;
-			self.complete(err);
+		var errorHandler = function(errorEvent, message, code, statusCode) {
+			var errorMessage = message + ', errorEvent was ' + Utils.inspect(errorEvent) + ', current statusText is ' + self.xhr.statusText;
+			Logger.logAction(Logger.LOG_ERROR, 'Request.on' + errorEvent.type + '()', errorMessage);
+			self.complete(new ErrorInfo(errorMessage, code, statusCode));
 		};
-		xhr.onabort = function() {
-			var err = new Error('Request cancelled');
-			err.statusCode = 400;
-			onerror(err);
+		xhr.onerror = function(errorEvent) {
+			errorHandler(errorEvent, 'XHR error occurred', 80000, 400);
+		}
+		xhr.onabort = function(errorEvent) {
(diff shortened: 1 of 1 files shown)


mattheworiordan:
Looks good, but I think needs @paddybyers's input for this one

browser/lib/transport/xhrrequest.js
```diff
@@ -122,19 +122,19 @@ var XHRRequest = (function() {
 		for(var h in headers)
 			xhr.setRequestHeader(h, headers[h]);
 
-		var onerror = xhr.onerror = function(err) {
-			err.code = 80000;
-			self.complete(err);
+		var errorHandler = function(errorEvent, message, code, statusCode) {
+			var errorMessage = message + ", errorEvent was " + Utils.inspect(errorEvent) + ", current statusText is " + self.xhr.statusText;
```
paddybyers: Single quotes please everywhere.

browser/lib/transport/xhrrequest.js
```diff
@@ -122,19 +122,19 @@ var XHRRequest = (function() {
...
+			Logger.logAction(Logger.LOG_ERROR, 'Request.on' + errorEvent.type + '()', errorMessage);
+			self.complete(new ErrorInfo(errorMessage, code, statusCode));
 		};
-		xhr.onabort = function() {
-			var err = new Error('Request cancelled');
-			err.statusCode = 400;
-			onerror(err);
+		xhr.onerror = function(errorEvent) {
+			errorHandler(errorEvent, "XHR error occurred", 8000, 400);
```
paddybyers: 80000, not 8000 (and below).

browser/lib/transport/xhrrequest.js
SimonWoolf: > Also, have we checked that Utils.inspect() gives us something meaningful for the event?

In my tests (just aborting or cutting the connection) it hasn't contained anything useful, no (the stringify was only `{isTrusted: true}`, and the rest of the event object didn't have anything better). Left it in there in the hope that maybe there's some more interesting error might happen in the wild where (a) the event object included some new fields, and (b) they'd be included in the stringify. Perhaps not very likely, but I figure leaving it there doesn't cost much. Want me to remove it?

paddybyers: > but I figure leaving it there doesn't cost much. Want me to remove it?

Looking at the event specs you're not going to get anything useful from the event object itself except the type of the event (ie error or abort). TBH I' remove the `inspect()` and just have `'type = ' + ev.type` or something like that in the log message.

paddybyers:
Thanks

Issue #203, opened 2016-01-19
Token renewal does not seem to be working


mattheworiordan:
See the log below from a client that I kept open for over 12 hours using the JSBin https://jsbin.ably.io/usebot/3/edit.  It seems that once the token expires and the connection is closed by the server correctly, the client does not attempt to retrieve a new token and reconnect causing the client library to go into a FAILED state.  

Looking in the console I can see that only one request to https://www.ably.io/ably-auth/token/docs?rnd=14745209529064596 is made at `Tue, 19 Jan 2016 02:10:39 GMT` with the response `xVLyHw.AvtQgVwYU2HNSfT9FrmE88vt_6QkKXZ_UkUKG3s7qCJRp4o6_E`.  

I believe there are two things that need to be fixed here:

* If a connection is closed with 40140, then a reauth should happen immediately
* If connecting with a token string the token does not exist i.e. 40400, then we should assume it has probably expired and we should reauth at least once and reattempt to connect

```
ably.js:4166 Ably: Rest(): started
2016-01-19 03:10:38.477 ably.js:4166 Ably: Auth(): using token auth with authUrl
2016-01-19 03:10:38.478 ably.js:4166 Ably: Realtime.ConnectionManager(): started
2016-01-19 03:10:38.478 ably.js:4166 Ably: Realtime.ConnectionManager(): requested transports = [web_socket,xhr,iframe,jsonp]
2016-01-19 03:10:38.479 ably.js:4166 Ably: Realtime.ConnectionManager(): available http transports = [xhr,iframe,jsonp]
2016-01-19 03:10:38.479 ably.js:4166 Ably: Realtime.ConnectionManager(): available transports = [web_socket,xhr,iframe,jsonp]
2016-01-19 03:10:38.479 ably.js:4166 Ably: Realtime.ConnectionManager(): http hosts = [rest.ably.io,A.ably-realtime.com,B.ably-realtime.com,C.ably-realtime.com]
2016-01-19 03:10:38.480 ably.js:4166 Ably: Realtime.connect(): 
2016-01-19 03:10:38.480 ably.js:4166 Ably: Connection.connect(): 
2016-01-19 03:10:38.480 ably.js:4166 Ably: ConnectionManager.requestState(): requested state: connecting
2016-01-19 03:10:38.481 ably.js:4166 Ably: ConnectionManager.cancelTransitionTimer(): 
[...]

PR #204, opened 2016-01-20
Force token rerequest if realtime indicates token problems

SimonWoolf:
Including adding a new ConnectionManager#disconnectReason so can remember whether the reason for the last disconnect was a token issue.

Also add isErrATokenProblem helper in anticipation of upcoming token code changes (uses `40140 <= err.code < 40150` so works with both old and new errors).

Fixes https://github.com/ably/ably-js/issues/203

### DIFF
--- a/common/lib/client/resource.js
+++ b/common/lib/client/resource.js
@@ -1,5 +1,8 @@
 var Resource = (function() {
 	var msgpack = (typeof(window) == 'object') ? window.Ably.msgpack : require('msgpack-js');
+	var isErrATokenProblem = function(error) {
+		return error.code && (error.code >= 40140) && (error.code < 40150);
+	};
 
 	function Resource() {}
 
@@ -104,7 +107,7 @@ var Resource = (function() {
 			}
 
 			Http.get(rest, path, headers, params, function(err, res, headers, unpacked) {
-				if(err && err.code == 40140) {
+				if(err && isErrATokenProblem(err)) {
 					/* token has expired, so get a new one */
 					rest.auth.authorise(null, {force:true}, function(err) {
 						if(err) {
@@ -146,7 +149,7 @@ var Resource = (function() {
 			}
 
 			Http.post(rest, path, headers, body, params, function(err, res, headers, unpacked) {
-				if(err && err.code == 40140) {
+				if(err && isErrATokenProblem(err)) {
 					/* token has expired, so get a new one */
 					rest.auth.authorise(null, {force:true}, function(err) {
 						if(err) {
--- a/common/lib/client/resource.js
+++ b/common/lib/client/resource.js
@@ -107,7 +107,7 @@ var Resource = (function() {
 			}
 
 			Http.get(rest, path, headers, params, function(err, res, headers, unpacked) {
-				if(err && isErrATokenProblem(err)) {
+				if(err && Auth.isTokenErr(err)) {
 					/* token has expired, so get a new one */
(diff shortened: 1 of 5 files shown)


SimonWoolf:
(Not included in this PR: changing expectation 40140 -> 40142 in the auth test. Commit that does that is https://github.com/ably/ably-js/commit/05c9f97c70d71fd401a4e413a11e4871c9a66fee , to be merged when realtime deploys that change to sandbox)

common/lib/transport/connectionmanager.js
```diff
@@ -8,12 +8,15 @@ var ConnectionManager = (function() {
 	var PendingMessage = Protocol.PendingMessage;
 	var noop = function() {};
 
+	var isErrATokenProblem = function(error) {
```
paddybyers: perhaps `isTokenError()` ?

paddybyers:
> Including adding a new ConnectionManager#disconnectReason so can remember whether the reason for the last disconnect was a token issue.

Isn't `ConnectionManager.error` (to be renamed `errorReason` in line with the spec) supposed to do this? I see that currently it is only set for transitions to terminal states but is that necessary?

Or more generically might it be more useful to keep the `StateChange` object for the most recent change?

SimonWoolf:
> Isn't ConnectionManager.error (to be renamed errorReason in line with the spec) supposed to do this? I see that currently it is only set for transitions to terminal states but is that necessary?

Good point. I saw that ConnectionManager.error was only set for terminal states and assumed there was a reason, so added a new field. But actually, looking now, the only thing it's used for any more is [as the error callback for #send if not queueing](https://github.com/ably/ably-js/blob/master/common/lib/transport/connectionmanager.js#L878), for which setting it on disconnected too would be no bad thing.

> Or more generically might it be more useful to keep the StateChange object for the most recent change?

That'd work for the token purpose, but not sure it'd be good for the send purpose - eg if I disable queueing, something disconnects me, connectionmanager automatically starts connecting again, then I try to send; I want to know in the err callback why I was disconnected, not the fact that I'm now connecting.

But keeping the most recent non-undefined `stateChange.reason` would work for both purposes, since 'good' statechanges don't have reasons! I'll give that a go.

SimonWoolf:
2faa927

SimonWoolf:
(and https://github.com/ably/ably-js/commit/c20888f5d7681deeca5832285bb56afa0e6b76a7 for connection#errorReason)

common/lib/client/resource.js
```diff
@@ -1,5 +1,8 @@
 var Resource = (function() {
 	var msgpack = (typeof(window) == 'object') ? window.Ably.msgpack : require('msgpack-js');
+	var isErrATokenProblem = function(error) {
```
paddybyers: This wasn't renamed, and is also a duplicate. Can we not put it into `ConnectionManager.isTokenErr()` ?

SimonWoolf: oops, forgot I'd put that in the other file

paddybyers:
Thanks

common/lib/client/resource.js
mattheworiordan: Looks like this was merged into `master`?  I have corrected in https://github.com/ably/ably-js/commit/07e6e3b9ca88f34f6a1ac5e6c657ce068f16788b, are you happy with that?

paddybyers: Sure, my mistake, thanks

Issue #205, opened 2016-01-21
Possible issue with serverTimeOffset calculation for people in other timezones


SimonWoolf:
Reported by alexg, who was having issues with `auth_token_expires`. Have given it to alexg to debug, as the serverTimeOffset mechanism should be fairly self-contained. Slack transcript:

```
...
Ably: Connection.connect(): 
Ably: Auth.requestToken(): Need a new token, but authOptions does not include any way to request one
Ably: ConnectionManager.connectImpl(): Connection attempt failed with error; err = [ErrorInfo: Need a new token, but authOptions does not include any way to request one; statusCode=401; code=40101]
Ably: Connection.close(): connectionKey = undefined
✖ auth_token_expires
Assertion Message: connection to server failed
AssertionError: connection to server failed
    at Object.ok (/home/alexg/workspace/ably/ably-js/node_modules/nodeunit/lib/types.js:83:39)
    at Object.<anonymous> (/home/alexg/workspace/ably/ably-js/spec/common/modules/shared_helper.js:26:11)
    at callListener (/home/alexg/workspace/ably/ably-js/common/lib/util/eventemitter.js:13:18)
    at Connection.EventEmitter.emit (/home/alexg/workspace/ably/ably-js/common/lib/util/eventemitter.js:131:5)
    at /home/alexg/workspace/ably/ably-js/common/lib/client/connection.js:19:10
    at nextTickCallbackWith0Args (node.js:452:9)
    at process._tickCallback (node.js:381:13)
Error: Expected 4 assertions, 1 ran
    at Object.test.done (/home/alexg/workspace/ably/ably-js/node_modules/nodeunit/lib/types.js:121:25)
    at Object.test.done (/home/alexg/workspace/ably/ably-js/spec/common/modules/shared_helper.js:96:22)
    at Object.<anonymous> (/home/alexg/workspace/ably/ably-js/spec/common/modules/shared_helper.js:27:11)
    at callListener (/home/alexg/workspace/ably/ably-js/common/lib/util/eventemitter.js:13:18)
    at Connection.EventEmitter.emit (/home/alexg/workspace/ably/ably-js/common/lib/util/eventemitter.js:131:5)
    at /home/alexg/workspace/ably/ably-js/common/lib/client/connection.js:19:10
    at nextTickCallbackWith0Args (node.js:452:9)
    at process._tickCallback (node.js:381:13)
Ably: Connection.connect(): 
Ably: ConnectionManager.chooseTransport(): 
[...]

PR #213, opened 2016-01-23
Fixing issue #205 - Failing test auth_token_expires

alex-georgiou:
The Rest client would fail to connect if the client's clock is ahead of
the server's clock by more than the auth token's TTL. The serverTimeOffset
was calculated only when the queryTime auth param was set, thus any connections
initiated without queryTime=true would not benefit from the serverTimeOffset
mechanism, and the following test from Auth.prototype.authorise would fail:

    if(token.expires === undefined || (token.expires > this.getTimestamp())) {

The proposed solution memoizes the time offset on the Rest object's prototype,
making sure that the offset is calculated once per library load, as opposed to
once per request.

Memoization behaviour can be overriden with the queryTime bool parameter, if set.
Thus the offset is calculated in the following cases:

                                     queryTime
                        |  truthy  |   falsy   |   not set
                 -------|----------|-----------|-----------
    offset===null  true |   yes    |    no     |    yes
                  false |   yes    |    no     |    no

When reviewing, please verify that this is the desired behaviour.

### DIFF
--- a/common/lib/client/auth.js
+++ b/common/lib/client/auth.js
@@ -432,7 +432,7 @@ var Auth = (function() {
 				authoriseCb();
 				return;
 			}
-			if(authOptions.queryTime || (typeof(authOptions.queryTime) === 'undefined' && Rest.prototype.serverTimeOffset === null )) {
+			if(authOptions.queryTime) {
 				rest.time(function(err, time) {
 					if(err) {callback(err); return;}
 					request.timestamp = time;
--- a/common/lib/client/auth.js
+++ b/common/lib/client/auth.js
@@ -432,7 +432,7 @@ var Auth = (function() {
 				authoriseCb();
 				return;
 			}
-			if(authOptions.queryTime) {
+			if(authOptions.queryTime || (typeof queryTime == 'undefined' && Rest.prototype.serverTimeOffset === null )) {
 				rest.time(function(err, time) {
 					if(err) {callback(err); return;}
 					request.timestamp = time;
@@ -507,7 +507,7 @@ var Auth = (function() {
 	};
 
 	Auth.prototype.getTimestamp = function() {
-		return Date.now() + (this.rest.serverTimeOffset || 0);
+		return Date.now() + (Rest.prototype.serverTimeOffset || 0);
 	};
 
 	Auth.prototype._tokenClientIdMismatch = function(tokenClientId) {
--- a/common/lib/client/auth.js
+++ b/common/lib/client/auth.js
@@ -432,7 +432,7 @@ var Auth = (function() {
 				authoriseCb();
 				return;
 			}
-			if(authOptions.queryTime || (typeof authOptions.queryTime == 'undefined' && Rest.prototype.serverTimeOffset === null )) {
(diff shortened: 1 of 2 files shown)


mattheworiordan:
> The Rest prototype in one VM does token creation and the other does authorisation. I'm currently looking more into why/how this happens.

Agree with @paddybyers, most people will only ever use one instance of the library so optimising for cases where multiple instances are used is not a concern in reality.

mattheworiordan:
@alex-georgiou is this ready for review now?

alex-georgiou:
@mattheworiordan No actually it is not. I have tried a number of things and can't get it to work. So I decided to move on with other tasks for now. I have submitted two other PRs which are ready for review, the Logger test and the test for tokenId precedence.

alex-georgiou:
@mattheworiordan As discussed, summarising the issue I face here: 

`Auth.prototype.authorise()` fails when TTL=5sec and my client's clock is about 20 seconds ahead of the server clock. This if condition is false
`if(token.expires === undefined || (token.expires > this.getTimestamp())) {`
because `rest.serverTimeOffset` is null when `getTimestamp()` is called, so effectively `getTimestamp()` becomes a call to `Date.now()`.

* Simon originally proposed that I do something to `Auth.getTimestamp()` so that it requests the server time if needed. So I had to turn it into something that takes a callback which is what I tried to do in this commit 8a5559991675208d994d213e1cbfc66b0a5ee051 . The test timeouts and I think the problem is with the altered flow in `Auth.authorise()`, which now does an asynchronous call to `getTimestamp()` but I couldn't pinpoint it.

* Paddy Byers suggested that since we only need an accurate offset in this particular test where the TTL is extremely low, I could explicitly make the rest client in the test to ask for it. Since `clientRealtime` inherits from `Rest`, I tried calling `time()` on it and I put the rest of the test body inside that callback. Again it didn't work. Inside the time callback the offset was being calculated from the request to `/time`, but then it wasn't available since the `Rest` object actually doing the authentication was a different one and actually lived in a different VM context than my `clientRealtime` object. Possibly something to do with how the tests are loaded?

I'm sure both solutions can work but I didn't get any of them to work. I was having a hard time debugging this since it's a time-sensitive thing. I looked a bit into how to disable the 60-second timeout for tests so I can step through the debugger more easily but didn't find a way and decided to work on the other tasks to make some progress.

I am very interested to see what I was missing.

SimonWoolf:
> Since clientRealtime inherits from Rest, I tried calling time() on it and I put the rest of the test body inside that callback. Again it didn't work. Inside the time callback the offset was being calculated from the request to /time, but then it wasn't available since the Rest object actually doing the authentication was a different one and actually lived in a different VM context than my clientRealtime object. Possibly something to do with how the tests are loaded?

Assuming you're still looking at the `auth_token_expires` test, the `clientRealtime` *is* the one doing the authentication. The `rest` instance's only job is to send a token request, the rest is done completely by `clientRealtime`. So the approach you describe should work - could you post the code you tried for us to see?

(To remove the 60 second test timeout, remove the `helper.withTimeout` wrapper from the return value of the test module)

mattheworiordan:
@SimonWoolf I am about to push a fix for all of this, found a few problems...

Issue #164, opened 2015-11-13
ably-js-browsers failing in mobile safari


SimonWoolf:
eg latest run https://ci.ably.io/job/ably-js-browsers-min/326/consoleText:
```
IE 11.0.0 (Windows 10 0.0.0): Executed 202 of 202 SUCCESS
Chrome 42.0.2311 (Windows 8.1 0.0.0): Executed 207 of 207 SUCCESS 
Mobile Safari 8.0.0 (iOS 8.4.0): Executed 12 of 207 (9 FAILED) ERROR
```
@mattheworiordan was looking into this (since he has a mac & iphone so can debug locally) -- just making an issue so it doesn't drop off our radar.

mattheworiordan:
Thanks @SimonWoolf. I will look at the issues once the XHR streaming issue in iOS is fixed by @paddybyers

mattheworiordan:
Reminder that we must add Safari to browsers min, not just Mobile Safari, see https://github.com/ably/ably-js/blob/master/test/bin/ci-saucelabs-min.sh

SimonWoolf:
`publishVariations` still consistently failing on mobile safari only (both 8.0 and 9.0) - https://ci.ably.io/job/ably-js-browsers-all/147/consoleText

mattheworiordan:
@SimonWoolf did we add normal Safari to the browser mix for min, and do you know if that passes?

SimonWoolf:
Normal safari isn't there, no.

SimonWoolf:
Have now added desktop safari to browsers-all.

Annoyingly, testing just publishVariations on saucelabs on mobile safari result in [it passing successfully](https://ci.ably.io/job/ably-js-browsers-min/432/console), though it still fails when run as part of browsers-min. Have added some logging as well.

@mattheworiordan any progress with trying it on an actual iphone?

SimonWoolf:
Closed by https://github.com/ably/ably-js/pull/215

PR #215, opened 2016-01-24
Publish messages serially

mattheworiordan:
Fixes a probable issue when publishing multiple messages over REST concurrently in mobile browsers.

In response to https://github.com/ably/ably-js/issues/164

### DIFF
--- a/spec/realtime/message.test.js
+++ b/spec/realtime/message.test.js
@@ -1,6 +1,6 @@
 "use strict";
 
-define(['ably', 'shared_helper'], function(Ably, helper) {
+define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
 	var exports = {},
 		displayError = helper.displayError,
 		closeAndFinish = helper.closeAndFinish,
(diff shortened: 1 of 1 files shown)


mattheworiordan:
@SimonWoolf this makes the tests pass on my mobile browser.

cc @paddybyers

SimonWoolf:
Thanks :)

SimonWoolf:
And ably-js-browsers-min is back to normal! \o/ https://ci.ably.io/job/ably-js-browsers-min/451/

SimonWoolf:
@mattheworiordan actually, quick question -- do you have a sense of what actually went wrong in mobile safari with the old code? It seems a bit worrying that waiting for the publish callback fixes things -- ably-js should be able to handle someone spewing 20 publishes at once without a problem (and does, on every other browser). (Even if they're published out of order, that won't matter for publishvariations since the received messages are all poured into the case statement).

mattheworiordan:
> quick question -- do you have a sense of what actually went wrong in mobile safari with the old code?

No, unfortunately not, I just found it odd that we were publishing that number of messages over REST in a short period of time which means that if a mobile browser was intentionally rate / concurrency limiting things, it could result in timeouts.  Nothing specific unfortunately

PR #216, opened 2016-01-25
RSC2,3,4 logger test

alex-georgiou:
This is a proposed solution for testing the Logger (RSC2, RSC3, RSC4).

The difficulty here is that the logger is not defined to be a module, and thus cannot be `require`d. Also the logger is unavailable outside the library's context due to closure, so tests cannot reach it.

A new script tag is added in `nodeunit.html`. This ensures that the Logger is loaded in a global context. This solution has the drawback that it pollutes our global namespace, but only for our tests, which shouldn't be a big problem.

I have exposed the entire `/common/lib/util` dir to the express webserver rather than just the logger file, because in the future we might want to test the other util/* files, which are also not modules.

If exposing these files to the webserver is not acceptable for whatever reason, another idea would be to use something like `grunt-preprocess` to conditionally expose the logger to the tests only if the environment is a test environment. IMO this is an even less pretty solution, and also seems more hacky and error-prone.

If we are running node, I mimic the mechanism used in the library to load the logger. Again this is done because we cannot require something that is not a module.

A possible issue with these tests is that they lack isolation. This can be fixed by loading the logger again and again, or perhaps with some other mechanism that clones the logger, such as `Object.create(Logger)`. The problem here is that the Logger is not being instantiated via a constructor. Let me know if you'd like me to look more into this issue.


alex-georgiou:
@mattheworiordan I know I said this was ready for review but today I improved it a lot after seeing the PHP logger tests.

Initially I had written code to hook into the `process.stdout` stream but since this doesn't work anywhere but in Node, I decided to intercept the `console.log` function instead, which should work everywhere but IE.

The tests will probably still not work in IE without polyfills as I depend on `Array.prototype.slice` and `console.log` being a function. I have left TODOs in the code for this.

I believe IE compatibility is all that remains to be done now for these tests.

alex-georgiou:
Have fired up a windows VM from https://dev.windows.com/en-us/microsoft-edge/ and ran the logger tests in IE7 and IE8 compatibility mode. Couldn't run the tests, since`browser_setup.js` is failing. The tests pass in IE9, IE10 and Edge.

mattheworiordan:
> Have fired up a windows VM from https://dev.windows.com/en-us/microsoft-edge/ and ran the logger tests in IE7 and IE8 compatibility mode. Couldn't run the tests, sincebrowser_setup.js is failing. The tests pass in IE9, IE10 and Edge.

Have you tried with the [`IE89` branch](https://github.com/ably/ably-js/compare/ie89)?

spec/rest/logger.test.js
```diff
@@ -0,0 +1,139 @@
...
+	 */
+
+	var logger;
+	
+	if (isBrowser) {
+		logger = Logger;
+	} else {
+
+		var
+			fs   = require('fs'),
```
mattheworiordan: Odd spacing with this variable defined in the line after the var?  Similar below iwth includeScript

mattheworiordan:
Looks good, thanks.  I see you had a lot of commits to get there!

SimonWoolf:
> The tests will probably still not work in IE without polyfills as I depend on Array.prototype.slice

[MDN says](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/slice) that Array.prototype.slice works in all versions of IE (and was part of javascript 1.2, so in theory should be supported from IE 4 onwards), so not sure why you don't expect it to work? (slice is used in plenty of other places in ably-js)

> and console.log being a function.

You can assume that console & console.log are defined during test runs (due to the way the browser tests work, they capture console output)

spec/rest/logger.test.js
alex-georgiou: > Odd spacing with this variable defined in the line after the var? Similar style issues below

@mattheworiordan Will fix it. By the way are we following some specific coding standard or hint tool? Or should I just try to mimic the existing code?

alex-georgiou: I notice that in some .js files indentation is done with a tab character (e.g. `karma.conf.js`) and in other files it's done with two (2) spaces (e.g. `auth.tests.js`).

alex-georgiou:
> Looks good, thanks. I see you had a lot of commits to get there!

@mattheworiordan I know it's confusing. I plan to cut down on those in the future. Or possibly use sub-branches locally. Also I want to adopt [this format](http://karma-runner.github.io/0.10/dev/git-commit-msg.html) for my commit messages.

> MDN says that Array.prototype.slice works in all versions of IE (and was part of javascript 1.2, so in theory should be supported from IE 4 onwards), so not sure why you expect it not to work? (slice is used in plenty of other places in ably-js)

@SimonWoolf That's great news. Last night I entered TODOs to check because I wasn't sure. I see now that it works. More importantly, substituting the `console.log` *object* with a *function* also works in IE7.

**Question**: Should I move my logger instantiation code from the top of my test file into the shared_helper file (or somewhere else)?

> Have you tried with the IE89 branch?

I did now. Still crashing for 7 and 8. Works fine for 9 10 and Edge. Logs from IE8 compatibility:

	HTML1300: Navigation occurred.
	File: nodeunit.html
	SCRIPT445: Object doesn't support this action
	File: logger.js, Line: 5, Column: 2
	SCRIPT445: Object doesn't support this action
	File: ably.js, Line: 4169, Column: 2
	TypeError: Unable to get property 'Utils' of undefined or null reference
	   {
		  description: "Unable to get property 'Utils' of undefined or null reference",
		  message: "Unable to get property 'Utils' of undefined or null reference",
		  name: "TypeError",
		  number: -2146823281
	   }

	TypeError: Unable to get property 'Ably' of undefined or null reference
	   {
		  description: "Unable to get property 'Ably' of undefined or null reference",
		  message: "Unable to get property 'Ably' of undefined or null reference",
		  name: "TypeError",
		  number: -2146823281
	   }

	TypeError: Unable to get property 'withTimeout' of undefined or null reference
	   {
		  description: "Unable to get property 'withTimeout' of undefined or null reference",
		  message: "Unable to get property 'withTimeout' of undefined or null reference",
		  name: "TypeError",
		  number: -2146823281
	   }

	TypeError: Unable to get property 'setUp' of undefined or null reference
	   {
		  description: "Unable to get property 'setUp' of undefined or null reference",
		  message: "Unable to get property 'setUp' of undefined or null reference",
		  name: "TypeError",
		  number: -2146823281
	   }

mattheworiordan:
http://karma-runner.github.io/0.10/dev/git-commit-msg.html looks good.

In terms of code styling, @paddybyers you will have to comment on that, it would be nice if we had a reference to an existing style guide that we could point people to.

> Question: Should I move my logger instantiation code from the top of my test file into the shared_helper file (or somewhere else)?

I don't think it's necessary unless other test are using it, but perhaps @SimonWoolf will have a different view.

> Still crashing for 7 and 8. Works fine for 9 10 and Edge. Logs from IE8 compatibility:

FYI, we don't aim to support IE7.  However, those errors seem like potentially new ones with the changes recently made by @SimonWoolf. @SimonWoolf any thoughts?

spec/rest/logger.test.js
mattheworiordan: In this repo, we should be using tabs.  Feel free to issue a separate PR to correct the space issues if you find them

alex-georgiou: I'm mainly asking because I like to set Eclipse to fix these things automatically on file save. Then I don't have to waste time on the trivial stuff. I currently have my tab set to 4 characters.

In the last project I worked on we followed Drupal coding standards, and there are publicly available settings that you can download and import into your favourite IDE. Perhaps we could do something similar. I can have a look later.

SimonWoolf:
> Still crashing for 7 and 8.

7 is expected, as Matt says. Rebasing this PR on ie89 results in it working fine in IE8 for me:

ie8logger

Could you push the branch you rebased which gives you those errors up somewhere.

> Question: Should I move my logger instantiation code from the top of my test file into the shared_helper file (or somewhere else)?

You mean the file loading mechanism for node? Since this is the only test that uses it right now I'm not bothered either way. If we find any more tests that need it then it should be extracted to a helper.

spec/rest/logger.test.js
```diff
@@ -0,0 +1,135 @@
+"use strict";
+
+define(['shared_helper'], function(helper) {
+	var exports = {},
+		isBrowser = !!(typeof window !== 'undefined' && typeof navigator !== 'undefined' && window.document);
```
SimonWoolf: There's already an `isBrowser` global, no need to shadow it with your own local

alex-georgiou:
[ie89_logger_test Branch](/alex-georgiou/ably-js/tree/ie89_logger_test)

I am running it on IE11 with IE8 compatibility on a Windows 8 VM. From past experience I know that IE compatibility mode is not always accurate. Some times it just adds another layer of version hell.

ably_logger_test_ie11 - win8 1

alex-georgiou:
@SimonWoolf Also, I just noticed that what I've set is **document** compatibility mode. I remember in past versions of IE you could set document *and* browser compatibility modes. If you're running it on an actual IE8 then your results might be more accurate.

SimonWoolf:
@alex-georgiou OK, it looks like you've found a bug in IE11's IE8 compatibility mode :)

More specifically: in actual IE8, `typeof window.console.log.apply` is `undefined`. In IE11's IE8 compatibility mode, that call throws an error.

I've pushed [a commit](https://github.com/ably/ably-js/commit/80c5e8b1846b1fc400eaba7d8a754fd401b64b4b) to `ie89a` which makes the actual library work for someone using the library in that mode. Your test still doesn't pass in IE11 in IE8 compatibility mode, but we don't care about that, since our IE8 browser tests are run in a proper IE8 VM (by saucelabs).

alex-georgiou:
> OK, it looks like you've found a bug in IE11's IE8 compatibility mode :)

Ah, I thought so. Hate it when that happens (it's not even the first time). Sorry for wasting your time, @SimonWoolf . I will download more VMs from http://modern.ie and in the future I will test on actual IE instances. Compatibility mode is unreliable, lesson learned.

alex-georgiou:
@mattheworiordan @SimonWoolf This is ready for review. Let me know what you think.

spec/rest/logger.test.js
```diff
@@ -0,0 +1,149 @@
...
+
+		test.equal(typeof(logger), 'function', 'Logger has been instantiated');
+
+		test.done();
+	};
+
+	/*
+	 * Check that the logger writes to stdout by default with no errors.
+	 */
+	exports.logger_writes_to_stdout = function(test) {
```
mattheworiordan: This test has nothing to do with stdout, so should be changed to describe what it's actually testing which is console.log

mattheworiordan:
One comment otherwise good.  I would suggest you consider squashing these commits though as they don't really tell a story worth keeping, and given there are 42 commits for 150 lines of code, that feels quite unnecessary unfortunately.

alex-georgiou:
> One comment otherwise good. I would suggest you consider squashing these commits though as they don't really tell a story worth keeping, and given there are 42 commits for 150 lines of code, that feels quite unnecessary unfortunately.

Of course you are quite right. I just noticed that I must have done something stupid with git, possibly when I rebased to the ie89 branch, because all my commits are doubled. The actual number was roughly half of 42 which is still a lot, I know and will squash them now.

spec/rest/logger.test.js
alex-georgiou: > This test has nothing to do with stdout, so should be changed to describe what it's actually testing which is console.log

@mattheworiordan Yes this description remained from when I had actually hooked to the stdout stream, because this is what requirement RSC2 is asking. I gave up on this becauseit only works on node and because I don't think it's what we want to test anyway. The logger doesn't know about standard I/O, only console.log.

Issue #165, opened 2015-11-13
Http fallback sequence tweaks


SimonWoolf:
See https://github.com/ably/ably-js/issues/160.

@paddybyers's thoughts:
> In the worst case we should have a 2-step fallback (ie best-available http transport -> jsonp). Trying all the options is bad I think, but I would prefer that we pick one and it works.

mattheworiordan:
Notes from our discussion on this issue:

The process for the `ably-js` library should be:

* Filter transports available 
* Try the preferred HTTP transport (XHR-streaming / iFrame for one edge case)
* If that fails to connect
  * If internet up try other hosts
  * If internet not up (using same transport to determine this) try fallback JSONP transport 
* Once connected, try and upgrade to WebSockets

The 15s disconnect timeout should only apply once all the fallbacks have been exhausted, currently it seems that if the request times out in 15s, no fallbacks are used because the realtime state timeout had passed.

SimonWoolf:
> The 15s disconnect timeout should only apply once all the fallbacks have been exhausted, currently it seems that if the request times out in 15s, no fallbacks are used because the realtime state timeout had passed.

Not sure about this part. This would mean that `realtimeRequestTimeout` no longer applies to establishing a connection, so there would be no enforced cap on the amount of time the client lib can be in the `connecting` state. (So eg any transport bug which results in it stalling while trying to connect and never actually failing, like https://github.com/ably/ably-js/issues/160, would jam things forever)

(Note: this is all separate from the `disconnectRetryTimeout`, which only starts once the client enters the `disconnected` state)

Possibly we could separate `realtimeOpenTimeout` back out from `realtimeRequestTimeout` and make it longer (maybe 30s), then have a shorter timeout for each transport connect attempt?

(Long-term, @mattheworiordan suggests we'll want to have multiple transports connecting at once (possibly staggered), choosing the first one that succeeds, to avoid a 15s wait if eg you're on a connection where the xhr attempt times out (eg https://github.com/ably/realtime/issues/371) but eg websockets or jsonp would succeed immediately)

mattheworiordan:
@SimonWoolf I would agree that we simply separate out, and in the case of JS which has more than one transport (other client libs don't offer this currently), we simply change the default timeouts.  I would personally opt for 10s per requrest and 20s total connection timeout

paddybyers:
I think in practice if it's going to succeed it will do so in much less than 10s. Maybe 5 isn't enough but it could be 7.5.

[...]

PR #218, opened 2016-01-26
Some proposed fixes for "Poor connection never attaches and reports connection to be connected #195"

SimonWoolf:
Addresses #195.

The questions from that issue, with discussion & questions:

> (1) we should change the Error to an ErrorInfo and verify we get a meaningful error message.

Already addressed by https://github.com/ably/ably-js/pull/197

> (2) does it correctly handle the timeout of the attach()? how come the channel remains in the attaching state?

Because previous strategy on timeout was just to retry. https://github.com/ably/ably-js/commit/96e65dee0f0d01bce8035683b7b429822fdbd17b changes that to fail channel ops (leaving the channel in the state it was in beforehand) if the request fails.

I'm not certain this is the correct thing to do (in particular, since realtimeRequestTimeout is 10s and httpRequestTimeout is 15s, this leaves a period of 5s after which the channel considers the request failed, but during which it could still succeed, which is kinda weird). Opinions?

> (3) what should be done about this ws transport that's orphaned, associated with an upgrade that's gone bad. Can we just arrange for this to become the active transport? Is the behaviour here actually ok, and the problem is that we keep going with the second xhr connection even though we're now connected?

https://github.com/ably/ably-js/commit/96f83a6921c1ed8f7fdf6055af8989368f20f625 changes the deactivateTransport logic such that if there is a pending transport which is connected and scheduled for activation -- just waiting for the active transport to finish what its
doing - then, if the current active transport fails, it just lets the pending one take over, avoiding going into the disconnected state. If that's not the case, we continue as we do at the moment.

(This technically leaves a period in which the state is `connected` but activeProtocol is null -- I think should be at most one cycle of the event loop -- not sure if this is a problem?)

> (4) why this error - is this reproducible?

Because transport#abort was calling `requestClose(true)`. I don't understand why that was like that -- AFAIK the only reason to do a /close (rather than a /disconnect) is if we want realtime to discard the connection state, which surely isn't something we should be doing unless the user requests a close. @paddybyers am I missing something?

https://github.com/ably/ably-js/commit/7a593e23236cc87f413b307ab438ac7b9243dbb6 changes it to `requestClose(false)`.

> (5) is it right that setSuspended() is called as a result of a non-fatal upgrade error?

In that case I think it was right -- realtime lost the connection state (due to the previous point), so Matt was no longer attached.

### DIFF
--- a/common/lib/client/realtimechannel.js
+++ b/common/lib/client/realtimechannel.js
@@ -172,6 +172,7 @@ var RealtimeChannel = (function() {
 							callback();
 							break;
 						case 'failed':
+						case 'attached':
 							callback(err || connectionManager.getStateError());
 							break;
 						default:
@@ -461,8 +462,7 @@ var RealtimeChannel = (function() {
 		this.stateTimer = setTimeout(function() {
 			Logger.logAction(Logger.LOG_MINOR, 'RealtimeChannel.setPendingState', 'timer expired');
 			self.stateTimer = null;
-			/* retry */
-			self.checkPendingState();
+			self.timeoutPendingState();
 		}, this.realtime.options.timeouts.realtimeRequestTimeout);
 	};
 
@@ -478,6 +478,23 @@ var RealtimeChannel = (function() {
 				/* resume any sync operation that was in progress */
 				this.sync();
 			default:
+				break;
+		}
+	};
+
+	RealtimeChannel.prototype.timeoutPendingState = function() {
+		switch(this.state) {
+			case 'attaching':
+				var err = new ErrorInfo('Channel attach timed out', 90000, 408);
+				this.setState('detached', err);
+				this.failPendingMessages(err);
+				break;
+			case 'detaching':
+				var err = new ErrorInfo('Channel detach timed out', 90000, 408);
+				this.setState('attached', err);
+				break;
+			default:
+				this.checkPendingState();
 				break;
 		}
 	};
--- a/common/lib/client/realtimechannel.js
+++ b/common/lib/client/realtimechannel.js
@@ -467,15 +467,12 @@ var RealtimeChannel = (function() {
 	};
 
 	RealtimeChannel.prototype.checkPendingState = function() {
-		var result = false;
 		switch(this.state) {
 			case 'attaching':
 				this.attachImpl();
-				result = true;
 				break;
 			case 'detaching':
 				this.detachImpl();
-				result = true;
 				break;
 			case 'attached':
 				/* resume any sync operation that was in progress */
@@ -483,7 +480,6 @@ var RealtimeChannel = (function() {
 			default:
 				break;
 		}
-		return result;
 	};
 
 	RealtimeChannel.prototype.clearStateTimer = function() {
--- a/common/lib/transport/comettransport.js
+++ b/common/lib/transport/comettransport.js
@@ -86,24 +86,33 @@ var CometTransport = (function() {
 
 	CometTransport.prototype.disconnect = function() {
 		Logger.logAction(Logger.LOG_MINOR, 'CometTransport.disconnect()', '');
-		this.requestClose(false);
+		this.requestDisconnect();
 		Transport.prototype.disconnect.call(this);
 	};
 
-	CometTransport.prototype.requestClose = function(closing) {
-		Logger.logAction(Logger.LOG_MINOR, 'CometTransport.requestClose()', 'closing = ' + closing);
-		var closeUri = this.closeUri;
-		if(closeUri) {
-			var self = this,
-				closeRequest = this.createRequest(closeUri(closing), null, this.authParams, null, REQ_SEND);
+	CometTransport.prototype.requestClose = function() {
+		Logger.logAction(Logger.LOG_MINOR, 'CometTransport.requestClose()');
+		this._requestCloseOrDisconnect(true);
+	};
 
-			closeRequest.on('complete', function (err) {
+	CometTransport.prototype.requestDisconnect = function() {
+		Logger.logAction(Logger.LOG_MINOR, 'CometTransport.requestDisconnect()');
+		this._requestCloseOrDisconnect(false);
+	};
+
+	CometTransport.prototype._requestCloseOrDisconnect = function(closing) {
+		var closeOrDisconnectUri = closing ? this.closeUri : this.disconnectUri;
+		if(closeOrDisconnectUri) {
+			var self = this,
+				request = this.createRequest(closeOrDisconnectUri, null, this.authParams, null, REQ_SEND);
+
+			request.on('complete', function (err) {
 				if(err) {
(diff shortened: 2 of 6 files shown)


mattheworiordan:
> I'm not certain this is the correct thing to do (in particular, since realtimeRequestTimeout is 10s and httpRequestTimeout is 15s, this leaves a period of 5s after which the channel considers the request failed, but during which it could still succeed, which is kinda weird). Opinions?

ok, but I assumed HTTP request timeout applies to normal REST HTTP requests as opposed to Comet based HTTP requests. Is that not the case?

Looks good :+1:

mattheworiordan:
Is there no way to simulate the problem we had with two transports by adding a test to prevent a regression?  I would prefer to see this if possible.

SimonWoolf:
> I assumed HTTP request timeout applies to normal REST HTTP requests as opposed to Comet based HTTP requests. Is that not the case?

Currently, no. At the moment in ably-js, the two timeouts work at different levels of the lib:

The httpRequestTimeout determines how long an http request in SEND mode lasts before giving up. At [that point](https://github.com/ably/ably-js/blob/ee12098810d1cf65dbc4903536011e9baba1ec4c/browser/lib/transport/xhrrequest.js#L93) the request object doesn't know or care if it's been created as part of a comet transport, a REST request, an authUrl call to a client server, or anything else, it just knows its url and request mode.

The realtimeRequestTimeout is at a much higher level, and determine how long the Connection stays in the `connecting` state before going to `disconnected`, how long the Channel stays in the `attaching` state, etc.

There's [an open issue](https://github.com/ably/ably-js/issues/165) which discusses changing how the timeouts interact in the case of opening a realtime connection specifically. We should probably discuss how timeouts should work more generally.

> Is there no way to simulate the problem we had with two transports by adding a test to prevent a regression? I would prefer to see this if possible.

I had a play around with trying to simulate the problem, but couldn't see any nice way of doing it from within ably-js (the failure is an http request that never responds, so would need to figure out the http transport being used, them shim XHRequest#exec for a version which figures out if a request is the right one replaces the xhr.onreadystatechange handler with a noop if so; and be able to do an equivalent thing for all the other comet transports so the test works everywhere).

So, yes, it's possible. But I figured not much point if we're shortly going to remove the upgrade mechanism anyway...

(Also: if we do want to do general realtime request failure tests: per Paddy's suggestion [here](https://github.com/ably/ably-js/issues/195#issuecomment-171368939), might it be better (more flexible, not so tied to the implementation of each transport) to do those by making realtime fail to respond, rather than fake it with complicated shims? Could even maybe have a set of bad-realtime-endpoints that fails in fixed ways simulating network conditions, and can test all the client libraries against them.)

mattheworiordan:
> So, yes, it's possible. But I figured not much point if we're shortly going to remove the upgrade mechanism anyway...

Didn't think we had decided on that, I thought we would see the impact of the new router changes first.

>  Could even maybe have a set of bad-realtime-endpoints that fails in fixed ways simulating network conditions

Sure, would be useful, perhaps even a header that can be passed in to `sandbox` instances that triggers this.  Saying that, I wouldn't want to add too much of that sort of thing into the realtime system adding more complexity unnecessarily.

An alternative would instead be to simply start up a local WS & HTTP server that proxies to realtime and achieves that thus keeping this complexity in the test suite itself where it belongs

SimonWoolf:
@paddybyers can you confirm whether you agree with the other changes in this? In particular, whether you think changing of the attaching timeout to revert the channel to `detached` rather than to retry indefinitely is correct. (I made the change per the discussion in #195, but I'm not at all sure that the retry-indefinitely behaviour (in the case of requests vanishing into the ether) wasn't better -- it's not like the user suddenly isn't going to want to be attached to the channel after all. (Also the weirdness caused by the difference between the two timeouts discussed above)).

SimonWoolf:
@paddybyers see prev comment

paddybyers:
> (in particular, since realtimeRequestTimeout is 10s and httpRequestTimeout is 15s, this leaves a period of 5s after which the channel considers the request failed, but during which it could still succeed, which is kinda weird). Opinions?

There's no real reason for them to be different timeouts. However, it doesn't really introduce any new case that needs to be handled, because of course the operation might have succeeded in the first 10s but the indication of that failed to reach us. So the 10s elapses, so we indicate failure and fail the pending ops; now in the next few seconds we find out that the operation did in fact succeed and we are attached after all. That's just too bad; the server response always takes precedence in determining our state, so we are attached, and it's just a shame that we failed those messages.

paddybyers:
> 96f83a6 changes the deactivateTransport logic such that if there is a pending transport which is connected and scheduled for activation -- just waiting for the active transport to finish what its
> doing - then, if the current active transport fails, it just lets the pending one take over, avoiding going into the disconnected state. If that's not the case, we continue as we do at the moment.

This looks ok but it would be even better if we could test it :)

> (This technically leaves a period in which the state is connected but activeProtocol is null -- I think should be at most one cycle of the event loop -- not sure if this is a problem?)

Is there no way to avoid this? We're synchronously deciding what the state should be, so why can't we make it so? I've not understood the problem here I think

> Because transport#abort was calling requestClose(true). I don't understand why that was like that -- AFAIK the only reason to do a /close (rather than a /disconnect) is if we want realtime to discard the connection state, which surely isn't something we should be doing unless the user requests a close.

Correct.

Otherwise LGTM, thanks.

SimonWoolf:
> We're synchronously deciding what the state should be, so why can't we make it so?

So: when this happens, [this closure](https://github.com/ably/ably-js/blob/632d9c3aec3330382f0b2bbb25412f3c62cb41da/common/lib/transport/connectionmanager.js#L370-L394) is registered as a listener for the `nopending` event. We could unregister that and run it manually (extracting it out from `scheduleTransportActivation`). The reason I didn't do that as I didn't want to bypass your `nopending` mechanism, as I wasn't sure what the consequences would be.

Would there be any bad consequences? Once the new transport is active, [Channels#onTransportActive](https://github.com/ably/ably-js/blob/632d9c3aec3330382f0b2bbb25412f3c62cb41da/common/lib/client/realtime.js#L47-L52) calls `checkPendingState` on all in-progress channels, which refreshes pending channel operations and presence syncs. If that's sufficient for continuity of pending channel ops and presence syncs across transport changes, then TBH I don't understand why the `nopending` mechanism is needed at all...?

paddybyers:
> So: when this happens, this closure is registered as a listener for the nopending event. We could unregister that and run it manually (extracting it out from scheduleTransportActivation). The reason I didn't do that as I didn't want to bypass your nopending mechanism, as I wasn't sure what the consequences would be.

What's asynchronous though?

SimonWoolf:
@paddybyers (reiterating slack discussion): Sorry, you're right, we could do it just be emitting `nopending` there and then.

But, again: that would bypass the nopending mechanism. Which I don't object to, except that I don't understand why it exists at all, given that Channels#onTransportActive deals with pending channel operations across transport changes (I'm obviously a bit wary of bypassing something whose purpose I don't understand...)

PR #220, opened 2016-01-27
test: clientId takes precedence over defaultCliendId for auth operations

alex-georgiou:
Adding auth test for RSA7a4.

### DIFF
--- a/spec/realtime/auth.test.js
+++ b/spec/realtime/auth.test.js
@@ -380,6 +380,28 @@ define(['ably', 'shared_helper'], function(Ably, helper) {
 	};
 
 	/*
+	 * When a clientId value is provided in both ClientOptions#clientId and ClientOptions#defaultTokenParams,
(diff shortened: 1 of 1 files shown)


SimonWoolf:
`defaultTokenParams` is not itself a `tokenParam` (or even an `authOption`), it's a `clientOption` -- ie a thing provided to the rest/realtime client initializer. This test currently is just showing that the `defaultTokenParams` tokenParam is discarded by realtime.

Incidentally, there's already a test for RSA7a4 --
https://github.com/ably/ably-js/blob/ee12098810d1cf65dbc4903536011e9baba1ec4c/spec/realtime/init.test.js#L147-L163, which it may be helpful to compare against.

alex-georgiou:
Ooops I missed that. Still learning...

mattheworiordan:
> Ooops I missed that. Still learning...

@alex-georgiou sorry about that, I should have checked with @SimonWoolf first...

PR #221, opened 2016-01-28
Issue205 auth token expires fails

mattheworiordan:
@alex-georgiou this builds on your PR https://github.com/ably/ably-js/pull/213

As you will see, there were in fact quite a few issues that was causing it to fail including:

* Token request in authorise was not run asynchronously once time was retrierved
* You were not getting the server time for both instances of the library so the Realtime library still had the wrong time
* Few bugs relating to expectations around objects existing such as authOptions
* Not using the client library configured queryTime option

cc @SimonWoolf @paddybyers

### DIFF
--- a/common/lib/client/auth.js
+++ b/common/lib/client/auth.js
@@ -146,7 +146,19 @@ var Auth = (function() {
 			authOptions = null;
 		}
 
-		var token = this.tokenDetails;
+		var self = this,
+			token = this.tokenDetails;
+
+		var requestToken = function() {
+			self.requestToken(tokenParams, authOptions, function(err, tokenResponse) {
+				if(err) {
+					callback(err);
+					return;
+				}
+				callback(null, (self.tokenDetails = tokenResponse));
+			});
+		}
+
 		if(token) {
 			if(this._tokenClientIdMismatch(token.clientId)) {
 				callback(new ErrorInfo('ClientId in token was ' + token.clientId + ', but library was instantiated with clientId ' + this.clientId, 40102, 401));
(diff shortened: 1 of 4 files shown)


SimonWoolf:
:+1: 

Though tbh, I still think it's kinda silly that we're forcing devs to choose between forcing a time request for *every* library-initiated `authorise` call (`querytime: true`), or none of them. Seems like a more-useful option would be a one that makes a time request at most once, iff serverTimeOffset is null, and rely on serverTimeOffset thereafter. (Or maybe once on library init? Or even forget the option and do it unconditionally on library *load* to save time, and store it statically).

If not doing that (or if having the option but not on by default) we should consider Paddy's suggestions [here](https://github.com/ably/ably-js/pull/213#commitcomment-15664406), last one especially seems very sensible.

mattheworiordan:
>  I still think it's kinda silly that we're forcing devs to choose between forcing a time request for every library-initiated authorise call (querytime: true)

Ah, I had assumed the time method when persisting the offset would prevent it being called again, see https://github.com/ably/ably-js/blob/issue205-auth_token_expires-fails/common/lib/client/rest.js#L101 and https://github.com/ably/ably-js/pull/221/files#diff-cc36f4a032c8d16ad5390d989524c305R517. That is not the case, so https://github.com/ably/ably-js/commit/6d2b374096c83177d2c03de720403996f2d77ab3 fixes that.

> Seems like a more-useful option would be a one that makes a time request at most once, iff serverTimeOffset is null, and rely on serverTimeOffset thereafter. 

I had assumed the existing code was doing that, but clearly not, so I will modify.  I agree that approach is best.  BTW. There is no mention of this behaviour in our spec anywhere, I will add this for a 0.9 release.

SimonWoolf:
Ah, fair enough then! 

I'll add a wiki issue for discussion of Paddy's suggestions.

Added a quick commit adding `queryTime: true` to the realtime constructors of the other auth tests that use tokens with small ttls. (Also removes it from the `rest` constructor of the test, just for consistency -- the rest one has 120s of leeway in both directions (unlike the realtime one using the token which has 5s of leeway in one direction), and if system clock is that far out, then most of the rest of the entire test suite will fail too (everything that gets a token). Not much point changing that for one test).

mattheworiordan:
> remember that realtime calculates the expiry time based on the requested ttl and it's own system time, not based on the timestamp in the token request

Yup, you are right, thanks for clarifying

PR #222, opened 2016-01-29
Safeguard Realtime Constructor

CrowdHailer:
Ably.Realtime is part of the public api to the whole library. I came across this because the presence docs have an example where the new keyword is not used to instantiate a client. Currently this throws an obscure error. `TypeError: this.connect is not a function(…)`

It seams that `new Ably.Realtime()` should never be called without the new keyword so this safeguard makes sense'

### DIFF
--- a/common/lib/client/realtime.js
+++ b/common/lib/client/realtime.js
@@ -1,6 +1,10 @@
 var Realtime = (function() {
 
 	function Realtime(options) {
+		if(!(this instanceof Realtime)){
+			return new Realtime(options);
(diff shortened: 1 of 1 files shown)


paddybyers:
Should we have this for `Rest` as well?

mattheworiordan:
Thanks @CrowdHailer, merged in with https://github.com/ably/ably-js/commit/6db7c2c221baebf33775b4f00ea849a09d42d45e + added support for Rest safeguard as well

PR #223, opened 2016-01-29
RSC2,RSC3,RSC4 logger test

alex-georgiou:
OK as I said before something very horrible has happened on my previous branch. After too much history hacking I decided to create a fresh branch.

I have squashed down the commits into three that capture the history of how I did the logger tests. I mainly did this in case I need to go back to the code where I hooked into the stdout stream.

The plus side of all this mess is that I got some excellent hands-on experience with the rebase command :)

### DIFF
--- a/spec/web_server
+++ b/spec/web_server
@@ -35,6 +35,7 @@ server.get('/support/browser_file_list.js', function(req, res) {
 server.use('/node_modules', express.static(__dirname + '/../node_modules'));
 server.use('/spec', express.static(__dirname));
 server.use('/browser', express.static(__dirname + '/../browser'));
+server.use('/common', express.static(__dirname + '/../common'));
 server.use(express.static(__dirname));
 
 var port = process.env.PORT || 3000;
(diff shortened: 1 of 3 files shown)


spec/web_server
```diff
@@ -32,6 +32,7 @@ server.get('/support/browser_file_list.js', function(req, res) {
 server.use('/node_modules', express.static(__dirname + '/../node_modules'));
 server.use('/spec', express.static(__dirname));
 server.use('/browser', express.static(__dirname + '/../browser'));
+server.use('/common/lib/util', express.static(__dirname + '/../common/lib/util'));
```
mattheworiordan: Could this not be `server.use('/common', express.static(__dirname + '/../common'));` so that any common file can be included more easily in future?

mattheworiordan:
This is failing so I cannot merge.  Given all of these commits should really be just merged into one (the history is not meaningful as we don't hook into `process.stdout`, please can you do that and rebase on `master` and see if it passes.  Master is passing, see https://travis-ci.org/ably/ably-js/builds/108239128

Issue #171, opened 2015-11-25
Update README 


mattheworiordan:
* [ ] Check that the README.md is complete with installation instructions
* [ ] Ensure all example code is up to date
* [ ] Ensure version number in README is up to date, and automatically updated in future version bumps
* [ ] Ensure repo is distributed to package manager

paddybyers:
Include a comment re the closure compiler installation.

SimonWoolf:
Most of this I've now done at various points: installation instructions in e3d82ec2, version number autoupdating in a49308ef, grunt:deploy task in 38ea16a0 (with CDN deployment much-improved by Matt in 2d0dbfa0).

Just done a few more tweaks to the readme: https://github.com/ably/ably-js/compare/readme-tweaks. Added a few more methods and explanations, added a note on the closure compiler, updated paginatedResults api, and removed a couple of things for simplicity (eg the list of what attributes a tokenRequest has). It's still far from a complete api documentation, but I guess it's not supposed to be, that's what the API documentation is for. I guess the readme should be the most common methods?

mattheworiordan:
Have reviewed @SimonWoolf, looks very good to me

SimonWoolf:
Closed by https://github.com/ably/ably-js/pull/232 (and previously: e3d82ec, a49308e, 38ea16a, 2d0dbfa)

PR #224, opened 2016-01-29
RSC18 client options tls

alex-georgiou:
This is my attempt at RSC18.

First I verified that the requirement for not sending a secret key over HTTP was NOT implemented. I used the stats request for this.

Then I modified the `Auth` object so that the error is generated exactly at the two functions where the key is added to the headers. This should ensure that all requests using the basic authentication headers will check for TLS.

Notes:
1. I have put in some other TLS-related tests in, perhaps some overlap with other tests in `/spec/rest/defaults.test.js`. Please review and let me know if it's overkill.
2. I am returning an `ErrorInfo`. I see we do not have any mechanism that will pull error message strings from `errors.json` so I repeated the strings in the code.
3. I am printing the error to the logger as it occurs, even though it is also printed when sent higher up to `PaginatedResource.get()`. I don't know how verbose we want our error reporting to be, but in theory the `Auth` functions could be called elsewhere, hence I'm printing out the error as it occurs for safety.

### DIFF
--- a/common/lib/client/auth.js
+++ b/common/lib/client/auth.js
@@ -483,8 +483,15 @@ var Auth = (function() {
 	 * based on the current auth parameters
 	 */
 	Auth.prototype.getAuthParams = function(callback) {
-		if(this.method == 'basic')
+		if(this.method == 'basic') {
+			if (!this.rest.options.tls) {
+				var msg = 'invalid use of Basic auth over non-TLS transport';
+				Logger.logAction(Logger.LOG_ERROR, 'Rest.getAuthParams()', msg );
+				callback(new ErrorInfo(msg, 40103, 401));
+				return;
+			}
 			callback(null, {key: this.key});
+		}
 		else
 			this.authorise(null, null, function(err, tokenDetails) {
 				if(err) {
@@ -501,6 +508,12 @@ var Auth = (function() {
 	 */
 	Auth.prototype.getAuthHeaders = function(callback) {
 		if(this.method == 'basic') {
+			if (!this.rest.options.tls) {
+				var msg = 'invalid use of Basic auth over non-TLS transport';
+				Logger.logAction(Logger.LOG_ERROR, 'Rest.getAuthHeaders()', msg );
+				callback(new ErrorInfo(msg, 40103, 401));
+				return;
+			}
 			callback(null, {authorization: 'Basic ' + this.basicKey});
 		} else {
 			this.authorise(null, null, function(err, tokenDetails) {
--- a/common/lib/client/auth.js
+++ b/common/lib/client/auth.js
@@ -484,15 +484,14 @@ var Auth = (function() {
 	 */
 	Auth.prototype.getAuthParams = function(callback) {
 		if(this.method == 'basic') {
-			if (!this.rest.options.tls) {
+			if (!this.client.options.tls) {
 				var msg = 'invalid use of Basic auth over non-TLS transport';
-				Logger.logAction(Logger.LOG_ERROR, 'Rest.getAuthParams()', msg );
+				Logger.logAction(Logger.LOG_ERROR, 'Rest.getAuthParams()', msg);
 				callback(new ErrorInfo(msg, 40103, 401));
 				return;
 			}
 			callback(null, {key: this.key});
-		}
-		else
+		} else
 			this.authorise(null, null, function(err, tokenDetails) {
 				if(err) {
 					callback(err);
@@ -508,9 +507,9 @@ var Auth = (function() {
 	 */
 	Auth.prototype.getAuthHeaders = function(callback) {
 		if(this.method == 'basic') {
-			if (!this.rest.options.tls) {
+			if (!this.client.options.tls) {
 				var msg = 'invalid use of Basic auth over non-TLS transport';
-				Logger.logAction(Logger.LOG_ERROR, 'Rest.getAuthHeaders()', msg );
+				Logger.logAction(Logger.LOG_ERROR, 'Rest.getAuthHeaders()', msg);
 				callback(new ErrorInfo(msg, 40103, 401));
 				return;
 			}
--- a/spec/rest/tls.test.js
+++ b/spec/rest/tls.test.js
@@ -51,7 +51,7 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
 	 *  A stats request is a good example of this.
 	 */
 	exports.basic_auth_no_tls_error = function(test) {
-		test.expect(5);
+		test.expect(6);
 
 		var rest,
 			keyStr = helper.getTestApp().keys[0].keyStr;
@@ -69,8 +69,9 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
 			},
 			function(cb) {
 				rest.stats(function(err, stats) {
-					test.ok(err, 'A stats request causes error with basic auth over https because it would have to send the secret key');
+					test.ok(err, 'A stats request causes error with basic auth over a non-tls connection because it would have to send the secret key');
 					test.equals(err.code, 40103, 'The error thrown is 40103');
+					test.ok(typeof(err.serverId) === 'undefined', 'The error did not come from a server');
 					cb();
 				});
 			}
--- a/spec/rest/tls.test.js
+++ b/spec/rest/tls.test.js
@@ -1,0 +1,87 @@
+"use strict";
+
+define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
+	var exports = {};
+
+	/*
+	 * Set up.
+	 */
+	exports.setup = function(test) {
+		test.expect(1);
+		helper.setupApp(function() {
+			var rest = helper.AblyRest();
+			test.ok(rest, 'Instantiated rest client');
+			test.done();
+		});
+	};
+
+	/*
+	 * tls:true initiates an https connection (RSC18)
+	 */
+	exports.tls_connection = function(test) {
+		test.expect(3);
+
+		var keyStr = helper.getTestApp().keys[0].keyStr,
+			rest = helper.AblyRest({ tls: true, key: keyStr, useTokenAuth: false });
+
+		async.parallel([
+			function(cb) {
+				rest.auth.requestToken(function(err, tokenDetails) {
+					test.ok(!err, 'A token request does not cause error with basic auth over tls');
+					cb();
+				});
+			},
+			function(cb) {
+				rest.stats(function(err, stats) {
+					test.ok(!err, 'A stats request does not cause error with basic auth over tls');
+					cb();
+				});
+			}
+		], function() {
+			test.equal(rest.baseUri().substr(0,5), "https", 'Requested token and stats over a tls connection');
+			test.done();
+		});
+	};
+
+	/*
+	 *  The combination of tls:false and basic auth results in error (see RSC18, RSA1).
+	 *  The error MUST NOT occur at instantiation, because we might need the Rest client 
+	 *  for things that don't require sending a key such as a token request.
+	 *  The error MUST occur on any request that tries to send a secret key over HTTP.
+	 *  A stats request is a good example of this.
+	 */
+	exports.basic_auth_no_tls_error = function(test) {
+		test.expect(5);
+
+		var rest,
+			keyStr = helper.getTestApp().keys[0].keyStr;
+
+		test.doesNotThrow(function() {
+			rest = helper.AblyRest({ tls: false, key: keyStr, useTokenAuth: false });
+		}, Error, 'Rest client does not throw error on instantiation, even if a secret key is specified along with tls:false');
+
+		async.parallel([
+			function(cb) {
+				rest.auth.requestToken(function(err, tokenDetails) {
+					test.ok(!err, 'A token request does not cause an error over a non-tls connection, because the secret key is not sent with the request');
+					cb();
+				});
+			},
+			function(cb) {
+				rest.stats(function(err, stats) {
+					test.ok(err, 'A stats request causes error with basic auth over https because it would have to send the secret key');
+					test.equals(err.code, 40103, 'The error thrown is 40103');
+					cb();
+				});
+			}
+		], function() {
+			test.equal(rest.baseUri().substr(0,5), "http:", 'Requested token and stats over a non-tls connection');
+			test.done();
+		});
+	};
+
+	return module.exports = helper.withTimeout(exports);
+});
+
+
+


common/lib/client/auth.js
```diff
@@ -476,8 +476,15 @@ var Auth = (function() {
 	 * based on the current auth parameters
 	 */
 	Auth.prototype.getAuthParams = function(callback) {
-		if(this.method == 'basic')
+		if(this.method == 'basic') {
+			if (!this.rest.options.tls) {
+				var msg = 'invalid use of Basic auth over non-TLS transport';
+				Logger.logAction(Logger.LOG_ERROR, 'Rest.getAuthParams()', msg );
```
mattheworiordan: Extra space after `msg`

common/lib/client/auth.js
```diff
@@ -476,8 +476,15 @@ var Auth = (function() {
...
+				callback(new ErrorInfo(msg, 40103, 401));
+				return;
+			}
 			callback(null, {key: this.key});
+		}
```
mattheworiordan: Everywhere else I believe we have the else on the same line so `} else`

common/lib/client/auth.js
```diff
@@ -494,6 +501,12 @@ var Auth = (function() {
 	 */
 	Auth.prototype.getAuthHeaders = function(callback) {
 		if(this.method == 'basic') {
+			if (!this.rest.options.tls) {
+				var msg = 'invalid use of Basic auth over non-TLS transport';
+				Logger.logAction(Logger.LOG_ERROR, 'Rest.getAuthHeaders()', msg );
```
mattheworiordan: I believe these are probably `LOG_FATAL` as it's a terminal condition i.e. the library can no longer be used. @paddybyers is that right?

spec/rest/tls.test.js
```diff
@@ -0,0 +1,118 @@
...
+		async.parallel([
+			function(cb) {
+				rest.auth.requestToken(function(err, tokenDetails) {
+					test.ok( tokenDetails.token, 'Verify that token is returned');
+					cb();
+				});
+			},
+			function(cb) {
+				rest.stats(function(err, stats) {
+					test.ok(err, 'Verify that an error was returned')
```
mattheworiordan: What has this to do with `tls` tests?  Why is this bundled into this test?

spec/rest/tls.test.js
```diff
@@ -0,0 +1,118 @@
...
+				}
+				currentTime = time;
+				test.ok(true, 'Obtained time');
+				test.done();
+			});
+		});
+	};
+
+	/*
+	 * tls:true initiates https connection
```
mattheworiordan: Can we include a reference to the spec ID it relates to?

spec/rest/tls.test.js
mattheworiordan: I am just not clear why you are testing a failing stats request?

spec/rest/tls.test.js
```diff
@@ -0,0 +1,118 @@
...
+			},
+			function(cb) {
+				rest.stats(function(err, stats) {
+					test.ok(stats, 'Verify that stats are returned');
+					cb(err);
+				});
+			}
+		], function(err) {
+			if(err)
+				test.ok(false, "Request returned an error: " + helper.displayError(err));
```
mattheworiordan: Instead `if(err)` why not `test.ok(!err, ...)`

spec/rest/tls.test.js
```diff
@@ -0,0 +1,118 @@
...
+			test.equal(rest.baseUri().substr(0,5), "http:", 'Requested token and stats fron non-tls endpoint');
+			test.done();
+		});
+	};
+
+	/*
+	 *  The combination of tls:false and basic auth results in error (see RSC18, RSA1).
+	 *  The error MUST NOT occur at instantiation, because we might need the Rest client 
+	 *  for things that don't require sending a key. The error MUST occur on any request
```
mattheworiordan: > we might need the Rest client for things that don't require sending a key.

Like what?

spec/rest/tls.test.js
```diff
@@ -0,0 +1,118 @@
+"use strict";
+
+define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
+	var currentTime, exports = {};
```
mattheworiordan: currentTime is not used in these tests so why do we request server time?

common/lib/client/auth.js
paddybyers: > I believe these are probably LOG_FATAL as it's a terminal condition i.e. the library can no longer be used. @paddybyers is that right?

No, we only have `LOG_ERROR`.

mattheworiordan: Ah, OK, my bad

spec/rest/tls.test.js
alex-georgiou: Quoting @SimonWoolf 

> (I believe currently neither rest nor realtime clients do a  tls/basic check on instantiation, only on use. In realtime's case at least, that's probably something we should change, to give the dev feedback earlier. For rest there's things you can do with a rest client with tls: false that don't require basic auth (time check, signed token requests) so I don't think we should check on init)

@mattheworiordan While I address the other issues you pointed out, let me know what you think about this.

mattheworiordan: Ah, that's a fair point, although I am not sure why anyone would explicitly ask for a client to not use TLS only to check the time, or for that matter to create a signed token request. Based on @SimonWoolf's feedback though leave as is, sorry for the mixed messages.

spec/rest/tls.test.js
alex-georgiou: > What has this to do with tls tests? Why is this bundled into this test?

I am quoting RSC18:

> (RSC18) If ClientOptions#tls is true, then all communication is over HTTPS. **If false, all communication is over HTTP** however Basic Auth over HTTP will raise an exception as private keys cannot be submitted over an insecure connection. See Auth below

I was testing the part in bold. My reasoning is that the case of `tls = false` should be tested to make sure that tls is not used when explicitly disabled. I thought I'd bundle it with other tls-related tests. Would you like me to move it to some other file or remove it?

> I am just not clear why you are testing a failing stats request?

Before starting to write these tests I made several attempts to force the client to send a secret key over unencrypted http. I was successful to do this using a stats request. Then I modified the code to make sure that this condition returns an error rather than sending the key unencrypted. Here I'm testing for this error. Did you have something else in mind?

spec/rest/tls.test.js
alex-georgiou: Fixed.

spec/rest/tls.test.js
alex-georgiou: Done.

common/lib/client/auth.js
alex-georgiou: Fixed.

common/lib/client/auth.js
alex-georgiou: Fixed.

spec/rest/tls.test.js
alex-georgiou: I thought about this myself but for some reason it didn't work for me without time. Thinking about it now, maybe it's because on this branch I did not have your fix on issue 205. I will calibrate my clock and try again but today I'm facing a different error when running these tests.

    Ably: PaginatedResource.get(): Unexpected error getting resource: err = {"statusCode":401,"code":40160,"message":"action not permitted, app = g8g-4A","serverId":"frontend.e821.eu-central-1-A.i-586ba0e5"}

Investigating.

alex-georgiou: Sorry for the confusion, I see now that this is the error you were asking about in your next comment. To be honest I'm not sure why I did that either. Will look at it again.

spec/rest/tls.test.js
alex-georgiou: OK looking at the code again I now remember what I did. My intention was to test that
* asking for stats over https using basic auth succeeds and
* asking for stats over http using basic auth fails
Unfortunately I mistakenly mixed the ordering of the tests. Silly mistake on my part, don't know how I missed this (and how it passed when I tested it before pushing).

mattheworiordan: It passed because the key `helper.getTestApp().keys[1].keyStr;` may not have had permissions to test for stats as it has less privileges, but that had nothing to do with the test you are executing unfortunately

alex-georgiou:
@mattheworiordan OK I did some major refactoring to the TLS tests. As I explained on the previous commit, I had mixed up some tests.

The logic here is that I specify a secret key and then test a) token requests and b) stats requests over a) tls and b) non-tls. The idea is that token requests should always succeed because they never pass the secret key, while stats requests should fail over non-tls because they would have to send the secret key.

I am now sending `helper.getTestApp().keys[0].keyStr;` while before I was sending `keys[1]`. It seems stats requests with the key `keys[0]` succeed over tls so I guess the key has the correct capabilities.

mattheworiordan:
Thanks, looks good, however the test suite fails now, see https://travis-ci.org/ably/ably-js/jobs/106957815. Do you know why?

alex-georgiou:
@mattheworiordan It fails on `auth.test.js`. I ran locally:

      grunt nodeunit --test spec/realtime/auth.test.js

and

      grunt nodeunit --test spec/restauth.test.js

and they both pass on `authbase` for me. My first guess was an issue with travis?

alex-georgiou:
I double-checked and the auth tests definitely pass `authsetup` and `authbaseX` for me at this branch. Only `auth_token_expires` fails as expected because your fix is not on this branch. Also, there are no pending code changes to push. I've seen CI errors due to connectivity issues before, that's why I didn't think much of it.

mattheworiordan:
Well @alex-georgiou your branch fails consistently, yet the recently merged PR https://travis-ci.org/ably/ably-js/jobs/107758228 builds and does not stop at that auth test.  Unfortunately this PR does not build, so there is something in this PR that causes the first test to fail.

alex-georgiou:
OK obviously travis is running different code than I am then, @mattheworiordan . Perhaps something to do with your recent patch. I'll have a look. Hang on while I'm context-switching from PHP... :-)

alex-georgiou:
OK I have merged the current upstream master into a sub-branch of my branch and strangely enough the realtime auth tests pass on the browser for me. 

While I investigate, I just wanted to let you know @mattheworiordan  that the closure compiler is failing, even on master branch:

I get this error every time I try to run the tests with nodeunit.
```
Running "closure-compiler:ably.js" (closure-compiler) task
Warning: Command failed: /bin/sh -c java -jar "/home/alexg/workspace/ably/ably-js/tools/closure-compiler/build/compiler.jar" --js "browser/static/ably.js" --js_output_file "/home/alexg/workspace/ably/ably-js/browser/static/ably.min.js" --compilation_level "SIMPLE_OPTIMIZATIONS"
Error: Unable to access jarfile /home/alexg/workspace/ably/ably-js/tools/closure-compiler/build/compiler.jar
 Use --force to continue.

Aborted due to warnings.
```

alex-georgiou:
For some reason `npm install` did not pull the `compiler.jar` file.

paddybyers:
> Unable to access jarfile /home/alexg/workspace/ably/ably-js/tools/closure-compiler/build/compiler.jar

Is it there?

Use `grunt compiler` to install the closure compiler if you haven't done so already.

paddybyers:
> For some reason npm install did not pull the compiler.jar file.

See above comment. AFAIK it's not available as an npm module.

alex-georgiou:
OK this is the first time I had to do this, perhaps it should be added to the instructions.

paddybyers:
> perhaps it should be added to the instructions.

https://github.com/ably/ably-js/issues/171#issuecomment-181366324

alex-georgiou:
Cool now I can reproduce the auth.test error! Fixing...

alex-georgiou:
OK it seems that with the recent merges something changed so that the `auth` object doesn't always have a reference to a `rest` object.

I added this simple fix 3f29b8d and then I inspected the network traffic while running the tests. I can't  find any case where the private key is leaked over http due to this fix, either in the rest or realtime tests.

SimonWoolf:
`this.rest` really shouldn't ever be null. @alex-georgiou could you push up a branch pointing to a commit that's failing due to this problem? I can't reproduce the problem even before the fix commit (eg with feaf392 checked out), and the commit shas that failed in travis are all orphaned and not fetchable from github.

alex-georgiou:
@SimonWoolf That's right, [like I said before](/ably/ably-js/pull/224#issuecomment-181334579), I couldn't reproduce the problem on my local machine either. feaf392 was passing the tests. I had to get somewhat "creative" with git until I managed to reproduce the problem that I saw on travis. See [this branch](/alex-georgiou/ably-js/tree/RSC18_ClientOptions_tls_rebase). I tried to apply my branch on to the latest changes because I assumed that's what travis did. Not sure, I might have done something stupid.

SimonWoolf:
@alex-georgiou OK, so the issue was that the latest master includes [a commit](https://github.com/ably/ably-js/commit/6c30f9816574bd53c1cd72a2ed477c845c7bf63a) that renames Auth#rest to Auth#client, but your commit still references rest. See the Auth constructor in the branch you just linked: [this.client = client](https://github.com/alex-georgiou/ably-js/blob/33a05b7524c65abf1226f8154524e7d7523de5da/common/lib/client/auth.js#L42).

Issue is replicable with a simple `git rebase master`, though your multimerge tree was admittedly creative :)

Incidentally, that means that with your fix commit, `this.rest && !this.rest.options.tls` is always false, so the two paths you added to auth.js both become noops. (Which makes it a bit worrying that your tests passed. Presumably it's because realtime returns the same error - would be nice if the tests could distinguish that, eg by spying on Http.get to ensure it isn't called. Tests are more useful as regression tests if they fail when you remove the code which solves the problem they're testing).

alex-georgiou:
> Issue is replicable with a simple git rebase master, though your multimerge tree was indeed creative :)

Thanks @SimonWoolf . I have now read a couple of articles on rebase and am not as scared of it as I was before :)

> Presumably it's because realtime returns the same error. Would be nice if the tests could distinguish that, eg by spying on Http.get to ensure it isn't called

You're quite right, fortunately it's quite simple to distinguish a server-side error

    Ably: PaginatedResource.get(): Unexpected error getting resource: err = {"statusCode":401,"code":40103,"message":"Invalid use of Basic authentication over non-TLS transport","serverId":"frontend.41f9.eu-central-1-A.i-5f0081e3"}

from a client-side error

    Ably: PaginatedResource.get(): Unexpected error getting resource: err = {"message":"invalid use of Basic auth over non-TLS transport","code":40103,"statusCode":401}

So now I'm checking for `err.serverId`, which I think is simpler. If you still think I should spy on Http.get let me know and I'll have a look.

alex-georgiou:
This CI build seems to be failing due to a timeout that I cannot reproduce. Closing and opening again to trigger a new build.

alex-georgiou:
Hmm... this time I get a timeout on a different test.

PR #225, opened 2016-01-31
Updated tests for spec

adamgiacomelli:
Work done:
1.)Missing tests 
 - Added fixtures for limit testing
 - Added missing tests for default & max limit
 - Added direction omitted test

2.)Various
- Fixed comment on query pagination forwards, 
- Renamed limit0 and limit1 tests to a more human readable name, fixed wrong direction specified on forwards test
- Tuned existing tests to the new fixtures

I have observed that while running the tests a message appears in the fixtures with current time/date and it caused my backwards tests to fail since it had 0 inbound/outbound messages with 0 data. When I test with an hour ago ending all tests pass.


mattheworiordan:
Thanks @adamgiacomelli, sorry for the delay in reviewing.

Over this looks good.  I do have a few comments:

* You may want to check your editor as the indentation is weird, we only use tabs in this repo.
* We typically provide references to the spec/test the test relates to as well as use as much of the copy from the 0.8 spec itself to make it clear what the test is doing.

adamgiacomelli:
Thank you for the review. 
I will go over the indentation and unify it to tabs - I had automatic indent detection on and it seemed to have gone haywire in some places. All the indentation comments are valid and I will fix them tonight,
I will check the try block and fix the declaration in the for loop.

I can put comments that point to the spec in every test header if that's alright? Would do so in the future for all tests I write (and all exisiting with missing spec references in the file)as well.

If there isn't an established way (I couldn't find an example) I would propose something along the lines of:
```
/**
 * Check default limit query param (backwards)
 * @spec : (RSC6b3) If omitted the limit defaults to the REST API default (100)
 */
```

mattheworiordan:
> If there isn't an established way (I couldn't find an example) I would propose something along the lines of:

Your suggestion looks great, thanks.

mattheworiordan:
This all looks really good, if you could fix my last comment I will merge in.

mattheworiordan:
@adamgiacomelli what is the status on this PR?

adamgiacomelli:
I have renamed the generic variable and as far as i'm concerned it can be merged - the failing test doesn't seem to have anything to do with the specs test (it says the publish_with_xhr_binary_transport fails - browser tests)

mattheworiordan:
Thanks @adamgiacomelli

PR #226, opened 2016-02-01
test(realtime): Testing echoMessages=true and echoMessages=false

alex-georgiou:
At first I was hoping that something could be done so I could somehow query the server and have it tell me that all pending messages up to now have been sent. I scanned the code and docs for the SYNC message and for the ping/HEARTBEAT mechanism. 

I thought I could send a message followed by a ping and then if I got the ping back but not the message, I could assume that the message was not echoed. Then I wouldn't have to stall the tests for an arbitrary amount of seconds.

When the above didn't work, I had a look at the implementations in other languages.
1. In .NET, there is a timer waiting for 10 seconds before asserting that the message was not returned.
2. In GO, there is a similar timer for 15 seconds.
3. In IOS, the wait is for 60 seconds, which sounds more correct given that other connection timeouts are also set to 60s, but slows down the tests.

I set my timeout to 15 seconds.

### DIFF
--- a/spec/realtime/message.test.js
+++ b/spec/realtime/message.test.js
@@ -137,8 +137,15 @@ define(['ably', 'shared_helper'], function(Ably, helper) {
 						return;
 					}
 
+					/* wait to see that event is not echoed */
+					var timeoutHandler = setTimeout(function() {
+						test.ok(true, 'No event received within a reasonable time');
+						closeAndFinish(test, realtime);
+					}, 15000);
+
 					/* subscribe to event */
 					rtChannel.subscribe('event0', function(msg) {
+						clearTimeout(timeoutHandler);
 						test.ok(false, 'Received event0 when it should not be echoed back');
 						closeAndFinish(test, realtime);
 					});
@@ -146,11 +153,6 @@ define(['ably', 'shared_helper'], function(Ably, helper) {
 					/* publish event */
 					rtChannel.publish('event0', testMsg);
 
-					/* wait to see that event is not echoed */
-					setTimeout(function() {
-						test.ok(true, 'No event received within a reasonable time');
-						closeAndFinish(test, realtime);
-					}, 15000);
 				});
 			});
 			monitorConnection(test, realtime);
--- a/spec/realtime/message.test.js
+++ b/spec/realtime/message.test.js
@@ -70,6 +70,96 @@ define(['ably', 'shared_helper'], function(Ably, helper) {
 		}
 	};
 
+	/*
+	 * Test that a message is sent back to the same realtime client
+	 * when echoMessages is true (RTC1a and RTL7f)
+	 */
+	exports.publishEchoTrue = function(test) {
+		test.expect(1);
+		try {
+			/* set up realtime */
+			var realtime = helper.AblyRealtime({ echoMessages: true });
+
+			/* connect and attach */
+			realtime.connection.on('connected', function() {
+				var testMsg = 'Hello world';
+				var rtChannel = realtime.channels.get('publishecho');
+				rtChannel.attach(function(err) {
+					if(err) {
+						test.ok(false, 'Attach failed with error: ' + displayError(err));
+						closeAndFinish(test, realtime);
+						return;
+					}
+
+					/* if event is not echoed within reasonable time, give up */
+					var timeoutHandler = setTimeout(function() {
+						test.ok(false, 'No event received within a reasonable time');
+						closeAndFinish(test, realtime);
+					}, 15000);
+
+					/* subscribe to event */
+					rtChannel.subscribe('event0', function(msg) {
+						clearTimeout(timeoutHandler);
+						test.ok(true, 'Received event0 as expected');
+						closeAndFinish(test, realtime);
+					});
+
+					/* publish event */
+					rtChannel.publish('event0', testMsg);
+
+				});
+			});
+			monitorConnection(test, realtime);
+		} catch(e) {
+			test.ok(false, 'Channel attach failed with exception: ' + e.stack);
+			closeAndFinish(test, realtime);
+		}
+	};
+
+	/*
+	 * Test that a message is not sent back to the same realtime client
+	 * when echoMessages is false (RTC1a and RTL7f)
+	 */
+	exports.publishEchoFalse = function(test) {
+		test.expect(1);
+		try {
+			/* set up realtime */
+			var realtime = helper.AblyRealtime({ echoMessages: false });
+
+			/* connect and attach */
+			realtime.connection.on('connected', function() {
+				var testMsg = 'Hello world';
+				var rtChannel = realtime.channels.get('publishecho');
+				rtChannel.attach(function(err) {
+					if(err) {
+						test.ok(false, 'Attach failed with error: ' + displayError(err));
+						closeAndFinish(test, realtime);
+						return;
+					}
+
+					/* subscribe to event */
+					rtChannel.subscribe('event0', function(msg) {
+						test.ok(false, 'Received event0 when it should not be echoed back');
+						closeAndFinish(test, realtime);
+					});
+
+					/* publish event */
+					rtChannel.publish('event0', testMsg);
+
+					/* wait to see that event is not echoed */
+					setTimeout(function() {
+						test.ok(true, 'No event received within a reasonable time');
+						closeAndFinish(test, realtime);
+					}, 15000);
+				});
+			});
+			monitorConnection(test, realtime);
+		} catch(e) {
+			test.ok(false, 'Channel attach failed with exception: ' + e.stack);
+			closeAndFinish(test, realtime);
+		}
+	};
+
 	exports.publishVariations = function(test) {
 		var testData = 'Some data';
 		var errorCallback = function(err){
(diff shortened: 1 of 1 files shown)


alex-georgiou:
I just thought of another solution that would avoid waiting for an arbitrary duration:
1. Realtime1 connects with `echoMessages = false`
2. Realtime2 connects
3. Realtime1 sends message1 to a channel
4. Immediately after, Realtime2 sends message2 to that same channel
5. Realtime1 waits to see if message2 arrives but not message1.

I'm not sure if this will work because I don't know if message ordering is guaranteed. Let me know if you'd like me to attempt this, but it would diverge from the tests I saw in the other languages.

mattheworiordan:
> I'm not sure if this will work because I don't know if message ordering is guaranteed

Yes it is for a channel in the same region, so that will work, that is definitely a better way of approaching it

alex-georgiou:
@mattheworiordan I have combined both tests into one so that there is no wait for an arbitrary amount of time.

I have removed some unnecessary tests as you suggested.

spec/realtime/message.test.js
```diff
@@ -70,6 +70,50 @@ define(['ably', 'shared_helper'], function(Ably, helper) {
...
+			rt2 = helper.AblyRealtime({ echoMessages: true }),
+
+			rt1Channel = rt1.channels.get('publishecho'),
+			rt2Channel = rt2.channels.get('publishecho'),
+
+			testMsg1 = 'Hello',
+			testMsg2 = 'World!';
+
+		monitorConnection(test, rt1);
+		monitorConnection(test, rt2);
```
mattheworiordan: BTW. I think the monitorConnection method supports `monitorConnection(test, rt1, rt2);`

spec/realtime/message.test.js
```diff
@@ -70,6 +70,50 @@ define(['ably', 'shared_helper'], function(Ably, helper) {
...
+	 * when echoMessages is false (RTC1a and RTL7f)
+	 *
+	 * Test that a message is sent back to the same realtime client
+	 * when echoMessages is true (RTC1a and RTL7f)
+	 */
+	exports.publishEcho = function(test) {
+		test.expect(3);
+
+		// set up two realtimes
+		var rt1 = helper.AblyRealtime({ echoMessages: false }),
```
mattheworiordan: Instead of rt1 and rt2, why not name them intuitively such as `realtimeNoEcho` and `realtimeEcho`?

spec/realtime/message.test.js
```diff
@@ -70,6 +70,50 @@ define(['ably', 'shared_helper'], function(Ably, helper) {
...
+			rt2 = helper.AblyRealtime({ echoMessages: true }),
+
+			rt1Channel = rt1.channels.get('publishecho'),
```
mattheworiordan: As above, this could be done with more intuitive naming

spec/realtime/message.test.js
```diff
@@ -70,6 +70,50 @@ define(['ably', 'shared_helper'], function(Ably, helper) {
...
+
+		monitorConnection(test, rt1);
+		monitorConnection(test, rt2);
+
+		// We expect to see testMsg2 on rt1 and testMsg1 on both rt1 and rt2
+		var expectedMessages = 3;
+
+		rt1Channel.subscribe('event0', function(msg) {
+			test.equal(msg.data, testMsg2, 'Received testMsg2 via rt1 instance');
+			if (!--expectedMessages)
```
mattheworiordan: I personally dislike checking the value of variables you are modifying in place with increment / decrement operators, from the Good Parts of Javascript:

> ++ and --
> The ++ (increment) and -- (decrement) operators have been known to contribute to bad code by encouraging excessive trickiness. They are second only to faulty architecture in enabling to viruses and other security menaces. There is a plusplus option that prohibits the use of these operators.****

It's also really tricky to know if the value is changed before or after evaluation so leads to unreadable code.  Why not put this check into a callback as it's shared by both subscribe callbacks anyway and simplify by first decrementing the value and then checking if all messages are received?

spec/realtime/message.test.js
```diff
@@ -70,6 +70,50 @@ define(['ably', 'shared_helper'], function(Ably, helper) {
...
+		});
+
+		rt2Channel.subscribe('event0', function(msg) {
+			test.ok( msg.data == testMsg1 || msg.data == testMsg2, 'Received testMsg1 or 2 via rt2 instance' );
+			if (!--expectedMessages)
+				closeAndFinish(test, [rt1, rt2]);
+		});
+
+		// publish testMsg1 on rt1 and testMsg2 on rt2
+		rt1Channel.publish('event0', testMsg1, function() {
```
mattheworiordan: This test is too brittle because rt2Channel.subscribe may attach and subscribe after rt1Channel has published it's message and thus never receive it.  You should wait until the attaches are finished or approach this differently.

alex-georgiou:
@mattheworiordan I have uploaded improved code.

I think this version is better because it also checks for messages arriving in the expected order.

There is a longer list of callbacks but hopefully with the comments I added it should be readable. Let me know.

spec/realtime/message.test.js
```diff
@@ -70,6 +70,68 @@ define(['ably', 'shared_helper'], function(Ably, helper) {
...
+				}
+				closeAndFinish(test, [realtimeNoEcho, realtimeEcho]);
+			}
+		}
+
+		// first subscribe to realtimeNoEcho's channel event0
+		realtimeNoEchoChannel.subscribe('event0', function(msg) {
+			receivedMessagesNoEcho.push(msg);
+			finishTest();
+		}, function() {
```
mattheworiordan: What is this second function?  The signature for subscribe is `subscribe(name, callback(msg))`, there is no third aargument?

alex-georgiou: Actually there is an optional third argument which is a callback:

      RealtimeChannel.prototype.subscribe = function(/* [event], listener, [callback] */) {

It allows me to do exactly what I wanted, i.e. to do stuff after the listener is subscribed to the channel.

mattheworiordan: Ok, that's an undocumented API and I expect may change in the future, not sure why that's even in there as it's rather unrelated to the subscribe method itself.  For clarity however, given the third callback is undocumented, I would prefer to simply use `channel.attach(function(err) { ... do somethign .. })` over the subscribe as it's a lot clearer what is happening.  Sorry once again, no idea why that argument exists.

SimonWoolf: > no idea why that argument exists.

FWIW you and Paddy have discussed this before -- see https://github.com/ably/ably-js/commit/1b151df5558339d8b2fd80bbadd2375d9a138b4a. Paddy argued strongly in favour of having a callback parameter (as opposed to throwing an exception) to notify a channel attach fail or a channel in a failed state, for a call which triggers an implicit attach.

mattheworiordan: But that is a different point entirely @SimonWoolf.  This is a second callback on a `subscribe` method which has in implicit attach, whereas that discussion was in relation to whether we throw an exception immediately or pass the exception to the callback. This callback is unrelated to that discussion as far as I can tell?

SimonWoolf: > that discussion was in relation to whether we throw an exception immediately or pass the exception to the callback. This callback is unrelated to that discussion as far as I can tell?

Sorry, I'm not following you? The callback that was being passed the error in that discussion was the second callback to `subscribe`, same as this one. There was never a suggestion that on an attach error we'd call the *first* callback (the onmessage event handler).

paddybyers: > This callback is unrelated to that discussion as far as I can tell?

The callback allows you to know whether or not any implicit attach succeeded or failed. It's not the same issue as the discussion on whether to throw or call the callback in the already-failed case.

SimonWoolf: > It's not the same issue as the discussion on whether to throw or call the callback in the already-failed case.

Apologies if I'm not being clear. Matt's comment was that he wasn't sure "why that argument exists". Since the outcome of the earlier discussion was that we should call the callback (rather than throw), that relies on that argument existing, hence why I linked to it.

alex-georgiou: @mattheworiordan OK I am trying to follow the discussion. I think I now understand the semantics of `attach` and `subscribe` after reading the docs and code more carefully. If I understand correctly `attach` is an event that fires once a connection has been established, and subscribing to a channel's event can cause an implicit attach if the channel is not already attached.

I think I got it now, will attempt again and let you know.

mattheworiordan: > Since the outcome of the earlier discussion was that we should call the callback (rather than throw), that relies on that argument existing, hence why I linked to it.

How does [`attach`](https://github.com/ably/ably-js/blob/master/common/lib/client/realtimechannel.js#L119-L145) rely on `subscribe`'s 2nd callback argument?

SimonWoolf: > How does `attach` rely on `subscribe`'s 2nd callback argument? 

The discussion linked was on a change in `RealtimeChannel.prototype.subscribe`...?

mattheworiordan: My bad, I am stupid, those pesky little plus signs are there to confuse people like me :)  I do think we should still avoid using the undocumented callback API though as it's likely to change and testing it simply obfuscates the code and is likely to break in future.  Sorry for the confusion.

alex-georgiou:
@mattheworiordan OK I believe I now call `attach` explicitly before subscribing. This should ensure correct execution order while avoiding the undocumented callback.

PR #227, opened 2016-02-02
enhance: Off removes all listeners for EventEmitter

mattheworiordan:
See https://github.com/ably/docs/issues/82

@paddybyers @SimonWoolf please review

### DIFF
--- a/common/lib/util/eventemitter.js
+++ b/common/lib/util/eventemitter.js
@@ -18,6 +18,31 @@ var EventEmitter = (function() {
 	}
 
 	/**
+	 * Remove listeners that match listener
+	 * @param targetListeners is an array of listener arrays or event objects with arrays of listeners
+	 * @param listener the listener to remove
+	 */
+	function removeListener(targetListeners, listener) {
+		var listeners, idx, eventName, targetListenersIndex;
+
+		for (targetListenersIndex = 0; targetListenersIndex < targetListeners.length; targetListenersIndex++) {
+			listeners = targetListeners[targetListenersIndex];
+			if (Utils.isArray(listeners)) {
+				while ((idx = Utils.arrIndexOf(listeners, listener)) !== -1) {
+					listeners.splice(idx, 1);
+				}
+			} else if (Utils.isObject(listeners)) {
+				/* events */
+				for (eventName in listeners) {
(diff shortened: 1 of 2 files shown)


mattheworiordan:
cc @tcard FYI in regards to the discussion at https://github.com/ably/ably-ios/pull/179/files#r51544655

common/lib/util/eventemitter.js
```diff
@@ -18,6 +18,31 @@ var EventEmitter = (function() {
...
+		for (targetListenersIndex = 0; targetListenersIndex < targetListeners.length; targetListenersIndex++) {
+			listeners = targetListeners[targetListenersIndex];
+			if (Utils.isArray(listeners)) {
+				while ((idx = Utils.arrIndexOf(listeners, listener)) !== -1) {
+					listeners.splice(idx, 1);
+				}
+			} else if (Utils.isObject(listeners)) {
+				/* events */
+				for (eventName in listeners) {
+					if (listeners.hasOwnProperty(eventName) && Utils.isArray(listeners[eventName])) {
```
paddybyers: This does not delete `listeners[eventName]` if it is empty after the remove operation.

paddybyers:
LGTM apart from the comment above.

Also, someone needs to take these tests: https://github.com/ably/realtime/blob/master/test/util/eventemitter.js and migrate them.

common/lib/util/eventemitter.js
mattheworiordan: Does it need to be deleted?

paddybyers: Yes of course. If an app for example continuously uses new randomly generated event names and then discards the listeners, the size of the `events` object will grow indefinitely. It's easy to re-add.

mattheworiordan: > Yes of course. If an app for example continuously uses new randomly generated event names and then discards the listeners, the size of the events object will grow indefinitely. It's easy to re-add.

Ah, good point, I will make that change.

mattheworiordan:
@paddybyers fixed with https://github.com/ably/ably-js/commit/4f50db9ae189f075b87c07e8e48fbec2f3ff122c

paddybyers:
> fixed with 4f50db9

LGTM

Issue #211, opened 2016-01-22
Multiple host mechanism to work around max-concurrent-connections limits?


SimonWoolf:
One of the problems I ran into with IE8 support is that: it turns out that when on connections that IE thinks aren't broadband (which I guess must sometimes be triggered running the tests in a vm..? who knows what heuristic they use), IE8 limits you to two max concurrent connections to the same host. Which is enough for one stable comet transport (one connection for recv, one for send and others). As soon as you add a second comet transport to the mix, everything becomes unreliable and flaky -- you can't rely on anything happening in bounded time.

The symptoms of this are annoyingly ambiguous -- the IE network tab seems to show 'not yet sent because of this limit' and 'server not yet replied' in the same way -- as "(pending)" -- so by appearances it seems like realtime just took 15 seconds to respond to a /send request. When actually AFAICT the send request was just waiting for 15s until a recv request gets a heartbeat and finishes, at which point it can jump in and actually execute.

This is a problem (a) for our tests, which on IE8 will be occasionally flaky and unreliable, and (b) for customers who for whatever reason have two ably connections happening at once who have users on narrowband connections on IE8.

The only test that was consistently failing because of this was the resume test, which [I've rewritten to do things more serially and only use one realtime connection](https://github.com/ably/ably-js/commit/0aa13154a78c4d271d08855080b06b76fd8032fb).

Matt suggests:

> Why not put each connection on  unique endpoint?
> So instead of sandbox-rest.ably.io, we can set up [1-9]-sandbox-rest.ably.io, and you have a global index that keeps increasing and then simply % 10
> If we have a helper to create connections, it could handle that automatically
> Setting up the DNS is piece of piss too, maybe just make it 20 unique domains to be safe?
> WDYT?

I thought it was a good idea, but that if we're going to put in that work, we could consider doing a mechanism that works in prod too (for customers who instantiate two realtime connections), rather than just for our tests.

In any case, decided it was outside the scope of https://github.com/ably/ably-js/pull/210 which at least gets one-connection-at-a-time working reliably in ie8, and the tests seem to mostly be okish. So adding this as a low-priority feature request to consider for the future.

paddybyers:
This does increase the risk that 1.rest.ably.io resolves to a different region from 2.rest.ably.io.

Nonetheless, I agree if we were going to do it we've do it for production. I think it would be implemented within the comet transport or the request primitive, rather than something outside that governs the host fed into the library `ClientOptions`.

[...]

PR #229, opened 2016-02-02
Ie8 10 test suite fixes

mattheworiordan:
This PR in response to discussion with @SimonWoolf on failing parts of the test suite cross-browser.

>  Had a look at some runs of browsers-all. Failures can be divided into:
> - Random http timeouts. Saucelabs internet connection seems to be occasionally flakey, and any of the VMs (inc firefox and chrome) sometimes get random timeouts with comet requests.

Ok, not sure we can do much about this.

> - Systematic saucelabs-only internet failures. Eg:
> - for IE9 and 9, the first attempt to set up an app (which happens to be the realtime auth tests) pretty much always times out, resulting in those tests all failing. Subsequent attempts succeed.

I hope I have fixed this by adding a retry into the app setup, 7a627c4

> - IE10 is almost the opposite: after 50 or so successful tests, starting usually somewhere the early channel tests the VM seems to stop being able to connect to ably in any way. All subsequent tests timeout.
>   (I've fired up local IE 9 and 10 virtual machines and double-checked that the test suite passes in both).

I've tried this in Saucelabs now with db56a1e and this appears to not be an issue now.  This also covers some of https://github.com/ably/ably-js/issues/211 although my commits are only aimed at fixing the tests.

> I spent some time trying to debug the IE10 one back when I did IE10, without success.
> Not really sure what to suggest tbh. Reliable cross-browser testing is something we're going to have to have eventually.(edited)
> OK -- I should stop darkly hinting that this is because saucelabs sucks, because I've just tried with testingbot, and got the same results, for both IE10 and IE8/9.
> - The IE10 problem (`network error 0x2ee4`) appears to be a bug in IE10 triggered by saucelabs/testingbot's use of a network proxy, which is why I'm not seeing it locally. Was reported to MS, closed as "fixed in latest version" (ie IE11) , which is unhelpful. (https://connect.microsoft.com/IE/feedback/details/852785/ie10-network-error-0x2ee4-https-proxy-ajax).

Ok, this may now be fixed in this PR

> The IE8/9 problem seems to only happen the first time the page is loaded after the VM is created. On subsequent page refeshes (even if cache cleared, 'always refresh from server' ticked etc.) it works fine; which is why I couldn't recreate it in my local VM.

As above, may now be fixed.

cc @SimonWoolf @paddybyers

### DIFF
--- a/common/lib/util/defaults.js
+++ b/common/lib/util/defaults.js
@@ -23,7 +23,7 @@ Defaults.apiVersion       = '0.8';
 
 Defaults.getHost = function(options, host, ws) {
 	if(ws)
-		host = host == options.restHost ? options.realtimeHost : host || options.realtimeHost;
+		host = (host == options.restHost ? options.realtimeHost : host) || options.realtimeHost;
 	else
 		host = host || options.restHost;
 
--- a/common/lib/util/defaults.js
+++ b/common/lib/util/defaults.js
@@ -23,7 +23,7 @@ Defaults.apiVersion       = '0.8';
 
 Defaults.getHost = function(options, host, ws) {
 	if(ws)
-		host = ((host == options.restHost) && options.realtimeHost) || host || options.realtimeHost;
+		host = host == options.restHost ? options.realtimeHost : host || options.realtimeHost;
 	else
 		host = host || options.restHost;
 
@@ -65,15 +65,13 @@ Defaults.normaliseOptions = function(options) {
 	if(!('queueMessages' in options))
 		options.queueMessages = true;
 
-	if(options.restHost) {
-		options.realtimeHost = options.realtimeHost || options.restHost;
-	} else {
-		var environment = (options.environment && String(options.environment).toLowerCase()) || Defaults.ENVIRONMENT,
-		production = !environment || (environment === 'production');
+	var environment = (options.environment && String(options.environment).toLowerCase()) || Defaults.ENVIRONMENT,
+	production = !environment || (environment === 'production');
+	options.fallbackHosts = production && !(options.restHost || options.realtimeHost) ? Defaults.FALLBACK_HOSTS : options.fallbackHosts;
+	if (!options.restHost)
 		options.restHost = production ? Defaults.REST_HOST : environment + '-' + Defaults.REST_HOST;
+	if (!options.realtimeHost)
 		options.realtimeHost = production ? Defaults.REALTIME_HOST : environment + '-' + Defaults.REALTIME_HOST;
-		options.fallbackHosts = production ? Defaults.FALLBACK_HOSTS : options.fallbackHosts;
-	}
 	options.port = options.port || Defaults.PORT;
 	options.tlsPort = options.tlsPort || Defaults.TLS_PORT;
 	if(!('tls' in options)) options.tls = true;
--- a/spec/common/modules/client_module.js
+++ b/spec/common/modules/client_module.js
@@ -34,7 +34,7 @@ define(['ably', 'globals', 'spec/common/modules/testapp_module'], function(Ably,
 		if (!options.restHost && ablyGlobals.environment) {
 			/* Use a new host name for each REST request to avoid old browser issues
 			   where HTTP connections are no longer available */
-			clientOptions.restHost = hostPrefixes[hostPrefixIndex] + "-" + ablyGlobals.environment + "-rest.ably.io";
+			clientOptions.restHost = hostPrefixes[hostPrefixIndex] + '-' + ablyGlobals.environment + '-rest.ably.io';
 			hostPrefixIndex = (hostPrefixIndex + 1) % hostPrefixes.length;
 		}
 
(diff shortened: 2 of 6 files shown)


common/lib/util/defaults.js
```diff
@@ -23,7 +23,7 @@ Defaults.apiVersion       = '0.8';
 
 Defaults.getHost = function(options, host, ws) {
 	if(ws)
-		host = ((host == options.restHost) && options.realtimeHost) || host || options.realtimeHost;
+		host = host == options.restHost ? options.realtimeHost : host || options.realtimeHost;
```
paddybyers: this needs parenthesis

spec/common/modules/client_module.js
```diff
@@ -28,6 +31,13 @@ define(['ably', 'globals', 'spec/common/modules/testapp_module'], function(Ably,
 			}
 		}
 
+		if (!options.restHost && ablyGlobals.environment) {
+			/* Use a new host name for each REST request to avoid old browser issues
+			   where HTTP connections are no longer available */
+			clientOptions.restHost = hostPrefixes[hostPrefixIndex] + "-" + ablyGlobals.environment + "-rest.ably.io";
```
paddybyers: single quotes pls

paddybyers:
LGTM but see comments.

spec/common/modules/client_module.js
mattheworiordan: Sorry, I keep forgetting...

common/lib/util/defaults.js
mattheworiordan: True...

SimonWoolf:
Result of running it through saucelabs on IE 8/9/10 on https://ci.ably.io/job/ably-js-browsers-min/472/console -- some good news, some bad

- Retrying app creation seems to have worked on IE9, but not IE8. Looks like on the problem is that on IE8 it times out rather than errors (so only caught by your 60s `withTimeout` helper), so this should be fixable by adding a timeout to `httpReqFunction`.

- The IE10 issue doesn't seem to be affected by cycling through hosts, sadly. (IE8 was the last one that limited you to 2 concurrent connections per host (and then only sometimes), so would have been odd if this caused issues on IE10 but not 8/9)

mattheworiordan:
> The IE10 issue doesn't seem to be affected by cycling through hosts, sadly. (IE8 was the last one that limited you to 2 concurrent connections per host (and then only sometimes), so would have been odd if this caused issues on IE10 but not 8/9)

Odd, I don't see the issue when I run it on Saucelabs with a manual test instance against my local server...  What do you think could be different, I saw very few failures on IE10 in fact.

SimonWoolf:
> Odd, I don't see the issue when I run it on Saucelabs with a manual test instance against my local server... What do you think could be different, I saw very few failures on IE10 in fact.

Weird. I've just tried with a manual instance and got the same result as the automated test https://ci.ably.io/job/ably-js-browsers-min/472/console -- everything after test 42ish fails.

The only thing I can see that my tests share with ci's that yours don't is that your tests are using ngrok; mine (and I think karma-saucelabs's) use sauceconnect proxy server. Which lends more evidence that it's the proxy bug mentioned [here](https://connect.microsoft.com/IE/feedback/details/852785/ie10-network-error-0x2ee4-https-proxy-ajax)

SimonWoolf:
Tried some likely looking options to sauceconnect (`--no-ssl-bump-domains sandbox-realtime.ably.io,sandbox-rest.ably.io,ably.io`, `--no-proxy-caching`), neither had any effect

mattheworiordan:
See my result below, not a perfect pass, but certainly well beyond test 42!

> The only thing I can see that my tests share with ci's that yours don't is that your tests are using ngrok; mine (and I think karma-saucelabs's) used the sauceconnect tunnel...

Mmm, sounds like we may need to not use their tunnel then :(

voila_capture 2016-02-03_01-19-30_am

mattheworiordan:
@SimonWoolf in theory we could use `ngrok` instead of a SauceLabs tunnel for SauceLabs, see https://www.npmjs.com/package/ngrok.  However, I expect this is a chunk of unnecessary work at this stage and we should simply put IE10, for now, in a list of browsers we'll come back to to ensure they pass in CI.  WDYT?

SimonWoolf:
@mattheworiordan Just as a sanity-check, I ran a test using ngrok, and --

sl

...every test after the 40th fails, just as with sauceconnect. So that wasn't the issue.

Bizarre. How on earth can there be a systematic difference between you and me both putting in an ngrok url and pressing 'start' with the same VM type?

> we should simply put IE10, for now, in a list of browsers we'll come back to to ensure they pass in CI. WDYT?

Yeah, I guess so. Anything which breaks IE10 will likely also break either IE11 or IE9, so we don't lose all that much.

mattheworiordan:
Just to complicate things further @SimonWoolf, I don't get that issue when using the SauceLabs tunnel against my local test server

voila_capture 2016-02-03_02-28-13_am

funkyboy:
I thought of updating this with the current situation in master.
Here's IE8 tests

screen shot 2018-03-01 at 12 33 50
screen shot 2018-03-01 at 12 34 21

funkyboy:
Much better situation with IE9
screen shot 2018-03-01 at 13 49 42

funkyboy:
And a few on IE10

screen shot 2018-03-01 at 14 21 23

SimonWoolf:
@funkyboy Thanks -- I look forward to your PR fixing all those bugs :slightly_smiling_face:

PR #231, opened 2016-02-09
Remove deprecated iFrame static files

mattheworiordan:
@paddybyers @SimonWoolf I assume we don't need these old files anymore? Please confirm


paddybyers:
Don't need until such time as we reinstate iframe transport.

PR #233, opened 2016-02-10
Presence action string

mattheworiordan:
Whilst working on the Realtime Presence documentation today, I found that the demo I was writing was not very intuitive, see https://jsbin.ably.io/oxituz/3/edit

The problem is simply that the action field is an integer value for the [`PresenceMessage.Action` enum like object](https://github.com/ably/ably-js/blob/master/common/lib/types/presencemessage.js#L14-L20), yet that enum is not available to the developer consuming the API, unlike pretty much all other client libraries we have implemented which expose the Enum or some constant to access the meaningful part of the enum.

I believe that in order to make our client library API easier for developers to use, we should change the `action` attribute of a PresenceMessage from an integer into a string when received (via REST or Realtime), and then convert back to an integer when published to Ably. I am sure we could come up with a far more robust way of doing this than this implementation, but as far as I can see this works, and I'd be happy to add an issue to improve this design later.

@paddybyers @SimonWoolf WDYT?

### DIFF
--- a/common/lib/client/realtimechannel.js
+++ b/common/lib/client/realtimechannel.js
@@ -264,7 +264,7 @@ var RealtimeChannel = (function() {
 		var msg = ProtocolMessage.fromValues({
 			action: actions.PRESENCE,
 			channel: this.name,
-			presence: [PresenceMessage.fromValuesWithNumericAction(presence)]
+			presence: [PresenceMessage.fromValues(presence)]
 		});
 		this.sendMessage(msg, callback);
 	};
--- a/common/lib/client/realtimechannel.js
+++ b/common/lib/client/realtimechannel.js
@@ -264,7 +264,7 @@ var RealtimeChannel = (function() {
 		var msg = ProtocolMessage.fromValues({
 			action: actions.PRESENCE,
 			channel: this.name,
-			presence: [PresenceMessage.fromValues(presence)]
+			presence: [PresenceMessage.fromValuesWithNumericAction(presence)]
 		});
 		this.sendMessage(msg, callback);
 	};
--- a/common/lib/client/realtimepresence.js
+++ b/common/lib/client/realtimepresence.js
@@ -1,6 +1,4 @@
 var RealtimePresence = (function() {
-	var presenceAction = PresenceMessage.Action;
-
 	var noop = function() {};
 
 	function memberKey(item) {
@@ -71,7 +69,7 @@ var RealtimePresence = (function() {
 		  action + 'ing; channel = ' + this.channel.name + ', client = ' + clientId || '(implicit) ' + getClientId(this));
 
 		var presence = PresenceMessage.fromValues({
-			action : presenceAction[action.toUpperCase()],
+			action : action,
 			data   : data
 		});
 		if (clientId) { presence.clientId = clientId; }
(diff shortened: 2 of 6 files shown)


paddybyers:
If we're going to do this then we should just do it properly. This is close but could be cleaner I think.

1) We say that `action` is a string value in the `PresenceMessage` type.
2) `PresenceMessage.fromValues()` expects to be given a string action, and returns a `PresenceMessage` with an `action` value.
3) All conversions to and from the numeric value should be at the point the message is packed and unpacked. This means it is in `toJSON()` (unconditionally assuming a string value and converting to the action value) and in `PresenceMessage.fromDecoded()` (which should only be passed a set of values obtained by unpacking (via msgpack or JSON) a wire value from the protocol or from rest, and so it should unconditionally assume a numeric value).
4) Conversion from the numeric to string value should be by looking up the value in a map without doing a `toUpperCase()` which is just unnecessary and gratuitously inefficient. Similarly there is no need to do a `toLowerCase()` anywhere.

mattheworiordan:
@paddybyers I agree with all of that, but I just think I have more important things to do right now. However, this is a bit of a blocker to readable demos, are you happy for me to merge and add an issue to be picked up a high priority item to address in nicer way?

SimonWoolf:
I don't mind hacking at this for a few mins this evening to free up Matt for more important stuff

mattheworiordan:
If you wouldn't mind @SimonWoolf I'd appreciate it, then at least there won't be the same of having my hacky code committed to the repo :)

paddybyers:
@SimonWoolf that would be great. And while you're at it, you can fix your own crime: https://github.com/ably/ably-js/blob/master/common/lib/client/realtimepresence.js#L74

:)

SimonWoolf:
Fixup commit: b9f7d5e
Branch with that squashed-down into Matt's commit: https://github.com/ably/ably-js/compare/presence-action-string2

mattheworiordan:
Looks good, thanks @SimonWoolf.  Out of interest, is `toJSON()` called when MsgPack encoding `PresenceMessages`?  If not, I'm not following how the action gets converted to a number over a binary transport.

SimonWoolf:
@mattheworiordan yep, https://github.com/ably/ably-js/blob/master/browser/lib/util/msgpack.js#L656-L657. PresenceMessage even takes advantage of msgpack and json-encoding call toJSON with different arguments to encode the data in different ways -- https://github.com/ably/ably-js/blob/master/common/lib/types/presencemessage.js#L38-L47

mattheworiordan:
Thanks @SimonWoolf, that makes sense, so it's not just magic :boom:

paddybyers:
> so it's not just magic

One man's magic is another man's hack :)

SimonWoolf:
Merged as a0dd60e

paddybyers:
@SimonWoolf: did you look at the associated changes to realtime tests?

SimonWoolf:
@paddybyers No, will do that now. (If you made a comment re that, I missed it, sorry :( )

SimonWoolf:
@paddybyers Looks like test/realtime/presence.js all pass with ably-js submodule up to date with no changes :) Don't think there are any realtime tests that look at the presence action.

paddybyers:
> Looks like test/realtime/presence.js all pass with ably-js submodule up to date with no changes :)

thanks

mattheworiordan:
@SimonWoolf can you bump the version and release this please?

SimonWoolf:
@mattheworiordan done

PR #234, opened 2016-02-15
Basic http test for version headers (rsc7a)

adamgiacomelli:
First try for this test - I created a new test file for it, I figure any additional tests regarding http requests could go here. 
Simple stuff, I replace the get&post functions with interceptors and test the headers there.
I call the minimal versions of the functions using the requests to test their responses.

If I call the presence function without arguments I get an error - noop() is not defined in presence.js.
I'll go through it again tomorrow, just creating the pull to get the ball rolling.

### DIFF
--- a/spec/rest/http.test.js
+++ b/spec/rest/http.test.js
@@ -1,0 +1,46 @@
+'use strict';
+
+define(['ably', 'shared_helper'], function(Ably, helper) {
+	var rest, exports = {},
+		Defaults = Ably.Rest.Defaults;
+
+	exports.setupHttp = function(test) {
+		test.expect(1);
+		helper.setupApp(function() {
+			rest = helper.AblyRest();
+			test.ok(true, 'App created');
+			test.done();
+		});
+	}
+
+	exports.apiVersionHeader = function(test) {
+
+		//Intercept get&post methods with test
+		Ably.Rest.Http.get_inner = Ably.Rest.Http.get;
+		Ably.Rest.Http.get = function (rest, path, headers, params, callback) {
+			test.ok(('X-Ably-Version' in headers), 'Verify version header exists');
+			test.equal(headers['X-Ably-Version'], Defaults.apiVersion, 'Verify current version number');
+			return this.get_inner(rest, path, headers, params, callback);
+		};
+
+		Ably.Rest.Http.post_inner = Ably.Rest.Http.post;
+		Ably.Rest.Http.post = function (rest, path, headers, body, params, callback) {
+			test.ok(('X-Ably-Version' in headers), 'Verify version header exists');
+			test.equal(headers['X-Ably-Version'], Defaults.apiVersion, 'Verify for current version number');
+			return this.post_inner(rest, path, headers, body, params, callback);
+		};
+
+		//Call all methods that use rest http calls
+		test.expect(8);
+
+		rest.auth.requestToken();
+		rest.time();
+		rest.stats();
+		rest.channels.get().publish();
+
+		test.done();
+
+	};
+
+	return module.exports = helper.withTimeout(exports);
+});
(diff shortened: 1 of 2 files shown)


mattheworiordan:
LGTM. @SimonWoolf @paddybyers want to comment before we merge?

spec/rest/http.test.js
```diff
@@ -0,0 +1,52 @@
...
+		rest.stats();
+
+		var channel = rest.channels.get('http_test_channel');
+		channel.publish('test', 'Testing http headers');
+		channel.presence.get();
+
+		test.done();
+	};
+
+	return module.exports = helper.withTimeout(exports);
```
mattheworiordan: This test unfortunately has side effects.  Every test run after this will be using the this post and get method instead of the original.  You need to ensure that the original methods are reinstated.

adamgiacomelli:
Made a simple addition to reset the intercepted functions - my only worry is that in the case of test failure the reset functions won't trigger. I will probably have to rewrite this one if that's a problem. I was thinking about setting/clearing the interceptors in the setup/teardown, but since I run test functions inside the intercepted functions I can't quite see that.

spec/rest/http.test.js
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	/**
+	 * Check presence of X-Ably-Version headers in get&post requests
+	 * @spec : (RSC7a)
+	 */
+	exports.apiVersionHeader = function(test) {
+
+		//Intercept get&post methods with test
+		Ably.Rest.Http.get_inner = Ably.Rest.Http.get;
```
SimonWoolf: Not sure why the original needs to be stored as a new attr on the Http object, why not just in a local variable?

SimonWoolf:
> my only worry is that in the case of test failure the reset functions won't trigger. I will probably have to rewrite this one if that's a problem

TBH I wouldn't worry about it too much -- I mean, if the test fails part way through then the test suite run's already failed. (AFAICT it shouldn't harm subsequent tests much anyway, probably just nodeunit logging a bunch of complaints about expectations running after test was finished)

adamgiacomelli:
@SimonWoolf I agree, no reason for originals to be stored as attributes on the object, fixed that one

spec/rest/http.test.js
```diff
@@ -0,0 +1,56 @@
...
+		rest.auth.requestToken();
+		rest.time();
+		rest.stats();
+
+		var channel = rest.channels.get('http_test_channel');
+		channel.publish('test', 'Testing http headers');
+		channel.presence.get();
+
+		//Clean interceptors from get&post methods
+		Ably.Rest.Http.get = get_inner;
```
mattheworiordan: If you're worried this won't get called, why not a try catch block?

paddybyers: This test ends (with `test.done()`) while there are requests still in progress, because it doesn't wait for callbacks/responses from the REST calls it makes. Is that an issue? I would prefer that it waits, so there is no other activity going on when later tests run.

In fact instead of that, why do you even need to call `post_inner()` and `get_inner()` ? It serves no purposes in the test.

adamgiacomelli:
@paddybyers I've noticed that in some of the tests - the tests don't fail, although they basically stopp before the expected assertions are made... can't quite figure this one out.

As far as post_inner() and get_inner() are concerned - you're right - I figured it would be good to include them, so the functions still return the same thing as before, but I see that's unnecessary in this case.

@mattheworiordan i think a try-catch block would only assure that the resets get called in case something goes "exception"ally wrong - the test cases don't raise exceptions

As far as i'm concerned I'm happy with this - if you are.

paddybyers:
LGTM

SimonWoolf:
Merged as 0078c4d

PR #239, opened 2016-02-24
RTN14a spec test update

adamgiacomelli:
Added errorReason check to invalid_cred_failure - as far as i could gather from the unified spec that's all thats needed.

### DIFF
--- a/spec/realtime/failure.test.js
+++ b/spec/realtime/failure.test.js
@@ -25,14 +25,15 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
 	 * Connect with invalid credentials on various transports; connection state should be 'failed'
 	 */
 	exports.invalid_cred_failure = function(test) {
-		test.expect((availableTransports.length + 1)*2);
+		test.expect((availableTransports.length + 1)*3);
 		try {
 			var failure_test = function(transports) {
 				return function(cb) {
 					var realtime = helper.AblyRealtime({key: "this.is:wrong", transports: transports});
-					realtime.connection.on('failed', function() {
+					realtime.connection.on('failed', function(connectionStateChange) {
 						test.ok(true, 'connection state for ' + transports + ' was failed, as expected');
-						test.notEqual(realtime.connection.errorReason, null, 'error reason was not set');
+						test.notEqual(realtime.connection.errorReason, null, 'error reason was not set on connection');
+						test.notEqual(connectionStateChange.reason, null, 'error reason was not set on connectionStateChange');
 						cb(null, realtime);
 					});
 					realtime.connection.on('disconnected', function() {
--- a/spec/realtime/failure.test.js
+++ b/spec/realtime/failure.test.js
@@ -32,8 +32,8 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
 					var realtime = helper.AblyRealtime({key: "this.is:wrong", transports: transports});
 					realtime.connection.on('failed', function(connectionStateChange) {
 						test.ok(true, 'connection state for ' + transports + ' was failed, as expected');
-						test.equal(realtime.connection.errorReason.constructor.name, 'ErrorInfo', 'error reason set on connection was not of type ErrorInfo.');
-						test.equal(connectionStateChange.reason.constructor.name, 'ErrorInfo', 'error reason set on connectionStateChange was not of type ErrorInfo');
+						test.equal(realtime.connection.errorReason.code, '40400', 'wrong error reason code on connection.');
+						test.equal(connectionStateChange.reason.code, '40400', 'wrong error reason code on connectionStateChange');
 						test.deepEqual(connectionStateChange.reason, realtime.connection.errorReason, 'error reason was not equally set on connection and connectionStateChange');
 						cb(null, realtime);
 					});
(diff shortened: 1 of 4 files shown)


spec/realtime/failure.test.js
```diff
@@ -25,13 +25,14 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
...
 	exports.invalid_cred_failure = function(test) {
-		test.expect(availableTransports.length + 1);
+		test.expect((availableTransports.length + 1)*2);
 		try {
 			var failure_test = function(transports) {
 				return function(cb) {
 					var realtime = helper.AblyRealtime({key: "this.is:wrong", transports: transports});
 					realtime.connection.on('failed', function() {
 						test.ok(true, 'connection state for ' + transports + ' was failed, as expected');
+						test.notEqual(realtime.connection.errorReason, null, 'error reason was not set');
```
mattheworiordan: Isn't this testing the wrong thing?  The spec says "and the Connection#errorReason will be set on the Connection object as well as the emitted ConnectionStateChange", so I would expect that you would add an argument to the function such as `realtime.connection.on('failed', function(connectionStateChange) {` and check that the error reason is present as well present on the errorReason attribute.  This test, as far as I can tell, proves that the spec is not curently met?

adamgiacomelli:
Added an additional test for the connectionStateChange - I hope i'm getting this right - I test if the errorReason is set on the connection and if it's set on the connectionStateChange.

spec/realtime/failure.test.js
```diff
@@ -25,13 +25,15 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
...
+		test.expect((availableTransports.length + 1)*3);
 		try {
 			var failure_test = function(transports) {
 				return function(cb) {
 					var realtime = helper.AblyRealtime({key: "this.is:wrong", transports: transports});
-					realtime.connection.on('failed', function() {
+					realtime.connection.on('failed', function(connectionStateChange) {
 						test.ok(true, 'connection state for ' + transports + ' was failed, as expected');
+						test.notEqual(realtime.connection.errorReason, null, 'error reason was not set on connection');
+						test.notEqual(connectionStateChange.reason, null, 'error reason was not set on connectionStateChange');
```
SimonWoolf: @adamgiacomelli looks fine as far as it goes, but would be even better to
- test that the `reason` is actually an ErrorInfo object as it should be and shows the expected code for that kind of failure, eg `test.equal(reason.code, 40400)`, rather than just not being null, and
- test that the `connection.errorReason` and the `connectionStateChange.reason` are `deepEqual` to each other, rather than just testing that they both exist

adamgiacomelli:
@SimonWoolf thank you for the comment - I updated the tests - and so it happens that the errorReason is infact not of ErrorInfo type - it is just an object.

mattheworiordan:
>  updated the tests - and so it happens that the errorReason is infact not of ErrorInfo type - it is just an object.

@adamgiacomelli so I assume you will need to implement a fix then?

adamgiacomelli:
@mattheworiordan @SimonWoolf added a fix, I have some questions though - the connectionStateChange object has the .reason attribute which I'm not entirely certain if its the same thing as an errorReason or its used for something else as well (a connection state change could have a non-error reason) - I test for errorReason specifically and leave the .reason alone - should I merge the properties and use only one of them or is this ok?

SimonWoolf:
@adamgiacomelli Thanks for the bug report!

I'm not sure I agree with your proposed fixes though, I'm afraid:

- Per the spec, a connectionStateChange shouldn't have an `errorReason` as well as a `reason`, just a `reason`, which will be an ErrorInfo (RTN4f). (Re your question: no, a reason is only set if it's an errorReason, and connectionStateChange.reason if present should always be equal to connection.errorReason).
- I don't think `enactStateChange` is the right place to construct the ErrorInfo; I reckon it makes more sense to do it when we first receive the error, and then pass that around.

For simplicity I've pushed up [a branch](https://github.com/ably/ably-js/compare/errorinfo?expand=1) with the fix I'd make on it (specifically [this commit](https://github.com/ably/ably-js/commit/c4c725a3a1db5fb25fdb31bd34d8d2dbd4138a34)), which happily seems to make your tests pass :)

SimonWoolf:
(Also `constructor.name` is non-standard - in particular, IE doesn't support it. TBH I don't even know if there is a reliable way of getting constructor name that works everywhere... Given this is javascript a duck-typed test is probably fine, e.g. test that the reason.code is 40400)

adamgiacomelli:
@SimonWoolf lovely - that's what I was going for - wasn't quite sure about the errorReason/reason since i was looking at this:
https://github.com/ably/ably-js/blob/master/common/lib/transport/connectionmanager.js#L607

How can I merge your branch to mine? (never had this situation before, I'd like to add your fix - not sure how to merge this)

@paddybyers No reason, should be unnecessary to construct anything after Simon's fix.

I can do the duck-typed test if thats ok - just testing the reason.code.

mattheworiordan:
@adamgiacomelli simply [`cherry-pick`](https://git-scm.com/docs/git-cherry-pick) @SimonWoolf's commits onto yours.  You will need to add this repo as a remote such as `git remote add ably git@github.com:ably/ably-js.git` and then `git fetch ably` so that you have those commits locally in your git repo before cherry picking.

adamgiacomelli:
@mattheworiordan cherry-picked @SimonWoolf 's fixes and updated the test to test for reason.code - if that looks ok, I'm happy with this one.

mattheworiordan:
LGTM

SimonWoolf:
Merged as 04e48aa

PR #240, opened 2016-02-26
Implement latest encryption spec

SimonWoolf:
Implement latest encryption spec per https://github.com/ably/docs/pull/89.


browser/lib/util/crypto.js
```diff
@@ -98,84 +115,102 @@ var Crypto = (function() {
...
+		var key;
+		/* Backward compatibility */
+		if((typeof(params) === 'function') || (typeof(params) === 'string')) {
+			Logger.deprecated('Crypto.getDefaultParams(key, callback)', 'Crypto.getDefaultParams({key: key})');
+			if(typeof(params) === 'function') {
+				Crypto.generateRandomKey(function(key) {
+					params(null, Crypto.getDefaultParams({key: key}));
+				})
+			} else {
+				callback = arguments[1];
```
mattheworiordan: You should probably check that this second argument is in fact a callback before calling it, else we will simply get a runtime error in our library without any suitable error message for the dev.

browser/lib/util/crypto.js
```diff
@@ -98,84 +115,102 @@ var Crypto = (function() {
...
+		} else {
+			key = BufferUtils.toWordArray(params.key); // Expect key to be an Array, ArrayBuffer, or WordArray at this point
+		}
+
+		var cipherParams = new CipherParams();
+		cipherParams.key = key;
+		cipherParams.algorithm = params.algorithm || DEFAULT_ALGORITHM;
+		cipherParams.keyLength = key.words.length * (4 * 8);
+		cipherParams.mode = params.mode || DEFAULT_MODE;
+		validateCipherParams(cipherParams);
```
mattheworiordan: What happens if a keyLength is provided explicitly?  I assume it's just ignored as the keyLength is derived from the key. In Ruby I've checked for inconsistencies and indicated an error. Not sure it's necessary, but just a comment

browser/lib/util/crypto.js
```diff
@@ -98,84 +115,102 @@ var Crypto = (function() {
...
+	Crypto.getCipher = function(params, callback) {
+		var cipherParams = (params instanceof CipherParams) ?
+		                   params :
+		                   Crypto.getDefaultParams(params);
+
+		if(params.iv) {
+			callback(null, cipherParams, new CBCCipher(cipherParams, params.iv));
+		} else {
+			generateRandom(DEFAULT_BLOCKLENGTH, function(err, iv) {
+				callback(null, cipherParams, new CBCCipher(cipherParams, iv));
```
mattheworiordan: @SimonWoolf I've not read this thoroughly, but this looks like the Cipher has an embedded `iv` and thus uses the same `iv` for every single request.  This is not the intention, the `iv` should be generated uniquely for every message encrypted, unless an `iv` is explicitly provided for testing purposes.

mattheworiordan:
Few comments, but otherwise LVGTM

browser/lib/util/crypto.js
SimonWoolf: :+1:

browser/lib/util/crypto.js
SimonWoolf: Currently just ignored, didn't think anyone'd bother passing one in. I can add a check for inconsistency easily enough if you think it'd be a useful sanity-check :+1:

mattheworiordan: I'm not sure it's going to be used that much in practice, but it is a supported and document `CipherParam` so I think a simple check is sensible

browser/lib/util/crypto.js
SimonWoolf: > uses the same iv for every single request

So this PR doesn't change how the iv works, I just moved the generation step from `getDefaultParams` to `getCipher`.

At the moment, the code is doing something clever that I don't understand: passing the generated iv to createEncryptor, then each new message iv is the result of the cipher (using the previous iv) encrypting an empty block (and is prepended to the ciphertext, but not actually passed to CryptoJS #process).

I'm wary of changing code I don't understand (especially crypto code), and you and Paddy both know way more about encryption than I do, so I'd appreciate @paddybyers input on this.

(also currently encrypting a message is synchronous, and requiring it to call an async function in the middle to generate an iv would need fairly deep surgery)

mattheworiordan: @paddybyers please comment then because perhaps I have misunderstood what this library is doing then when reviewing this PR.

SimonWoolf:
Added some more commits:

- 6c39bad: Use a new random iv for each new message (by generating a random block and encrypting it, which will return (and set as the iv for the next block) `AES(randomBlock xor lastCipherText`) (which theoretically is just as random as `randomBlock`, and means we avoid creating a new cipher object for every message))
- 723a00d Add the extra checks suggested by Matt to getDefaultParams
- 806496b Add a short crypto blurb to the readme
- More controversially: 95c95ca (generateRandom is now purely internal so we can remove the fake-async api) and 3d2ea87 (make getCipher synchronous).

Why the last two? Because since channels.get is sync (which presumably is not something we want to change at this stage), having getCipher be async means that, for node, anything you do in the same cycle of the event loop that uses encryption will fail, as the cipher will not have initialized. This isn't a problem for subscribes (as by the time the channel attaches it's fine), but is for 'immediate' publishes, as the encoding of the message happens synchronously with the `publish` call. The simplest solution seems to me to just be to use the sync version of `randomBytes` (which in practice should never fail for lack of entropy except in the first few milliseconds after boot, see https://cs.fit.edu/code/projects/teamcomefrom/repository/revisions/f68a116c3c061151ab120f10fdc9230192d6b157/diff/ -- it uses OpenSSL's RAND_bytes, which unlike /dev/random, doesn't block on depletion after it's initially seeded), which is what I've done here.

Note that this doesn't affect `generateRandomKey`, which still presents an async api (and in node, still uses the async version of `randomBytes`). But even so, I know @paddybyers has expressed reservations about this sort of thing, so I'll offer [this commit as an alternative](https://github.com/ably/ably-js/commit/3f13361#diff-64cdbbf2e2596c594549b7cfbdb408b0R81) (having `publish` delay until the encryption is initialized - I'm not a fan of this as it's horribly hacky).

mattheworiordan:
LGTM bar one comment, I think we should merge, tag and release ASAP as I am pushing up docs now with this new API

mattheworiordan:
@paddybyers you ok with this?

paddybyers:
> you ok with this?

Just one comment above but otherwise LGTM

SimonWoolf:
Squashed down and merged as 0d32938

PR #241, opened 2016-02-26
Upgrade when closing

SimonWoolf:
Fix to minor race condition: if connection was closing/closed when an upgrade happens, `activateTransport` would return without doing anything, but `scheduleTransportActivation` assumed there was a new active transport and tries to sync it, set the state, etc, leading to the occasional `Unexpected exception in transport.send()` exception happening right after a test `closeAndFinish`'s.

Also syntax fix to a closeAndFinish call that was leaving a realtime connection open.

### DIFF
--- a/common/lib/transport/connectionmanager.js
+++ b/common/lib/transport/connectionmanager.js
@@ -378,26 +378,28 @@ var ConnectionManager = (function() {
 
 			/* make this the active transport */
 			Logger.logAction(Logger.LOG_MINOR, 'ConnectionManager.scheduleTransportActivation()', 'Activating transport; transport = ' + transport);
-			self.activateTransport(transport, self.connectionKey, self.connectionSerial, self.connectionId);
+			/* if activateTransport returns that it has not done anything (eg because the connection is closing), don't bother syncing */
+			if(self.activateTransport(transport, self.connectionKey, self.connectionSerial, self.connectionId)) {
+				Logger.logAction(Logger.LOG_MINOR, 'ConnectionManager.scheduleTransportActivation()', 'Syncing transport; transport = ' + transport);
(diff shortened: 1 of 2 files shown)


common/lib/transport/connectionmanager.js
```diff
@@ -378,26 +378,28 @@ var ConnectionManager = (function() {
...
-				if(err) {
-					Logger.logAction(Logger.LOG_ERROR, 'ConnectionManager.scheduleTransportActivation()', 'Unexpected error attempting to sync transport; transport = ' + transport + '; err = ' + err);
-					return;
-				}
-				Logger.logAction(Logger.LOG_MINOR, 'ConnectionManager.scheduleTransportActivation()', 'sync successful upgraded transport; transport = ' + transport + '; connectionSerial = ' + connectionSerial + '; connectionId = ' + connectionId);
+			/* if activateTransport returns that it has not done anything (eg because the connection is closing), don't bother syncing */
+			if(self.activateTransport(transport, self.connectionKey, self.connectionSerial, self.connectionId)) {
+				Logger.logAction(Logger.LOG_MINOR, 'ConnectionManager.scheduleTransportActivation()', 'Syncing transport; transport = ' + transport);
+				self.sync(transport, function(err, connectionSerial, connectionId) {
+					if(err) {
```
paddybyers: This indentation looks wrong

paddybyers:
LGTM

common/lib/transport/connectionmanager.js
SimonWoolf: More specifically? Looks right to me

paddybyers: ok, my mistake

PR #249, opened 2016-03-31
Fixes for a couple of bugs triggered by closing the connection during a transport upgrade

SimonWoolf:
Fixes for a couple bugs turned up in the course of making the longrunning qos tests, triggered by the connection being closed at certain times during a transport upgrade


paddybyers:
LGTM

Issue #247, opened 2016-03-22
Inconsistent connection state following an 80006 error


SimonWoolf:
From a sandbox I'd left running last night:

- ws connection was disconnected presumably due to a deploy
- reconnected as xhr (succeeds but with `Unable to recover connection`)
- attaches to a channel over xhr
- starts ws upgrade, doesn't succeed (`transport.inactive - WebSocketTransport`)
- 6 seconds pass
- channel attach gets an ERROR action reply with `{"statusCode":404,"code":80006,"message":"Unable to recover channel (messages expired)"}`
- `transport.inactive - WebSocketTransport`
- XHRTransport is disposed
- connection state is `connected`, but there are no connected transports

```
12741773ms - state: connected
12741774ms - state: channel event attaching
12742281ms - state: connectionmanager - transport.pending - WebSocketTransport
12742285ms - state: connectionmanager - transport.inactive - WebSocketTransport
12749025ms - state: connectionmanager - transport.active - WebSocketTransport
12749029ms - state: XHRTransport - disconnected
12749030ms - state: connectionmanager - transport.inactive - XHRTransport
12749033ms - state: attach callback called, err = {"statusCode":404,"code":80006,"message":"Unable to recover channel (messages expired)"}
12749038ms - state: XHRTransport - disposed
```

Full log:
```
 ably-sandbox.js:170:3
Ably: WebSocketTransport.onWsData(): data received; length = 12; type = string ably.js:4199:32
Ably: Transport.onProtocolMessage(): received; [ProtocolMessage; action=HEARTBEAT] ably.js:4199:32
Ably: Transport.onProtocolMessage(): heartbeat; connectionKey = b510e!i9Dd_8SFltkcZqtS-383b510e ably.js:4199:32
WebSocketTransport - heartbeat ably-sandbox.js:170:3
Ably: WebSocketTransport.onWsClose(): closed WebSocket; wasClean = false; code = 1006 ably.js:4199:32
Ably: Transport.onDisconnect(): err = undefined ably.js:4199:32
WebSocketTransport - disconnected ably-sandbox.js:170:3
Ably: ConnectionManager.deactivateTransport(): transport = WebSocketTransport; uri=wss://sandbox-realtime.ably.io:443/?key=mz3G9w.G3yQww:Rjo9T6rLAKaK64wu&upgrade=b510e!i9Dd_8SFltkcZqtS-383b510e&format=json&v=0.8 ably.js:4199:32
[...]

PR #255, opened 2016-04-05
Don’t activate a transport that isn’t connected

SimonWoolf:
Fixes https://github.com/ably/ably-js/issues/247

### DIFF
--- a/common/lib/transport/connectionmanager.js
+++ b/common/lib/transport/connectionmanager.js
@@ -425,6 +425,13 @@ var ConnectionManager = (function() {
 		/* remove this transport from pending transports */
 		Utils.arrDeleteValue(this.pendingTransports, transport);
 
+		/* if the transport is not connected (eg because it failed during a
+		 * scheduleTransportActivation#onceNoPending wait) then don't activate it */
+		if(!transport.isConnected) {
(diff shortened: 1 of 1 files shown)


common/lib/transport/connectionmanager.js
```diff
@@ -422,6 +422,14 @@ var ConnectionManager = (function() {
 		if(existingState == this.states.closing || existingState == this.states.closed)
 			return false;
 
+		/* if the transport is not connected (eg because it failed during a
+		 * scheduleTransportActivation#onceNoPending wait) then don't activate it */
+		if(!transport.isConnected) {
+			Logger.logAction(Logger.LOG_MINOR, 'ConnectionManager.activateTransport()', 'Declining to activate transport ' + transport + ' since it appears to no longer be connected');
+			Utils.arrDeleteValue(this.pendingTransports, transport);
```
paddybyers: I think I would have reversed the order of these things; ie unconditionally remove from the `pendingTransports` list, then  decline to activate + return if not connected.

paddybyers:
LGTM but see comment above.

common/lib/transport/connectionmanager.js
SimonWoolf: good point

SimonWoolf:
Merged as c2a4912

Issue #252, opened 2016-04-01
Should not reject presence enter if clientId is unknown


mattheworiordan:
If the client library does not know if a clientId may be determined following auth, then it should not raise an exception when performing presence actions as it has to assume that it's possible following auth a clientId will be configured, and thus, let the server handle the request if in valid.

See https://jsbin.ably.io/ivezaj/2/edit

Error:
```
ably.js:8818 Uncaught [ErrorInfo: clientId must be specified to enter a presence channel; statusCode=400; code=40012]**
```

SimonWoolf:
Fixed by https://github.com/ably/ably-js/issues/256

PR #256, opened 2016-04-05
Don't reject a presence enter unless definitely anonymous

SimonWoolf:
Fixes https://github.com/ably/ably-js/issues/252

### DIFF
--- a/common/lib/client/realtimepresence.js
+++ b/common/lib/client/realtimepresence.js
@@ -9,6 +9,15 @@ var RealtimePresence = (function() {
 		return realtimePresence.channel.realtime.auth.clientId;
 	}
 
+	function isAnonymous(realtimePresence) {
+		var realtime = realtimePresence.channel.realtime;
+		/* If not currently connected, we can't assume that we're an anonymous
+		 * client, as realtime may inform us of our clientId in the CONNECTED
+		 * message. So assume we're not anonymous and leave it to realtime to
+		 * return an error if we are */
+		return !realtime.auth.clientId && realtime.connection.state === 'connected';
+	}
+
 	function waitAttached(channel, callback, action) {
 		switch(channel.state) {
 			case 'attached':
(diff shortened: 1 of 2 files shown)


common/lib/client/realtimepresence.js
```diff
@@ -9,6 +9,15 @@ var RealtimePresence = (function() {
 		return realtimePresence.channel.realtime.auth.clientId;
 	}
 
+	function definitelyAnonymous(realtimePresence) {
+		var realtime = realtimePresence.channel.realtime;
+		// If not currently connected, we can't assume that we're an anonymous
```
paddybyers: Comment style?

I would have preferred `isAnonymous()` but whatever.

paddybyers:
LGTM but see comment above.

SimonWoolf:
Merged as d5e6fa1

common/lib/client/realtimepresence.js
```diff
@@ -9,6 +9,15 @@ var RealtimePresence = (function() {
 		return realtimePresence.channel.realtime.auth.clientId;
 	}
 
+	function isAnonymous(realtimePresence) {
+		var realtime = realtimePresence.channel.realtime;
+		/* If not currently connected, we can't assume that we're an anonymous
+		 * client, as realtime may inform us of our clientId in the CONNECTED
+		 * message. So assume we're not anonymous and leave it to realtime to
+		 * return an error if we are */
+		return !realtime.auth.clientId && realtime.connection.state === 'connected';
```
mattheworiordan: This does work, but I am not sure I agree entirely with this logic.  If for example, a user has provided a `clientId` in the `ClientOptions`, but not yet connected, then we can assume the `clientId` is specified anyway because if the `CONNECTED` message comes back with a different `clientId`, the connection fails at that point anyway.  I think this will do, but it's not 100% right.

SimonWoolf:
> If for example, a user has provided a clientId in the ClientOptions, but not yet connected, then...

...then `isAnonymous` will (correctly) return false. If someone specifies `auth.clientId` in clientOptions, then [auth.clientId will be set on client lib init](https://github.com/ably/ably-js/blob/e6319f9dc8d5ea802cdbfb6b24b09175231f7300/common/lib/client/auth.js#L47-L50), so `!realtime.auth.clientId` will be false. (so the function doesn't bother checking the connection state due to short-circuiting)

The function might return false even if the user will end up being anonymous after all (which is fine, as then realtime can reject the ENTER), but it should never return true when the user might not be. (I originally called it `isDefinitelyAnonymous` to make that clear, but Paddy requested `isAnonymous`, see other thread)

common/lib/client/realtimepresence.js
SimonWoolf: > If for example, a user has provided a clientId in the ClientOptions, but not yet connected, then...

...then `isAnonymous` will (correctly) return false. If someone specifies `auth.clientId` in clientOptions, then [auth.clientId will be set on client lib init](https://github.com/ably/ably-js/blob/e6319f9dc8d5ea802cdbfb6b24b09175231f7300/common/lib/client/auth.js#L47-L50), so `!realtime.auth.clientId` will be false. (so the function doesn't bother checking the connection state due to short-circuiting)

The function might return false even if the user will end up being anonymous after all (which is fine, as then realtime can reject the ENTER), but it should never return true when the user might not be. (I originally called it `isDefinitelyAnonymous` to make that clear, but Paddy requested `isAnonymous`, see other thread)

mattheworiordan: Ok, thanks for clarifying. :+1:

PR #257, opened 2016-04-07
Tests for RSC10 & RSC14d

adamgiacomelli:
Inital attempt on these tests - still checking if everything is ok, would love some comments.

### DIFF
--- a/spec/rest/auth.test.js
+++ b/spec/rest/auth.test.js
@@ -630,8 +630,6 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
 			if(single_token == null) {
 				rest.auth.requestToken(tokenParams, function(err, tokenDetails) {
 					if(err) {
-						console.log("Token error");
-
 						test.ok(false, helper.displayError(err));
 						test.done();
 						return;
--- a/spec/rest/auth.test.js
+++ b/spec/rest/auth.test.js
@@ -465,7 +465,7 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
 	};
 
 	/*
-	 * createTokenRequest uses the key it was initialized with if authOptions is null
+	 * createTokenRequest uses the key it was initialised with if authOptions is null
 	 */
 	exports.auth_createTokenRequest_given_key = function(test) {
 		test.expect(1);
@@ -513,7 +513,7 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
 	};
 
 	/*
-	 * createTokenRequest uses the key it was initialized with if authOptions does not have a "key" key
+	 * createTokenRequest uses the key it was initialised with if authOptions does not have a "key" key
 	 */
 	exports.auth_createTokenRequest_given_key2 = function(test) {
 		test.expect(1);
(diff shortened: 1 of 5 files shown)


paddybyers:
@SimonWoolf can you have a look pls?

SimonWoolf:
:+1: looks good to me as far as it goes.

Would be nice to have a test for the second half of RSC14d as well, as that's the part that I'm not sure is implemented ("then the client should automatically make a single attempt to reissue the token and resend the request using the new token. If the token creation failed or the subsequent request with the new token failed due to a token error, then the request should result in an error")

eg you could have a test with authCallback that always returns the same token it got originally, and check that it's only called one additional time after it expires when the client lib errors. (And maybe another test that does the exact same thing but with an authCallback that returns just the token string rather than tokenDetails, so the client lib can't see that it's expired).

Mind adding those?

adamgiacomelli:
@SimonWoolf Will do, thanks for the review and the example

adamgiacomelli:
@mattheworiordan @SimonWoolf added the first test for the single retry, seems the functionality is missing - if you could take a look it would be lovely. I have to check the code to see if its implemented.

spec/rest/auth.test.js
```diff
@@ -615,5 +615,57 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
...
+		var single_token = null;
+		var request_counter = 0;
+
+		var staticTokenCallback = function(tokenParams, callback) {
+			tokenParams.ttl = 1;
+			//Always return the same token
+			if(single_token == null) {
+				rest.auth.requestToken(tokenParams, function(err, tokenDetails) {
+					if(err) {
+						console.log("Token error");
```
mattheworiordan: Is this needed given the error is going to be displayed by assertion below?

mattheworiordan:
>  added the first test for the single retry, seems the functionality is missing - if you could take a look it would be lovely.

What error are you getting @adamgiacomelli?

spec/rest/auth.test.js
adamgiacomelli: The debug printout was a leftover from debugging - I have removed it .

adamgiacomelli:
@mattheworiordan I am not getting an error at all - the token callback is getting called in a loop - I just stop the test after it gets called more than once after its reissued (per @SimonWoolf 's example) .

mattheworiordan:
ok thanks @adamgiacomelli, we'll get that fixed then and see if the tests passes.

SimonWoolf:
@adamgiacomelli added a fix (https://github.com/ably/ably-js/commit/3190ec175f5444ef645e41ab7d86fd78166864cb) to the `reauthOnlyOnce` branch, could you rebase on that and retest, please.

I believe to make it pass you'll have to make staticTokenCallback call back with `single_token.token` rather than `single_token` (if auth has the full tokenDetails then getAuthHeaders notices that the token has expired and takes the opportunity to request a new one, which throws the counting off. Technically that's not quite spec-compliant, but not in a way I really care about - whether it's once or twice doesn't matter that much, main point is that it doesn't do an infinite loop, which was the previous behaviour).

I also included a blind fix for the realtime equivalent (RTN14b and RTN15h): https://github.com/ably/ably-js/commit/5ac9c70d96a1113454a3fea2bc7cdf7dd4f853f6. Haven't tested. Could you please write a couple test for the single-attempt part of those specs to make sure the fix is correct, checking that they fail without that commit and pass with it. Thanks :)

adamgiacomelli:
@SimonWoolf tested the reauthOnlyOnce branch and did the necessary modifications - the test passes easily, will get to the realtime equivalents shortly

PR #258, opened 2016-04-12
Added test for unsecure basic auth failure (RSA1)

adamgiacomelli:
Simple test for RSA1

### DIFF
--- a/spec/realtime/auth.test.js
+++ b/spec/realtime/auth.test.js
@@ -608,10 +608,24 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
 		test.expect(2);
 
 		var realtime = helper.AblyRealtime({ tls: false });
+		var def_chooseTransportForHost = realtime.connection.connectionManager.chooseTransportForHost;
+
+		realtime.connection.connectionManager.chooseTransportForHost = function(transportParams, candidateTransports, callback) {
+			test.ok(false, 'Attempt to establish connection');
+
+			//Clean intercepted method
+			realtime.connection.connectionManager.chooseTransportForHost = def_chooseTransportForHost;
+			test.done();
+			return;
+		};
+
 		realtime.connection.once('failed', function(stateChange){
 			test.ok(true, 'Verify connection failed');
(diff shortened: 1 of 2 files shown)


adamgiacomelli:
In the similar test for rest it seems no errors are thrown when connecting with tls: false ... researching issue

SimonWoolf:
Your realtime test is actually passing even though ably-js has not implemented this yet, since the Ably server replies with the error. (Try grepping the library for "40103", no results). Your test will need to distinguish between Ably-generated and client-lib-generated errors, or alternatively could check that no actual connection attempt is made (and currently should fail).

> In the similar test for rest it seems no errors are thrown when connecting with tls: false

Your REST test uses `createTokenRequest`. That doesn't use basic auth -- it doesn't communicate with Ably at all, it just creates a `tokenRequest` object locally. So RSA1 does not apply.

RSA1 also wouldn't apply to eg `requestToken`, since while that does communicate with ably, it doesn't use basic auth: token requests work by the client lib signing the request with the secret key, so it does not need to be sent with the request. Try eg a stats request.

adamgiacomelli:
Ok great, i'll revise the test to differentiate the errors and i'll use a stats request to check the connection.

@SimonWoolf When I initialize the rest lib with tls:false I was expecting to get an error, the token request was just to double check if the connection was established.

adamgiacomelli:
@SimonWoolf Ok, both tests now pass because of the server response, should I implement in ably-js so the client lib throws an error before a connection attempt is made?

SimonWoolf:
@adamgiacomelli Sure, go for it. (only for realtime, rest should be initializable with no token with tls: false). Would be nice to have a test that fails before you implement that and succeeds after, though -- eg could check that `connect()` is never called or something.

adamgiacomelli:
@SimonWoolf I agree, I'll make a test that fails and a separate PR for the functionality.

adamgiacomelli:
@SimonWoolf updated the realtime test - now it's failing if the connectionManager's chooseTransportForHost is called - it's where the connection is attempted. If that's ok, i'll get started on throwing a client lib error before that happens.

Issue #237, opened 2016-02-19
authorise should store AuthOptions and TokenParams as defaults for subsequent requests


mattheworiordan:
See http://docs.ably.io/client-lib-development-guide/features/#RSA10g

I believe [this functionality is not implemented](https://github.com/ably/ably-js/blob/master/common/lib/client/auth.js#L139-L187), yet is documented in our documentation, so we should fix quickly

mattheworiordan:
This no longer applies due to changes in 0.9

PR #261, opened 2016-04-18
Second proposal for connection reauthing: using auth#authorise()

SimonWoolf:
Alternative to https://github.com/ably/ably-js/pull/260/files, as requested

Also fixes https://github.com/ably/ably-js/issues/237

One of the tests added (`reauth_tokenDetails`) currently fails due to https://github.com/ably/realtime/issues/534

### DIFF
--- a/common/lib/client/auth.js
+++ b/common/lib/client/auth.js
@@ -38,12 +38,16 @@ var Auth = (function() {
 		return JSON.stringify(c14nCapability);
 	}
 
-	function containsAuthInfo(options) {
-		return options.authUrl ||
-		  options.authCallback ||
-		  options.token        ||
-		  options.tokenDetails ||
-		  options.key;
+	/* RSA10j/d */
+	function persistAuthOptions(options) {
+		for(var prop in options) {
+			if(!(prop === 'force'       ||
+			     options[prop] === null ||
+			     options[prop] === undefined)) {
+				return true;
+			}
+		}
+		return false;
 	}
 
 	function logAndValidateTokenAuthMethod(authOptions) {
@@ -188,6 +192,9 @@ var Auth = (function() {
 		logAndValidateTokenAuthMethod(this.authOptions);
 
 		this._ensureValidAuthCredentials(function(err, tokenDetails) {
+			/* RSA10g */
+			self.tokenParams.timestamp = null;
+			/* RTC8 */
 			if(self.force && !err && (self.client instanceof Realtime)) {
 				self.client.connection.connectionManager.onAuthUpdated();
 			}
@@ -557,6 +564,9 @@ var Auth = (function() {
 	Auth.prototype._saveTokenOptions = function(tokenParams, authOptions) {
 		this.method = 'token';
 
+		/* We temporarily persist tokenParams.timestamp in case a new token needs
+		 * to be requested, then null it out in the callback of
+		 * _ensureValidAuthCredentials for RSA10g compliance */
 		this.tokenParams = tokenParams || this.tokenParams || {};
 
 		/* If an authOptions object is passed in that contains new auth info (ie
@@ -567,30 +577,26 @@ var Auth = (function() {
 		if(authOptions) {
 			this.force = authOptions.force;
 
-			if(containsAuthInfo(authOptions)) {
-				this.authOptions = authOptions || {};
-				this.authOptions.force = false;
-			} else if('queryTime' in authOptions) {
-				/* queryTime isn't an authInfo so can't replace the current
-				* authOptions, but per RSA10g, should still be stored */
-				this.authOptions.queryTime = authOptions.queryTime;
-			}
-
 			if(this.force) {
 				/* get rid of current token even if still valid */
 				this.tokenDetails = null;
 			}
 
-			if(authOptions.token) {
-				/* options.token may contain a token string or, for convenience, a TokenDetails */
-				authOptions.tokenDetails = (typeof(authOptions.token) === 'string') ? {token: authOptions.token} : authOptions.token;
-			}
-			if(authOptions.tokenDetails) {
-				this.tokenDetails = authOptions.tokenDetails;
-			}
+			if(persistAuthOptions(authOptions)) {
+				this.authOptions = authOptions;
+				this.authOptions.force = false;
 
-			if('clientId' in authOptions) {
-				this._userSetClientId(authOptions.clientId);
+				if(authOptions.token) {
+					/* options.token may contain a token string or, for convenience, a TokenDetails */
+					this.authOptions.tokenDetails = (typeof(authOptions.token) === 'string') ? {token: authOptions.token} : authOptions.token;
+				}
+				if(authOptions.tokenDetails) {
+					this.tokenDetails = authOptions.tokenDetails;
+				}
+
+				if('clientId' in authOptions) {
+					this._userSetClientId(authOptions.clientId);
+				}
 			}
 		}
 	};
--- a/common/lib/client/auth.js
+++ b/common/lib/client/auth.js
@@ -62,40 +62,46 @@ var Auth = (function() {
 		}
 	}
 
+	function basicAuthForced(options) {
+		return 'useTokenAuth' in options && !options.useTokenAuth;
+	}
+
+	/* RSA4 */
+	function useTokenAuth(options) {
+		return options.useTokenAuth ||
+			(!basicAuthForced(options) &&
+			 (options.clientId     ||
+			  options.authCallback ||
+			  options.authUrl      ||
+			  options.token        ||
+			  options.tokenDetails))
+	}
 
 	function Auth(client, options) {
 		this.client = client;
 		this.tokenParams = options.defaultTokenParams || {};
 
-		/* decide default auth method */
-		var key = options.key;
-		if(key) {
-			if(!options.clientId && !options.useTokenAuth) {
-				/* we have the key and do not need to authenticate the client,
-				 * so default to using basic auth */
-				Logger.logAction(Logger.LOG_MINOR, 'Auth()', 'anonymous, using basic auth');
-				this._saveBasicOptions(options);
-				return;
-			}
-			/* token auth, but we have the key so we can authorise
-			 * ourselves */
-			if(!hmac) {
+		if(useTokenAuth(options)) {
+			/* Token auth */
+			if(options.key && !hmac) {
 				var msg = 'client-side token request signing not supported';
 				Logger.logAction(Logger.LOG_ERROR, 'Auth()', msg);
 				throw new Error(msg);
 			}
+			this._saveTokenOptions(options.defaultTokenParams, options);
+			logAndValidateTokenAuthMethod(this.authOptions);
+		} else {
+			/* Basic auth */
+			if(options.clientId || !options.key) {
+				var msg = 'Cannot authenticate with basic auth' +
+					(options.clientId ? ' as a clientId implies token auth' :
+					 (!options.key ? ' as no key was given' : ''));
+					 Logger.logAction(Logger.LOG_ERROR, 'Auth()', msg);
+					 throw new Error(msg);
+			}
+			Logger.logAction(Logger.LOG_MINOR, 'Auth()', 'anonymous, using basic auth');
+			this._saveBasicOptions(options);
 		}
-		if('useTokenAuth' in options && !options.useTokenAuth) {
-			var msg = 'option useTokenAuth was falsey, but basic auth cannot be used' +
-				(options.clientId ? ' as a clientId implies token auth' :
-				(!options.key ? ' as a key was not given' : ''));
-			Logger.logAction(Logger.LOG_ERROR, 'Auth()', msg);
-			throw new Error(msg);
-		}
-
-		/* using token auth */
-		this._saveTokenOptions(options.defaultTokenParams, options);
-		logAndValidateTokenAuthMethod(this.authOptions);
 	}
 
 	/**
--- a/common/lib/client/auth.js
+++ b/common/lib/client/auth.js
@@ -194,8 +194,10 @@ var Auth = (function() {
 		this._ensureValidAuthCredentials(function(err, tokenDetails) {
 			/* RSA10g */
 			self.tokenParams.timestamp = null;
-			/* RTC8 */
-			if(self.force && !err && (self.client instanceof Realtime)) {
+			/* RTC8
+			 * use self.client.connection as a proxy for (self.client instanceof Realtime),
+			 * which doesn't work in node as Realtime isn't part of the vm context for Rest clients */
+			if(self.force && !err && self.client.connection) {
 				self.client.connection.connectionManager.onAuthUpdated();
 			}
 			callback(err, tokenDetails);
--- a/common/lib/client/auth.js
+++ b/common/lib/client/auth.js
@@ -38,16 +38,34 @@ var Auth = (function() {
 		return JSON.stringify(c14nCapability);
 	}
 
+	function containsAuthInfo(options) {
+		return options.authUrl ||
+		  options.authCallback ||
+		  options.token        ||
+		  options.tokenDetails ||
+		  options.key;
+	}
+
+	function logAndValidateTokenAuthMethod(authOptions) {
+		if(authOptions.authCallback) {
+			Logger.logAction(Logger.LOG_MINOR, 'Auth()', 'using token auth with authCallback');
+		} else if(authOptions.authUrl) {
+			Logger.logAction(Logger.LOG_MINOR, 'Auth()', 'using token auth with authUrl');
+		} else if(authOptions.key) {
+			Logger.logAction(Logger.LOG_MINOR, 'Auth()', 'using token auth with client-side signing');
+		} else if(authOptions.tokenDetails) {
+			Logger.logAction(Logger.LOG_MINOR, 'Auth()', 'using token auth with supplied token only');
+		} else {
+			var msg = 'authOptions must include valid authentication parameters';
+			Logger.logAction(Logger.LOG_ERROR, 'Auth()', msg);
+			throw new Error(msg);
+		}
+	}
+
+
 	function Auth(client, options) {
 		this.client = client;
 		this.tokenParams = options.defaultTokenParams || {};
-
-		/* RSA7a4: if options.clientId is provided and is not
-		 * null, it overrides defaultTokenParams.clientId */
-		if(options.clientId) {
-			this.tokenParams.clientId = options.clientId;
-			this.clientId = options.clientId
-		}
 
 		/* decide default auth method */
 		var key = options.key;
@@ -56,9 +74,7 @@ var Auth = (function() {
 				/* we have the key and do not need to authenticate the client,
 				 * so default to using basic auth */
 				Logger.logAction(Logger.LOG_MINOR, 'Auth()', 'anonymous, using basic auth');
-				this.method = 'basic';
-				this.key = key;
-				this.basicKey = toBase64(key);
+				this._saveBasicOptions(options);
 				return;
 			}
 			/* token auth, but we have the key so we can authorise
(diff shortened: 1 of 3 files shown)


common/lib/client/auth.js
```diff
@@ -534,6 +519,144 @@ var Auth = (function() {
 		}
 	};
 
+	Auth.prototype._saveBasicOptions = function(defaultTokenParams, authOptions) {
+		this.method = 'basic';
+		this.key = authOptions.key;
+		this.basicKey = toBase64(authOptions.key);
+		this.tokenParams = defaultTokenParams || {};
```
paddybyers: why this?

common/lib/client/auth.js
```diff
@@ -534,6 +519,144 @@ var Auth = (function() {
...
+		this.authOptions = authOptions || {};
+		delete this.authOptions.force;
```
paddybyers: I don't think we should use `delete` - why not `this.authOptions.force = false` ?

common/lib/client/auth.js
```diff
@@ -38,27 +38,25 @@ var Auth = (function() {
 		return JSON.stringify(c14nCapability);
 	}
 
+	function containsAuthInfo(options) {
+		var authMethods = ['authUrl', 'authCallback', 'token', 'tokenDetails', 'key'];
```
paddybyers: I think I would have simply put

    return options.key || options.token || options.tokenDetails || .....

common/lib/client/auth.js
```diff
@@ -38,27 +38,25 @@ var Auth = (function() {
...
+		}));
+	}
+
 	function Auth(client, options) {
 		this.client = client;
 		this.tokenParams = options.defaultTokenParams || {};
 
-		/* RSA7a4: if options.clientId is provided and is not
-		 * null, it overrides defaultTokenParams.clientId */
-		if(options.clientId) {
```
mattheworiordan: Why is this deleted?  As far as I can see, the `clientId` if passed in options will now not override `tokenParams.clientId`

common/lib/client/auth.js
```diff
@@ -76,35 +74,17 @@ var Auth = (function() {
...
-			Logger.logAction(Logger.LOG_MINOR, 'Auth()', 'using token auth with supplied token only');
-		} else {
-			var msg = 'options must include valid authentication parameters';
-			Logger.logAction(Logger.LOG_ERROR, 'Auth()', msg);
-			throw new Error(msg);
-		}
+
+		/* using token auth */
+		this._saveTokenOptions(options.defaultTokenParams, options);
+		this._logTokenAuthMethod();
```
mattheworiordan: Given this is log & validate valid authentication params, it should have an appropriate name.

spec/realtime/auth.test.js
```diff
@@ -600,5 +600,113 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
...
+				realtime.auth.authorise(null, {force: true}, function(err) {
+					test.ok(!err, err && displayError(err));
+				})
+			}, 3000)
+			test.ok(true, 'Verify connection connected');
+
+			/* statechanges due to the reauth */
+			realtime.connection.once('disconnected', function(stateChange){
+				test.ok(true, 'Verify connection disconnected');
+				test.equal(stateChange.reason.code, 80003, 'Verify disconnect was voluntary, not server-triggered (40142)');
```
paddybyers: s/voluntary/client-initiated

common/lib/client/auth.js
```diff
@@ -534,6 +519,144 @@ var Auth = (function() {
...
+				this.tokenDetails = authOptions.tokenDetails;
+			}
+
+			if('clientId' in authOptions) {
+				this._userSetClientId(authOptions.clientId);
+			}
+		}
+	};
+
+	Auth.prototype._logTokenAuthMethod = function() {
```
mattheworiordan: This function does not need to be exposed as part of the prototype and instead could be called with `authOptions` as an argument.  The less exposed functions on the object the better IMHO

paddybyers:
A few comments but otherwise lgtm

common/lib/client/auth.js
```diff
@@ -38,27 +38,25 @@ var Auth = (function() {
...
-		 * null, it overrides defaultTokenParams.clientId */
-		if(options.clientId) {
-			this.tokenParams.clientId = options.clientId;
-			this.clientId = options.clientId
-		}
-
 		/* decide default auth method */
 		var key = options.key;
 		if(key) {
 			if(!options.clientId && !options.useTokenAuth) {
```
mattheworiordan: Shouldn't this be checking if any token auth params are provided as well. Also `clientId` could be passed in with `defaultTokenParams.clientId` which would then be ignored and basic auth assumed.

common/lib/client/auth.js
```diff
@@ -124,16 +104,38 @@ var Auth = (function() {
...
+	 *                  params, to obtain a signed token request.
+	 *
+	 * - authHeaders:   (optional) a set of application-specific headers to be added to any request
+	 *                  made to the authUrl.
+	 *
+	 * - authParams:    (optional) a set of application-specific query params to be added to any
+	 *                  request made to the authUrl.
+	 *
+	 *
+	 * - requestHeaders (optional, unsuported, for testing only) extra headers to add to the
```
mattheworiordan: Typo, `unsupported`

common/lib/client/auth.js
```diff
@@ -534,6 +519,144 @@ var Auth = (function() {
...
+		delete this.authOptions.force;
+		if('clientId' in authOptions) {
+			this._userSetClientId(authOptions.clientId);
+		}
+	}
+
+	Auth.prototype._saveTokenOptions = function(tokenParams, authOptions) {
+		this.method = 'token';
+
+		this.tokenParams = tokenParams || this.tokenParams || {};
```
mattheworiordan: I believe tokenParams could be merged not overwritten.

common/lib/client/auth.js
```diff
@@ -534,6 +519,144 @@ var Auth = (function() {
...
+		/* If an authOptions object is passed in that contains new auth info (ie
+		* isn't just {force: true} or something), it becomes the new default, with
+		* the exception of the force attribute (RSA10g), which is set anew on each
+		* call to authorise (defaulting to false) */
+		this.force = false;
+		if(authOptions) {
+			this.force = authOptions.force;
+
+			if(containsAuthInfo(authOptions)) {
+				this.authOptions = authOptions || {};
```
mattheworiordan: As above, I believe provide authOptions should be merged into existing authOptions as opposed to replacing them

common/lib/client/auth.js
```diff
@@ -145,45 +147,28 @@ var Auth = (function() {
...
-					/* expired, so remove */
-					Logger.logAction(Logger.LOG_MINOR, 'Auth.getToken()', 'deleting expired token');
-					self.tokenDetails = null;
-				}
-				requestToken();
-			});
-		} else {
-			requestToken();
-		}
+		this._ensureValidAuthCredentials(function() {
```
mattheworiordan: Isn't there an `err` argument for this function callback?

common/lib/transport/connectionmanager.js
```diff
@@ -876,6 +871,15 @@ var ConnectionManager = (function() {
 		this.notifyState({state: 'closed'});
 	};
 
+	ConnectionManager.prototype.onAuthUpdated = function() {
+		var state = this.state.state;
```
mattheworiordan: Gotta love `state.state` :)

common/lib/client/auth.js
paddybyers: > Isn't there an err argument for this function callback?

yes, but all arguments are just forwarded with `apply()` below.

That said, unless it is called with different numbers of arguments in different places, I would have preferred to see the args explicitly.

mattheworiordan:
Few comments, but on the whole I think it's a good solution and sets the groundwork for an improved authentication upgrade process in future.

common/lib/client/auth.js
SimonWoolf: > Shouldn't this be checking if any token auth params are provided as well.

The logic of when to use basic vs token auth isn't something I changed here - ably-js has always used basic if a key is given as long as `useTokenAuth` isn't set. That said, looking at `RSA4` now, you're right, the logic not spec-compliant. I'll add a separate commit to fix that.

> Also clientId could be passed in with defaultTokenParams.clientId which would then be ignored and basic auth assumed.

I'm not sure I agree with this bit - should merely providing `defaultTokenParams` really force token auth used if it wouldn't otherwise have been used? That seems surprising. If someone wants an identified client they should pass a `clientId` in `clientOptions` directly

common/lib/client/auth.js
SimonWoolf: fair enough

common/lib/client/auth.js
SimonWoolf: It still will, it's just now done by the `_setClientId` stuff which is called by the `_save` methods. DRYer this way, don't need to have RSA7a4 logic in three places. (note that the `init_conflicting_clientids` test still passes)

common/lib/client/auth.js
SimonWoolf: I dunno, merging seems unintuitive to me. Especially in the case of authOptions: say someone initializes the library with an authCallback. then they call authorise() with an authUrl. should the library really keep using the authCallback just because it now has both an authCallback happens to come first in the elfsif logic?

common/lib/client/auth.js
SimonWoolf: fair enough.

<sup>grumble javascript grumble private methods grumble</sup>

common/lib/client/auth.js
SimonWoolf: Fair enough. (also allows for a !err check before disconnecting, which is probably a good idea)

common/lib/client/auth.js
mattheworiordan: > I'm not sure I agree with this bit - should merely providing defaultTokenParams really force token auth used if it wouldn't otherwise have been used?

True, you are right

common/lib/client/auth.js
mattheworiordan: I disagree unfortunately and believe we have merged in all other implementations.  I think a common use case would be to initialise the library, and then later add a `clientId` for an explicit authorise or change the `ttl`.  Assuming therefore the user did the following, I assume it would effectively nullify the `clientId` and the `authUrl`.

```javascript
client = new Ably.Realtime({ clientId: 'bob', authUrl: '/auth' });
client.authorise({ ttl: 60 }, {});
```

As we have defaults currently that should not be overwritten with each call to authorise, I personally prefer that if a user wants to remove an option they provide a `null` value which would work fine with a merge. Either way, we need to update the spec, but if we don't support merge, we're going to need to modify Python, PHP, Ruby, and iOS I believe.  I am not sure how Java approaches it.

@paddybyers thoughts?

SimonWoolf: > Assuming therefore the user did the following, I assume it would effectively nullify the clientId and the authUrl

In the implementation in this PR: no, to both. auth.authOptions is only overwritten if it `containsAuthInfo` (ie an authUrl, authCallback, token, tokenDetails, or key), to avoid this problem (see the line above the line this comment is on). `auth.clientId` is only ever touched if authOptions has a clientId key ([see here](https://github.com/ably/ably-js/pull/261/files#diff-cc36f4a032c8d16ad5390d989524c305R586)) so it would not get blanked out even for an authOptions that containsAuthInfo.

That said, I take the point, and don't mind changing them both to merge if @paddybyers agrees. (Though we'd have to make sure that's documented well, especially since with requestToken and createTokenRequest, a passed in tokenParams is explicitly not merged, but replaces any stored tokenParams)

paddybyers: Of course this is why it was a bad idea all along to have `authorise()` persist the overrides, and a `setOptions()`, which fully supersedes the previously set options, would have been more intuitive IMO :)

Merging (and then persisting) is inconsistent with the other APIs and doesn't let you unset things once they're set in java and similar languages. Unsetting them by including `authURL: undefined` etc is a bit horrible anyway.

So, if we're not going to have a `setOptions()` which explicitly and fully supersedes the set options, then what @SimonWoolf has done here (which is piecemeal and arbitrary) may well deal with the typical use-cases as well as possible.

> believe we have merged in all other implementations

It's not as if it's a huge change, so unfortunately I think that's not a good enough reason by itself to go further along the road of making an inferior API.

mattheworiordan: > Of course this is why it was a bad idea all along to have authorise() persist the overrides, and a setOptions(), which fully supersedes the previously set options, would have been more intuitive IMO :)

I am not sure how it changes anything really.  Either `authorise` persists new changes or does not persist, but whether it merges is not relevant as far as I can tell because we'll still need to decide on merging options or not.

> Merging (and then persisting) is inconsistent with the other APIs and doesn't let you unset things once they're set in java and similar languages. Unsetting them by including authURL: undefined etc is a bit horrible anyway.

Yes, it is in typed langs.

> So, if we're not going to have a setOptions() which explicitly and fully supersedes the set options, then what @SimonWoolf has done here (which is piecemeal and arbitrary) may well deal with the typical use-cases as well as possible.

How?  If you provide a set of options to `authorise` regardless of `setOptions` existing will either merge with existing options or not.  

I assume a merge will happen regardless based on the default values anyway i.e. whatever you provide will merge with `TokenParam` and `AuthOption` defaults.

>>  believe we have merged in all other implementations

> It's not as if it's a huge change, so unfortunately I think that's not a good enough reason by itself to go further along the road of making an inferior API.

It's big enough that we'll need to change tests in every library now, so I agree going down the wrong road is not necessarily the right approach, but this is not trivial given the number of client libraries and the test coverage that exists in each.

The problem now though if we did go down this route is that if you wanted to reauth without changing any options and thus set the `force: true` key/value of `AuthOptions`, how will the client library now know that you the rest of the `AuthOptions` should be ignored.  There is no non-ugly way to do that, especially in typed languages.  Also, if a user currently provides `{}` as an AuthOption arg in JS, it will reset `AuthOption` to empty, which IMHO, could be a little confusing for users.

SimonWoolf: > if a user currently provides {} as an AuthOption arg in JS, it will reset AuthOption to empty

No it won't (with the implementation in this PR), see my previous comment.

Look, authOptions currently consists of only three things:

1. something which decides how to get a token (authUrl, authCallback, token, tokenDetails, or key). (what I called 'authInfo' above)
2. `force`
3. `queryTime`

For 1, there is only one operative way of getting a token at any one time, so IMHO merging just doesn't make sense - you never want an `authUrl` and an `authCallback` at the same time.

For 2, `force` explicitly doesn't persist either way, so merging vs replacing is irrelevant.

For 3, `queryTime` has no effect once the `serverTimeOffset` has been set once, so merging vs replacing is irrelevant.

So currently, the practical difference between "replace authOptions if passed-in once contains authInfo" and "only replace authInfo, merge all other options" is literally none at all.

Which means the only thing we're changing by making this policy decision now is the default for how any new authOptions we may add in the future will behave. Which I have no strong opinions on, I'll implement whatever you two decide. (Ditto on whether to replace/merge tokenParams).

mattheworiordan:
So to be clear, are we all in agreement that:

* If you provide null as either or both arguments to `authorise` the default settings will remain and an authorise request will be made if a valid token does not currently exist
* If you provide an object for either argument, the attributes of the `TokenParams` or `AuthOptions` will be merged with the client library initial defaults (not the current configured values of the library) and used for that authorise request and all future authorise requests
* If `AuthOptions` has all attributes null apart from `force: true` then it will not change the configured `AuthOptions` for the library, but instead simply force an authorise.  I don't think `queryTime` should fall into this category personally and should be treated the same as all other `AuthOptions` as why would someone want to call `authorise({}, { queryTime: true })` after a previous authentication had been made.  They should do that when instancing the library as if the time is incorrect when creating the library there is no point trying to correct that later with an `authorise`
* I realise some time ago we moved `force: true` into the `AuthOptions`, but given that options are NOT merged, I think this is now a mistake.  I propose for spec `0.9`, we move `force: true` out of AuthOptions, and perhaps go back to a simple boolean argument i.e. `authorise(TokenParams, AuthOptions, boolean force)` or go for something more flexible such as `authorise(TokenParams, AuthOptions, AuthoriseOptions)` where `AuthoriseOptions` only supports `force: true` for now.  Whilst this is more flexible, the naming is terribly confusing so we'd need a better name for the class.  Thoughts on this please.

SimonWoolf:
> If you provide null as either or both arguments to authorise the default settings will remain and an authorise request will be made if a valid token does not currently exist

s/default/previous, but yes

> If you provide an object for either argument, the attributes of the TokenParams or AuthOptions will be merged with the client library initial defaults (not the current configured values of the library)

There are no client library initial defaults though, for either tokenParams or authOptions. If you don't supply any authOptions, the library will just throw an exception; if you don't supply any tokenParams, it won't pass any to realtime, so realtime will use its own defaults. (I can't find it now, but IIRC we had a long discussion about this, and [we changed the spec](https://github.com/ably/docs/commit/ab8307b0) such that the client lib would not supply its own defaults). So this is equivalent to "replaces the current values". 

> If AuthOptions has all attributes null apart from force: true then it will not change the configured AuthOptions for the library, but instead simply force an authorise. I don't think queryTime should fall into this category personally and should be treated the same as all other AuthOptions

All the other authOptions apart from `force` and `queryTime` provide a way to authenticate, that's why they're mutually exclusive. Treating `queryTime` like token/key/authCallback/etc, just means an `authorise({}, { queryTime: true })` call would result in the library raising a "no valid authentication parameters" exception next time it needed a new token, rather than just being pointless-but-harmless. Though I guess it doesn't really matter as you're right, no-one's going to do it.

> I realise some time ago we moved force: true into the AuthOptions, but given that options are NOT merged, I think this is now a mistake. I propose for spec 0.9, we move force: true out of AuthOptions, and perhaps go back to a simple boolean argument i.e. authorise(TokenParams, AuthOptions, boolean force) or go for something more flexible such as authorise(TokenParams, AuthOptions, AuthoriseOptions) where AuthoriseOptions only supports force: true for now. Whilst this is more flexible, the naming is terribly confusing so we'd need a better name for the class. Thoughts on this please.

Yeah, I guess that would make sense (since {authUrl, authCallback, token, tokenDetails, key} are special: they're not independent settings, they're alternative ways to authenticate. Cleaner to split out actual toggleable settings). I don't have a good answer to the naming/organisation thing though. I dunno, maybe an `auth` attribute to `authOptions` (ie `authOptions = {auth: {authUrl: "..."}, force: true}`), with single-level merge semantics, so the auth attribute has replace semantics? Bit complicated though, and not very good for java etc. I'll wait & see what Paddy suggests.

mattheworiordan:
> There are no client library initial defaults though

There are, see https://www.ably.io/documentation/realtime/types#auth-options.  All I am saying is that the default behaviour of the client lib remains if an option is not provided in the Hash with loosely typed languages.

> All the other authOptions apart from force and queryTime provide a way to authenticate, that's why they're mutually exclusive

I am not sure I understand what you're saying here because I don't think either `queryTime` or `force` are mutually exclusive, and the other attributes are not mutually exclusive either i.e. you can provide `authUrl` and `authParams` or you can provide `useTokenAuth` and `authCallback`.  I also don't see why we would disallow providing `token` and `authCallback` so that the lib could use the token provided but has a means to renew.

> I don't have a good answer to the naming/organisation thing though. I dunno, maybe an auth attribute to authOptions (ie authOptions = {auth: {authUrl: "..."}, force: true})

If we were starting from scratch perhaps, but I don't think that big a change is a good idea.  Simply removing `force` and moving it into a new argument would for most users, probably not be breaking, whereas your change would certainly break things for everyone.

common/lib/client/auth.js
SimonWoolf: Added 998b6a4 for RSA4 compliance

paddybyers:
@SimonWoolf Can you update this please to reflect the latest spec proposal, and then add some tests so I can test my realtime changes?

SimonWoolf:
Added 83b563b for latest reauth spec compliance.

@paddybyers the new `reauth_tokenDetails` test in spec/realtime/auth.test.js currently fails and should pass against a realtime that allows changing token capabilities. That currently just tests ability to upgrade capabilities, since iirc that's what we decided was the minimum necessary change. Let me know if you also want tests for downgraded capabilities and/or going from anonymous to identified.

SimonWoolf:
also 1516466. grumble node vm contexts grumble

mattheworiordan:
Just a note. Before this lands, we need to ensure the connection recovery documentation states that capabilities can be changed because currently it's not

paddybyers:
Closing in favour of https://github.com/ably/ably-js/issues/269

SimonWoolf:
Merged as 31f70cb

PR #264, opened 2016-04-29
Add onPageRefresh option

### LINKED ISSUE #259: TODO: consider new `onPageRefresh: close | persist | none` connection option
in place of `recover: true`

also make docs clearer

SimonWoolf:
326e78d Fixes #259.

Currently this leaves `none` as the default, same behaviour as now.

Also includes  c62fc42 as a proposed solution to https://github.com/ably/realtime/issues/540: turn off persisting the connection onConnect, so that it's only persisted on page refresh; and unpersist after a successful recovery. Former change minimises the time the cookie is persisted to the duration of the refresh/reconnect cycle, rather than the whole time. Latter change prevents a recovery key from being used twice (unless there's a race condition between tabs, which is much less likely). The downside is we lose the fallback recover-from-key-only (no serial) functionality - but afaict that's really only useful if onPageRefresh isn't supported, which at this point is very few browsers (safari < 3 and that's about it), so I don't think we're losing much.

### DIFF
--- a/common/lib/transport/connectionmanager.js
+++ b/common/lib/transport/connectionmanager.js
@@ -4,6 +4,7 @@ var ConnectionManager = (function() {
 	var eraseCookie = (typeof(Cookie) !== 'undefined' && Cookie.erase);
 	var connectionKeyCookie = 'ably-connection-key';
 	var connectionSerialCookie = 'ably-connection-serial';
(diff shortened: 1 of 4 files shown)


SimonWoolf:
Closing in favour of https://github.com/ably/ably-js/pull/266

PR #266, opened 2016-05-05
Recover: implement callback proposal

SimonWoolf:
Fixes https://github.com/ably/realtime/issues/541

Replaces https://github.com/ably/ably-js/pull/264


SimonWoolf:
Also, just noticed that persistence was set to last for only 15s (`connectionPersistTimeout`) @paddybyers why was that so low - is there a reason to have it be any lower than the realtime connection ttl? Or did the realtime connection ttl used to be only 15s? (On the assumption it's just out of date, added d86686d to change it to connectionStateTtl)

common/lib/transport/connectionmanager.js
```diff
@@ -165,51 +158,77 @@ var ConnectionManager = (function() {
...
-			this.chooseTransportForHost(transportParams, self.transports.slice(), callback);
-			return;
+			if(typeof self.options.recover === 'string') {
+				modeCb('recover');
+				return;
+			}
+
+			var recoverFn = self.options.recover,
+				lastSessionData = getFromSession && getFromSession(sessionRecoveryName);
+			if(lastSessionData && lastSessionData.host === window.location.host && typeof(recoverFn) === 'function') {
```
mattheworiordan: Why do we need to check `lastSessionData.host == window.location.host`, I would have thought the security model in the browser would look after that for us?

SimonWoolf: Er, yep, you're completely right. Not sure why I was thinking.  :persevere:

paddybyers:
> @paddybyers why was that so low - is there a reason to have it be any lower than the realtime connection ttl? Or did the realtime connection ttl used to be only 15s?

Because it was only  intended to cover the time for the refresh to happen; not the time for the connection in the refreshed page to succeed.

However, if there is a loss of connectivity it's equally possible that the page refresh stalls so perhaps it could be the same as the connection ttl.

common/lib/transport/connectionmanager.js
```diff
@@ -1,9 +1,8 @@
 var ConnectionManager = (function() {
-	var readCookie = (typeof(Cookie) !== 'undefined' && Cookie.read);
-	var createCookie = (typeof(Cookie) !== 'undefined' && Cookie.create);
-	var eraseCookie = (typeof(Cookie) !== 'undefined' && Cookie.erase);
-	var connectionKeyCookie = 'ably-connection-key';
-	var connectionSerialCookie = 'ably-connection-serial';
+	var getFromSession    = (typeof(SessionStorage) !== 'undefined' && SessionStorage.get);
```
mattheworiordan: I find this approach a little odd.  Why are we creating a `SessionStorage` object that sometimes has an interface with `get`, `set` and `remove`, and other times without the method.  I would prefer we either:

* Have a method that on SessionStorage such as `isSupported()` but the interface remains the same.
* My preference: If `sessionStorage` is not supported, we should have `noop` methods in their place that simply don't do anything.

common/lib/util/defaults.js
```diff
@@ -11,10 +11,9 @@ Defaults.TIMEOUTS = {
 	suspendedRetryTimeout      : 30000,
 	httpRequestTimeout         : 15000,
 	/* Not documented: */
-	connectionStateTtl         : 60000,
+	connectionStateTtl         : 120000,
```
mattheworiordan: Oops, ignore previous comment, I see this in a later commit :/

common/lib/util/defaults.js
```diff
@@ -62,6 +61,16 @@ Defaults.normaliseOptions = function(options) {
 		options.queueMessages = options.queueEvents;
 	}
 
+	if(options.recover === true) {
+		Logger.deprecated('{recover: true}', '{recover: function(lastConnectionDetails, cb) { cb(true); }}');
+		options.recover = function(lastConnectionDetails, cb) { cb(true); };
+	}
+
+	if(typeof options.recover === 'function' && options.closeOnUnload === true) {
+		Logger.logAction(LOG_ERROR, 'Defaults.normaliseOptions', 'closeOnUnload was true and a session recovery function was set - these are mutually exclusive, so unsetting the latter');
```
mattheworiordan: Nice :)

mattheworiordan:
Nice work, looks great. 👍

paddybyers:
Closing in favour of https://github.com/ably/ably-js/issues/269

common/lib/transport/connectionmanager.js
SimonWoolf: Sure, happy to do that. Though will still need to do `(typeof(SessionStorage) !== 'undefined' && ...` because the `sessionstorage.js` file isn't even loaded in node, so the `SessionStorage` object doesn't exist

mattheworiordan: Why not just load it and have noop methods like unsupported browsers?

SimonWoolf: Actually on second thoughts I think it'd be simpler to leave it as-is. You still need the sessionStorageSupported checks because of node (eg because window.addEventListener isn't defined in node), so the change doesn't really let you simplify things much, sadly. https://github.com/ably/ably-js/commit/285ca827787aa7cf23c9c0b7007015ea1309a3b3

paddybyers: > Why not just load it and have noop methods like unsupported browsers?

Personally I would prefer to test for the existence of the interface, and use it if present.

SimonWoolf: @mattheworiordan sorry, github hadn't loaded your comment when I posted my last one.

The library uses a pattern of browser-specific helpers in `browser/lib` and node ones in `node/lib` for a bunch of things. I wouldn't be against just having both load all files and using tests & noops, but doing that just for SessionStorage would make that inconsistent with the other browser-specific/node-specific libs, and doing it for everything would be a major refactor at this stage. So I'm inclined to leave it as-is for now

mattheworiordan: Ok, fine, leave as-is

SimonWoolf:
Merged as 4c2c0d8

Issue #270, opened 2016-05-10
ConnectionManager.sendImpl(): Unexpected exception in transport.send(): TypeError: Cannot read property 'send' of null


mattheworiordan:
A user has hit a runtime error in the JS library which should not happen even if the underlying transports fail.  Separately we are investigating the root cause of the transport failure, however this issue is about the realtime library not correctly handling the failure.

```
Ably: Request.onabort(): Request cancelled, errorEvent type was abort, current statusText is 
ably.min.js?_=1462617215292:81 Ably: CometTransport.sendItems(): on complete: err = {"message":"Request cancelled, errorEvent type was abort, current statusText is ","code":80000,"statusCode":400}
ably.min.js?_=1462617215292:81 Ably: Transport.onProtocolMessage(): error; connectionKey = 8e985!Z2pEuPu18QgOEl85-100328e985; err = {"message":"Request cancelled, errorEvent type was abort, current statusText is ","code":80000,"statusCode":400}
rest.ably.io/comet/8e985!Z2pEuPu18QgOEl85-100328e985/send?access_token=[REMOVED]&rnd=9733905395760611:1 POST https://rest.ably.io/comet/8e985!Z2pEuPu18QgOEl85-100328e985/send?access_to…[REMOVED]&rnd=9733905395760611 404 (Not Found)
ably.min.js?_=1462617215292:81 Ably: CometTransport.sendItems(): on complete: err = {"statusCode":404,"code":40400,"message":"Invalid transport id: 100328e985","serverId":"frontend.9aff.eu-central-1-A.i-1a59c8a6"}
ably.min.js?_=1462617215292:81 Ably: Transport.onProtocolMessage(): error; connectionKey = 8e985!Z2pEuPu18QgOEl85-100328e985; err = {"message":"Invalid transport id: 100328e985","code":40400,"statusCode":404,"serverId":"frontend.9aff.eu-central-1-A.i-1a59c8a6"}
chat:1081 function::convBinding();
ably.min.js?_=1462617215292:81 Ably: ConnectionManager.sendImpl(): Unexpected exception in transport.send(): TypeError: Cannot read property 'send' of null
    at b.sendImpl (http://cdn.ably.io/lib/ably.min.js?_=1462617215292:147:361)
    at b.send (http://cdn.ably.io/lib/ably.min.js?_=1462617215292:146:460)
    at f.sendMessage (http://cdn.ably.io/lib/ably.min.js?_=1462617215292:216:252)
[...]

PR #274, opened 2016-05-17
If wasActive & transport drops when another pending, go to connecting

SimonWoolf:
Blind fix to https://github.com/ably/ably-js/issues/270

### DIFF
--- a/common/lib/transport/connectionmanager.js
+++ b/common/lib/transport/connectionmanager.js
@@ -551,6 +551,11 @@ var ConnectionManager = (function() {
 				state = 'disconnected';
 			}
 			this.notifyState({state: state, error: error});
+		} else if(wasActive) {
+			/* If we were active but there is another transport scheduled for
+			* activation, go into to the connecting state until that transport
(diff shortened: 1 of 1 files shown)


common/lib/transport/connectionmanager.js
```diff
@@ -551,6 +551,11 @@ var ConnectionManager = (function() {
 				state = 'disconnected';
 			}
 			this.notifyState({state: state, error: error});
+		} else if(wasActive) {
```
paddybyers: Personally I'd prefer a single `if(wasActive)` with both of these cases as sub-cases

paddybyers:
LGTM but do we have outstanding issues that this fixes?

common/lib/transport/connectionmanager.js
SimonWoolf: First case applies also when wasActive is false (when the transport being deactivated was pending and there is no active and no further pending transports), that change would break that case

SimonWoolf:
Merged as b0bb9a0

PR #276, opened 2016-05-18
ErrorInfo: arg order, coercing Error objects

SimonWoolf:
See https://app.intercom.io/a/apps/ua39m1ld/inbox/all/conversations/4737349518


paddybyers:
lgtm

Issue #277, opened 2016-05-18
Bitdefender "Scan SSL" setting strips transfer-encoding headers from streaming responses


mattheworiordan:
A customer could not make XHR requests to Ably as it appears BitDefender Firewall for PC was interfering with the requests and possibly stripping out some headers.

We should see if we can reproduce and implement a workaround if possible.

SimonWoolf:
fwiw looking at the xhr code, my guess is we'll find it's stripping the transfer-encoding header. That is workaroundable, but should reproduce in a VM to make sure it's that first

SimonWoolf:
Installed Bitdefender in a Win10 VM, and: Confirmed. Bitdefender is stripping transfer-encoding, and adding content-length.

Before:
```
HTTP/1.1 200 OK
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: http://localhost:8000
Access-Control-Expose-Headers: Link,Transfer-Encoding,Content-Length,X-Ably-ErrorCode,X-Ably-ErrorMessage,X-Ably-ServerId
Content-Type: application/json
Date: Wed, 18 May 2016 19:04:50 GMT
Set-Cookie: ably_sessionid=1082b; path=/comet; httponly
Vary: Origin
X-Ably-Serverid: frontend.1be5.eu-central-1-A.i-305dca8c
X-Content-Type-Options: nosniff
Transfer-Encoding: chunked
```
After
```
HTTP/1.1 200 OK
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: http://10.0.2.2:8000
Access-Control-Expose-Headers: Link,Transfer-Encoding,Content-Length,X-Ably-ErrorCode,X-Ably-ErrorMessage,X-Ably-ServerId
Content-Type: application/json
Date: Wed, 18 May 2016 19:04:19 GMT
Set-Cookie: ably_sessionid=1082b; path=/comet; httponly
Vary: Origin
X-Ably-Serverid: frontend.1be5.eu-central-1-A.i-305dca8c
X-Content-Type-Options: nosniff
Content-Length: 396
```
Experimenting further: the culprit is the "Scan SSL" setting in the "Web Protection module". This setting appears to work by MITMing all ssl traffic through a local proxy (using a new root-level certificate it installs). Aside from messing with headers, said proxy also breaks EV certificates, and uses an outdated and insecure set of cyphersuites (howsmyssl.com goes from a 'probably okay' rating with it off to a 'bad' rating with it on).

[...]

Issue #182, opened 2015-12-14
connectionManager should probably keep track of proposed transports


SimonWoolf:
currently connectionManager doesn't have a reference to not-yet-proved-viable transports, which leads to having to reimplement various connectionManager functionality (eg timing out, reacting to disconnect/close events) in the transport itself (eg https://github.com/ably/ably-js/commit/5f700cc569a76efc1b012440126253e523aedc8e), to avoid race conditions if those events happen before the transport is proved viable. would be neater if connectionManager did this.

SimonWoolf:
Fixed by https://github.com/ably/ably-js/pull/279

PR #279, opened 2016-05-28
Trasnport fallback sequence rewrite

SimonWoolf:
Per https://github.com/ably/wiki/issues/76

Fixes https://github.com/ably/ably-js/issues/277
Fixes https://github.com/ably/ably-js/issues/165
Fixes https://github.com/ably/ably-js/issues/182
Fixes https://github.com/ably/realtime/issues/557

Basic idea is that there are now three stages in connecting:

- preference: if there is a cached transport preference, we try to connect on that. If that fails or times out we abort the attempt,  remove the preference and fall back to base. If it succeeds, we try upgrading it if needed (will only be in the case where the preference is xhrs and the browser supports ws).

- base: we try to connect with the best transport that we think will never fail for this browser (usually this is xhr_polling; for very old browsers will be jsonp, for node will be comet). If it doesn't work, we try fallback hosts.

- upgrade: once connected, we see if there are any better transports, and if so, try to upgrade to them.

We can work out which stage we're at as a function just of the current connection state (and whether there are any stored preferences). So this PR moves connection flow away from a model where the connection stage is carried around implicitly in the callback heirarchy, to one where connectImpl explicitly dispatches to the appropriate action based on the current state (and setTransportPending just calls connectImpl again).

I've given connection states a bit more bite. Eg previously, going into the 'disconnected' state did not affect transport attempts: pending transports could still be activated, and proposed transports could still become pending (and connectionManager didn't even have a reference to
the latter). This caused various issues - eg comet transports that froze on connecting would stick around for 90s, with new ones accumulating every time the transport went into 'connecting'. It also meant that eg the websocket transport had to implement its own connection state
watchers to guard against state changes that happened before connectionManager got a reference to them (eg https://github.com/ably/ably-js/commit/5f700cc569a76efc1b012440126253e523aedc8e).  With this pr, connectionManager tracks transports from their inception, and pending and unviable transports are aborted on expiration of the connecting timeout, so connection attempts only take place in the `connecting` state.

I've also split the xhr transport into separate `xhr_polling` and `xhr_streaming` transports. They still use the same `XhrRequest` under the hood (so technically you can still hand `stream: false` to `xhr_streaming` and get a polling transport) but having them nominally be separate transports makes the new fallback tree much clearer, given that it generally starts from a polling-only transport.


paddybyers:
I've yet to review in detail, but looks good.

Thanks

spec/realtime/presence.test.js
```diff
@@ -694,7 +694,7 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
 	 * Attach to channel, enter+leave presence, detatch again, and get presence history
 	 */
 	exports.presenceHistory = function(test) {
-		test.expect(4);
+		test.expect(3);
```
mattheworiordan: @SimonWoolf not sure I understand this commit, can you explain?

mattheworiordan:
@SimonWoolf this is great work, 👍 

Not sure how best to test this before we roll out.  If we're confident it works we could just push to the CDN. Equally, we may want to tag it as beta, and then try it in few of our apps first?

spec/realtime/presence.test.js
SimonWoolf: Just a quick workaround for https://ably-real-time.slack.com/archives/tech/p1464371716000794

mattheworiordan: Ok, not sure I agree we should implement a workaround for a bug...

SimonWoolf: Fair point. Though since the 'buggy' behaviour isn't that unreasonable and is unlikely to break anyone's code, it's low-priority so probably won't be fixed very soon, so it's a bit annoying to have ably-js ci be permanently failing for the next N months because of it. But happy to remove the commit if you'd prefer

mattheworiordan: Ok leave in and add an issue to fix

SimonWoolf:
Checked everything out in a bunch of VMs manually (mostly old versions of IE). I'm satisfied that they all work at least no worse than before. (Also discovered an IE8 crypto bug, addressed separately).

I'm happy for this to be merged and released once @paddybyers gives the go-ahead.

paddybyers:
We should have some documentation somewhere that captures the description above - where should that be?

Is it possible that the node comet transport will run behind a proxy that breaks streaming? If so, do we also need separate polling and streaming transports for node? (I realise we don't do that for other non-browser libs, eg Ruby or Java, but we do eventually want to upgrade all libs to support proxies in all their modes and breakageness.)

browser/lib/transport/xhrpollingtransport.js
```diff
@@ -0,0 +1,37 @@
...
+		transport.connect();
+	};
+
+	XHRPollingTransport.prototype.toString = function() {
+		return 'XHRPollingTransport; uri=' + this.baseUri + '; isConnected=' + this.isConnected;
+	};
+
+	XHRPollingTransport.prototype.createRequest = XHRRequest.createRequest;
+
+	if(typeof(ConnectionManager) !== 'undefined' && XHRPollingTransport.isAvailable()) {
```
paddybyers: What's the purpose of this check for definedness of `ConnectionManager` ?

browser/lib/util/defaults.js
```diff
@@ -1,6 +1,21 @@
...
+	transportPreferenceOrder: ['jsonp', 'xhr_polling', 'xhr_streaming', 'web_socket'],
+	upgradeTransports: ['xhr_streaming', 'web_socket'],
 	minified: !(function _(){}).name
 };
+
+/* If using IE8, don't attempt to upgrade from xhr_polling to xhr_streaming -
+* while it can do streaming, the low max http-connections-per-host limit means
+* that the polling transport is crippled during the upgrade process. So just
+* leave it at the base transport */
+if(navigator.userAgent.toString().match(/MSIE\s8\.0/)) {
```
paddybyers: Why isn't this logic in `XHRStreamingTransport.isAvailable()` ?

common/lib/transport/connectionmanager.js
```diff
@@ -95,18 +121,22 @@ var ConnectionManager = (function() {
...
 		this.connectionSerial = undefined;
 
-		this.httpTransports = Utils.intersect((options.transports || Defaults.httpTransports), ConnectionManager.httpTransports);
-		this.transports = Utils.intersect((options.transports || Defaults.transports), ConnectionManager.transports);
-		this.upgradeTransports = Utils.arrSubtract(this.transports, this.httpTransports);
+		this.transports = Utils.intersect((options.transports || Defaults.transports), ConnectionManager.supportedTransports);
+		/* baseTransports selects the leftmost transport in the Defaults.transports list
+		* that's both requested and supported. Normally this will be xhr_polling;
+		* if xhr isn't supported it will be jsonp. If the user has forced a
+			* transport, it'll just be that one. */
```
paddybyers: whitespace?

common/lib/transport/connectionmanager.js
```diff
@@ -117,11 +147,11 @@ var ConnectionManager = (function() {
...
+		if(haveWebStorage && typeof options.recover === 'function' && window.addEventListener)
 			window.addEventListener('beforeunload', this.persistConnection.bind(this));
 
-		if(setInSession && options.closeOnUnload === true && window.addEventListener)
-			window.addEventListener('beforeunload', function() { self.requestState({state: 'closing'})});
+		if(options.closeOnUnload === true && window.addEventListener)
+			window.addEventListener('beforeunload', function() { self.requestState({state: 'closing'}); });
 
 		/* Listen for online and offline events */
 		if(typeof window === "object" && window.addEventListener) {
```
paddybyers: single quotes pls

common/lib/transport/connectionmanager.js
```diff
@@ -188,187 +211,94 @@ var ConnectionManager = (function() {
...
+		Utils.arrDeleteValue(this.proposedTransports, transport);
 		this.pendingTransports.push(transport);
 
 		var self = this;
-		transport.on('connected', function(error, connectionKey, connectionSerial, connectionId, clientId) {
+		transport.once('connected', function(error, connectionKey, connectionSerial, connectionId, clientId) {
 			if(mode == 'upgrade' && self.activeProtocol) {
-				self.scheduleTransportActivation(transport, connectionKey);
+				/*  if ws and xhrs are connecting in parallel, delay xhrs activation to let ws go ahead */
+				if(transport.shortName !== 'web_socket' && Utils.arrIn(self.getUpgradePossibilities(), 'web_socket')) {
```
paddybyers: I'd prefer if this literal use of `web_socket` is removed from here. Maybe we define `var OPTIMAL_TRANSPORT = 'web_socket'` at the top of the file? Or it's even taken to be the last element of the transport short names array?

common/lib/transport/connectionmanager.js
```diff
@@ -809,70 +785,217 @@ var ConnectionManager = (function() {
...
-		var self = this;
-		var auth = this.realtime.auth;
-		var connectErr = function(err) {
-			err = ErrorInfo.fromValues(err);
-			Logger.logAction(Logger.LOG_ERROR, 'ConnectionManager.connectImpl()', 'Connection attempt failed with error; err = ' + err.toString());
-			var state = self.state;
-			if(state == self.states.closing || state == self.states.closed || state == self.states.failed) {
-				/* do nothing */
-				return;
+		if(auth.method == 'basic') {
```
paddybyers: You've been pretty consistently using `===`

common/lib/transport/connectionmanager.js
```diff
@@ -809,70 +785,217 @@ var ConnectionManager = (function() {
...
+			Logger.logAction(Logger.LOG_MINOR, 'ConnectionManager.connectImpl()', 'Transports ' + this.pendingTransports[0].toString() + ' currently pending; taking no action');
+		} else if(state == this.states.connected.state) {
+			this.upgradeIfNeeded(transportParams);
+		} else if(this.transports.length > 1 && getTransportPreference()) {
+			this.connectPreference(transportParams);
+		} else {
+			this.connectBase(transportParams);
+		}
+	};
+
```
paddybyers: extra line

common/lib/transport/connectionmanager.js
```diff
@@ -809,70 +785,217 @@ var ConnectionManager = (function() {
...
-					tryConnect();
-			});
+		this.tryATransport(transportParams, this.baseTransport, hostAttemptCb);
+	};
+
+
+	ConnectionManager.prototype.getUpgradePossibilities = function() {
+		var current = this.activeProtocol.getTransport().shortName;
+
+		if(current === 'web_socket') {
```
paddybyers: I'm wondering if there's a way for this part of the logic to be less hard-coded.

spec/browser/connection.test.js
```diff
@@ -218,6 +224,61 @@ define(['ably', 'shared_helper'], function(Ably, helper) {
...
 
+	exports.persist_preferred_transport = function(test) {
+		test.expect(1);
+		/* ensure we start with a blank slate */
+		helper.clearTransportPreference();
+
+		var realtime = helper.AblyRealtime();
+
+		realtime.connection.connectionManager.on(function(transport) {
+			if(this.event === "transport.active" && transport.shortName === 'web_socket') {
```
paddybyers: single quotes

paddybyers:
A few comments but otherwise LGTM.

SimonWoolf:
> We should have some documentation somewhere that captures the description above - where should that be? ETA: the handling of fallback hosts needs to be added to that description.

I've put most of that description above `connectImpl` in connectionmanager.js. I figure if anyone's trying to work out how connections work they'll come across it. I'll add some fallback host docs to connectBase.

> Is it possible that the node comet transport will run behind a proxy that breaks streaming?

It's a good question. I don't know. For the PR I took the view that we haven't had any reports of it, so no point increasing the complexity of the library to solve a problem that might not exist. Wouldn't be hard to add if it ever becomes necessary.

browser/lib/transport/xhrpollingtransport.js
SimonWoolf: Not a clue. Not my addition, it's been part of xhrtransport.js [since the initial commit](https://github.com/ably/ably-js/commit/8006e0c#diff-1b55013b80a5912a628e0c5b392b4d6cR36)

browser/lib/util/defaults.js
SimonWoolf: Because we want it to work in IE8 if someone specifies `transports: ['xhr_streaming']`. IE [no pun intended] we only want to stop upgrading from polling to streaming, not using streaming-only

SimonWoolf: > [no pun intended]

(ok that was a lie)

common/lib/transport/connectionmanager.js
SimonWoolf: :+1:

common/lib/transport/connectionmanager.js
SimonWoolf: Oops, yep

(it's the switching back and forth from js to elixir (which uses == and double-quotes for utf8 strings) that does it :) )

common/lib/transport/connectionmanager.js
SimonWoolf: I left an extra line between all the new methods I added, just to make it a bit clearer. Happy to remove if you prefer

common/lib/transport/connectionmanager.js
SimonWoolf: :+1:

common/lib/transport/connectionmanager.js
SimonWoolf: :+1: Added `var optimalTransport = transportPreferenceOrder[transportPreferenceOrder.length - 1];`

common/lib/transport/connectionmanager.js
SimonWoolf: Hmm. The underlying logic is: "return the subset of upgradeTransports to the right of the current transport (if it's there, if not return upgradeTransports)". So maybe something like
```
var currentPosition = Utils.arrIndexOf(this.upgradeTransports, current);
return this.upgradeTransports.slice(Math.max(currentPosition, 0));
```
?

SimonWoolf: Just tried it out, above is off-by-one, correct code would be
```
var currentPosition = Utils.arrIndexOf(this.upgradeTransports, current);
return this.upgradeTransports.slice(currentPosition + 1)
```
Which, yeah, is nicer than hardcoding, so I'll add that in.

SimonWoolf:
Added 4c8b087 to address Paddy's feedback

browser/lib/util/defaults.js
mattheworiordan: > (ok that was a lie)

:)

SimonWoolf:
Merged as 3c6edb9

PR #283, opened 2016-06-08
Disable JSONP transport if document is undefined.

### LINKED ISSUE #280: React-native
Can we support React Native using our ably-js lib, or do we need to use the native libraries under the hood.

Given `async` HTTP requests and `WebSockets` are supported https://facebook.github.io/react-native/docs/network.html, it should, in theory, be quite straightforward to add react native support.

One to review.

tcard:
This makes the library compatible with React Native.

Fixes #280.

### DIFF
--- a/browser/lib/transport/jsonptransport.js
+++ b/browser/lib/transport/jsonptransport.js
@@ -8,7 +8,8 @@ var JSONPTransport = (function() {
 	 */
 	_._ = function(id) { return _['_' + id] || noop; };
 	var idCounter = 1;
-	var head = document.getElementsByTagName('head')[0];
+	var isSupported = (typeof(document) !== 'undefined');
+	var head = isSupported ? document.getElementsByTagName('head')[0] : null;
 
 	/* public constructor */
 	function JSONPTransport(connectionManager, auth, params) {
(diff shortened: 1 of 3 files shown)


browser/lib/transport/jsonptransport.js
```diff
@@ -17,8 +18,10 @@ var JSONPTransport = (function() {
 	}
 	Utils.inherits(JSONPTransport, CometTransport);
 
-	JSONPTransport.isAvailable = function() { return true; };
-	ConnectionManager.httpTransports['jsonp'] = ConnectionManager.transports['jsonp'] = JSONPTransport;
+	JSONPTransport.isAvailable = function() { return isSupported; };
+	if(isSupported) {
```
paddybyers: If we do not make this change (ie we still add the `jsonp` member) then if anyone attempts to request jsonp they will get an error saying it isn't supported (which I think is preferable), instead of saying it's unrecognised.

paddybyers:
One comment but otherwise LGTM.

@SimonWoolf ?

browser/lib/transport/jsonptransport.js
SimonWoolf: I disagree, I think what's been done is correct. All other transports only add themselves to the `transports` array if they're supported (and connectionManager occasionally assumes that). If someone using React uses `transports: ['jsonp']` (though I don't imagine anyone would), they'll get a "*no requested transports available*" error, same as if someone using IE8 specified `transports: ['web_socket']`.

SimonWoolf:
LGTM (though @tcard as a rule we generally don't check in generated files until we release a new version, eg so merge conflicts etc only have to be solved in one place rather than many).

I'll merge this in once the transport fallback changes have been merged in, as I'll need to rebase this on top of that (transport array names have changed).

browser/lib/transport/jsonptransport.js
paddybyers: > I disagree, I think what's been done is correct.

Fair enough. Maybe I should go back and read my own code :(

SimonWoolf:
Merged as 79f93f0

mattheworiordan:
👍

mattheworiordan:
@SimonWoolf mind adding something to Readme about the library supporting React Native & also Cordova?

SimonWoolf:
@mattheworiordan done https://github.com/ably/ably-js/commit/372b62e02d6e3740c8a42f1f56504cfbdb6e1c62#diff-04c6e90faac2675aa89e2176d2eec7d8R7

mattheworiordan:
Thanks @SimonWoolf

PR #291, opened 2016-06-21
Pass along CONNECTED errors

SimonWoolf:
Was putting this off until the untilAttach, but pulled forward due to https://github.com/ably/realtime/issues/589

For connected errors in upgrades, I did the workaround described in https://github.com/ably/realtime/issues/587#issuecomment-227158648 - ie syncing with the new (-1) connectionSerial after a failed upgrade. This is a bit hacky, but fixes the problem from that issue for now. If we decide to change realtime behaviour (eg to not require a sync in the case of a failed upgrade), then we can change it then.

### DIFF
--- a/spec/realtime/connection.test.js
+++ b/spec/realtime/connection.test.js
@@ -104,5 +104,25 @@ define(['ably', 'shared_helper'], function(Ably, helper) {
 		}
 	};
 
+	exports.unrecoverableConnection = function(test) {
+		test.expect(4);
+		var realtime,
+			fakeRecoveryKey = '_____!ablyjs_test_fake-key____:5';
+		try {
+			realtime = helper.AblyRealtime({recover: fakeRecoveryKey});
+			realtime.connection.on('connected', function(stateChange) {
+				console.log(JSON.stringify(stateChange));
+				test.equal(stateChange.reason.code, 80008, "verify unrecoverable-connection error set in stateChange.reason");
+				test.equal(realtime.connection.errorReason.code, 80008, "verify unrecoverable-connection error set in connection.errorReason");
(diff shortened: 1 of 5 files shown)


spec/realtime/connection.test.js
```diff
@@ -104,5 +104,25 @@ define(['ably', 'shared_helper'], function(Ably, helper) {
...
+		var realtime,
+			fakeRecoveryKey = '_____!ablyjs_test_fake-key____:5';
+		try {
+			realtime = helper.AblyRealtime({recover: fakeRecoveryKey});
+			realtime.connection.on('connected', function(stateChange) {
+				console.log(JSON.stringify(stateChange));
+				test.equal(stateChange.reason.code, 80008, "verify unrecoverable-connection error set in stateChange.reason");
+				test.equal(realtime.connection.errorReason.code, 80008, "verify unrecoverable-connection error set in connection.errorReason");
+				test.equal(realtime.connection.serial, -1, "verify serial is -1 (new connection), not 5");
+				test.equal(realtime.connection.key.indexOf('aaaaaaaaaaaaaaaa'), -1, "verify connection using a new connectionkey");
```
paddybyers: I don't understand the `aaaaaaaaaaaaaa`

paddybyers:
LGTM except for one comment

spec/realtime/connection.test.js
SimonWoolf: argh oops. Right before I put this up I decided to change the fake recovery key from `aaaaa!aaaaaaaaaaaaaaaa-aaaaaaa:5` to `_____!ablyjs_test_fake-key____:5` (to make it obvious to anyone looking in the realtime sandbox logs and seeing failed resumes what was going on). Forgot to change that line :(

Issue #285, opened 2016-06-12
Quick detach never detaches


mattheworiordan:
See https://github.com/ably/ably-js/issues/284, I had to place a timeout in the test as without a timeout the channel never became detached.

See https://jsbin.ably.io/amisux/4/edit

Note how the channel never becomes detached, and there are Comet request errors.  I assume this issue has something to do with the DETACH being sent during upgrade?

SimonWoolf:
@mattheworiordan probably, though in theory, activating the upgrade transport is supposed to be postponed until attaches and detaches are complete (the `nopending` mechanism). Unfortunately I can't replicate the problem from here (even forcing ably-0.8.19.js to avoid the new preferredTransport stuff). Could you post a complete console log of it happening?

mattheworiordan:
I can't replicate this either n ow :(

SimonWoolf:
Ah well. If it happens again, reopen with a log.

SimonWoolf:
Reopening - I managed to reproduce this. The bug occurs when the presence set is nonempty, so there is a presence sync that overlaps with the detach. The `nopending` mechanism is effectively just a boolean lock per channel, so when the presence sync finishes it unlocks the lock, even though the detach hasn't finished.

(Fixing https://github.com/ably/realtime/issues/522#issuecomment-225596458 will probably require major changes to the `nopending` stuff anyway, so I'll roll the fix for this into whatever we decide on there)

SimonWoolf:
Fixed by https://github.com/ably/ably-js/pull/292

Issue #286, opened 2016-06-12
Invalid transport every time


mattheworiordan:
See https://jsbin.ably.io/amisux/3/edit

Note in the logs every time the page is loaded I see:

```
VM1460 ably.js:4245 Ably: ConnectionManager.sendQueuedMessages(): sending 0 queued messages
2016-06-13 00:31:34.375 rest.ably.io/comet/1087cd81!lDHhsQiS1JKadyKd-30ab1087cd81/send?key=Q2ZabQ.CQcRYg%3AN17LB4Ave2eOYnfA&rnd=5236846233075232:1 POST https://rest.ably.io/comet/1087cd81!lDHhsQiS1JKadyKd-30ab1087cd81/send?key=Q2ZabQ.CQcRYg%3AN17LB4Ave2eOYnfA&rnd=5236846233075232 404 (Not Found)
2016-06-13 00:31:34.376 VM1460 ably.js:4245 Ably: CometTransport.sendItems(): on complete: err = {"statusCode":404,"code":40400,"message":"Invalid transport id: 30ab1087cd81","serverId":"frontend.24ab.eu-west-1-A.i-f2c5567e"}
2016-06-13 00:31:34.377 VM1460 ably.js:4245 Ably: Transport.onProtocolMessage(): received; [ProtocolMessage; action=ERROR; error=[ErrorInfo: Invalid transport id: 30ab1087cd81; statusCode=404; code=40400]]
2016-06-13 00:31:34.377 VM1460 ably.js:4245 Ably: Transport.onProtocolMessage(): received error action; connectionKey = 1087cd81!lDHhsQiS1JKadyKd-30ac1087cd81; err = {"message":"Invalid transport id: 30ab1087cd81","code":40400,"statusCode":404,"serverId":"frontend.24ab.eu-west-1-A.i-f2c5567e"}
```

This may very well be a realtime issue, but I am raising here first @SimonWoolf as I think you should investigate this.  Having 404 errors as part of a normal upgrade process clearly does not look good i.e. even if we know this has no effect on the upgrade, the process should not issue 4xx errors.

SimonWoolf:
This is a realtime error: https://github.com/ably/realtime/issues/522 . Tldr: comet requests sent before the upgrade process that arrive after are rejected by realtime, which disposes of the comet transport immediately on upgrading.

[...]

PR #292, opened 2016-06-22
Rework upgrade mechanism

SimonWoolf:
Implements option 3 from https://github.com/ably/realtime/issues/522#issuecomment-226157187 , per @paddybyers .

Fixes https://github.com/ably/realtime/issues/522 and https://github.com/ably/ably-js/issues/286 (from the client-side end)
Fixes https://github.com/ably/ably-js/issues/285

This PR stalls messages from the moment the sync is sent until the old transport is idle. Once/if the realtime changes in https://github.com/ably/realtime/issues/522 are done, we'll be able to do a further change to only stall once the sync is done.

The change to separate the upgrade sync and the activation of the upgrade transport (delaying the latter until the old transport is idle) required some other changes in the library. Eg previously a message that arrived on a transport that was not the active transport could be safely discarded; this is no longer true as messages will arrive on the upgrade transport as soon as the sync is done.

The channel operation inprogress mechanism needed enough surgery to accommodate the change that I put it in a separate commit:
- It onlys sets the lock once have it's actually sent an attaching or detaching  message. The purpose of the mechanism is for connectionManager to decide when it's safe to upgrade. Before, channel might be inProgress and in the attaching state without any pending operations, eg if attached while the connection state is connecting. Since the upgrade transport now waits for nopending before activating, this caused a deadlock if an attach request timed out: checkPendingState is waiting for transport.active, which is waiting for nopending. Now, setting inProgress is only done once the a message has been sent.
- It now starts the statetimer when the message is sent (if it's not already running) - before, if an attach was queued due to not being connected, the timer would never be set, even on connected.
- The on('transport.active') checkpendingstate now happens a tick after transport.active, to give connectionManager time to finish activating the transport and setting the connection state.
- It also now treats channelstate operations (attach/detach) and sync separately, which solves the bug where a sync ending would incorrectly remove the lock of a pending detach (#285).

### DIFF
--- a/common/lib/client/realtime.js
+++ b/common/lib/client/realtime.js
@@ -30,7 +30,12 @@ var Realtime = (function() {
 		this.all = {};
 		this.inProgress = {};
 		var self = this;
-		realtime.connection.connectionManager.on('transport.active', function(transport) { self.onTransportActive(transport); });
+		realtime.connection.connectionManager.on('transport.active', function() {
+			/* nextTick to allow connectionManager to set the connection state to 'connected' if necessary */
+			Utils.nextTick(function() {
+				self.onTransportActive();
+			});
+		});
 	}
 	Utils.inherits(Channels, EventEmitter);
 
@@ -48,11 +53,17 @@ var Realtime = (function() {
 		channel.onMessage(msg);
 	};
 
-	/* called when a transport becomes connected; reattempt attach()
-	 * for channels that may have been inProgress from a previous transport */
+	/* called when a transport becomes connected; reattempt attach/detach
+	 * for channels that are attaching or detaching.
+	 * Note that this does not use inProgress as inProgress is only channels which have already made
+	* at least one attempt to attach/detach */
 	Channels.prototype.onTransportActive = function() {
-		for(var channelId in this.inProgress)
-			this.inProgress[channelId].checkPendingState();
+		for(var channelName in this.all) {
+			var channel = this.all[channelName];
+			if(channel.state === 'attaching' || channel.state === 'detaching') {
+				channel.checkPendingState();
+			}
+		}
 	};
 
 	Channels.prototype.setSuspended = function(err) {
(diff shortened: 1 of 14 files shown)


SimonWoolf:
@paddybyers I've rebased this now the other's been merged, so no longer have the extra commits

common/lib/client/realtime.js
```diff
@@ -48,11 +53,18 @@ var Realtime = (function() {
...
-	/* called when a transport becomes connected; reattempt attach()
-	 * for channels that may have been inProgress from a previous transport */
+	/* called when a transport becomes connected; reattempt attach/detach
+	 * for channels that are attaching or detaching.
+	 * Note that this does not use inProgress as inProgress is only channels which have already made
+	* at least one attempt to attach/detach */
 	Channels.prototype.onTransportActive = function() {
-		for(var channelId in this.inProgress)
-			this.inProgress[channelId].checkPendingState();
+		var statechangeOp = RealtimeChannel.progressOps.statechange;
```
paddybyers: This is redundant?

common/lib/client/realtime.js
```diff
@@ -80,24 +92,32 @@ var Realtime = (function() {
...
-			if(Utils.isEmpty(this.inProgress)) {
-				this.emit('nopending');
-			}
+	/* Records operations currently pending on a transport; used by connectionManager to decide when
+	 * it's safe to upgrade. Note that a channel might be in the attaching state without any pending
+	 * operations (eg if attached while the connection state is connecting) - such a channel must not
+	 * hold up an upgrade, so is not considered inProgress.
+	 * Operation is currently one of either 'statechange' or 'sync' */
+	Channels.prototype.setInProgress = function(channel, operation, bool) {
+		this.inProgress[channel.name] = this.inProgress[channel.name] || {};
```
paddybyers: what's `bool` ? Can't we have a meaningful identifier?

paddybyers:
This looks good, although I'd be more confident if we had a bit more test coverage

common/lib/client/realtime.js
SimonWoolf: Oops, thanks. (That line was left from an earlier revision where I was checking `inProgress[channeld][statechangeOp]` in this method, rather than just the channel state)

common/lib/client/realtime.js
SimonWoolf: Fair enough, I guess my reasoning for changing it away from `inProgress` was a bit flimsy. Changed back: 63be463

SimonWoolf:
Added https://github.com/ably/ably-js/pull/292/commits/1a2aa5339344bbfe549c2a3ad60f6db46d2fcb3a with a couple additional tests. Ci's still getting random tests failing with 'Unable to recover connection: not found', but I'm pretty sure at this point that that's sandbox-specific and due to https://github.com/ably/infrastructure/issues/711 - everything passing when run against local realtime.

paddybyers:
> Ci's still getting random tests failing with 'Unable to recover connection: not found', but I'm pretty sure at this point that that's sandbox-specific and due to ably/infrastructure#711 - everything passing when run against local realtime

If it's incorrectly routed to a different frontend in the same region, it should still be found, so we do need to get to the bottom of this. However, that's not a regression with this PR.

paddybyers:
> Added 1a2aa53 with a couple additional tests.

I'm happy for this to land.

Issue #297, opened 2016-06-29
Safari can throw an exception when you use localstorage, even if it looks enabled


SimonWoolf:
https://stackoverflow.com/questions/21159301/quotaexceedederror-dom-exception-22-an-attempt-was-made-to-add-something-to-st

https://app.intercom.io/a/apps/ua39m1ld/inbox/all/conversations/5196809049

wtf safari

SimonWoolf:
Fixed by 88fe5ab

mattheworiordan:
```
I still see the same error on safari I’m afraid. Opening [customer site] in a safari private window results in this error:

Ably: EventEmitter.emit(): Unexpected listener exception: Error: QuotaExceededError: DOM Exception 22; stack = setItem@[native code]
```

SimonWoolf:
Been wrestling with saucelabs' bloody slow interactive VMs connection to try this in a couple versions of safari before the 10 minute hard limit on manual sessions expires, but haven't been ably to reproduce it. All seems to work for me. Could you give some more details? Safari version, OS X version, line number from a nonminified lib, anything..?

mattheworiordan:
FYI, @SimonWoolf looks like the CDN had 0.8.22 on it, so that was prob the problem.  Waiting to hear from the customer as I could not replicate it.

wichert:
I can confirm this is fixed now!

SimonWoolf:
>  looks like the CDN had 0.8.22 on it

Ah bugger, I thought I'd deployed it on Friday, guess I didn't!

> I can confirm this is fixed now!

Glad to hear :)

PR #298, opened 2016-06-29
Fix local/sessionstorage may throw on access

SimonWoolf:
Fixes https://github.com/ably/ably-js/issues/297

This should work, but if anyone has a mac and can check it in safari in private mode, that'd be good.


SimonWoolf:
Merged as 88fe5ab

PR #304, opened 2016-07-06
Add "lib=js-x.x.xx" qs param to connectParams

SimonWoolf:


paddybyers:
LGTM

mattheworiordan:
@SimonWoolf any reason not to simply add the header at the same time, see https://github.com/ably/ably-js/blob/be8ac97a6b86df90805ceb740564915f446f95ee/common/lib/util/utils.js#L294?

SimonWoolf:
@mattheworiordan None at all (I vaguely thought we'd decided not to add the header for now on the call, but I think I just misremembered). https://github.com/ably/ably-js/pull/304/commits/c624fbdb17a7f4f7a8c9567deebd3481e24e1687

SimonWoolf:
Reverted adding the X-Ably-Lib header - 400d251. Needs to be added to the Access-Control-Allow-Headers list [here](https://github.com/ably/realtime/blob/master/common/lib/rest/router2.js#L51) first, or browsers will show CORS preflight errors.

mattheworiordan:
Have you got an issue to add that back in @SimonWoolf?

SimonWoolf:
looks like the frontend change made it into production at the weekend (though not sandbox), so I'll just add it back in master now

mattheworiordan:
> so I'll just add it back in master now

Ok, great, please do a release as well then.

SimonWoolf:
Released 0.8.26

PR #308, opened 2016-07-12
When retrying previously-attempted messages on a new transport, keep the same msgSerial

SimonWoolf:
Also stop newer messages from being merged into previously-attempted messages. (I originally just checked whether msg.msgSerial was defined to see if it'd previously been attempted, but having an explicit boolean field in the pendingMessage object is neater & more robust in case we change how msgSerials are done in the future)


paddybyers:
LGTM

SimonWoolf:
Merged as f035cd2

PR #313, opened 2016-07-13
Make server-sent DETACHED always detach, not fail a channel

SimonWoolf:
Lib originally assumed that a DETACHED with an error property should fail the channel


mattheworiordan:
Thanks, looks good, we should bump and get this out as soon as possible.

SimonWoolf:
released 0.8.28

paddybyers:
LGTM, thanks

Issue #311, opened 2016-07-13
Quick reconnect after token TTL expires gets stuck in connecting


mattheworiordan:
See https://jsbin.ably.io/icapud/2/edit

It connects fine, gets a `message: "Key/token status changed (expire)", code: 40142, statusCode: 401}` as expected, but then when it returns to the connecting state there does not appear to be a connection attempt made and it just stays in that state for a short period of time?

SimonWoolf:
@mattheworiordan did you link the wrong jsbin? That one seems to just be demonstrating that if you change the clientId and resume, realtime'll reject it with a `Mismatched clientId for existing connection` error; the connection seems to be made fine.

```
Created token request: {"keyName":"xVLyHw.V4EYrw","ttl":5000,"clientId":"client1","log":{"level":3},"capability":"","timestamp":1468410087097,"nonce":"4015014724873799","mac":"1gPpP3Bu0SCMGG6UcobJ5FwLPQNVizrk5LEAk21r9fA="}
Connecting
Connected
Created token request: {"keyName":"xVLyHw.V4EYrw","ttl":5000,"clientId":"client2","log":{"level":3},"capability":"","timestamp":1468410093006,"nonce":"3928446953677815","mac":"UMXfqBlDjZYbjRhHMBpXqWmZml5bi7lNdGb4k/xlRlw="}
Connecting
Connection failed: {"previous":"connecting","current":"failed","reason":{"message":"Mismatched clientId for existing connection","code":40102,"statusCode":401}}
Current client ID: null
Removed clientId and reconnecting
Connecting
Connected
Created token request: {"keyName":"xVLyHw.V4EYrw","ttl":5000,"clientId":"client3","log":{"level":3},"capability":"","timestamp":1468410098498,"nonce":"6700454427210315","mac":"IgSb9TRoo834hc3KF47EhfJyZjFgkVSVfz3o5VCsAuA="}
Connecting
Connection failed: {"previous":"connecting","current":"failed","reason":{"message":"Mismatched clientId for existing connection","code":40102,"statusCode":401}}
Current client ID: null
Removed clientId and reconnecting
Connecting
Connected
Created token request: {"keyName":"xVLyHw.V4EYrw","ttl":5000,"clientId":"client4","log":{"level":3},"capability":"","timestamp":1468410104402,"nonce":"2753727702408557","mac":"ulirMn97PHY8+oRTCaYaW8X2nnS0SEE0owLckNZeXok="}
Connecting
[...]

PR #314, opened 2016-07-13
Emit any pre-connnect auth errors in a way the onConnect listeners will handle

SimonWoolf:
Fixes https://github.com/ably/ably-js/issues/311

@paddybyers would you mind explaining a couple things when you have a sec?

- What's the semantic difference between finishing a transport with `failed` (which is what aborting does) and finishing it with `error`? Any reason not to merge the two events, and avoid bugs like this where something's listening for one but not the other?

- You have a `/* FIXME: this should not be fatal */` in the websocket `onerror` handler. Why not? Is it possible it'll ever be called with a nonfatal error? (the w3 spec is [very vague](https://www.w3.org/TR/websockets/#handler-websocket-onerror), and the websocket spec isn't much help either - it defines a lot of codes for the close frame, but doesn't mention any nonfatal errors that I can find). And if nonfatal errors are possible, why have the abort there at all -- why not just log them and rely on the `onclose` being called immediately after if the error was fatal?

### DIFF
--- a/browser/lib/transport/jsonptransport.js
+++ b/browser/lib/transport/jsonptransport.js
@@ -54,10 +54,10 @@ var JSONPTransport = (function() {
 	JSONPTransport.tryConnect = function(connectionManager, auth, params, callback) {
 		var transport = new JSONPTransport(connectionManager, auth, params);
 		var errorCb = function(err) { callback(err); };
-		transport.on('error', errorCb);
+		transport.on('failed', errorCb);
 		transport.on('preconnect', function() {
 			Logger.logAction(Logger.LOG_MINOR, 'JSONPTransport.tryConnect()', 'viable transport ' + transport);
-			transport.off('error', errorCb);
+			transport.off('failed', errorCb);
 			callback(null, transport);
 		});
 		transport.connect();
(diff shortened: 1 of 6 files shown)


mattheworiordan:
Nice work @SimonWoolf

paddybyers:
> Any reason not to merge the two events, and avoid bugs like this where something's listening for one but not the other?

No, I don't think so. I might have had something in mind at the time, but it was a long time ago :)

> Is it possible it'll ever be called with a nonfatal error?

Again, I was obviously thinking of something at the time, but now I can't remember. Maybe some error conditions such as an individual frame exceeding maximum length or something like that needn't necessarily mean the transport needs to die. But in our protocol it would mean that, so I can't think of any reason now to make it non-fatal.

paddybyers:
LGTM BTW

SimonWoolf:
Replaced previous commit with 598fc8d which switches everything to `failed`, per discussion above.

@paddybyers done some more experimenting, and as far as I can tell, onerror is: almost always useless, and if the websocket was connected, always followed by an onclose which contains the actual reason for the error (eg in the exceeded-max-frame-size case it'd be 1009). Aborting the transport in an onerror was if anything harmful, as that disposed of it before the onclose event fires, so we don't find out what the ws close code was. Added de3de72, which puts off the abort into a nexttick (so if it was followed by an onClose, it'll already have been disposed on the onClose's terms by then), and changed onClose to fail the transport if the close was unclean (to match the current behaviour).

paddybyers:
ok, LGTM, thanks

PR #319, opened 2016-07-18
Fix presence messages ending an in-progress sync

SimonWoolf:
Issue reported by Larry where a large room would occasionally show far too few presence members (within a region - separate from inter-region sync issues). Turns out that a presence message arriving in the middle of a sync would interrupt the sync.

(I reused the `broadcast` parameter from `setPresence`, since it wasn't actually used anywhere in the function. I'm assuming it's leftover from some previous design for presence and is now defunct)

### DIFF
--- a/spec/realtime/presence.test.js
+++ b/spec/realtime/presence.test.js
@@ -1426,5 +1426,67 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
 		});
 	};
 
+	/*
+	 * Do a 110-member sync, so split into two sync messages. Inject a normal
+	 * presence enter between the syncs. Check everything was entered correctly
+	 */
+	exports.presence_sync_interruptus = function(test) {
+		test.expect(1);
+		var channelName = "presence_sync_interruptus";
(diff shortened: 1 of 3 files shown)


mattheworiordan:
LGTM

spec/realtime/presence.test.js
```diff
@@ -1426,5 +1426,67 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
 		});
 	};
 
+	/*
```
paddybyers: Should we maybe include tests where the mid-sync message is one of the sync'd members, including both the case that the sync for that member was received already, and the case that it is yet to be received?

paddybyers:
LGTM with a comment

spec/realtime/presence.test.js
SimonWoolf: Yep, we should, and we do need more tests for sync. Not directly relevant to this change, so since we're about to change how syncing works for the 0.9 spec, I'll add some more sync tests then.

Issue #318, opened 2016-07-17
Webpack support


SimonWoolf:
Need to:
- Check whether ably-js works with webpack, and if not, fix it. (May involve making a browser npm package)
- Add instructions to readme for using with webpack

patie:
@SimonWoolf https://gist.github.com/patie/59913678c178a7fa331ec91f21d6093f

PR #321, opened 2016-07-20
Webpack support + remove legacy compat libraries

mattheworiordan:
@SimonWoolf this resolves https://github.com/ably/ably-js/issues/318 and https://github.com/ably/ably-js/issues/307

I'm yet to see what impact it has on the CI build, but from testing locally it seems to be working fine so far.

cc @paddybyers


browser/lib/transport/jsonptransport.js
```diff
@@ -1,6 +1,6 @@
 var JSONPTransport = (function() {
 	var noop = function() {};
-	var _ = window.Ably._ = {};
+	var _ = Ably._ = {};
```
SimonWoolf: This will break the jsonp transport unfortunately, it currently relies on there being a `window.Ably` object.

common/lib/client/auth.js
```diff
@@ -1,7 +1,7 @@
 var Auth = (function() {
 	var isBrowser = (typeof(window) == 'object');
 	var crypto = isBrowser ? null : require('crypto');
-	var msgpack = isBrowser ? window.Ably.msgpack : require('msgpack-js');
+	var msgpack = (typeof require !== "function") ? Ably.msgpack : require('msgpack-js');
```
paddybyers: Can we use single quotes?

paddybyers:
> single quotes for paddy

yay. Great to know I added some value there :)

SimonWoolf:
Thanks Matt :)

:+1: except for the jsonp issue. Not sure what the best solution is there. I've added commit 84ddef1 which fixes the issue, though is a bit ugly (uses `window._ablyjs_jsonp` if we're using msgpack so we don't control the `window.Ably` object). WDYT?

browser/lib/transport/jsonptransport.js
```diff
@@ -94,7 +105,7 @@ var JSONPTransport = (function() {
 			params = this.params,
 			self = this;
 
-		params.callback = 'Ably._._(' + id + ')';
+		params.callback = jsonpcb + '._(' + id + ')';
```
mattheworiordan: Nice.  I had a niggling feeling the JSONP needed to be treated differently, but in the end skipped over it :/

mattheworiordan:
Great. @SimonWoolf do you want to merge & publish?

mattheworiordan:
@SimonWoolf should we merge this?  can you do it?

SimonWoolf:
Sorry, forgot about this

Issue #253, opened 2016-04-04
Allow >1 pending presence event


SimonWoolf:
Currently, `realtimePresence` assumes at most one presence event is pending (ie awaiting channel attach). For most usecases this is fine as clients will only enter themselves into presence, but technically the assumption is invalid since one client can enter many clientIds if using basic auth or a wildcard identity. So should probably be an array, ala `realtimeChannel#pendingEvents`

paddybyers:
Yes, well spotted.

SimonWoolf:
This is fixed in 0.9.

Issue #289, opened 2016-06-15
Presence subscribe with multiple events does not work


mattheworiordan:
See https://www.ably.io/documentation/realtime/presence#subscribe-event-array

Docs state it supports an array.

SimonWoolf:
Not sure why the docs say it supports an array, the spec specifies "a single action argument" http://docs.ably.io/client-lib-development-guide/features/#RTP6. If we want to support an array argument for subscribes, we should add it to the spec (for both presence and normal subscribes for consistency).

SimonWoolf:
This is fixed in 0.9

PR #324, opened 2016-08-01
0.9 spec

SimonWoolf:
First and largest tranche of 0.9 spec updates. Includes most things from https://github.com/ably/docs/pull/127 except for presence/sync changes (waiting for https://github.com/ably/wiki/issues/119 to be decided and for realtime changes). Also not included: AUTH message (waiting for https://github.com/ably/docs/pull/132 to be finalised and for realtime changes), heartbeat timeouts (waiting for https://github.com/ably/docs/pull/130 to be finalised). Only one test needed to be stubbed out to make the test suite pass: `channel_resumed_flag`. Other changes are done backwards-compatibly (eg it accepts both new and old comet error message styles).

Note that it is now a lot more strict about failing the connection for non-token ERROR messages than before, per the spec. So before we deploy this, need to make sure that all realtime errors that should not result in the client connection failing permanently are sent as DISCONNECTEDs.

Edit: this now contains the complete 0.9 spec

### DIFF
--- a/spec/realtime/presence.test.js
+++ b/spec/realtime/presence.test.js
@@ -1666,6 +1666,7 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
 			},
 			function(cb) {
 				channel.presence.subscribe(function(presmsg) {
+					channel.presence.unsubscribe();
 					test.equal(presmsg.action, 'leave', 'Check action was leave');
 					test.equal(presmsg.clientId, goneClientId, 'Check goneClient has left');
 					cb();


SimonWoolf:
Note: since this hasn't been reviewed yet, I've added the spec/presence updates to this PR rather than making a new one. bf7f5b0, d5e1945, 3782e4f. (Presence tests failing atm mostly due to message id format change not on sandbox, they pass against a local realtime with that change)

SimonWoolf:
(also now fixes https://github.com/ably/ably-js/issues/253 and https://github.com/ably/ably-js/issues/289)

SimonWoolf:
Rest of 0.9 now added to to this PR as well (authorize, heartbeats/maxidletimeout, AUTH, rest#request).

There'll need to be some test changes & a few more new tests once the realtime side of the AUTH changes is done; I'm leaving those till that (as hard to write tests before can actually try the behaviour - in particular not sure yet how the realtime implementation's going to behave with the sort of ultra-short-ttl tokens used in the test suite).

The rest#request spec required HttpPaginatedResponse#errorCode and HttpPaginatedResponse#errorMessage, which implies you get one of those whether on failure or success; but in every other js api in the lib, you get a (err, result) tuple, and errors are indicated by an ErrorInfo in err with result being null. Wasn't sure whether I should interpret the spec strictly or  more loosely and js-idiomatically. In the end went for the more loose & js-idiomatic way, let me know if you disagree @mattheworiordan / @paddybyers

SimonWoolf:
threw in the encoding/decoding interoperability tests for good measure: 9ae41cf

SimonWoolf:
Authorize is now compliant with https://github.com/ably/docs/pull/148 and works with all transports, plus added a few more reauth tests. (And connectionManager#onAuthUpdated is now almost a hundred lines long due to all the crap needed to make it work at any stage of transport upgrading etc. :sob: )

mattheworiordan:
@SimonWoolf sorry I've not reviewed, although arguably nothing to do right now as we're blocked I assume by @paddybyers until he is able to do the resume stuff yes?

SimonWoolf:
We can't release 0.9 until the serverside attached-flag work is done, but there's no reason not to review the code before then, if you have time :). (Attach flags aren't currently working, but the clientside code is done, and everything except attached flags is working against sandbox. Besides, s'just one fairly small commit out of 40).

mattheworiordan:
> but there's no reason not to review the code before then, if you have time

Time... yes, what a luxury

mattheworiordan:
e8f1564 👍

SimonWoolf:
Rebased on 0.8.41

common/lib/client/realtime.js
```diff
@@ -62,14 +62,39 @@ var Realtime = (function() {
...
 			}
 		}
 	};
 
-	Channels.prototype.setSuspended = function(err) {
+	Channels.prototype.reattach = function(reason) {
 		for(var channelId in this.all) {
 			var channel = this.all[channelId];
-			channel.setSuspended(err);
+      if(channel.state === 'attaching' || channel.state === 'attached') {
```
mattheworiordan: Odd spacing here

common/lib/client/realtimechannel.js
```diff
@@ -183,7 +181,18 @@ var RealtimeChannel = (function() {
 					}
 				});
 		}
-		this.setSuspended(RealtimeChannel.channelDetachedErr, true);
+	};
+
+	RealtimeChannel.prototype.autonomousAttach = function() {
+		console.log("Trying to autoreattaach")
```
mattheworiordan: Think this shouldn't have been committed

common/lib/client/realtimechannel.js
```diff
@@ -409,68 +423,47 @@ var RealtimeChannel = (function() {
...
 		this.presence.setAttached();
-		this.setState('attached');
+		this.notifyState('attached');
 	};
 
-	RealtimeChannel.prototype.setDetached = function(message) {
-		this.clearStateTimer();
+	RealtimeChannel.prototype.notifyState = function(state, reason) {
+		Logger.logAction(Logger.LOG_MICRO, 'RealtimeChannel.notifyState', 'name = ' + this.name + ', current state = ' + this.state + ', notifying state ' + state);
+    this.clearStateTimer();
```
mattheworiordan: Odd spacing here too

common/lib/client/realtimechannel.js
```diff
@@ -409,68 +423,47 @@ var RealtimeChannel = (function() {
...
-	RealtimeChannel.prototype.setSuspended = function(err, suppressEvent) {
-		if(this.state !== 'detached' && this.state !== 'failed') {
-			Logger.logAction(Logger.LOG_MINOR, 'RealtimeChannel.setSuspended', 'deactivating channel; name = ' + this.name + ', err ' + (err ? err.message : 'none'));
-			this.clearStateTimer();
-			this.presence.setSuspended(err);
-			if(!suppressEvent) {
-				this.setState('detached');
-			}
-			this.failPendingMessages(err);
+    if(state !== 'attached' && state !== 'attaching') {
```
mattheworiordan: And spacing here

common/lib/client/realtimechannel.js
```diff
@@ -409,68 +423,47 @@ var RealtimeChannel = (function() {
...
+      this.presence.setSuspended(reason || RealtimeChannel.invalidStateError);
+      this.failPendingMessages(reason || RealtimeChannel.invalidStateError);
+    }
+		if(state === 'suspended' && this.connectionManager.state.sendEvents) {
```
mattheworiordan: Out of interest, what is `state.sendEvents` representing?

common/lib/client/realtimechannel.js
```diff
@@ -409,68 +423,47 @@ var RealtimeChannel = (function() {
...
-	};
-
-	RealtimeChannel.prototype.setState = function(state, err) {
-		this.state = state;
-		if(err) {
-			this.errorReason = err;
+		if(reason) {
+			this.errorReason = reason;
 		}
+    var change = new ChannelStateChange(this.state, state, false /* TODO resumed */, reason);
```
mattheworiordan: Sorry, more spacing issues

common/lib/client/realtimechannel.js
```diff
@@ -409,68 +423,47 @@ var RealtimeChannel = (function() {
...
-
-		/* notify the state change, but don't set inProgress until the request is actually in progress */
-		this.setState(state);
+		this.state = state;
+		this.emit(state, change);
+	};
 
+	RealtimeChannel.prototype.requestState = function(state, reason) {
+		Logger.logAction(Logger.LOG_MINOR, 'RealtimeChannel.requestState', 'name = ' + this.name + ', state = ' + state);
+    this.notifyState(state, reason);
```
mattheworiordan: and more spacing... won't comment on any more from this commit...

common/lib/client/realtimechannel.js
```diff
@@ -505,12 +498,12 @@ var RealtimeChannel = (function() {
...
 				var err = new ErrorInfo('Channel attach timed out', 90000, 408);
-				this.setState('detached', err);
-				this.failPendingMessages(err);
+				this.notifyState('suspended', err);
+				this.failPendingMessages(err); // RTP5f
 				break;
 			case 'detaching':
 				var err = new ErrorInfo('Channel detach timed out', 90000, 408);
-				this.setState('attached', err);
+				this.notifyState('attached', err);
```
mattheworiordan: I vaguely recall discussing this, but wonder if we should modify the spec to state that going from detaching to attached always resends the `ATTACH` protocol message in case somehow, the `DETACH` arrived but the client never got the `DETACHED`.  By simply resending `ATTACH` for this edge case we rule out any possibility that the two are not kept in sync.  WDYT? cc @paddybyers

common/lib/client/realtimepresence.js
```diff
@@ -360,6 +360,23 @@ var RealtimePresence = (function() {
 		return result;
 	};
 
+	function newerThan(item, existing) {
+		/* RTP2b1: if either is synthesised, compare by timestamp */
+		if(item.isSynthesized() || existing.isSynthesized()) {
+			console.log("one is synth, returning ", item.timestamp > existing.timestamp)
```
mattheworiordan: Did you mean to commit this log?

spec/realtime/failure.test.js
```diff
@@ -333,5 +333,31 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
...
+		});
+
+		realtime.connection.once('connected', function() {
+			realtime.connection.once(function(statechange) {
+				test.equal(statechange.current, 'disconnected', 'check connection goes to disconnected');
+				test.equal(statechange.reason.code, 80003, 'check code');
+				test.equal(statechange.reason.statusCode, 408, 'check statusCode');
+				closeAndFinish(test, realtime);
+			});
+		});
```
mattheworiordan: Might be nice to have one test that ensures that we do get the heartbeat from Ably in a browser and in Node, WDYT?

common/lib/client/rest.js
```diff
@@ -108,6 +108,38 @@ var Rest = (function() {
...
+		}, /* useHttpPaginatedResponse: */ true);
+
+		switch(method) {
+			case 'get':
+				paginatedResource.get(params, callback);
+				break;
+			case 'post':
+				paginatedResource.post(params, body, callback);
+				break;
+			default:
```
mattheworiordan: Why not `put`?

SimonWoolf: https://github.com/ably/docs/pull/181#discussion_r85719182

common/lib/client/realtimechannel.js
SimonWoolf: all this stuff was already fixed in later commits. IIRC I tried to squash it down, but it didn't squash cleanly, didn't bother spending too much time working out why

common/lib/client/realtimechannel.js
SimonWoolf: Sounds reasonable to me.

spec/realtime/failure.test.js
SimonWoolf: Ideally sure, but I'd rather not add an ably-js test that takes 15s minimum to complete, (been trying to rewrite tests that require waits, not add more, test suite takes too long as it is). Might be a thing to add if in the future we add a connectparam that allows the lib to request a custom heartbeat period from realtime. (Anyway, if realtime stops sending heartbeats we'll know soon enough, given ably-js now fails the transport if it doesn't get a heartbeat...)

mattheworiordan: @SimonWoolf a simpler alternative would be to launch a simple WS server that sends the heartbeats and connection details with a very low heartbeat interval.   You could then very quickly test heartbeat protocol messages or native WS heartbeats.

common/lib/client/realtimechannel.js
mattheworiordan: @paddybyers WDYT?

common/lib/client/realtimechannel.js
SimonWoolf: whether you can send events in that state. (i.e. basically just `connected`). Leftover from the days when 'messages' were called 'events' (with a `queueEvents` client option, etc)

common/lib/client/realtimechannel.js
mattheworiordan: Channel state is decoupled from connection state (in most cases unless it becomes suspended / failed).  As such, it's very plausible that an `ATTACH` / `DETACH` could be sent to realtime, actioned by realtime, but the client never receives the ack on the current transport.  In the case that the transport and thus connection becomes disconnected, when the ACK has not been received, the connection will then reestablish a transport and continue from where it left off, yet now the client library does not know if the channel is `ATTACHED` or `DETACHED`. We have discussed this before, and I would suggest that any pending channel state request is resent immediately when the connection becomes connected again. I don't see any downside to this, and given there is no serial number change when for channel state changes, there is simply no way for a client library to know if the ATTACH / DETACH request was actually accepted or not.

paddybyers: https://github.com/ably/realtime/issues/751

spec/realtime/message.test.js
```diff
@@ -690,6 +690,34 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
 			});
 	};
 
+	/* For now realtime doesn't yet support the extras field, so stub transport.send to test it */
```
mattheworiordan: Perhaps worth a `TODO` to come back and fix this once supported?

SimonWoolf: https://github.com/ably/ably-js/issues/367

mattheworiordan: @SimonWoolf I would really like to see the spec codes in all tests if possible.  We do this now for iOS and Android, and I've done it for all new Ruby tests and retrospectively went back and added a few for existing tests when I touched those files / sections.  It makes it so much easier to figure out if a test exists for a test spec item, and later when specs change, it's easy again to search, find the test, and amend if necessary.  I am not sure how much work it would be, but it would be good to a) update the tests in `0.9 branch` with the spec codes, b) make sure the `ably-js` spreadsheet is up to date.  The reported test coverage still feels quite low, but expect that is not accurate?

mattheworiordan:
fc940ec looks good, nice 👍

PR #327, opened 2016-08-16
Fix rest fallback hosts functionality

SimonWoolf:
@mattheworiordan PR relevant to https://support.ably.io/helpdesk/tickets/1242 . Turns out ably-js never properly implemented the rest fallback spec in a bunch of ways. Fixed and added some tests.

Did it against the 0-9-spec branch because the new rest#request functionality was useful for writing the tests. I can backport the fixes to 0.8 you think we should do that, once the PR's been reviewed.


browser/lib/util/http.js
```diff
@@ -3,6 +3,16 @@ var Http = (function() {
 
 	function Http() {}
 
+	function shouldFallback(err) {
+		var statusCode = err.statusCode;
+		/* 400 + no code = a generic xhr onerror. Browser doesn't give us enough
+		 * detail to know whether it's fallback-fixable, but it may be (eg if a
+		 * network issue), so try just in case */
+		return (statusCode === 408 && !err.code) ||
+			(statusCode === 400 && !err.code)      ||
```
mattheworiordan: Silly comment, but odd spacing here

mattheworiordan:
@SimonWoolf this looks good, but would like @paddybyers to review.  I think given this is a pretty significant bug, we should port this back to `0.8`.

One comment, host fallbacks should be randomised.  Probably easiest to fix in https://github.com/ably/ably-js/blob/master/common/lib/util/defaults.js#L49, WDYT?

mattheworiordan:
Oh, and thanks, 👍

paddybyers:
LGTM, thanks

browser/lib/util/http.js
SimonWoolf: Heh. Lines up correctly if viewed in an editor with tabstop=2 :P

lining up things on different levels of indentation when indented with tabs and the tabwidth might change isn't possible in general, but I'll kludge it with spaces...

mattheworiordan: > lining up things on different levels of indentation when indented with tabs and the tabwidth might change isn't possible in general, but I'll kludge it with spaces...

But I am pretty sure the consensus is that as a result you use tabs to left align your code based on the indentation, and spaces when indenting relative to another tabbed indentation for exactly this reason?

SimonWoolf:
> One comment, host fallbacks should be randomised. Probably easiest to fix in https://github.com/ably/ably-js/blob/master/common/lib/util/defaults.js#L49, WDYT?

:+1:  	8d85d91

Will backport to 0.8 now

browser/lib/util/http.js
paddybyers: Tabs are for leading indentation only; never anywhere except at the start of a line. Then, spaces only where whitespace is needed for syntax/readability.

All other spaces are vanity spacing only :) If you do want to do that, it only ever applies among lines that have equal numbers of initial tabs. If things don't line up the way you like, deal with it :)

SimonWoolf: That's what I meant by 'kludge it with spaces' (and what https://github.com/ably/ably-js/commit/68cf9ca1896015240e3db375566ccabd10e8f8f5 did), sorry if that wasn't clear.

IMHO it still a kludge because you end up doing
```
<tab>return foo ||
<tab>  bar      ||
```
only when you do want to align ||'s; if not, then you leave it at what your editor does automatically, ie
```
<tab>return foo ||
<tab><tab>bar;
```
so the beginning-of-line whitespace on the remaining lines changes depending on whether there's something you want to align at the end of the line, hence kludgy

SimonWoolf: > Tabs are for leading indentation only; never anywhere except at the start of a line.

They were at the start of the line. Since the return statement was split over multiple lines, all except the first line were indented one level relative to the first, which was done with tabs. (But isn't any longer after the fixup)

mattheworiordan:
👍

browser/lib/util/http.js
paddybyers: > hence kludgy

> all except the first line were indented one level relative to the first, which was done with tabs. 

.. which was correct

> (But isn't any longer after the fixup)

Well that's wrong. The number of tabs is defined by indentation level, as you had it, not by the demands of vanity spacing. If your vanity spacing needs are then unsatisfiable, then just invest your aesthetic energy somewhere else :)

SimonWoolf:
Merged into 0.8 as 6965584

Issue #330, opened 2016-09-06
Review agent options for http


paddybyers:
Please check what we're doing with http agent options in the nodejs library. It looks like rest requests use the default global agent which has persistent connections off by default. If that's the case, we need to set up an agent that enables them by default.

PR #331, opened 2016-09-07
Use foreverAgent for node http requests

SimonWoolf:
The lib will generally be making requests repeatedly to one or two servers only (usually just Ably), so for efficiency, can use the foreverAgent to keep the TCP stream alive between requests.

Fixes https://github.com/ably/ably-js/issues/330

(hard to test, but I've checked in wireshark and the option seems to do what it should be doing)

### DIFF
--- a/nodejs/lib/util/http.js
+++ b/nodejs/lib/util/http.js
@@ -89,7 +89,10 @@ this.Http = (function() {
 	 * @param callback (err, response)
 	 */
 	Http.getUri = function(rest, uri, headers, params, callback) {
-		var getOptions = {headers:headers, encoding:null};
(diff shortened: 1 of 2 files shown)


paddybyers:
LGTM, assuming of course we've tested it :)

mattheworiordan:
Nice, and is this "forever" or does forever have a timeout eventually?

SimonWoolf:
Updated the `request` module, which fixes the odd behaviour I've been seeing with this.

The client never closes it (unless/until the server closes it or it receives a response with `connection: close`). But ELB terminates TCP stream after 90s of inactivity.

PR #340, opened 2016-10-11
Node build without crypto and buffertools

SimonWoolf:
That was a lot easier than I was expecting. Turns out node wasn't actually using either buffertools or crypto-js at all at runtime. So could basically just move them both from dependencies to devDependencies

### DIFF
--- a/nodejs/lib/transport/nodecomettransport.js
+++ b/nodejs/lib/transport/nodecomettransport.js
@@ -1,6 +1,5 @@
 "use strict";
 var NodeCometTransport = (function() {
-	var buffertools = require('buffertools');
 	var http = require('http');
 	var https = require('https');
 	var url = require('url');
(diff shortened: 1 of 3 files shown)


spec/common/modules/shared_helper.js
```diff
@@ -11,6 +11,31 @@ define(['spec/common/modules/testapp_module', 'spec/common/modules/client_module
...
+				cmp = buf1[i] - buf2[i];
+				if(cmp != 0) return cmp;
+			}
+			return 0;
+		};
+
+		var nodeBufferCompare = function(buf1, buf2) {
+			if(!buf1) return -1;
+			if(!buf2) return 1;
+			return require('buffertools').compare(buf1, buf2);
```
paddybyers: But since this was written we have https://nodejs.org/api/buffer.html#buffer_buf_compare_target_targetstart_targetend_sourcestart_sourceend

for node > 0.11.13

SimonWoolf:
even easier then. We can even keep the min required node version at 0.10.x, since bufferCompare is only called by tests.

SimonWoolf:
Merged as 2ec5f45

mattheworiordan:
Nice 👍

PR #342, opened 2016-10-13
Disable upgrade to xhr_streaming if cloudflare headers detected

SimonWoolf:


paddybyers:
> Do we? I'm not sure how we're planning on using cloudflare -- mightn't we want to enable it on all endpoints in some circumstances (eg DDOS attack)?

We might, but my point is that once you have detected it on one host it doesn't then mean you want that restriction, forever, on all hosts.

> Are we anticipating that people might have multiple instances of the library connecting to different endpoints?

I'm more concerned about the possibility that it never gets re-enabled, than I am about the issue of multiple instances sharing state.

SimonWoolf:
@paddybyers 

> We might, but my point is that once you have detected it on one host it doesn't then mean you want that restriction, forever, on all hosts.

Fair enough. Though given you have to reinitialize the lib if you want to change host, making it per-instance should suffice, right? 406c475

paddybyers:
> no need to also make it re-enablable for a given lib instance

But a single library instance might be required to connect to a fallback, and then later be able to use the normal host. You don't think we should try to make streaming work again in that case?

SimonWoolf:
Right, yes, of course. Sorry, was being stupid. 5005a70

SimonWoolf:
That would work, but TBH I quite like that at the moment the connectionManager instance doesn't know anything about the current host (it's in the transportParams instance which is created by startConnect and passed around to the things that need it). So I'd rather not set properties on connectionManager that depend on what the current host being passed around; this seems cleaner even if it's a couple lines more code

PR #344, opened 2016-10-20
Add Message#fromEncoded

SimonWoolf:
No new tests as made the crypto `testEachFixture` helper use it, which exercises it pretty thoroughly on the crypto fixture data.


mattheworiordan:
Would be nice to test the array method too no?

mattheworiordan:
Otherwise LGTM, I'll reproduce this in Ruby REST 0.9 as well

SimonWoolf:
Merged as 04bada6

Issue #333, opened 2016-09-13
useBinaryProtocol defaults


SimonWoolf:
Currently (since https://github.com/ably/ably-js/commit/b94bdaf) ably-js defaults to json rather than msgpack unless you explicitly set useBinaryProtocol to true. Paddy made the change as msgpack was slow in browsers. http://docs.ably.io/client-lib-development-guide/features/#TO3f wants the default to be binary. Options are:
1. default to binary everywhere it's supported
2. default to binary for node and json for browsers
3. leave as-is (default to json for everyone)

paddybyers:
I vote 2

SimonWoolf:
then we'll do option 2, docs pr: https://github.com/ably/docs/pull/161

SimonWoolf:
(reopening until actually implemented)

SimonWoolf:
This was fixed by 12e1f2f

Issue #346, opened 2016-10-31
Platform object


mattheworiordan:
As we are starting to extend out ably-js to support lots more platforms (NativeScript, Titanium, etc), I think our logic that is scattered around (where necessary) that evaluates statements like `if (isBrowser) { ... }` should be feature driven instead.

So I propose that prior to the NativeScript work being done, we consider introducing a new Platform abstract / default object, that is then extended based on the detected platform.  So instead of `isBrowser` we could do something like `if (Platform.supportsXHR())`, `if (Platform.supportWebsocketPings())`, if `Platform.supportsPushNotifications())` etc.

WDYT?

BTW. We could even make this object exposed (pseudo-privately) so that we could use it to more easily enable / disable functionality in tests.

paddybyers:
We already have abstracted at lot of this with APIs such as `BufferUtils.supportsBinary()` and eg `WebSocketTransport.isAvailable()` for each of the transports but there's no harm in collocating them.

mattheworiordan:
> We already have abstracted at lot of this with APIs such...

Yup, but unfortunately still [a lot of `isBrowser` checks](https://github.com/ably/ably-js/search?utf8=%E2%9C%93&q=isBrowser), and I expect other similar feature checks scattered throughout.

SimonWoolf:
Sure, I've no objection to this

mattheworiordan:
We now have a Platform object, see https://github.com/ably/ably-js/pull/349

PR #349, opened 2016-11-08
React support using Platform object

SimonWoolf:
- Hopefully adds react native support per https://github.com/ably/ably-js/issues/280
- Adopts @mattheworiordan 's suggestion in https://github.com/ably/ably-js/issues/346 of a `Platform` object to abstract the differences between browser and react-native
- Also fixes https://github.com/ably/ably-js/issues/333

@tcard if you could try this out and tell me if the react native build works, that'd be good (to try it out, run `grunt concat:ably-reactnative`, it'll output `browser/static/ably-reactnative.js`) -- I've done all your suggestions, but haven't actually tested it with an android/ios environment.

I've removed crypto support from the react native build, since as far as I can tell from the docs, there is no `global.crypto` object or any attempt at an equivalent, so the library has no access to cryptographically strong random numbers. There are projects that add that, e.g. https://www.npmjs.com/package/react-native-randombytes, but they're not 'pure js', rather wrappers over ios and android PRNGs, so not shippable in a pure js file. We can maybe look further into this if anyone asks for it - unless @tcard you have any ideas on this?

@paddybyers quick question -- what's the reasoning behind the presence of `TextDecoder` being the test for whether we support the binary protocol, given the lib doesn't actually use `TextDecoder` anywhere?


paddybyers:
> @paddybyers quick question -- what's the reasoning behind the presence of TextDecoder being the test for whether we support the binary protocol, given the lib doesn't actually use TextDecoder anywhere?

If you wanted to optimise a JS implementation of msgpack then you need to have support for binary arrays, and you need an efficient way to unpack utf8 slices of those arrays into strings. Currently that's hopelessly inefficient and you'd use TextDecoder. I was leaving it as an outstanding TODO to implement that, although from memory I thought I saw someone who'd forked the msgpack-js implementation to do that.

tcard:
Thanks @SimonWoolf ! I've added a small change and now it works :)

I'll see if I can make it so that you can just do `import Ably from 'ably'` instead of importing the concrete `ably-reactnative.js` file.

tcard:
I've managed to get crypto working at fe9718a. Unfortunately, I had to rewrite a lot of code to callback-passing style because React Native's `getRandomBytes` is async.

paddybyers:
> Unfortunately, I had to rewrite a lot of code to callback-passing style because React Native's getRandomBytes is async.

It was like that originally: https://github.com/ably/ably-js/commit/6d80c27f02705defeb6e63653cc01cc8c4fbd405

SimonWoolf:
> It was like that originally

Somewhat, though originally `getIv`, which is called on every message encrypt call, didn't require randomness (it now does, for good reasons), so generateRandom being async only affected channel setOptions rather than publishing etc. (And we had race conditions which were the reason I made it sync back then.)

@tcard's changes here address both of those (by making message encoding etc. async, and this `_waitingOptions`/`_optionsSet`/`afterOptions` mechanism to avoid the race conditions). And if we have to do that, so be it. But it looks like we might not have to -- @tcard since you did this work, the react-native-randombytes guy replied to say the sync api is actually supported now -- https://github.com/mvayngrib/react-native-randombytes/commit/7baa1745a4c6a87f1252ce95e08907c7d7bede68 . Do you think we should just use that? I know it seems a shame after you just wrote did all this work, but I'd rather not add complexity to the lib unless it's actually needed.

tcard:
@SimonWoolf Yes, I considered that, but it's a no-go because of https://github.com/mvayngrib/react-native-randombytes/issues/2 . As mentioned there that would've force our users to do a hack instead of just doing `npm install ably` and `require('ably')`. Not the best impression to offer just to spare us some callbacks I think. (In fact, I had to [fork the library](https://github.com/ably-forks/react-native-randombytes) and remove the synchronous API for it to work.)

SimonWoolf:
@tcard ah, that's unfortunate. OK, fair enough. I'll review this properly soon.

mattheworiordan:
Looks good guys, lot more work than I had expected

paddybyers:
It's really unfortunate that everything now needs a `_afterOptionsSet()` when afaict the only thing that is potentially async is generating a key, to create the CipherParams; when for all practical purposes that won't happen because people will pass in a key.

Is there no way we can have the async step just the creation of the `CipherParams` and/or the `ChannelOptions`, and once these are created they are passed synchronously to `Channels.get()` or `setOptions()` ?

tcard:
@paddybyers The key isn't generated, I recall we got rid of that functionality since it didn't make sense. It's the IV which is generated on `setOptions`, which I guess can be deferred until `encrypt` or `decrypt` is called. (Although I've managed to keep `decrypt` sync, so I would have to make it async if we make `setOptions` sync.)

paddybyers:
> It's the IV which is generated on setOptions, which I guess can be deferred until encrypt or decrypt is called.

There's no longer any need to create the IV when you create the cipher, because a new one is needed fo every message.

> (Although I've managed to keep decrypt sync, so I would have to make it async if we make setOptions sync.)

No, because `decrypt` gets the IV from the message itself - it doesn't need to generate it.

tcard:
@paddybyers OK, done that.

paddybyers:
> @paddybyers OK, done that.

Looks good, thanks. @SimonWoolf you need to review as well pls.

package.json
```diff
@@ -3,12 +3,19 @@
...
+    "./nodejs/index.js": "browser/static/ably-commonjs.js"
+  },
   "dependencies": {
     "async": "git://github.com/ably-forks/async#requirejs",
     "hexy": "~0.2",
     "msgpack-js": "git://github.com/paddybyers/msgpack-js.git",
     "request": "^2.74.0",
-    "ws": "^1.1.1"
+    "ws": "^1.1.1",
+    "react-native-randombytes": "git+https://github.com/ably-forks/react-native-randombytes.git"
```
SimonWoolf: I think this should be in devDependencies, since it's not a dependency of the node version and is only used to compiling ably-reactnative.js

tcard: This is actually used in runtime, and since it has native Obj-C and Java files we can't just concat the dependency on ably-reactnative.js. It's unfortunate that we force the dependency to every platform using this library, but as there's no way I know of of having platform-specific dependencies the only alternative I can think of is to have a separate `ably-react-native` library that wraps `ably-js` and provides the Platform object.

SimonWoolf:
@tcard :+1: I'm good with this - thanks. Just a couple of comments, and get the tests passing (as you fixed some tests but not others, I'm guessing you haven't been running the browser tests -- you can do that locally with `grunt test:karma` (or `grunt test:webserver` and go to localhost:3000), and you can run only the test you're interested in by editing karma.conf.js line 115).

browser/lib/util/crypto.js
```diff
@@ -207,34 +207,52 @@ var Crypto = (function() {
...
+		if (!this.encryptCipher) {
+			generateRandom(DEFAULT_BLOCKLENGTH, function(err, iv) {
+				if (err) {
+					callback(err);
+					return;
+				}
+				this.encryptCipher = CryptoJS.algo[this.cjsAlgorithm].createEncryptor(this.key, { iv: iv });
+				this.iv = iv;
+				then();
+			}.bind(this));
```
SimonWoolf: b32e67f

(`encrypt`'s getting a bit messy, might refactor it a bit later)

package.json
SimonWoolf: ah, that's unfortunate. We may end up going the dedicated-separate-library route eventually, but for now, s'probably fine -- at least it's only 188K, could be worse

mattheworiordan:
What is the status on this now in terms of being able to be merged in and released?

package.json
mattheworiordan: This is quite unfortunate.  I think it's OK for this release, but we should add a task to create a new repo for ably-reactive-native that simply depends on ably-js and adds the necessary dependency, WDYT?

paddybyers: > This is quite unfortunate. I think it's OK for this release, but we should add a task to create a new repo for ably-reactive-native that simply depends on ably-js and adds the necessary dependency, WDYT?

Maybe I'm not following but I still don't understand why it's not a dev dependency.

Since this only a spurious dependency in node deployments I don't think the impact of including it is that bad, even if it's a bit crap.

tcard:
I'm running the browser tests right now. Once they pass I guess they can be merged.

tcard:
The tests pass on my machine.

I'm OK with creating a new repo now, BTW. Then we can do this clean from the beginning.

SimonWoolf:
> The tests pass on my machine.

@tcard yep, I ended up adding https://github.com/ably/ably-js/commit/b32e67f1df6a742a5361bb6951ac7be0bec173f0 (allowing a specified iv to be used) to fix the browser tests, see comment [here](https://github.com/ably/ably-js/pull/349#discussion_r87634113)

> What is the status on this now in terms of being able to be merged in and released?

If we want to go with the suggestion of a new repo, then actually @tcard's right, may as well do it now. In that case I'll merge this (minus the dependency) into 0-9-spec branch and tag that with a beta tag on npm. Then we can add a new repo that depends on that and adds the dependency. (I'll also add some logic to platform-reactnative.js that points you to the right repo if you try to use it without having racct-native-randombytes available). Sound good?

(Note that this is all against 0.9 -- we don't want people starting new projects against 0.8 at this stage..)

paddybyers:
> Since this only a spurious dependency in node deployments I don't think the impact of including it is that bad, even if it's a bit crap.

Do we really need to do this? Now?

SimonWoolf:
> Do we really need to do this? Now?

No, we don't - as I said, it's only 188K. But if we decide that we are going to do it at some point (per Matt's suggestion above that "*it's OK for this release, but we should add a task to create a new repo*"), then we should just do it now, and avoid having to find a way to tell everyone using react-native when we do make the change to change dependency to stop things breaking after they do an `npm update ably`. I'm fine with either option.

tcard:
> If we want to go with the suggestion of a new repo, then actually @tcard's right, may as well do it now. In that case I'll merge this (minus the dependency) into 0-9-spec branch and tag that with a beta tag on npm. Then we can add a new repo that depends on that and adds the dependency. (I'll also add some logic to platform-reactnative.js that points you to the right repo if you try to use it without having racct-native-randombytes available). Sound good?

@SimonWoolf Sounds good to me :)

mattheworiordan:
@SimonWoolf sounds good to me too

SimonWoolf:
Rebased, fixup-squashed, and merged into 0-9-spec as e3d616e.

PR #359, opened 2016-12-12
Add TypeScript typings support

NathanaelA:
Added Typings file; ably.ts.d, updated package.json to point to it and updated docs on how to use it.

### DIFF
--- a/ably.ts.d
+++ b/ably.ts.d
@@ -1,457 +1,394 @@
 //declare module Ably {
 
-  namespace ChannelState {
-    export type INITIALIZED = 'initialized';
-    export type ATTACHING = 'attaching';
-    export type ATTACHED = "attached";
-    export type DETACHING = "detaching";
-    export type DETACHED = "detached";
-    export type SUSPENDED = "suspended";
-    export type FAILED = "failed";
-  }
-  type ChannelState = ChannelState.FAILED | ChannelState.INITIALIZED | ChannelState.SUSPENDED | ChannelState.ATTACHED | ChannelState.ATTACHING | ChannelState.DETACHED | ChannelState.DETACHING;
-
-  namespace ConnectionState {
-    export type INITIALIZED = "initialized";
-    export type CONNECTING = "connecting";
-    export type CONNECTED = "connected";
-    export type DISCONNECTED = "disconnected";
-    export type SUSPENDED = "suspended";
-    export type CLOSING = "closing";
-    export type CLOSED = "closed";
-    export type FAILED = "failed";
-  }
-  type ConnectionState = ConnectionState.INITIALIZED | ConnectionState.CONNECTED | ConnectionState.CONNECTING | ConnectionState.DISCONNECTED | ConnectionState.SUSPENDED | ConnectionState.CLOSED | ConnectionState.CLOSING | ConnectionState.FAILED;
-
-  namespace ConnectionEvent {
-    export type INITIALIZED = "initialized";
-    export type CONNECTING = "connecting";
-    export type CONNECTED = "connected";
-    export type DISCONNECTED = "disconnected";
-    export type SUSPENDED = "suspended";
-    export type CLOSING = "closing";
-    export type CLOSED = "closed";
-    export type FAILED = "failed";
-    export type UPDATE = "update";
-  }
-  type ConnectionEvent = ConnectionEvent.INITIALIZED | ConnectionEvent.CONNECTED | ConnectionEvent.CONNECTING | ConnectionEvent.DISCONNECTED | ConnectionEvent.SUSPENDED | ConnectionEvent.CLOSED | ConnectionEvent.CLOSING | ConnectionEvent.FAILED | ConnectionEvent.UPDATE;
-
-  namespace PresenceAction {
-    export type ABSENT = "absent";
-    export type PRESENT = "present";
-    export type ENTER = "enter";
-    export type LEAVE = "leave";
-    export type UPDATE = "update";
-  }
-  type PresenceAction = PresenceAction.ABSENT | PresenceAction.PRESENT | PresenceAction.ENTER | PresenceAction.LEAVE | PresenceAction.UPDATE;
-
-  namespace StatsIntervalGranularity {
-    export type MINUTE = "minute";
-    export type HOUR = "hour";
-    export type DAY = "day";
-    export type MONTH = "month";
-  }
-  type StatsIntervalGranularity = StatsIntervalGranularity.MINUTE | StatsIntervalGranularity.HOUR | StatsIntervalGranularity.DAY | StatsIntervalGranularity.MONTH;
-
-  namespace ProtocolMessageAction {
-    export type HEARTBEAT = "heartbeat";
-    export type ACK = "ack";
-    export type NACK = "nack";
-    export type CONNECT = "connect";
-    export type CONNECTED = "connected";
-    export type DISCONNECT = "disconnect";
-    export type DISCONNECTED = "disconnected";
-    export type CLOSE = "close";
-    export type CLOSED = "closed";
-    export type ERROR = "error";
-    export type ATTACH = "attach";
-    export type ATTACHED = "attached";
-    export type DETACH = "detach";
-    export type DETACHED = "detached";
-    export type PRESENCE = "presence";
-    export type MESSAGE = "message";
-    export type SYNC = "sync";
-    export type AUTH = "auth";
-  }
-  type ProtocolMessageAction = ProtocolMessageAction.ACK | ProtocolMessageAction.ATTACH | ProtocolMessageAction.ATTACHED | ProtocolMessageAction.AUTH | ProtocolMessageAction.CLOSE | ProtocolMessageAction.CLOSED | ProtocolMessageAction.CONNECT | ProtocolMessageAction.CONNECTED | ProtocolMessageAction.DETACH | ProtocolMessageAction.DETACHED | ProtocolMessageAction.DISCONNECT | ProtocolMessageAction.DISCONNECTED | ProtocolMessageAction.ERROR | ProtocolMessageAction.HEARTBEAT | ProtocolMessageAction.MESSAGE | ProtocolMessageAction.NACK | ProtocolMessageAction.PRESENCE | ProtocolMessageAction.SYNC;
-
-
-  // Interfaces
-  interface AuthDetails {
-    accessToken: String;
-  }
-
-  interface ClientOptions extends AuthOptions {
-    autoConnect?: boolean;
-    clientId?: String;
-    defaultTokenParams?: TokenParams;
-    echoMessages?: boolean;
-    environment?: String;
-    logHandler?: any;
-    logLevel?: any;
-    port?: number;
-    queueMessages?: boolean;
-    restHost?: String;
-    realtimeHost?: String;
-    fallbackHosts?: Array<String>
-    recover?: String
-    tls?: boolean;
-    tlsPort?: number;
-    useBinaryProtocol?: boolean;
-    disconnectedRetryTimeout?: number;
-    suspendedRetryTimeout?: number;
-    channelRetryTimeout?: number;
-    httpOpenTimeout?: number;
-    httpRequestTimeout?: number;
-    httpMaxRetryCount?: number;
-    httpMaxRetryDuration?: number;
-  }
-
-  interface AuthOptions {
-    authCallback?: (data: String | TokenDetails | TokenRequest) => void;
-    authHeaders?: Array<any>;
-    authMethod?: String;
-    authParams?: Array<any>;
-    authUrl?: String;
-    key?: String;
-    queryTime?: boolean;
-    token?: String | TokenDetails;
-    tokenDetails?: TokenDetails;
-    useTokenAuth?: boolean;
-  }
-
-  interface TokenParams {
-    capability: String;
-    clientId?: String;
-    nonce?: String;
-    timestamp: Number;   // Time?
-    ttl: Number;
-  }
-
-  interface CipherParams {
-    algorithm: String;
-    key: any;
-    keyLength: number;
-    mode: String;
-  }
-
-  interface ErrorInfo {
-    code: number;
-    message: String;
-    statusCode: number;
-  }
-
-  interface ConnectionDetails {
-    clientId: String;
-    connectionKey: String;
-    connectionStateTtl: number;
-    maxFrameSize: number;
-    maxInboundRate: number;
-    maxMessageSize: number;
-    serverId: String;
-    maxIdleInterval: number;
-  }
-
-  interface StatsMessageTypes {
-    all: number,
-    messages: number,
-    presence: number;
-  }
-
-  interface StatsRequestCount {
-    failed: number,
-    refused: number,
-    succeeded: number
-  }
-
-  interface StatsResourceCount {
-    mean: number,
-    min: number,
-    opened: number,
-    peak: number,
-    refused: number
-  }
-
-  interface StatsConnectionTypes {
-    all: number,
-    plain: number,
-    tls: number
-  }
-
-  interface StatsMessageTraffic {
-    all: number,
-    realtime: number,
-    rest: number,
-    webhook: number
-  }
-
-  // Listeners
-  type ListenerFn = (...args: Array<any>) => void;
-
-
-  // Internal Classes
-  declare class EventEmitter {
-    on: (eventOrCallback: String | ListenerFn, callback?: ListenerFn) => void;
-
-    once: (eventOrCallback: String | ListenerFn, callback?: ListenerFn) => void;
-
-    off: (eventOrCallback?: String | ListenerFn, callback?: ListenerFn) => void;
-
-    emit: (Event: String, ...args: Array<any>) => void;
-  }
-
-  // Classes
-  export declare class Auth {
-    clientId: String;
-    authorize: (tokenParams?: TokenParams, authOptions?: AuthOptions) => TokenDetails;
-    createTokenRequest: (tokenParams?: TokenParams, authOptions?: AuthOptions) => TokenRequest;
-    requestToken: (TokenParams?: TokenParams, authOptions?: AuthOptions) => TokenDetails;
-  }
-
-  export declare class TokenDetails {
-    fromJson: (JsonData: string) => TokenDetails;
-    capability: String;
-    clientId: String;
-    expires: number; // Time
-    issued: number; // Time
-    token: String;
-  }
-
-  export declare class TokenRequest {
-    fromJson: (JsonData: String) => TokenRequest;
-    capability: String;
-    clientId: String;
-    keyName: String;
-    mac: String;
-    nonce: String;
-    timestamp: number; // Time
-    ttl: number;
-  }
-
-  export declare class ChannelOptions {
-    withCipherKey: (key: any) => ChannelOptions;
-    cipher: (params: CipherParams) => void;
-  }
-
-  export declare class RestPresence {
-    get: (limit: number, clientId?: String, connectionId?: String) => PaginatedResult;
-
-    history: (start: number, end: number, direction: String, limit: number) => PaginatedResult;
-  }
-
-  export declare class RealtimePresence {
-    syncComplete: boolean;
-
-    get: (waitForSync: boolean, clientId?: String, connectionId?: String) => PresenceMessage;
-
-    history: (start: number, end: number, direction: string, limit: number, untilAttach: boolean) => PaginatedResult;
-
-    subscribe: (presense: PresenceMessage | PresenceAction, presenceMessage?: PresenceMessage) => void;
-
-    unsubscribe: (presense?: PresenceMessage | PresenceAction, presenceMessage?: PresenceMessage) => void;
-
-    enter: (Data: any) => void;
-
-    update: (Data: any) => void;
-
-    leave: (Data: any) => void;
-
-    enterClient: (clientId: String, Data: any) => void;
-
-    updateClient: (clientId: String, Data: any) => void;
-
-    leaveClient: (clientId: String, Data: any) => void;
-  }
-
-  type MessageFn = (msg: Message) => void;
-
-  export declare class RestChannel {
-    name: String;
-    presence: RestPresence;
-    history: (start: number, // TIME
-              end: number, // time
-              direction: String,
-              limit: number) => PaginatedResult;
-
-    publish: (name?: Array<String> | String, data?: any, clientId?: String, extras?: any) => any;
-  }
-
-  export declare class RealtimeChannel extends EventEmitter {
-    name: String;
-    errorReason: ErrorInfo;
-    state: ChannelState;
-    presence: RealtimePresence;
-
-    attach: () => any;
-
-    detach:() => any;
-
-    history: (start: number,
-            end: number,
-            direction: String,
-            limit: number,
-            untilAttach: boolean) => PaginatedResult;
-
-    subscribe: (Msg: String | Message | MessageFn, ExtraMsg?: Message) => any;
-
-    unsubscribe: (Msg?: String | Message | MessageFn, ExtraMsg?: Message) => any;
-
-    publish: (name?: Array<String> | String, data?: any, clientId?: String, extras?: any) => any;
-
-  }
-
-  export declare class Channels {
-    exists: (name: String) => boolean;
-
-    get: (name: String, channelOptions?: ChannelOptions) => RealtimeChannel;
-    iterate: () => RealtimeChannel;
-
-    release: (name: String) => void;
-  }
-
-  export declare class Message {
-    constructor(name?: String, data?: any, clientId?: String);
-
-    fromEncoded: (JsonObject: String, channelOptions: ChannelOptions) => Message;
-
-    fromEncodedArray: (JsonArray: String, channelOptions: ChannelOptions) => Message;
-
-    clientId: String;
-    connectionId: String;
-    data: any;
-    encoding: String;
-    extras: String;
-    id: String;
-    name: String;
-    timestamp: number; // Time
-  }
-
-  export declare class PresenceMessage {
-    fromEncoded: (JsonObject: any, channelOptions?: ChannelOptions) => PresenceMessage;
-
-    fromEncodedArray: (JsonArray: Array<any>, channelOptions?: ChannelOptions) => Array<PresenceMessage>;
-
-    action: PresenceAction;
-    clientId: String;
-    connectionId: String;
-    data: any;
-    encoding: String;
-    id: String;
-    timestamp: number; // Time
-    memberKey: () => String;
-  }
-
-  export declare class ProtocolMessage {
-    action: ProtocolMessageAction;
-    auth: AuthDetails;
-    channel: String;
-    channelSerial: String;
-    connectionDetails: ConnectionDetails;
-    connectionId: String;
-    connectionKey: String;
-    connectionSerial: number;
-    count: number;
-    error: ErrorInfo;
-    flags: String;
-    id: String;
-    messages: Array<Message>;
-    msgSerial: number;
-    presence: Array<PresenceMessage>
-    timestamp: number; // Time
-  }
-
-  export declare class Rest {
-    constructor(options: String | ClientOptions);
-    auth: Auth;
-    channels: Channels;
-    request: (method: string, path: string, parmas?: any, body?: Array<any> | any, headers?: any) => HttpPaginatedResponse;
-    stats: ( start: number, // Time
-           end: number, // Time
-           direction: String,
-           limit: number,
-           unit: String) => PaginatedResult;
-    time: () => number; // time
-  }
-
-  export declare class Realtime {
-    constructor(options: String | ClientOptions);
-
-    auth: Auth;
-    channels: Channels;
-    clientId: String;
-    connection: Connection;
-
-    request: (method: string, path: string, parmas?: any, body?: Array<any> | any, headers?: any) => HttpPaginatedResponse;
-
-    stats: (start: number, // Time
-          end: number, // Time
-          direction: String,
-          limit: number,
-          unit: String) => PaginatedResult;
-
-    close: () => void;
-
-    connect: () => void;
-
-    time: () => number;
-
-  }
-
-  export declare class Connection extends EventEmitter {
-    errorReason: ErrorInfo;
-    id: String;
-    key: String;
-    recoveryKey: String;
-    serial: number;
-    state: ConnectionState;
-
-    close: () => void;
-
-    connect: () => void;
-
-    ping: () => void;
-  }
-
-  export declare class Stats {
-    all: StatsMessageTypes;
-    apiRequests: StatsRequestCount;
-    channels: StatsResourceCount;
-    connections: StatsConnectionTypes;
-    inbound: StatsMessageTraffic;
-    intervalGranularity: StatsIntervalGranularity;
-    intervalId: String;
-    intervalTime: number; // type
-    outbound: StatsMessageTraffic;
-    persisted: StatsMessageTypes;
-    tokenRequests: StatsRequestCount;
-  }
-
-  export declare class ConnectionStateChange {
-  current: ConnectionState;
-  previous: ConnectionState;
-  reason?: ErrorInfo;
-  retryIn: number;
+namespace ChannelState {
+  export type INITIALIZED = 'initialized';
+  export type ATTACHING = 'attaching';
+  export type ATTACHED = "attached";
+  export type DETACHING = "detaching";
+  export type DETACHED = "detached";
+  export type SUSPENDED = "suspended";
+  export type FAILED = "failed";
+}
+type ChannelState = ChannelState.FAILED | ChannelState.INITIALIZED | ChannelState.SUSPENDED | ChannelState.ATTACHED | ChannelState.ATTACHING | ChannelState.DETACHED | ChannelState.DETACHING;
+
+namespace ConnectionState {
+  export type INITIALIZED = "initialized";
+  export type CONNECTING = "connecting";
+  export type CONNECTED = "connected";
+  export type DISCONNECTED = "disconnected";
+  export type SUSPENDED = "suspended";
+  export type CLOSING = "closing";
+  export type CLOSED = "closed";
+  export type FAILED = "failed";
+}
+type ConnectionState = ConnectionState.INITIALIZED | ConnectionState.CONNECTED | ConnectionState.CONNECTING | ConnectionState.DISCONNECTED | ConnectionState.SUSPENDED | ConnectionState.CLOSED | ConnectionState.CLOSING | ConnectionState.FAILED;
+
+namespace ConnectionEvent {
+  export type INITIALIZED = "initialized";
+  export type CONNECTING = "connecting";
+  export type CONNECTED = "connected";
+  export type DISCONNECTED = "disconnected";
+  export type SUSPENDED = "suspended";
+  export type CLOSING = "closing";
+  export type CLOSED = "closed";
+  export type FAILED = "failed";
+  export type UPDATE = "update";
+}
+type ConnectionEvent = ConnectionEvent.INITIALIZED | ConnectionEvent.CONNECTED | ConnectionEvent.CONNECTING | ConnectionEvent.DISCONNECTED | ConnectionEvent.SUSPENDED | ConnectionEvent.CLOSED | ConnectionEvent.CLOSING | ConnectionEvent.FAILED | ConnectionEvent.UPDATE;
+
+namespace PresenceAction {
+  export type ABSENT = "absent";
+  export type PRESENT = "present";
+  export type ENTER = "enter";
+  export type LEAVE = "leave";
+  export type UPDATE = "update";
+}
+type PresenceAction = PresenceAction.ABSENT | PresenceAction.PRESENT | PresenceAction.ENTER | PresenceAction.LEAVE | PresenceAction.UPDATE;
+
+namespace StatsIntervalGranularity {
+  export type MINUTE = "minute";
+  export type HOUR = "hour";
+  export type DAY = "day";
+  export type MONTH = "month";
+}
+type StatsIntervalGranularity = StatsIntervalGranularity.MINUTE | StatsIntervalGranularity.HOUR | StatsIntervalGranularity.DAY | StatsIntervalGranularity.MONTH;
+
+// Interfaces
+
+interface ClientOptions extends AuthOptions {
+  autoConnect?: boolean;
+  clientId?: String;
+  defaultTokenParams?: TokenParams;
+  echoMessages?: boolean;
+  environment?: String;
+  logHandler?: any;
+  logLevel?: any;
+  port?: number;
+  queueMessages?: boolean;
+  restHost?: String;
+  realtimeHost?: String;
+  fallbackHosts?: Array<String>
+  recover?: String
+  tls?: boolean;
+  tlsPort?: number;
+  useBinaryProtocol?: boolean;
 }
 
-  export declare class PaginatedResult {
-    items: Array<any>;
+interface AuthOptions {
+  authCallback?: (data: String | TokenDetails | TokenRequest) => void;
+  authHeaders?: Array<any>;
+  authMethod?: String;
+  authParams?: Array<any>;
+  authUrl?: String;
+  key?: String;
+  queryTime?: boolean;
+  token?: String | TokenDetails;
+  tokenDetails?: TokenDetails;
+  useTokenAuth?: boolean;
+}
 
-    first: () => any;
+interface TokenParams {
+  capability: String;
+  clientId?: String;
+  nonce?: String;
+  timestamp: Number;   // Time?
+  ttl: Number;
+}
 
-    hasNext: () => boolean;
+interface CipherParams {
+  algorithm: String;
+  key: any;
+  keyLength: number;
+  mode: String;
+}
 
-    isLast: () => boolean;
+interface ErrorInfo {
+  code: number;
+  message: String;
+  statusCode: number;
+}
 
-    next: () => any;
-  }
+interface StatsMessageTypes {
+  all: number,
+  messages: number,
+  presence: number;
+}
 
-  export declare class HttpPaginatedResponse extends PaginatedResult {
-    items: Array<String>;
-    statusCode: number;
-    success: boolean;
-    errorCode: number;
-    errorMessage: String;
-    headers: any;
-  }
+interface StatsRequestCount {
+  failed: number,
+  refused: number,
+  succeeded: number
+}
+
+interface StatsResourceCount {
+  mean: number,
+  min: number,
+  opened: number,
+  peak: number,
+  refused: number
+}
+
+interface StatsConnectionTypes {
+  all: number,
+  plain: number,
+  tls: number
+}
+
+interface StatsMessageTraffic {
+  all: number,
+  realtime: number,
+  rest: number,
+  webhook: number
+}
+
+interface TokenDetails {
+  capability: String;
+  clientId: String;
+  expires: number; // Time
+  issued: number; // Time
+  token: String;
+}
+
+interface TokenRequest {
+  capability: String;
+  clientId: String;
+  keyName: String;
+  mac: String;
+  nonce: String;
+  timestamp: number; // Time
+  ttl: number;
+}
+
+interface ChannelOptions {
+  cipher: any;
+}
+
+interface RestPresenceHistoryParams {
+  start: number;
+  end: number;
+  direction: String;
+  limit: number;
+}
+
+
+interface RestPresenceParams {
+  limit: number;
+  clientId: String;
+  connectionId: String;
+}
+
+interface RealtimePresenceParams {
+  waitForSync: boolean;
+  clientId: String;
+  connectionId: String;
+}
+
+interface RealtimePresenceHistoryParams {
+  start: number;
+  end: number;
+  direction: String;
+  limit: number;
+  untilAttach: boolean
+}
+
+// Common Listeners
+type ListenerFn = (...args: Array<any>) => void;
+type PaginatedResultCallback = (Error: ErrorInfo, results: PaginatedResult ) => void;
+type standardCallback = (Error: ErrorInfo, results: any) => void;
+type messageCallback = (Msg: any) => void;
+
+
+
+// Internal Classes
+declare class EventEmitter {
+  on: (eventOrCallback: String | ListenerFn, callback?: ListenerFn) => void;
+
+  once: (eventOrCallback: String | ListenerFn, callback?: ListenerFn) => void;
+
+  off: (eventOrCallback?: String | ListenerFn, callback?: ListenerFn) => void;
+}
+
+// Classes
+export declare class Auth {
+  clientId: String;
+  authorize: (tokenParams?: TokenParams, authOptions?: AuthOptions, callback?: (Error: ErrorInfo, Results: TokenDetails) => void) => void;
+  createTokenRequest: (tokenParams?: TokenParams, authOptions?: AuthOptions, callback?: (Error: ErrorInfo, Results: TokenRequest) => void) => void;
+  requestToken: (TokenParams?: TokenParams, authOptions?: AuthOptions, callback?: (Error: ErrorInfo, Results: TokenDetails) => void) => void;
+}
+
+export declare class Presence {
+  get: (params: RestPresenceParams, PaginatedResultCallback) => void;
+
+  history: (params: RestPresenceHistoryParams, PaginatedResultCallback) => void;
+}
+
+export declare class RealtimePresence {
+  syncComplete: () =>  boolean;
+
+  get: (Params: RealtimePresenceParams, callback?: PaginatedResultCallback) => void;
+
+  history: (ParamsOrCallback: RealtimePresenceHistoryParams | PaginatedResultCallback, callback?: PaginatedResultCallback) => void;
+
+  subscribe: (presenceOrCallback: PresenceAction | messageCallback, listener?: standardCallback, callback?: standardCallback) => void;
+
+  unsubscribe: (presence?: PresenceAction, listener?: standardCallback, callback?: standardCallback) => void;
+
+  enter: (Data: any, callback?: standardCallback) => void;
+
+  update: (Data: any, callback?: standardCallback) => void;
+
+  leave: (Data: any, callback?: standardCallback) => void;
+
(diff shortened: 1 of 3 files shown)


ably.ts.d
```diff
@@ -0,0 +1,457 @@
...
+  current: ConnectionState;
+  previous: ConnectionState;
+  reason?: ErrorInfo;
+  retryIn: number;
+}
+
+  export declare class PaginatedResult {
+    items: Array<any>;
+
+    first: () => any;
```
mattheworiordan: This method takes a function that has the signature `err ErrorInfo, result PaginatedResult`.  Should this not be defined?

ably.ts.d
```diff
@@ -0,0 +1,457 @@
...
+
+    hasNext: () => boolean;
+
+    isLast: () => boolean;
+
+    next: () => any;
```
mattheworiordan: As above for `first`

ably.ts.d
```diff
@@ -0,0 +1,457 @@
...
+    intervalGranularity: StatsIntervalGranularity;
+    intervalId: String;
+    intervalTime: number; // type
+    outbound: StatsMessageTraffic;
+    persisted: StatsMessageTypes;
+    tokenRequests: StatsRequestCount;
+  }
+
+  export declare class ConnectionStateChange {
+  current: ConnectionState;
```
mattheworiordan: Indentation for this one is odd

ably.ts.d
```diff
@@ -0,0 +1,457 @@
...
+    next: () => any;
+  }
+
+  export declare class HttpPaginatedResponse extends PaginatedResult {
+    items: Array<String>;
+    statusCode: number;
+    success: boolean;
+    errorCode: number;
+    errorMessage: String;
+    headers: any;
```
mattheworiordan: I think this is a always an object with key/value string pairs.

ably.ts.d
```diff
@@ -0,0 +1,457 @@
...
+    key: String;
+    recoveryKey: String;
+    serial: number;
+    state: ConnectionState;
+
+    close: () => void;
+
+    connect: () => void;
+
+    ping: () => void;
```
mattheworiordan: This function has a callback, why is that definition missing?

ably.ts.d
```diff
@@ -0,0 +1,457 @@
...
+
+  export declare class Realtime {
+    constructor(options: String | ClientOptions);
+
+    auth: Auth;
+    channels: Channels;
+    clientId: String;
+    connection: Connection;
+
+    request: (method: string, path: string, parmas?: any, body?: Array<any> | any, headers?: any) => HttpPaginatedResponse;
```
mattheworiordan: Out of interest, I think the callback def is actually `request => (err, response)` so is `err` not needed as it's "normal" for JS callbacks?

NathanaelA:
Any mistakes like this are because either the IDL was wrong (as this is a direct by hand translation.   Or I misunderstood the IDL.  :-)  

  So I can fix any areas you find.    As it was there was a couple places I did go back to the source to figure things out as the IDL was incomplete or wrong.   So I'll look at each of these places and fix them if need be.

ably.ts.d
NathanaelA: <any> is used for any object, including a empty key/pair object.  :grinning:

ably.ts.d
NathanaelA: Agreed, I misread the IDL.  I've fixed in the next push.

ably.ts.d
NathanaelA: Agreed, I misread the IDL.  I've fixed in the next push.

ably.ts.d
NathanaelA: Fixed in next push.

ably.ts.d
NathanaelA: IDL is wrong.  No callback is listed.    But I fixed in the next push.

ably.ts.d
NathanaelA: IDL is written such that it returns a HTTPPaginateResponse, not that it is a Callback with a HTTPPaginatedResponse..    But looking at the source I can see what you are saying.    I'll get this fixed in the next push.

NathanaelA:
The IDL is wrong in a several areas, unless I am totally misunderstanding reading the IDL.  Multiple places you have in the IDL it is returning the result set; but in every one of those areas it is actually an additional callback parameter that has the results set when completed.     My understanding of an IDL is the callback should be listed, not that the function should list that it is directly returning the results.      

So this is why a lot of callbacks are missing.  I am attempting to update the Typings since I now know the IDL is wrong in this areas...

SimonWoolf:
> The IDL is wrong in a several areas ... My understanding of an IDL is the callback should be listed, not that the function should list that it is directly returning the results.

Hi Nathanael. The reason for the confusion is that the IDL is not javascript-specific. It's a 'lowest-common-denominator' IDL that covers all of our client libraries. However, programming languages work differently and have different idioms, so our client libraries are a compromise between trying to keep things reasonably consistent while still keeping the interface idiomatic for the language.

(We could make a separate IDL for each language, but that would defeat the point of having an IDL, which is to guide client library devs develop new libraries and try to have at least a broad level of interface consistency).

For example, for some things that in the python library are done synchronously and return a value or throw an exception, in the javascript library may be async and take a callback that is called with `(err, res)`. Making the python function take a two-valued callback would be very unidiomatic; making the js one synchronous would be impossible.

For the typescript typings, the source of truth for the interface is the javascript lib source.

Apologies for the confusion.

ably.ts.d
```diff
@@ -0,0 +1,457 @@
...
+  }
+
+  interface ClientOptions extends AuthOptions {
+    autoConnect?: boolean;
+    clientId?: String;
+    defaultTokenParams?: TokenParams;
+    echoMessages?: boolean;
+    environment?: String;
+    logHandler?: any;
+    logLevel?: any;
```
SimonWoolf: this is wrong for js. log is an object, has keys `level` and `handler`, level is an integer from 0 to 4 inclusive, handler is a function

ably.ts.d
```diff
@@ -0,0 +1,457 @@
...
+    realtimeHost?: String;
+    fallbackHosts?: Array<String>
+    recover?: String
+    tls?: boolean;
+    tlsPort?: number;
+    useBinaryProtocol?: boolean;
+    disconnectedRetryTimeout?: number;
+    suspendedRetryTimeout?: number;
+    channelRetryTimeout?: number;
+    httpOpenTimeout?: number;
```
SimonWoolf: this is not implemented in the js lib

ably.ts.d
```diff
@@ -0,0 +1,457 @@
...
+    httpRequestTimeout?: number;
+    httpMaxRetryCount?: number;
+    httpMaxRetryDuration?: number;
```
SimonWoolf: this is not implemented in the js lib

ably.ts.d
```diff
@@ -0,0 +1,457 @@
...
+
+  // Internal Classes
+  declare class EventEmitter {
+    on: (eventOrCallback: String | ListenerFn, callback?: ListenerFn) => void;
+
+    once: (eventOrCallback: String | ListenerFn, callback?: ListenerFn) => void;
+
+    off: (eventOrCallback?: String | ListenerFn, callback?: ListenerFn) => void;
+
+    emit: (Event: String, ...args: Array<any>) => void;
```
SimonWoolf: emit isn't a public api

ably.ts.d
```diff
@@ -0,0 +1,457 @@
...
+    fromJson: (JsonData: string) => TokenDetails;
+    capability: String;
+    clientId: String;
+    expires: number; // Time
+    issued: number; // Time
+    token: String;
+  }
+
+  export declare class TokenRequest {
+    fromJson: (JsonData: String) => TokenRequest;
```
SimonWoolf: fromJson isn't used in js, as the tokenDetails and tokenRequest objects are just basic js objects, so it's equivalent to JSON.parse()

ably.ts.d
```diff
@@ -0,0 +1,457 @@
...
+
+  // Classes
+  export declare class Auth {
+    clientId: String;
+    authorize: (tokenParams?: TokenParams, authOptions?: AuthOptions) => TokenDetails;
+    createTokenRequest: (tokenParams?: TokenParams, authOptions?: AuthOptions) => TokenRequest;
+    requestToken: (TokenParams?: TokenParams, authOptions?: AuthOptions) => TokenDetails;
+  }
+
+  export declare class TokenDetails {
```
SimonWoolf: sorry for my ignorance of typescript, but: does this imply that the js lib has a TokenDetails (similarly TokenRequest) prototype? If so, we don't, the js lib uses basic objects for those

ably.ts.d
NathanaelA: Thanks, I haven't got to this point yet.  I'll eliminate it.  :-)

ably.ts.d
```diff
@@ -0,0 +1,457 @@
...
+    clientId: String;
+    connectionId: String;
+    data: any;
+    encoding: String;
+    id: String;
+    timestamp: number; // Time
+    memberKey: () => String;
+  }
+
+  export declare class ProtocolMessage {
```
NathanaelA: I was going off the IDL to make the typings.  Anything in the IDL was translated.   If nothing uses the ProtocolMessage then it is pointless to have in the Typings.

ably.ts.d
```diff
@@ -0,0 +1,457 @@
...
+    timestamp: number; // Time
+    ttl: number;
+  }
+
+  export declare class ChannelOptions {
+    withCipherKey: (key: any) => ChannelOptions;
+    cipher: (params: CipherParams) => void;
+  }
+
+  export declare class RestPresence {
```
SimonWoolf: > RestPresence

Again, sorry for my ignorance of nativescript, but: should the class names match up with the js lib?

ably.ts.d
```diff
@@ -0,0 +1,457 @@
...
+    get: (limit: number, clientId?: String, connectionId?: String) => PaginatedResult;
+
+    history: (start: number, end: number, direction: String, limit: number) => PaginatedResult;
+  }
+
+  export declare class RealtimePresence {
+    syncComplete: boolean;
```
SimonWoolf: this is a function

ably.ts.d
NathanaelA: Any is also used for a blank customizable Object in TypeScript.     If you have the actual definition of what keys are allowed we can create a custom LogObject, with those keys instead of using Any.

ably.ts.d
```diff
@@ -0,0 +1,457 @@
...
+
+    get: (waitForSync: boolean, clientId?: String, connectionId?: String) => PresenceMessage;
```
SimonWoolf: This takes a params object, not three separate parameters. (similarly for lots of other fns here)

ably.ts.d
```diff
@@ -0,0 +1,457 @@
...
+
+    history: (start: number, end: number, direction: string, limit: number, untilAttach: boolean) => PaginatedResult;
+
+    subscribe: (presense: PresenceMessage | PresenceAction, presenceMessage?: PresenceMessage) => void;
```
SimonWoolf: s/presense/presence/

SimonWoolf: also not sure what's going on with the function signature? subscribe() has the same interface as on()

ably.ts.d
NathanaelA: In the IDL, means it was translated.  :grinning:.   But I've dropped them from the next version of the typigns.  Thanks.

ably.ts.d
```diff
@@ -0,0 +1,457 @@
...
+    export type ERROR = "error";
+    export type ATTACH = "attach";
+    export type ATTACHED = "attached";
+    export type DETACH = "detach";
+    export type DETACHED = "detached";
+    export type PRESENCE = "presence";
+    export type MESSAGE = "message";
+    export type SYNC = "sync";
+    export type AUTH = "auth";
+  }
```
SimonWoolf: This is all wrong (they're integers, not strings), but luckily not needed as not a public api

ably.ts.d
NathanaelA: Actually basically this is saying that there is a Object that follows this basic outline and has these properties.  This will allow me to take the Results of the callback and it will know that the only valid properties should be:
```
    capability: String;
    clientId: String;
    expires: number; 
    issued: number; 
    token: String;
```

Since these are pure property classes now; these actually will be converted to interfaces.  (not a lot of difference internally, but cleaner in the typings).

ably.ts.d
NathanaelA: Actually they don't have to match, but for "our" sake, they should, so that we don't get confused.    The only thing that has to match is that the callback sends a RestPresence object then TS will use the RestPresence definition.

Is this supposed to be named something else, I just copied the IDL...

SimonWoolf:
Hi @NathanaelA . I haven't finished going through everything, but given you've said you're now checking everything against the source, I'll stop there, don't want to pile on to you :slightly_smiling_face: 

Sorry for the confusion re the IDL in the client lib development guide, see [previous comment](https://github.com/ably/ably-js/pull/359#issuecomment-266532878).

If in the js source it's not clear whether some function is intended to be public, go to our [official documentation](https://www.ably.io/documentation/); if a method isn't there, it isn't public. (Note that on our website right now is 0.8 documentation, so it's wrong in some areas for 0.9. I'll make sure docs.ably.io (our beta documentation) has 0.9 up tomorrow.)

ably.ts.d
NathanaelA: Thanks. Fixed.

ably.ts.d
NathanaelA: Yeah, I noticed.  Not very happy -- Lots of fixing.  Apparently using the IDL was not a great idea for converting to the Typescript Typings.  I should have just gone through the JS files...

ably.ts.d
NathanaelA: Thanks.

tcard:
I just want to add to what @SimonWoolf said that, as a rule of thumb, return types in the IDL prefixed with `io` are implemented as callbacks in JavaScript other and platforms (iOS, for instance) that present a similar callback-based async interface to input/output and similarly "effectful" actions (HTTP requests, timers, etc.). So it's not arbitrary whether a return type in a IDL translates into a callback or not.

So e. g. `f(arg: Type) => io Type` would translate into TypeScript-ish as `f: (arg: Type, callback: (Error, Type) => ())`. (`io` implies `Error` there, because I/O can always fail.)

NathanaelA:
@SimonWoolf @mattheworiordan - New typings in place.  Let me know if you see anything that I missed.  I took them and then compared it against the source code (along with your prior suggestions).  
@tcard - Thanks, didn't know that.  That is why some things were apparently correct in the idl, but I didn't translate it correctly.  However hopefully this version has everything fixed.  :grinning:

ably.ts.d
```diff
@@ -0,0 +1,394 @@
...
+// Interfaces
+
+interface ClientOptions extends AuthOptions {
+  autoConnect?: boolean;
+  clientId?: String;
+  defaultTokenParams?: TokenParams;
+  echoMessages?: boolean;
+  environment?: String;
+  logHandler?: any;
+  logLevel?: any;
```
SimonWoolf: it's a `log` object - feel free to just go for `log?: any;` if you don't want to enumerate `handler` and `level`

ably.ts.d
```diff
@@ -0,0 +1,394 @@
...
+  port?: number;
+  queueMessages?: boolean;
+  restHost?: String;
+  realtimeHost?: String;
+  fallbackHosts?: Array<String>
+  recover?: String
```
SimonWoolf: String or function. (This is something specific to the js lib, which is why it's not in the main idl -- see the example in https://www.ably.io/documentation/realtime/connection/#connection-state-recover-options with your language choice set to `javascript`)

ably.ts.d
```diff
@@ -0,0 +1,394 @@
...
+  tls?: boolean;
+  tlsPort?: number;
+  useBinaryProtocol?: boolean;
+}
+
+interface AuthOptions {
+  authCallback?: (data: String | TokenDetails | TokenRequest) => void;
```
SimonWoolf: This should be something like (sorry if syntax is wrong)
`(tokenParams: TokenParams, callback(err: ErrorInfo | String, tokenRequestOrDetails: String | TokenDetails | TokenRequest) => void))`

ably.ts.d
```diff
@@ -0,0 +1,394 @@
...
+  tokenDetails?: TokenDetails;
+  useTokenAuth?: boolean;
+}
+
+interface TokenParams {
+  capability: String;
+  clientId?: String;
+  nonce?: String;
+  timestamp: Number;   // Time?
+  ttl: Number;
```
SimonWoolf: so in any of the places a user can feed in a tokenParams, all the fields are optional, so I think these should all have question marks?

ably.ts.d
```diff
@@ -0,0 +1,394 @@
...
+interface StatsMessageTraffic {
+  all: number,
+  realtime: number,
+  rest: number,
+  webhook: number
+}
+
+interface TokenDetails {
+  capability: String;
+  clientId: String;
```
SimonWoolf: clientId might not be present

ably.ts.d
```diff
@@ -0,0 +1,394 @@
...
+  expires: number; // Time
+  issued: number; // Time
+  token: String;
+}
+
+interface TokenRequest {
+  capability: String;
+  clientId: String;
```
SimonWoolf: clientId and ttl I think are both optional

ably.ts.d
```diff
@@ -0,0 +1,394 @@
...
+interface ChannelOptions {
+  cipher: any;
+}
+
+interface RestPresenceHistoryParams {
+  start: number;
+  end: number;
+  direction: String;
+  limit: number;
+}
```
SimonWoolf: similarly, users don't need to provide all of these, they can give any subset. If these aren't marked optional, won't that mean that their program won't compile unless they give all? Or am I misunderstanding how ts works?

NathanaelA: I think you are misunderstanding how TS works.   ? to my knowledge only works for parameters; they are not used in Interfaces or Classes.  Interfaces and Classes are all valid options.  You can "optionally" use any of them.

SimonWoolf: ah, fair enough then - my bad, apologies :)

NathanaelA: I'm wrong.  LOL, ``?`` are valid in interface for optional.    For some reason I thought they were always optional.  I was unaware of that language feature.

ably.ts.d
```diff
@@ -0,0 +1,394 @@
...
+
+  attach: (callback?: standardCallback) => void;
+
+  detach:(callback?: standardCallback) => void;
+
+  history: (ParamsOrCallback?: RealtimePresenceHistoryParams | PaginatedResultCallback, callback?: PaginatedResultCallback) => void;
+
+  subscribe: (presenceOrCallback: PresenceAction | messageCallback, listener?: standardCallback, callback?: standardCallback) => void;
+
+  unsubscribe: (presence?: PresenceAction, listener?: standardCallback, callback?: standardCallback) => void;
```
SimonWoolf: similarly

ably.ts.d
NathanaelA: Unknown, there is no real details in the JS what are provided; these must all come from the server.

ably.ts.d
NathanaelA: Marked as such.  Thanks.

ably.ts.d
NathanaelA: Thx, that was pretty close.  :grinning:

ably.ts.d
NathanaelA: Fixed

ably.ts.d
NathanaelA: Ah, now I see what you are saying.  Fixed.

ably.ts.d
```diff
@@ -0,0 +1,394 @@
...
+  state: ChannelState;
+  presence: RealtimePresence;
+
+  attach: (callback?: standardCallback) => void;
+
+  detach:(callback?: standardCallback) => void;
+
+  history: (ParamsOrCallback?: RealtimePresenceHistoryParams | PaginatedResultCallback, callback?: PaginatedResultCallback) => void;
+
+  subscribe: (presenceOrCallback: PresenceAction | messageCallback, listener?: standardCallback, callback?: standardCallback) => void;
```
NathanaelA: Thanks, Fixed.

ably.ts.d
```diff
@@ -0,0 +1,394 @@
...
+  updateClient: (clientId: String, Data: any, callback?: standardCallback) => void;
+
+  leaveClient: (clientId: String, Data: any, callback?: standardCallback) => void;
+}
+
+export declare class Channel {
+  name: String;
+  presence: Presence;
+  history: (ParamsOrCallback?: RestPresenceHistoryParams | PaginatedResultCallback, callback?: PaginatedResultCallback ) => void;
+  publish: (messages: any, MessageDataOrCallback?: any | PaginatedResultCallback, callback?: PaginatedResultCallback) => void;
```
NathanaelA: Thanks.  Fixed.

NathanaelA:
@SimonWoolf - ok, I think I got all your changes.  Thank you for reviewing them.  It really helps to have someone who knows the library to see where I messed up IDL -> Typings conversion.  ;-)
Let me know if you see anything else out of place.  :grinning:

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+  fallbackHosts?: Array<String>;
+  recover?: String | standardCallback;
+  tls?: boolean;
+  tlsPort?: number;
+  useBinaryProtocol?: boolean;
+}
+
+interface AuthOptions {
+  authCallback?: (data: TokenParams,callback: (Error: ErrorInfo | String, tokenRequestOrDetails: String | TokenDetails | TokenRequest) => void) => void;
+  authHeaders?: Array<any>;
```
mattheworiordan: Perhaps I am misunderstanding this, but `Array<any>` implies that an Array is being passed in whereas in fact an Object with string key/value pairs is expected.  `Map<string, string>` is what it should be, but we do not support `Map` for backwards compatibility reasons.  I am not sure what the answer is in TypeScript world, should we perhaps simply define a new type called Params with key/vaue string pairs?

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+  authMethod?: String;
+  authParams?: Array<any>;
```
mattheworiordan: As above.

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+  recover?: String | standardCallback;
+  tls?: boolean;
+  tlsPort?: number;
+  useBinaryProtocol?: boolean;
+}
+
+interface AuthOptions {
+  authCallback?: (data: TokenParams,callback: (Error: ErrorInfo | String, tokenRequestOrDetails: String | TokenDetails | TokenRequest) => void) => void;
+  authHeaders?: Array<any>;
+  authMethod?: String;
```
mattheworiordan: Not essential, but as we only support GET & POST, perhaps we should create a type for this like we have done for `StatsIntervalGranularity`?

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+  token?: String | TokenDetails;
+  tokenDetails?: TokenDetails;
+  useTokenAuth?: boolean;
+}
+
+interface TokenParams {
+  capability?: String;
+  clientId?: String;
+  nonce?: String;
+  timestamp?: Number;   // Time?
```
mattheworiordan: What's the `Time?` for?  This is milliseconds since the epoch. For various reasons, we decided against a Time/Date type because of various issues across platforms and devices, whereas milliseconds since the epoch is pretty reliable.

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+}
+
+interface ErrorInfo {
+  code: number;
+  message: String;
+  statusCode: number;
+}
+
+interface StatsMessageTypes {
+  all: number,
```
mattheworiordan: this and all other attributes of this type are `StatsMessageCount`, see http://www.rubydoc.info/gems/ably/Ably/Models/Stats/MessageTypes

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+interface StatsResourceCount {
+  mean: number,
+  min: number,
+  opened: number,
+  peak: number,
+  refused: number
+}
+
+interface StatsConnectionTypes {
+  all: number,
```
mattheworiordan: These attributes are all StatsResourceCount types, see http://www.rubydoc.info/gems/ably/Ably/Models/Stats/ConnectionTypes

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+  plain: number,
+  tls: number
+}
+
+interface StatsMessageTraffic {
+  all: number,
```
mattheworiordan: These attributes are all `StatsMessageTypes`, see http://www.rubydoc.info/gems/ably/Ably/Models/Stats/MessageTraffic

ably.ts.d
mattheworiordan: Yes, but I think what @SimonWoolf is saying is that the `clientId` field may be null or a string, so should it not be `clientId?`

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+  token: String;
+}
+
+interface TokenRequest {
+  capability: String;
+  clientId?: String;
+  keyName: String;
+  mac: String;
+  nonce: String;
+  timestamp: number; // Time
```
mattheworiordan: as before, milliseconds since the epoch

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+
+interface LogInfo {
+  level?: number;
+  handler?: (...args) => void;
+}
+
+
+// Common Listeners
+type ListenerFn = (...args: Array<any>) => void;
+type PaginatedResultCallback = (Error: ErrorInfo, results: PaginatedResult ) => void;
```
mattheworiordan: Out of interest, why `Error: ErrorInfo` as opposed to `error: ErrorInfo` given all other args are lowercase?

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+type standardCallback = (Error: ErrorInfo, results: any) => void;
+type messageCallback = (Msg: any) => void;
+type errorCallback = (Error: ErrorInfo) => void;
+
+
+
+// Internal Classes
+declare class EventEmitter {
```
mattheworiordan: Could we not use a generic here so that we know what type of event the eventemitter is expecting for the event argument as well as the ListenerFn?  For example, when used with channels it would be `EventEmitter<ChannelEvent>`.

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+export declare class Presence {
+  get: (params: RestPresenceParams, PaginatedResultCallback) => void;
+
+  history: (params: RestPresenceHistoryParams, PaginatedResultCallback) => void;
+}
+
+export declare class RealtimePresence {
+  syncComplete: () =>  boolean;
+
+  get: (Params: RealtimePresenceParams, callback?: PaginatedResultCallback) => void;
```
mattheworiordan: This does not return a paginated result, it simply returns an array of `PresenceMessage` objects, definition is:

```
get(
    waitForSync: Bool default true, // RTP11c1
    clientId: String?, // RTP11c2
    connectionId: String?, // RTP11c3
  ) => io [PresenceMessage] // RTP11
```

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+}
+
+export declare class Presence {
+  get: (params: RestPresenceParams, PaginatedResultCallback) => void;
+
+  history: (params: RestPresenceHistoryParams, PaginatedResultCallback) => void;
+}
+
+export declare class RealtimePresence {
+  syncComplete: () =>  boolean;
```
mattheworiordan: I believe this is an attribute as opposed to a method, @SimonWoolf can you confirm?

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+  connections: StatsConnectionTypes;
+  inbound: StatsMessageTraffic;
+  intervalId: String;
+  outbound: StatsMessageTraffic;
+  persisted: StatsMessageTypes;
+  tokenRequests: StatsRequestCount;
+}
+
+
+export declare class PaginatedResult {
```
mattheworiordan: We could use a generic here as well as `PaginatedResult` only supports item arrays of `Message`, `PresenceMessage` or `Stats` objects

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+}
+
+export declare class RealtimePresence {
+  syncComplete: () =>  boolean;
+
+  get: (Params: RealtimePresenceParams, callback?: PaginatedResultCallback) => void;
+
+  history: (ParamsOrCallback: RealtimePresenceHistoryParams | PaginatedResultCallback, callback?: PaginatedResultCallback) => void;
+
+  subscribe: (presenceOrCallback: PresenceAction | messageCallback, listener?: standardCallback, callback?: standardCallback) => void;
```
mattheworiordan: This is not a standardCallback, `err` argument does not exist first (as there is no failure condition, it is only called when a new PresenceMessage arrives), so it's a `PresenceMessage` callback.

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+  level?: number;
+  handler?: (...args) => void;
+}
+
+
+// Common Listeners
+type ListenerFn = (...args: Array<any>) => void;
+type PaginatedResultCallback = (Error: ErrorInfo, results: PaginatedResult ) => void;
+type standardCallback = (Error: ErrorInfo, results: any) => void;
+type messageCallback = (Msg: any) => void;
```
mattheworiordan: Why not `Message` type?  Then perhaps add a `presenceMessageCallback` too?  Alternatively, perhaps better, create a new `subscribeCallback` using a generic that supports any type and specify the type later.

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+export declare class RealtimePresence {
+  syncComplete: () =>  boolean;
+
+  get: (Params: RealtimePresenceParams, callback?: PaginatedResultCallback) => void;
+
+  history: (ParamsOrCallback: RealtimePresenceHistoryParams | PaginatedResultCallback, callback?: PaginatedResultCallback) => void;
+
+  subscribe: (presenceOrCallback: PresenceAction | messageCallback, listener?: standardCallback, callback?: standardCallback) => void;
+
+  unsubscribe: (presence?: PresenceAction, listener?: standardCallback, callback?: standardCallback) => void;
```
mattheworiordan: As above.

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+
+  history: (params: RestPresenceHistoryParams, PaginatedResultCallback) => void;
+}
+
+export declare class RealtimePresence {
+  syncComplete: () =>  boolean;
+
+  get: (Params: RealtimePresenceParams, callback?: PaginatedResultCallback) => void;
+
+  history: (ParamsOrCallback: RealtimePresenceHistoryParams | PaginatedResultCallback, callback?: PaginatedResultCallback) => void;
```
mattheworiordan: If we supported generics, we could have a `PaginatedResultCallback<PresenceMessage>` here

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+
+  subscribe: (presenceOrCallback: PresenceAction | messageCallback, listener?: standardCallback, callback?: standardCallback) => void;
+
+  unsubscribe: (presence?: PresenceAction, listener?: standardCallback, callback?: standardCallback) => void;
+
+  enter: (Data: any, callback?: standardCallback) => void;
```
mattheworiordan: this is just an `errCallback`, same applies to all other `enter*`, `update*`, `leave*`

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+  updateClient: (clientId: String, Data: any, callback?: standardCallback) => void;
+
+  leaveClient: (clientId: String, Data: any, callback?: standardCallback) => void;
+}
+
+export declare class Channel {
+  name: String;
+  presence: Presence;
+  history: (ParamsOrCallback?: RestPresenceHistoryParams | PaginatedResultCallback, callback?: PaginatedResultCallback ) => void;
+  publish: (messages: any, MessageDataOrCallback?: any | errorCallback, callback?: errorCallback) => void;
```
mattheworiordan: first agument is `messagesOrName`

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+
+  updateClient: (clientId: String, Data: any, callback?: standardCallback) => void;
+
+  leaveClient: (clientId: String, Data: any, callback?: standardCallback) => void;
+}
+
+export declare class Channel {
+  name: String;
+  presence: Presence;
+  history: (ParamsOrCallback?: RestPresenceHistoryParams | PaginatedResultCallback, callback?: PaginatedResultCallback ) => void;
```
mattheworiordan: Could we type this for `Message`? Extra whitespace too at the end.

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+  request: (method: string, path: string, parmas?: any, body?: Array<any> | any, headers?: any, callback?: (Error: ErrorInfo, results: HttpPaginatedResponse) => void) => void;
+  stats: (ParamsOrCallback?: any, callback?: PaginatedResultCallback) => void;
+  time: (paramsOrCallback?: any, callback?: (Error: ErrorInfo, time: number ) => void) => void; // time
+}
+
+export declare class Realtime {
+  constructor(options: String | ClientOptions);
+
+  auth: Auth;
+  channels: Channels;
```
mattheworiordan: This should reference `Channels<RealtimeChannel>` as it does not return a REST `Channel`

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+
+  detach:(callback?: standardCallback) => void;
+
+  history: (ParamsOrCallback?: RealtimePresenceHistoryParams | PaginatedResultCallback, callback?: PaginatedResultCallback) => void;
+
+  subscribe: (eventOrCallback: String | messageCallback, listener?: messageCallback) => void;
+
+  unsubscribe: (eventOrCallback?: String | messageCallback, listener?: messageCallback) => void;
+
+  publish: (messages: any, MessageDataOrCallback?: any | errorCallback, callback?: errorCallback) => void;
```
mattheworiordan: `messages: any` should be `messagesOrName: any`

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+  name: String;
+  errorReason: ErrorInfo;
+  state: ChannelState;
+  presence: RealtimePresence;
+
+  attach: (callback?: standardCallback) => void;
+
+  detach:(callback?: standardCallback) => void;
+
+  history: (ParamsOrCallback?: RealtimePresenceHistoryParams | PaginatedResultCallback, callback?: PaginatedResultCallback) => void;
```
mattheworiordan: As above, would be better to type to `Message`

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+
+  subscribe: (eventOrCallback: String | messageCallback, listener?: messageCallback) => void;
+
+  unsubscribe: (eventOrCallback?: String | messageCallback, listener?: messageCallback) => void;
+
+  publish: (messages: any, MessageDataOrCallback?: any | errorCallback, callback?: errorCallback) => void;
+
+}
+
+export declare class Channels {
```
mattheworiordan: This needs to support generics so that `Channels` can return `RealtimeChannel` or `Channel` objects

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+
+  fromEncoded: (JsonObject: String, channelOptions: ChannelOptions) => Message;
+
+  fromEncodedArray: (JsonArray: String, channelOptions: ChannelOptions) => Array<Message>;
+
+  clientId: String;
+  connectionId: String;
+  data: any;
+  encoding: String;
+  extras: String;
```
mattheworiordan: This is not a string, it's an Object `any`

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+  id: String;
+  name: String;
+  timestamp: number; // Time
```
mattheworiordan: As before, milliseconds since epoch

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+}
+
+export declare class Message {
+  constructor();
+
+  fromEncoded: (JsonObject: String, channelOptions: ChannelOptions) => Message;
+
+  fromEncodedArray: (JsonArray: String, channelOptions: ChannelOptions) => Array<Message>;
+
+  clientId: String;
```
mattheworiordan: Should we specify that this is optional or is this unnecessary?

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+  timestamp: number; // Time
+}
+
+export declare class PresenceMessage {
+  fromEncoded: (JsonObject: any, channelOptions?: ChannelOptions) => PresenceMessage;
+
+  fromEncodedArray: (JsonArray: Array<any>, channelOptions?: ChannelOptions) => Array<PresenceMessage>;
+
+  action: PresenceAction;
+  clientId: String;
```
mattheworiordan: As above, should we make this optional?

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+  data: any;
+  encoding: String;
+  id: String;
+  timestamp: number; // Time
+}
+
+export declare class Rest {
+  constructor(options: String | ClientOptions);
+  auth: Auth;
+  channels: Channels;
```
mattheworiordan: See above, this should refer to `Channels<Channel>` so that it returns REST channels

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+  request: (method: string, path: string, parmas?: any, body?: Array<any> | any, headers?: any, callback?: (Error: ErrorInfo, results: HttpPaginatedResponse) => void) => void;
```
mattheworiordan: `s/parmas/params/`

mattheworiordan: Personally I would rename `results` to `response` as that is what it is, i.e. a single response with multiple results.

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+  stats: (ParamsOrCallback?: any, callback?: PaginatedResultCallback) => void;
```
mattheworiordan: Should this not be `ParamsOrCallback?: any | PaginatedResultCallback<Stats>`?  Also, again, why upper case here and then lower case in the next line?  Can we make sure lowercase is used consistently?

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+  time: (paramsOrCallback?: any, callback?: (Error: ErrorInfo, time: number ) => void) => void; // time
```
mattheworiordan: extra whitespace after `time: number`

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+
+export declare class Realtime {
+  constructor(options: String | ClientOptions);
+
+  auth: Auth;
+  channels: Channels;
+  clientId: String;
+  connection: Connection;
+
+  request: (method: string, path: string, parmas?: any, body?: Array<any> | any, headers?: any, callback?: (Error: ErrorInfo, results: HttpPaginatedResponse) => void) => void;
```
mattheworiordan: `s/params/params`.  All other comments from above apply too

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+
+  stats: (ParamsOrCallback?: any, callback?: PaginatedResultCallback) => void;
```
mattheworiordan: Comments from above apply too

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+
+  close: () => void;
+
+  connect: () => void;
+
+  time: (paramsOrCallback?: any, callback?: (Error: ErrorInfo, time: number ) => void) => void; // time
```
mattheworiordan: comments from above apply too

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+  key: String;
+  recoveryKey: String;
+  serial: number;
+  state: ConnectionState;
+
+  close: () => void;
+
+  connect: () => void;
+
+  ping: (callback?: (Error: ErrorInfo, ResponseTime: number ) => void ) => void;
```
mattheworiordan: Upper case vars here, extra whitespace

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+}
+
+export declare class Stats {
+  all: StatsMessageTypes;
+  apiRequests: StatsRequestCount;
+  channels: StatsResourceCount;
+  connections: StatsConnectionTypes;
+  inbound: StatsMessageTraffic;
+  intervalId: String;
```
mattheworiordan: Not sure if someone else added a comment here, but `intervalTime` is part of the IDL, but not sure if it's supported by the JS lib, @SimonWoolf?

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+
+
+export declare class PaginatedResult {
+  items: Array<any>;
+
+  first: (PaginatedResultCallback) => void;
+
+  next: (PaginatedResultCallback) => void;
+
+  current: (PaginatedResultCallback) => void;
```
mattheworiordan: This is not part of the IDL, @SimonWoolf @paddybyers any reason to support it?

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+
+  hasNext: () => boolean;
+
+  isLast: () => boolean;
+
+}
+
+export declare class HttpPaginatedResponse extends PaginatedResult {
+  items: Array<String>;
```
mattheworiordan: This is a JsonOjbect not a string, so perhaps `Array<any>`

ably.ts.d
```diff
@@ -0,0 +1,400 @@
...
+
+  current: (PaginatedResultCallback) => void;
+
+  hasNext: () => boolean;
+
+  isLast: () => boolean;
+
+}
+
+export declare class HttpPaginatedResponse extends PaginatedResult {
```
mattheworiordan: If the type is included using generics, then the `items` won't need to be overwritten, but more importantly, according to this, `next()` will now return a `PaginatedResult` instead of a `HttpPaginatedResponse` which is wrong.

ably.ts.d
SimonWoolf: No, in ably-js it's a method, per http://docs.ably.io/client-lib-development-guide/features/#RTP13

ably.ts.d
NathanaelA: Added

ably.ts.d
NathanaelA: Fixed

ably.ts.d
NathanaelA: Fixed

ably.ts.d
NathanaelA: Fixed

ably.ts.d
NathanaelA: Fixed

ably.ts.d
NathanaelA: Fixed

ably.ts.d
NathanaelA: Changed

ably.ts.d
NathanaelA: Switched to Generics.

ably.ts.d
NathanaelA: See Line 10390 in ably.js -- It is a function that returns a boolean.  :grinning:

ably.ts.d
NathanaelA: Ah, I missed that -- The Presence -> Get returns a PaginatedResultSet of PresenceMessage.  Seems weird that this one doesn't also...

ably.ts.d
NathanaelA: Fixed

ably.ts.d
NathanaelA: Fixed

ably.ts.d
NathanaelA: Sorry, I'm not following you; are you meaning Generics, if so fixed.

ably.ts.d
NathanaelA: Fixed, and typed.

ably.ts.d
NathanaelA: Generics?

ably.ts.d
NathanaelA: Fixed

ably.ts.d
NathanaelA: Fixed

ably.ts.d
NathanaelA: This is a class definition.  Everything is optional.

ably.ts.d
NathanaelA: Fixed

ably.ts.d
NathanaelA: Fixed/Changed.

ably.ts.d
NathanaelA: Fixed.

ably.ts.d
NathanaelA: Fixed

ably.ts.d
NathanaelA: Fixed

ably.ts.d
NathanaelA: Fixed

ably.ts.d
NathanaelA: Fixed

ably.ts.d
NathanaelA: Its part of the IDL but not part of the JS lib.

ably.ts.d
NathanaelA: Fixed.

ably.ts.d
NathanaelA: This was my "Addition" as I was going through the JS lib to fix issues that the IDL didn't have or specify correctly I figured this was another one that was missed in the IDL.  If you don't want it; it is trivial to remove.

ably.ts.d
NathanaelA: Fixed

ably.ts.d
NathanaelA: Fixed

NathanaelA:
@mattheworiordan @SimonWoolf - Ok, another version.  Let me know if you see anything out of place now.
Generics are in place; and all the ``// Time `` comments are gone.   Everything should be cased properly; and the only item that I am unsure if you want to leave in the typings is the ``current()``

ably.ts.d
tcard: The IDL for stats is somewhat less reliable since it comes from the Ruby lib. The Stats type is not defined in detail in the spec.

ably.ts.d
```diff
@@ -195,72 +205,90 @@ interface LogInfo {
...
+  current: ConnectionState;
+  previous: ConnectionState;
+  reason?: ErrorInfo;
+  retryIn?: number;
+}
+
+// Common Listeners
+type PaginatedResultCallback<T> = (error: ErrorInfo, results: PaginatedResult<T> ) => void;
+type standardCallback = (error: ErrorInfo, results: any) => void;
+type messageCallback = (message: Message) => void;
```
mattheworiordan: We either need a presenceCallback or we could make this a generic such as `messageCallback<T> = (message: <T>)`, see below for why

ably.ts.d
```diff
@@ -195,72 +205,90 @@ interface LogInfo {
...
   syncComplete: () =>  boolean;
 
-  get: (Params: RealtimePresenceParams, callback?: PaginatedResultCallback) => void;
+  get: (Params: RealtimePresenceParams, callback?: (error: ErrorInfo, messages: Array<PresenceMessage>) => void) => void;
 
-  history: (ParamsOrCallback: RealtimePresenceHistoryParams | PaginatedResultCallback, callback?: PaginatedResultCallback) => void;
+  history: (ParamsOrCallback: RealtimePresenceHistoryParams | PaginatedResultCallback<PresenceMessage>, callback?: PaginatedResultCallback<PresenceMessage>) => void;
 
-  subscribe: (presenceOrCallback: PresenceAction | messageCallback, listener?: standardCallback, callback?: standardCallback) => void;
+  subscribe: (presenceOrCallback: PresenceAction | messageCallback, listener?: messageCallback) => void;
```
mattheworiordan: this should be a presenceCallback as it returns a `PresenceMessage`

ably.ts.d
```diff
@@ -195,72 +205,90 @@ interface LogInfo {
...
 
-  unsubscribe: (presence?: PresenceAction, listener?: standardCallback, callback?: standardCallback) => void;
+  unsubscribe: (presence?: PresenceAction, listener?: messageCallback) => void;
```
mattheworiordan: as above

ably.ts.d
mattheworiordan: Fine, OK to leave in

ably.ts.d
mattheworiordan: Well it doesn't actually come from the Ruby lib btw. It comes from the Java lib, which I copied in Ruby, and simply reference the Ruby in our docs because of rubydocs which makes it easier to browse the definitions.  Anyway, @SimonWoolf can we add an issue to add that support in the JS lib down the line because it should match the IDL.

ably.ts.d
mattheworiordan: Yes, for a generic

ably.ts.d
mattheworiordan: @NathanaelA the reason it doesn't is because the Realtime client keeps an up to date set of all presence members in memory, whereas the REST version obtains it from the server upon request.  So in the case of Realtime there is no need for paging given the set is in memory anyway

ably.ts.d
NathanaelA: Changed.

NathanaelA:
Let me know if you have any other changes.  Seems like we are finally almost done.  :grinning:

On the Property vs Routine ``SyncComplete``; if you want it to be a property, it is trivial to change the JS lib into a property by using Object.defineProperty(...);   However this obviously would be a breaking change for anyone using the old function...   If you guys have done any other breaking changed in 0.9 then it might be worth getting it into place.

mattheworiordan:
@NathanaelA LGTM.

On the property, over to @SimonWoolf and @paddybyers to decide if the ably-js lib should conform or we should change the spec, I suspect the former.  Either way I think what you've done is right and that would be treated separately.  👍 to merge into 0-9

mattheworiordan:
@paddybyers @SimonWoolf I would like to merge, any objections?

paddybyers:
> I would like to merge, any objections?

LGTM

PR #362, opened 2016-12-20
Add NativeScript support

NathanaelA:


Gruntfile.js
```diff
@@ -78,6 +78,10 @@ module.exports = function (grunt) {
 			dest: '<%= dirs.dest %>/ably-commonjs.js',
 			nonull: true
 		},
+        'ably-nativescript': {
```
paddybyers: Spaces vs tabs whitepace issue here?

common/lib/util/utils.js
```diff
@@ -129,10 +129,16 @@ var Utils = (function() {
...
-	Utils.inherits = (typeof(require) === 'function' && require('util') && require('util').inherits) || function(ctor, superCtor) {
+
+	Utils.inherits = function(ctor, superCtor) {
 		ctor.super_ = superCtor;
 		ctor.prototype = Utils.prototypicalClone(superCtor.prototype, { constructor: ctor });
 	};
+	try {
+        if (typeof(require) === 'function' && require('util') && require('util').inherits) {
+        	Utils.inherits = require('util').inherits;
+		}
```
paddybyers: Whitespace issues

browser/lib/util/webstorage.js
```diff
@@ -1,6 +1,7 @@
 var WebStorage = (function() {
 	var sessionSupported,
 		localSupported,
+        mainClass = (typeof window !== 'undefined' ? window : global),
```
paddybyers: Wouldn't `globalObject` be more descriptive than `mainClass` ?

browser/fragments/platform-nativescript.js
```diff
@@ -0,0 +1,71 @@
...
+            buffer = chars.indexOf(buffer);
+        }
+        return output;
+    };
+
+var randomBytes;
+if (global.android) {
+    randomBytes = function(size) {
+        var sr = new java.security.SecureRandom();
+        var buffer = java.lang.reflect.Array.newInstance(java.lang.Byte.class.getField("TYPE").get(null), size);
```
paddybyers: Isn't there now a more efficient way of doing this? (https://github.com/NativeScript/android-runtime/issues/70)

browser/fragments/platform-nativescript.js
```diff
@@ -0,0 +1,71 @@
+require('nativescript-websockets');
+
+/**
+ * atob polyfill
+ * @param input
+ * @returns {string}
+ * Based on From: https://github.com/davidchambers/Base64.js/blob/master/base64.js
+ * License is Apache 2.0 also
+ */
+global.atob = function(input) {
```
paddybyers: Ably-js already has a base64 decode in `browser/lib/util`. We should avoid duplication and choose a single version.

browser/fragments/platform-nativescript.js
NathanaelA: LOL, I was even subscribed to that thread, but I didn't see it actually get finally resolved, with updated code,   I'll update this pull request.

browser/lib/util/webstorage.js
NathanaelA: I have no issue switching this.   And yeah I agree `globalObject` does make more sense.

browser/fragments/platform-nativescript.js
NathanaelA: Thanks, I'll look at how to use that atob function properly.

browser/lib/util/webstorage.js
```diff
@@ -9,25 +10,26 @@ var WebStorage = (function() {
...
 		localSupported = false;
 	}
 
-	function WebStorage() {}
+
+    function WebStorage() {}
 
 	function storageInterface(session) {
-		return session ? window.sessionStorage : window.localStorage;
+		return session ? globalObject.sessionStorage : globalObject.localStorage;
```
SimonWoolf: Will any of this do anything? At a quick google, [nativescript doesn't support either localstorage or sessionstorage APIs](https://stackoverflow.com/questions/39463157/is-there-localstorage-in-nativescrpt). It does have an `application-settings` api which would broadly be equivalent to localStorage, but you're not using that here. Am I missing something?

(In any case, sessionstorage in ably-js is used for persisting connection state across page refreshes -- is that useful for nativescript apps?)

common/lib/util/utils.js
```diff
@@ -129,10 +129,16 @@ var Utils = (function() {
...
+    } catch (err) {  /* Do Nothing. */ }
```
NathanaelA: The issue is that both `NativeScript` and `Node` both will throw an exception if a module does not exist, when you try to require it. 

 In this case this code assumes that if `require` exists then the module "util" exists.   It doesn't on all platforms.   (I'm not sure what react-native does here, or maybe a util module exists in RN?)    So since I have no idea how RN works, the simplest solution to this "issue" is to catch an expected error on platforms that might not have a `util` module, and do nothing, as `inherits` by default gets set first, then overrode by the `util` if it exists.

browser/lib/util/webstorage.js
SimonWoolf: Thanks. I get that `window` won't be defined, but if we just wanted to not support webstorage for nativescript, we could just put `if(typeof window == 'undefined') return;` at the top rather than bother having logic to make it look in the global object which won't have the required properties anyway. (Unless someone installed your plugin, which I don't think is something we should bet on). (Node has `WebStorage` undefined too, so other parts of the library expect that).

Re how webstorage is used: sessionstorage is used to persist connection state across page refreshes; localstorage is used to store transport preferences.

Questions:
- I'm guessing persisting connection state across page refreshes isn't a thing in nativescript, and once the client is initialized, it will exist for the duration that the app is open, so sessionstorage won't be useful. Is that right?
- But nativescript does support both xhr and websocket, so persisting transport preferences in localstorage is useful - right?

If that's the case, then perhaps we should have `sessionsSupported` be false, but `localSupported` be true and `WebStorage.get/set/remove` implemented here as a layer around the nativescript `application-settings` mechanism. What do you think?

For cleanness, if we're doing that, then should probably implement a separate `webstorage.js` for nativescript and include that separately in the grunt concatenator (ie split it out of `ablyFiles`).

common/lib/util/utils.js
SimonWoolf: Thanks. I think it's cleaner to support this by adding `inspects` and `inherits` keys to the `Platform` object, which will avoid the need to have a bunch of try-catches. WDYT?

browser/lib/util/webstorage.js
NathanaelA: When I wrote the patch around WebStorage, I wasn't sure if we needed it for anything.  So it made more sense to make the webStorage check the global object, this also allows this code to work inside browser webworkers as a webworker also has no `window` variable; but does have a `global` object with webstorage attached; so I was attempting to not only fix it so if NS needed it, but make Ably also be a bit more WebWorker friendly.  :-D

1. Yes, once something is created in NativeScript it exists forever or until GC'd (just like Node). 
2. Yes, NS has support for both xhr and WebSocket.  Persisting the info might be useful in the cases where WebSocket is always failing, but doesn't ably by default try WS then XHR; so in most cases I wouldn't think this info would not need to be persisted between app runs.

common/lib/util/utils.js
NathanaelA: Would definitely make it cleaner and eliminate two try/catch's.    Is this something where an issue needs to be created to fix after the pull is accepted?   Or how do you want to handle this?

browser/lib/util/webstorage.js
SimonWoolf: > a webworker also has no window variable; but does have a global object with webstorage attached

Are you sure? I've never used webworkers, but from what I can tell from a quick google (eg [this](http://stackoverflow.com/a/6179599)), that doesn't seem to be correct?

> doesn't ably by default try WS then XHR

No, it starts with the lowest-common-denominator thing (xhr) then upgrades to ws where it can. That avoids having a delay before connection on old platforms. Then once it does upgrade, it stores websocket as a preference, so next time it knows it can go straight for that. That's why I'm keen to have localstorage if possible.

Any thoughts on my earlier suggestion:

> If that's the case, then perhaps we should have `sessionsSupported` be false, but `localSupported` be true and `WebStorage.get/set/remove` implemented here as a layer around the nativescript `application-settings` mechanism. What do you think?

> For cleanness, if we're doing that, then should probably implement a separate `webstorage.js` for nativescript and include that separately in the grunt concatenator (ie split it out of `ablyFiles`).

common/lib/util/utils.js
SimonWoolf: Nah, you can just add the change to this PR

NathanaelA: Other than don't we need to change the platforms on all of them, Browser, encryption, node, react, and nativescript.   That seems out of place for this pull request.

browser/lib/util/webstorage.js
SimonWoolf: Including your emulation layer seems unnecessary given that the WebStorage module is already intended as an abstraction layer; not to mention it requires modifying the global object. Could we not just pass on to [application-settings](http://docs.nativescript.org/cookbook/application-settings)? `appSettings.setString`, `getString` and `remove` look like exact analogues of `localStorage.setItem`, `getItem` and `removeItem`, so should be pretty straightforward?

common/lib/util/utils.js
SimonWoolf: > That seems out of place for this pull request.

...I mean, if I was fussy about keeping PRs strictly single-purpose, then I wouldn't have suggested it :)

If you'd really rather not do it yourself, I can make the change and you can rebase on that?

browser/lib/util/webstorage.js
NathanaelA: I have no issue with creating a new WebStorage wrapper, other than time -- it is way quicker and simpler to just include LocalStorage and be done with it.   But if your preference is to switch this over, I can do that.

common/lib/util/utils.js
NathanaelA: Ok.  Wasn't sure -- some people are VERY picky about it.  ;-)      I have no issues making the changes.

Gruntfile.js
NathanaelA: That is weird, normally my editor is very picky about this and tells me when there is mixed spaces vs tabs.   (

 And my second editor I looked at didn't show anything, and GitDiff didn't show it.   But I have an editor that can show tabs vs spaces as characters.   I'll double check.   

Do you guys prefer spaces or tabs in the code?

browser/lib/util/webstorage.js
NathanaelA: It is trivial to enable localstorage as I have a plugin that does it -- and quicker too; but for long term maintenance Simon is correct; it would be better to eliminate the any additional plugins/modules that you can.      

@mattheworiordan I will defer to whatever you prefer:
1. Use my localstorage plugin, no other changes needed.
2. Create a new WebStorage wrapper for NativeScript that uses the internal appsettings.

Gruntfile.js
paddybyers: Tabs pls, unless editing an existing file that doesn't use tabs (eg `package.json`)

common/lib/util/utils.js
SimonWoolf: This is a 5 minute change; really not much point spending more than that debating who's going to do it ☺
I've gone ahead done it on 0-9-spec branch; please rebase on that

browser/lib/util/webstorage.js
SimonWoolf: >  it is way quicker and simpler to just include LocalStorage and be done with

Sorry -- I'm not sure if I'm missing something, but I'm not sure I see the difficulty. Here's a version of WebStorage tweaked to use application-settings based on the nativescript docs; please test and let me know if it works.

(To expand on the reasoning above -- as well as the double abstraction layer objection, Paddy's gone to some lengths to avoid modifying global objects in the lib (see eg most of utils.js), so I'd prefer not to use the lib that monkeypatches localstorage onto the global object. So I'd prefer this solution).

```javascript
var WebStorage = (function() {
	var appSettings = require("application-settings");
		test = 'ablyjs-storage-test';

	function WebStorage() {}

	function set(name, value, ttl) {
		var wrappedValue = {value: value};
		if(ttl) {
			wrappedValue.expires = Utils.now() + ttl;
		}
		return appSettings.setString(name, JSON.stringify(wrappedValue));
	}

	function get(name) {
		var rawItem = appSettings.getString(name);
		if(!rawItem) return null;
		var wrappedValue = JSON.parse(rawItem);
		if(wrappedValue.expires && (wrappedValue.expires < Utils.now())) {
			appSettings.remove(name);
			return null;
		}
		return wrappedValue.value;
	}

	WebStorage.set    = function(name, value, ttl) { return set(name, value, ttl); };
	WebStorage.get    = function(name) { return get(name); };
	WebStorage.remove = function(name) { return appSettings.remove(name); };

	return WebStorage;
})();
```

NathanaelA: @SimonWoolf - You will notice I said the best course was the one YOU suggested; but adding only two lines of code to the ns wrapper and not modifying any other files is WAY quicker than having to update the gradle build to eliminate webstorage from the general files and add it to each of the separate build files.    That was my only point.  :-D     In terms of QUICKNESS; 2 lines in one place is always going to be faster than any other solution.  :grinning:

NathanaelA: Btw, I don't see anything wrong with that code; should work great -- the "test" value isn't used.  :grinning:

common/lib/util/utils.js
NathanaelA: Awesome, thanks, I apologize if it seemed like I was being a pain; I love programming and to fix/debug things -- I also have to work very hard to stay inside scope sometimes...  :grinning:      

@mattheworiordan has only contracted a certain number of hours with me and stuff that occurs potentially outside of scope of my changes will impact the number of hours he has remaining.   So unfortunately unless he says yeah lets do this and Nathanael does it, I'm trying very hard not to use any of the time he has allocated for any other items without his permission.    If he want to convey to me that you have the ability to direct me also; that also would suffice as permission.  :grinning:

SimonWoolf:
> mattheworiordan has only contracted a certain number of hours with me and stuff that occurs potentially outside of scope of my changes will impact the number of hours he has remaining

Ah, fair enough then :)

I've pushed up a [0.9-nativescript](https://github.com/ably/ably-js/tree/0.9-nativescript) branch that rebases latest 0-9-spec, includes your changes and the Platform object changes, fixes the whitespaces issues, adds a nativescript-specific WebStorage wrapper, and hopefully fixes the msgpack issue you were running into. (There's already a version of msgpack that doesn't use bops included in the library, you just need to select it -- in the platform object, you had `msgpack: (typeof require === 'function') ? require('msgpack-js') : Ably.msgpack`, I've changed it to `msgpack: Ably.msgpack`. (I also changed the inclusion mechanism for that lib to ensure `Ably.msgpack` is always defined if that lib is included in the grunt concat; previously it tried to be clever and detect if it was running in a commonjs env and define it is a module if so so that `msgpack-js` would get that rather than the node/bops one, but presumably something wasn't working with that if you were running into issues)).

Could you test all that on nativescript and let me know if it works (assuming that's in scope - if not poke Matt on slack). Thanks :)

browser/lib/util/webstorage.js
SimonWoolf: > the "test" value isn't used. :grinning:

Fixed, thanks :)

NathanaelA:
@SimonWoolf - AWESOME!   You just saved me a lots of time.  :-)      I'll pull it and test it tonight and verify everything is good still.

NathanaelA:
@SimonWoolf - I tested your NS branch; only one small change appears to be needed.   I'm assuming we are closing this pull request in favor of your refactored NS branch, since it has everything we need.     

The atob in the NativeScript platform needs to wrap the Base64.decode in a function.  I didn't realize that the Base64 wouldn't be defined by that point.    So currently by the time it initiates the platform module the global "Base64" object is still undefined.    So the change is:   `atob: function(f) { return Base64.decode(f); },` and everything appears to be working fine.  ;-)

SimonWoolf:
Replaced by https://github.com/ably/ably-js/pull/366

PR #366, opened 2017-01-02
Add NativeScript support v2

SimonWoolf:
Replaces https://github.com/ably/ably-js/pull/362

@NathanaelA hope you had a good xmas! Made the atob change. Just to check -- if you haven't already, could you test on nativescript with the `useBinaryProtocol` clientOption set to `true` explicitly, since it's defaulting to `false` for nativescript?


SimonWoolf:
Merged into 0-9-spec as e1092b9. @NathanaelA if you're around, would still appreciate that useBinaryProtocol manual test, but since it's off by default, don't think I need to wait for that before merging

NathanaelA:
Sorry I had some vehicle problems the last couple days; I'll try and do this test shortly.  :grinning:

PR #376, opened 2017-02-09
TypeScript + JSPM + DefintelyTyped

mattheworiordan:
A customer got in touch as they had a problem getting Aurelia to work.  Aurelia uses JSPM (for package management) and is most often used with TypeScript.  As a result, I tried to set up a similar environment to see what issue they were having, and discovered that our ably-js lib needed a few tweaks.  I also discovered our TypeScript definition was not working at all, it was named incorrectly :(

Also see:
* https://github.com/ably/demo-typescript-jspm
* https://github.com/DefinitelyTyped/DefinitelyTyped/pull/14524

### DIFF
--- a/browser/fragments/ably-epilogue.js
+++ b/browser/fragments/ably-epilogue.js
@@ -22,11 +22,14 @@ if (typeof exports === "object" && exports) {
 			exports[obj] = window.Ably[obj];
 		}
 	}
+	/* SystemJS support for default exports to be added to the root of the module
+	   https://github.com/frankwallis/plugin-typescript/issues/185 */
+	exports.__esModule = true;
 }
 
 /* AMD support */
 if (typeof define === "function" && define.amd) {
   define("ably", [], function() {
-    return Ably;
+    return window.Ably;
   });
 }
--- a/browser/fragments/ably-epilogue.js
+++ b/browser/fragments/ably-epilogue.js
@@ -1,16 +1,32 @@
-if(typeof Realtime !== 'undefined') {
-	Ably.msgpack = msgpack;
-	Ably.Rest = Rest;
-	Ably.Realtime = Realtime;
-	Realtime.ConnectionManager = ConnectionManager;
-	Realtime.BufferUtils = Rest.BufferUtils = BufferUtils;
-	if(typeof(Crypto) !== 'undefined') Realtime.Crypto = Rest.Crypto = Crypto;
-	Realtime.Defaults = Rest.Defaults = Defaults;
-	Realtime.Http = Rest.Http = Http;
-	Realtime.Utils = Rest.Utils = Utils;
-	Realtime.Http = Rest.Http = Http;
-	Realtime.Message = Rest.Message = Message;
-	Realtime.PresenceMessage = Rest.PresenceMessage = PresenceMessage;
-	Realtime.ProtocolMessage = Rest.ProtocolMessage = ProtocolMessage;
-}
+	if(typeof Realtime !== 'undefined') {
+		Ably.msgpack = msgpack;
+		Ably.Rest = Rest;
+		Ably.Realtime = Realtime;
+		Realtime.ConnectionManager = ConnectionManager;
+		Realtime.BufferUtils = Rest.BufferUtils = BufferUtils;
+		if(typeof(Crypto) !== 'undefined') Realtime.Crypto = Rest.Crypto = Crypto;
+		Realtime.Defaults = Rest.Defaults = Defaults;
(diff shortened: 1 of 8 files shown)


browser/fragments/ably-epilogue.js
```diff
@@ -1,16 +1,35 @@
...
+	   https://github.com/frankwallis/plugin-typescript/issues/185 */
+	exports.__esModule = true;
+}
+
+/* AMD support */
+if (typeof define === "function" && define.amd) {
+  define("ably", [], function() {
+    return Ably;
+  });
+}
```
SimonWoolf: Shouldn't this all be in ably-commonjs-epilogue.js, not ably-epilogue.js? Wouldn't someone using commonjs be using ably-commonjs.js?

mattheworiordan: Well I know our approach before was to create separate versions, but is that necessary?  Why not add CommonJS support when CommonJS is being used?

mattheworiordan:
@paddybyers @SimonWoolf you happy for this to be merged then?

browser/fragments/ably-epilogue.js
SimonWoolf: > Well I know our approach before was to create separate versions, but is that necessary? Why not add CommonJS support when CommonJS is being used?

If that works, sure. I'm a bit fuzzy on how commonjs/amd/etc work, so happy to leave to your judgement.

But if we're doing that, we should probs remove the separate ably-commonjs.js (or rather, remove the separate generator and make it another version of the same file) -- we don't want to have  two different versions of ably-js which both support all module systems but do so in slightly different ways...

(don't mind if you'd rather merge this now and make an issue to sort it out later if you prefer)

mattheworiordan: > But if we're doing that, we should probs remove the separate ably-commonjs.js

True, although because we can't break backwards compatibility, the primary version always sets a global `Ably`, whereas the CommonJS one doesn't.  So I get what you're saying, but if we make `ably-commonj.sj` just use this ver then we're going to start polluting people's global namespace.

browser/fragments/ably-epilogue.js
```diff
@@ -1,16 +1,35 @@
...
+	}
+	/* SystemJS support for default exports to be added to the root of the module
+	   https://github.com/frankwallis/plugin-typescript/issues/185 */
+	exports.__esModule = true;
+}
+
+/* AMD support */
+if (typeof define === "function" && define.amd) {
+  define("ably", [], function() {
+    return Ably;
```
SimonWoolf: Not sure `Ably` is defined at this point? Isn't it scoped to the IIFE that was closed on line 16?

browser/fragments/ably-epilogue.js
```diff
@@ -1,16 +1,35 @@
...
+		Realtime.Utils = Rest.Utils = Utils;
+		Realtime.Http = Rest.Http = Http;
+		Realtime.Message = Rest.Message = Message;
+		Realtime.PresenceMessage = Rest.PresenceMessage = PresenceMessage;
+		Realtime.ProtocolMessage = Rest.ProtocolMessage = ProtocolMessage;
+	}
 }).call({});
+
+/* CommonJS support */
+if (typeof exports === "object" && exports) {
```
SimonWoolf: Not sure ` && exports` does anything, afaik objects are always truthy in JS?

browser/fragments/ably-epilogue.js
SimonWoolf: :+1: OK, fair enough.

Sorry, couple more questions though, see other comments

browser/fragments/ably-epilogue.js
mattheworiordan: Yes, that should be `window.Ably`, thanks

browser/fragments/ably-epilogue.js
mattheworiordan: Sorry, but you're wrong here.
`typeof(null) == "object" => true`

browser/fragments/ably-epilogue.js
mattheworiordan: Ok, fixed in a06b463

mattheworiordan:
Happy to merge now?

browser/fragments/ably-epilogue.js
SimonWoolf: ...Wow. [Right you are](https://tc39.github.io/ecma262/#sec-typeof-operator). I did not know that. That an... _interesting_ design choice.

SimonWoolf:
yep

PR #377, opened 2017-02-20
Simple logic to allow requesting and viewing channel attach flags

SimonWoolf:
So with debug log level, the attached message now looks like e.g. `[ProtocolMessage; action=ATTACHED; channel=channel; channelSerial=whatever; flags=RESUMED,SUBSCRIBE,PUBLISH,PRESENCE_SUBSCRIBE]` for ease of debugging


SimonWoolf:
Merged in 1cfe662

Issue #384, opened 2017-03-10
ably-js can send adouble attaches after a failed resume


SimonWoolf:
the first one is a forced reattach of all channels due to the failed resume, as expected. The second is `on(transport.active)` logic that calls `checkPendingState` for any channels that are in the `attaching` state when any new transport activates -- and the channel is now in the attaching state due to the first one.

the ordering for this is a bit of a mess since both are done in nextTicks, to allow connection state to finish being set before doing anything with channels

```
Ably: 2017-03-10T10:16:14.446Z: ConnectionManager retry timer expired: retrying
Ably: 2017-03-10T10:16:14.446Z: ConnectionManager.requestState(): requested state: connecting; current state: disconnected
Ably: 2017-03-10T10:16:14.446Z: ConnectionManager.cancelTransitionTimer():
Ably: 2017-03-10T10:16:14.446Z: Connection state: connecting
Ably: 2017-03-10T10:16:14.446Z: ConnectionManager.enactStateChange: setting new state: connecting; reason = undefined
Ably: 2017-03-10T10:16:14.446Z: ConnectionManager.startConnect(): starting connection
Ably: 2017-03-10T10:16:14.446Z: ConnectionManager.startTransitionTimer(): transitionState: connecting
Ably: 2017-03-10T10:16:14.446Z: Auth.getToken(): using cached token; expires = 1489144526847
Ably: 2017-03-10T10:16:14.446Z: ConnectionManager.getTransportParams(): Transport recovery mode = resume; connectionKey = 108rLpZCgA4pM9!xEpDxUPGUhFrDDC0-3108rLpZCgA4pM9; connectionSerial = -1
Ably: 2017-03-10T10:16:14.446Z: ConnectionManager.connectBase(): Trying to connect with base transport web_socket
Ably: 2017-03-10T10:16:14.446Z: ConnectionManager.tryATransport(): trying web_socket
Ably: 2017-03-10T10:16:14.446Z: WebSocketTransport.connect(): starting
Ably: 2017-03-10T10:16:14.446Z: WebSocketTransport.connect(): uri: wss://local-realtime.ably.io:443/
Ably: 2017-03-10T10:16:14.447Z: Auth.getToken(): using cached token; expires = 1489144526847
[...]

PR #385, opened 2017-03-10
Fix multiple attaches after a failed resume

SimonWoolf:
so channels#onTransportActive is now called synchronously with transport activation (but after connection state is changed), and any forced reattach happens in the next tick, and only for ATTACHED channels

Fixes https://github.com/ably/ably-js/issues/384


paddybyers:
Looks like it should work, thanks

PR #387, opened 2017-03-13
Undocumented feature: support transient channels

SimonWoolf:
See https://support.ably.io/helpdesk/tickets/1657 . Customer using ably-js serverside to publish to lots of different channels, never needs to subscribe, and needs lower latency than REST publishes give. Transient publishes seem like the appropriate solution, and now it's supported by realtime, seems a shame not to support it in ably-js, albeit undocumented for now (will need to explain to the customer that the api may change once we introduce it officially in 1.1/2.0).

API design: adds a `transient` channelOption, i.e.
`realtime.channels.get(channelName, {transient: true});`


SimonWoolf:
Replaced with https://github.com/ably/ably-js/pull/430

Issue #388, opened 2017-03-16
Keep track of last-known activity time, use it to discard connection state


SimonWoolf:
to avoid issues where computer goes to sleep or something and ends up 2 hours later trying to resume without the connection state having transitioned to `suspended`. See https://github.com/ably/realtime/issues/940#issuecomment-286985784

(NB can address https://github.com/ably/ably-js/issues/358 at the same time, find a way to use one mechanism for this and for transport idle timeout for efficiency)

mattheworiordan:
Sounds good, but expect this is a broader docs / spec issue and not just a JS issue as devices have similar problem no?

SimonWoolf:
:+1: https://github.com/ably/docs/issues/287

Issue #358, opened 2016-12-08
Revisit resetIdleTimeout() handling


paddybyers:
Currently this, which unsets and sets a timer, is called on receipt of every protocol message. This is needlessly expensive.

A much more lightweight approach would be to set the timer once, and remember the time of the last received message on each message. When the timer expires, the actual idle time can be calculated knowing the present time and the time of the last message. If this interval is less than the specified idle time, then start a new timer with the balance of the idle time.

This way if there is a continuous stream of messages we would be setting a timer once every 15s, say, instead of hundreds of times a second.

PR #389, opened 2017-03-20
Various connection state and idle timeout handling changes

SimonWoolf:
- Send the platform in the lib version string
- Remember lastActivity; use that for idle transport timeout (fixes https://github.com/ably/ably-js/issues/358)
- Don’t resume if last known activity was > connectionStateTtl ago (fixes https://github.com/ably/ably-js/issues/388)
- Use connectionStateTtl from connectionDetails

### DIFF
--- a/common/lib/transport/connectionmanager.js
+++ b/common/lib/transport/connectionmanager.js
@@ -102,6 +102,7 @@ var ConnectionManager = (function() {
 		this.connectionId = undefined;
 		this.connectionKey = undefined;
 		this.connectionSerial = undefined;
+		this.connectionStateTtl = timeouts.connectionStateTtl;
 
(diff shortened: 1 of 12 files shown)


common/lib/transport/connectionmanager.js
```diff
@@ -680,6 +682,14 @@ var ConnectionManager = (function() {
 		this.unpersistConnection();
 	};
 
+	ConnectionManager.prototype.checkConnectionStateFreshness = function() {
+		var sinceLast = Utils.now() - this.lastActivity;
+		if(sinceLast > this.connectionStateTtl) {
```
mattheworiordan: Shouldn't this be `connectionStateTtl + maxIdleInterval` though as there is a small window in which the no update may have been received by the client but the connection is still active?

paddybyers:
LGTM

PR #390, opened 2017-03-21
Remove unreachable GET /requesttoken branch

SimonWoolf:
There's no situation where Http.get is defined but Http.post isn't. (If using JSONP, Http.post still exists, it just happens to use a GET behind the scenes; the Http.Request interface abstracts it away)

cf https://github.com/ably/realtime/pull/957


paddybyers:
lgtm

PR #392, opened 2017-03-29
XHR Content TypeFixes, NativeScript Changes/Testing fixes

NathanaelA:
Here is the set of changes that fixes some actual NativeScript issues inside Ably, and changes that allows the nodeunit testing framework to work inside NativeScript.

### DIFF
--- a/browser/lib/transport/xhrrequest.js
+++ b/browser/lib/transport/xhrrequest.js
@@ -115,7 +115,7 @@ var XHRRequest = (function() {
 
 		if(body) {
 			var contentType = headers['content-type'] || (headers['content-type'] = 'application/json');
-			if(contentType == 'application/json' && typeof(body) != 'string')
+			if(contentType.indexOf('application/json') > -1 && typeof(body) != 'string')
 				body = JSON.stringify(body);
 		}
 
@@ -172,7 +172,7 @@ var XHRRequest = (function() {
 				var contentType = getHeader(xhr, 'content-type'),
 					headers,
 					server,
-					json = contentType ? (contentType == 'application/json') : (xhr.responseType == 'text');
+					json = contentType ? (contentType.indexOf('application/json') >= 0) : (xhr.responseType == 'text');
 
 				responseBody = json ? xhr.responseText : xhr.response;
 
--- a/spec/common/modules/testapp_manager.js
+++ b/spec/common/modules/testapp_manager.js
@@ -1,15 +1,18 @@
 "use strict";
+/* global define, isNativescript, fetch */
 
 /* testapp module is responsible for setting up and tearing down apps in the test environment */
 define(['globals', 'browser-base64', 'ably'], function(ablyGlobals, base64, ably) {
 	var restHost = ablyGlobals.restHost || prefixDomainWithEnvironment('rest.ably.io', ablyGlobals.environment),
 			tlsPort  = ablyGlobals.tlsPort;
 
+
 	var isBrowser = (typeof(window) === 'object'),
 			httpReq   = httpReqFunction(),
-			toBase64  = base64Function(),
-			loadJsonData = isBrowser ? loadJsonDataBrowser : loadJsonDataNode,
-			testResourcesPath = (isBrowser && window.__karma__ && window.__karma__.start ? 'base/' : '') + 'spec/common/ably-common/test-resources/';
+			toBase64  =  base64Function(),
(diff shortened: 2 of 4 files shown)


browser/lib/transport/xhrrequest.js
```diff
@@ -115,7 +115,7 @@ var XHRRequest = (function() {
 
 		if(body) {
 			var contentType = headers['content-type'] || (headers['content-type'] = 'application/json');
-			if(contentType == 'application/json' && typeof(body) != 'string')
```
SimonWoolf: `contentType` on this line is the content type of the body that the client is POSTing, so is set by us (ie ably-js) -- at this stage the request hasn't actually started yet. So there's no opportunity for it to be anything other than what we set it to, so this change isn't necessary.

browser/lib/transport/xhrrequest.js
```diff
@@ -172,7 +172,7 @@ var XHRRequest = (function() {
 				var contentType = getHeader(xhr, 'content-type'),
 					headers,
 					server,
-					json = contentType ? (contentType == 'application/json') : (xhr.responseType == 'text');
```
SimonWoolf: :+1:

spec/common/modules/testapp_manager.js
```diff
@@ -1,15 +1,18 @@
...
 
+
 	var isBrowser = (typeof(window) === 'object'),
 			httpReq   = httpReqFunction(),
-			toBase64  = base64Function(),
-			loadJsonData = isBrowser ? loadJsonDataBrowser : loadJsonDataNode,
-			testResourcesPath = (isBrowser && window.__karma__ && window.__karma__.start ? 'base/' : '') + 'spec/common/ably-common/test-resources/';
+			toBase64  =  base64Function(),
+			loadJsonData = isNativescript ? loadJsonNativescript : isBrowser ? loadJsonDataBrowser : loadJsonDataNode,
+			testResourcesPath = isNativescript ? '~/ably/ably-common/test-resources/' :
```
SimonWoolf: - This should be using a relative path, not an absolute one -- we can't assume that the person running the tests happens to have a folder called `ably` in their home dir and has cloned the ably-common repo to it.

- the ably-js repo has ably-common as a submodule (it's in `spec/common/ably-common/test-resources/`), so as well as being a relative path, it should be a relative path to that instance of ably-common, not a separately cloned instance.

- This assumes the existance of a global variable called `isNativescript`. If we're not using nativescript, that variable isn't false, it just doesn't exist, so this will crash with a reference error.

- Minor point, but not a huge fan of nested ternaries

NathanaelA: Sorry, I think you misunderstand where this will be running...  :-)

  `~` = the devices running application home path.  i'e.  If the app is named `com.ably.demo`  The files will be located at `/data/data/com.ably.demo/files/app/...` on an android device and on an iOS device the path is also a big old deep path that I don't really want to type  ;-) .     That is NOT the home path on linux/mac.     This is the correct, and it is a fixed path, and is ONLY for testing inside a NativeScript app.    However your second point; is accurate; now that I understand the layout of how your testing system works; I can actually direct the test data to the node_modules/ably folder without an issue.   Good catch, "my" layout has been in place since I first cloned all your tests into the app.

-  Your second point, part of it is correct some of these files are cloned due to the limitation for how NativeScript test framework works, and also the way you use nodeunit&define.   But we can do away with the testing data; those can be directed to the originals and not copied, along with pretty much everything else...   

3. You are correct, I can't believe I did that, I know better -- I'll update that code.  Thanks.

4. I hate them also; I just was quickly trying to fix it, months ago. :)  But your right technical debt is not a good thing.  I'll clean that area up...

browser/lib/transport/xhrrequest.js
NathanaelA: Ok, well you if you go up to line 112 you see the same type of check that someone at ably did.   Figured this spot needed it too...

SimonWoolf: Line 112 is the `accept` header, not `content-type`; it's specifying what content types it would be happy for the server to return, rather than the content type of the body it's sending. For auth requests ably-js sets the `accept` header to `application/json, text/plain` (since some auth servers return a json tokenDetails and some return just a token string), hence the use of indexOf there.

SimonWoolf:
Looking much nicer after the fixups, thanks @NathanaelA :+1: 

Quite busy today, but will run through ci and merge on mon or tues

SimonWoolf:
Merged as f7e28f9

PR #398, opened 2017-04-21
fix idle timeout handling on node with heartbeats=false

SimonWoolf:
https://github.com/ably/ably-js/pull/389 slightly broke idle timeout handling on node with heartbeats=false, not caught since realtime didn't act on that yet.

Not a blocker for realtime as heartbeats=false was previously disabled.

Will wait till realtime with heartbeats=false is on sandbox and this passes ci with that before merging.

### DIFF
--- a/common/lib/transport/comettransport.js
+++ b/common/lib/transport/comettransport.js
@@ -109,9 +109,7 @@ var CometTransport = (function() {
 					err = err || new ErrorInfo('Request cancelled', 80000, 400);
 				}
 				self.recvRequest = null;
-				if(this.maxIdleInterval) {
-					this.setIdleTimer();
-				}
+				self.onActivity();
 				if(err) {
 					if(err.code) {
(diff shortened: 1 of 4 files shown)


mattheworiordan:
@SimonWoolf does this mean if people are on old versions and we enable native heartbeats it may break?

paddybyers:
Native pings are already sent.

This change means that if your library explicitly says `?heartbeats=false` then you won't get protocol heartbeats. So we need to check that's ok with all libraries that currently do that. Apart from java, and nodejs, which others currently do that?

SimonWoolf:
> does this mean if people are on old versions and we enable native heartbeats it may break?

Nope, that's what I meant by `Not a blocker for realtime as heartbeats=false was previously disabled.` -- sorry if was unclear. (ie prior to this pr ably-js didn't send heartbeats=false even for node, as we were being cautious)

SimonWoolf:
Merged as 90cabd3

PR #404, opened 2017-06-21
Use relative path for browser alias to fix webpack module resolution.

mckaydavis:


SimonWoolf:
Hi @mckaydavis, sorry for the delay, this slipped under the radar -- I'll have a look at this today or tomorrow

mckaydavis:
@SimonWoolf No problem. Sorry for the somewhat terse pull message.  Here's some more info on why I made the pull request.

I can't share the code-base for a repro, but this is the original build error:
```
ERROR in ./src/sync/AblySyncTransport.ts
Module not found: Error: Can't resolve 'ably' in '/tmp/myproject/src/sync'
 @ ./src/sync/AblySyncTransport.ts 12:0-29
 @ ./src/store/StateSync.ts
 @ ./src/Session.ts
 @ ./~/babel-loader/lib!./~/vue-loader/lib/selector.js?type=script&index=0!./src/vue/App.vue
 @ ./src/vue/App.vue
 @ ./src/index.js
 @ multi webpack-hot-middleware/client?reload=true ./src/index.js
```
The project's TypeScript `tsconfig.json` has `"moduleResolution": "node"` and correctly finds Ably in `node_modules`, but webpack's module resolution does not.

I've attached [ably.webpack.error.log.txt](https://github.com/ably/ably-js/files/1113115/ably.webpack.error.log.txt), a log of the relevant Ably module resolution path failure output provided by `webpack --display-error-details`.

After looking at the log I prefixed the `"browser"` alias in `ably-js/package.json` with `./` and everything worked.

Looking at the other paths, the 'react-native' alias may have the same issue.

I checked out the [package-browser-field-spec](https://github.com/defunctzombie/package-browser-field-spec#alternate-main---basic) where it states "All paths for browser fields are relative to the `package.json` file location (and usually project root as a result)." -- and it seems to me that the non-prefixed non-absolute path should work, but I guess it fails in whatever code implements it.  Unfortunately the spec doesn't specify the resolution behavior if the path provided is not prefixed with `./`.

Hopefully this info helps future searchers (even outside Ably-js) who run into this error!

SimonWoolf:
Thanks @mckaydavis! Made the equivalent change to other paths in 465cdda.

PR #405, opened 2017-06-25
Impllement Channels#release

SimonWoolf:
to avoid a memory leak with loadtesting when doing a REST publish to a new channel every time


SimonWoolf:
Merged as b9e90fd

PR #406, opened 2017-06-28
Http requests with useBinaryProtocol: true

SimonWoolf:
There were several issue with using `useBinaryProtocol: true` with rest requests in a browser environment. Main one was that the `accept` header was being set to `application/x-msgpack,application/json`, and xhrrequest was treating any accept header with `application/json` in as setting the responseType to be set to `text`, but we need to set the responseType before exec-ing the request so we can't do that on the basis of the content-type of the response, and there's no reliable way of converting a string to an ArrayBuffer. Aargh.

So this fixes that and simplifies a few areas (only setting a single value in 'accept', given we know realtime supports both; removes support for a `format` authOption (it was undocumented and pretty unlikely anyone used it); makes sure that responseType is set to 'arraybuffer' if we request msgpack (and handling it if the server responds with json, since we can convert an arraybuffer that's valid utf8 to a string easily enough), etc.). And fixes a couple other bugs, and adds a bit of test coverage for REST requests with format=msgpack.

### DIFF
--- a/browser/lib/transport/xhrrequest.js
+++ b/browser/lib/transport/xhrrequest.js
@@ -165,17 +165,23 @@ var XHRRequest = (function() {
 			try {
 				var contentType = getHeader(xhr, 'content-type'),
 					headers,
-					server,
+					responseBody,
+					/* Be liberal in what we accept; buggy auth servers may respond
+					 * without the correct contenttype, but assume they're still
+					 * responding with json */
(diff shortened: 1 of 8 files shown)


SimonWoolf:
@paddybyers I'm plan to merge this and do a release in the next couple days, shout if you have any objections

paddybyers:
LGTM

PR #412, opened 2017-09-08
Include msgSerial in the recover data

SimonWoolf:
Fixes https://github.com/ably/realtime/issues/1224#issuecomment-327140332 for ably-js

### DIFF
--- a/common/lib/transport/connectionmanager.js
+++ b/common/lib/transport/connectionmanager.js
@@ -206,7 +206,15 @@ var ConnectionManager = (function() {
 		}
 
 		decideMode(function(mode) {
-			Logger.logAction(Logger.LOG_MINOR, 'ConnectionManager.getTransportParams()', 'Transport recovery mode = ' + mode + (mode == 'clean' ? '' : '; connectionKey = ' + self.connectionKey + '; connectionSerial = ' + self.connectionSerial));
+			if(mode === 'recover') {
+				Logger.logAction(Logger.LOG_MINOR, 'ConnectionManager.getTransportParams()', 'Transport recovery mode = recover; recoveryKey = ' + self.options.recover);
+				var match = self.options.recover.split(':');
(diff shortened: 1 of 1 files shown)


common/lib/transport/connectionmanager.js
```diff
@@ -652,13 +660,16 @@ var ConnectionManager = (function() {
...
-		if(this.connectionId !== connectionId)  {
-			Logger.logAction(Logger.LOG_MINOR, 'ConnectionManager.setConnection()', 'New connectionId; resetting msgSerial and reattaching any attached channels');
+		/* If no previous connectionId, don't reset the msgSerial as it may have been set by recover data */
+		if(this.connectionId && (this.connectionId !== connectionId))  {
+			Logger.logAction(Logger.LOG_MINOR, 'ConnectionManager.setConnection()', 'Resetting msgSerial');
 			this.msgSerial = 0;
+		}
+		/* but do need to reattach channels, for channels that were previously in
+		 * the attached state even though the connection mode was 'clean' due to a
+		 * freshness check - see https://github.com/ably/ably-js/issues/394 */
```
mattheworiordan: @SimonWoolf is this behaviour / bug definitely covered by the spec?  I suspect other libraries could have this same bug and thus attached channels prior to losing connection state will remain connected in the lib.

mattheworiordan:
LGTM. I think we should bump the version, release, and ask the customer who had an issue with this to test.

SimonWoolf:
Merged as 610014c

common/lib/transport/connectionmanager.js
SimonWoolf: @mattheworiordan it isn't in the spec because it's an artifact of the freshness check, which isn't yet in the spec -- the docs issue to do so is https://github.com/ably/docs/issues/287

PR #415, opened 2017-09-25
xhrrequest: send as POST if body is not null/undefined (even if empty string)

SimonWoolf:
In particular, Utils.toQueryString (used for token requests if authMethod is POST) returns an empty string if no authParams, and we want that to use POST


paddybyers:
LGTM

SimonWoolf:
Merged as 1d868e1

Issue #413, opened 2017-09-13
Token requests with no capabilities specified still include a `capabilities` key with an empty string


SimonWoolf:
see https://github.com/ably/ably-java/issues/349#issuecomment-329077763

SimonWoolf:
Fixed by #416

PR #416, opened 2017-09-25
Don't include empty-string capability in token request

SimonWoolf:
Fixes https://github.com/ably/ably-js/issues/413


SimonWoolf:
Merged as 1386f13

PR #424, opened 2017-10-20
Only fail the connection for token auth err if unrecoverable or 403

SimonWoolf:
Implementation of https://github.com/ably/docs/pull/322

### DIFF
--- a/common/lib/client/auth.js
+++ b/common/lib/client/auth.js
@@ -359,7 +359,7 @@ var Auth = (function() {
 		} else {
 			var msg = "Need a new token, but authOptions does not include any way to request one";
 			Logger.logAction(Logger.LOG_ERROR, 'Auth.requestToken()', msg);
-			callback(new ErrorInfo(msg, 40101, 401));
+			callback(new ErrorInfo(msg, 40101, 403));
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


common/lib/client/auth.js
```diff
@@ -437,11 +437,6 @@ var Auth = (function() {
 			tokenRequest(tokenRequestOrDetails, function(err, tokenResponse, headers, unpacked) {
 				if(err) {
 					Logger.logAction(Logger.LOG_ERROR, 'Auth.requestToken()', 'token request API call returned error; err = ' + Utils.inspectError(err));
-					if(!err.code) {
-						/* network errors don't have an error code, so assign them
-						 * 40170 so they'll be seen by connectionManager as nonfatal */
-						err = new ErrorInfo(Utils.inspectError(err), 40170, 401);
-					}
```
SimonWoolf: Note: removed this as errors.json technically reserves 40170 for errors from the client token callback, which this isn't. (Having only added it recently to stop network errors getting a token being fatal, which is no longer needed)

SimonWoolf:
...apparently I did this twice. oh well.

closing in favour of https://github.com/ably/ably-js/pull/434/files

PR #430, opened 2017-11-06
Enable transient publishes per new spec

SimonWoolf:
Spec: https://github.com/ably/docs/pull/323

Took a bit longer than expected, as one of the test failures turned out to be a realtime bug, https://github.com/ably/realtime/pull/1367 . To be safe, this shouldn't be released until that fix is deployed.


SimonWoolf:
Merged as 891a37c

PR #434, opened 2017-11-21
Make auth failures only send realtime conns to FAILED if 403

SimonWoolf:
Implements https://github.com/ably/docs/pull/322


spec/realtime/auth.test.js
```diff
@@ -8,7 +8,9 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
 		closeAndFinish = helper.closeAndFinish,
 		monitorConnection = helper.monitorConnection,
 		testOnAllTransports = helper.testOnAllTransports,
-		mixin = helper.Utils.mixin;
+		mixin = helper.Utils.mixin,
+		echoServer = "http://echo.ably.io";
+		//echoServer = "http://localhost:5000";
```
mattheworiordan: Not sure this is necessary, but I suppose it doesn't matter

Issue #429, opened 2017-10-30
Support for receiver-side browser push notifications.


tcard:
We currently have server-side support for receiving push notifications in the browser [as implemented in Firefox](https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API). We'd like to integrate this into the client-side API, following [the push receiver spec](https://docs.ably.io/client-lib-development-guide/features/#RSH2) as closely as possible.

We'd also need to make sure it works for Chrome, Safari and others. As of today, I've been using [this little Go program](https://gist.github.com/tcard/04e4371326d5a9108444b515db02ce19) to test device registration, channel subscription and receiving notifications, but it only works in Firefox. The current server-side implementation expects a `targetUrl` and an `encryptionKey`, both unique per device (browser), but this might need to be adapted too. (I hope not, since that's supposed to be [a W3C standard](https://w3c.github.io/push-api).)

tcard:
Missing:

* Ably's application server key is currently hardcoded. Probably needs to be served. https://github.com/ably/ably-js/blob/183a62dd7cb381262ad884677d126a966c8cb64e/browser/fragments/platform-browser.js#L77
* Safari support. See https://github.com/ably/wiki/issues/242 https://github.com/ably/ably-js/blob/183a62dd7cb381262ad884677d126a966c8cb64e/browser/fragments/platform-browser.js#L123
* Form factor needs to be detected. https://github.com/ably/ably-js/blob/183a62dd7cb381262ad884677d126a966c8cb64e/browser/fragments/platform-browser.js#L30
* Documentation for `ClientOptions.pushServiceWorkerUrl`, the argument passed to [`navigator.serviceWorker.register`](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer/register). For W3C push only.
* Push activation documentation for JS.
* Tests. See [ably-java tests](https://github.com/ably/ably-java/blob/push/android/src/androidTest/java/io/ably/lib/test/android/AndroidPushTest.java) for inspiration.

tcard:
Also, PushChannel is not implemented at all.

Peter-Maguire:
Got it - this is blocked due to the backend web push API support not being complete. The backend requires a public/private ECDSA keypair to be generated, but browsers do not support exporting the keys in a format that the backend can understand. The general consensus seems to be that work on JS push would be better directed towards react-native, as there is more of a demand and already proven backend support for FCM/APN

jamienewcomb:
[...]

PR #435, opened 2017-12-06
WIP: Push activation

tcard:
Fixes https://github.com/ably/ably-js/issues/429.


browser/lib/util/ulid.js
```diff
@@ -0,0 +1,185 @@
...
+		}
+		if (allowInsecure) {
+			try {
+				console.error("secure crypto unusable, falling back to insecure Math.random()!");
+			} catch (e) {}
+			return function () {
+				return Math.random();
+			};
+		}
+		throw createError("secure crypto unusable, insecure Math.random not allowedW");
```
mattheworiordan: Typo, trailing `W`

tcard: Well, that makes for a juicy PR upstream :)

https://github.com/ulid/javascript/pull/59

paddybyers:
Looking good so far.

I don't like the fact that we don't have the state machine in node. I know there's no transport currently that we can use for this, but I'd prefer that the state machine implementation was just a generic part of `Push`, and it calls the initialiser of some underlying object/transport when activated, which throws an exception in node to say it's not supported.

In a similar vein, the actual code implementing the subscription (https://github.com/ably/ably-js/blob/dc984ae15e3d993986127f1e6f9416c9e6448823/browser/lib/client/push.js#L132-L168) is embedded within the state machine, whereas I'd prefer that all of these platform-specific primitive operations were encapsulated in a `PushTransport` class or something, so the state machine could be used with some other transport. For example we could, for testing purposes, implement a transport that uses the `ablyChannel` mechanism so we can exercise the state machine in node tests. Or, is it possible for us to have a fcm client in node?

Also in js can we use `const` where possible pls.

SimonWoolf:
> Also in js can we use const where possible pls.

ably-js nominally still supports back to IE8 and node 0.12, we're a long way from being able to use `const`

paddybyers:
> ably-js nominally still supports back to IE8 and node 0.12, we're a long way from being able to use const

Oops, yes. In that case there's lots of `let` and `const` in `Push.js` that needs fixing.

tcard:
Sorry, fixed those.

Do we have some kind of syntax checker in place?

paddybyers:
> Do we have some kind of syntax checker in place?

I don't see any `jshint` task or equivalent in the gruntfile. @SimonWoolf ?

SimonWoolf:
nope, as far as I know we've never had a linter beyond the syntax checking provided by the closure-compiler minifier. I agree that it would be a good idea.

funkyboy:
Up to this point push tests pass both on `nodeunit` and `karma`. 
Tests have been run on Chrome 63.0.3239.132 and Firefox 57.0.4 (both on macOS 10.12.6).
Tests are still heterogeneous tho, some in `JSON` some in `msgpack`.

QuintinWillison:
Replaced by #648

jessica02200:
How does work?

Issue #409, opened 2017-08-02
Change closeOnUnload to default to true


mattheworiordan:
We have lots of customers get confused as to why their connection counts are going high as they refresh browsers. As we anecdotally expect most people are not using resume anyway between refreshes, to avoid this common issue, should we enable close on unload by default?

See support tickets 2120 for a recent occurrence of this.

@paddybyers WDYT?

paddybyers:
> using resume

it would be recover

but yes, I agree ... pretty much all apps will reinitialise in a way that makes recovery not sensible or useful. If they go out of their way to support that, then we can show them how.

mattheworiordan:
>> using resume

> it would be recover

Yes, sorry, wrong terminology.

Glad you agree though. @SimonWoolf perhaps we could change when you're back and notify our customers in a newsletter.

mattheworiordan:
Or hold off until 1.1 so that we'd expect customers to read the release notes?

SimonWoolf:
> Or hold off until 1.1 so that we'd expect customers to read the release notes?

Agreed, flipping a default probably isn't something we should do in a patch release (it is technically a breaking change, if only for people who are extracting the recoveryKey manually and passing that in as a string)

SimonWoolf:
Fixed by https://github.com/ably/ably-js/pull/439

PR #439, opened 2017-12-11
default to closeOnUnload unless we have some indication that a user may want to recover the connection

SimonWoolf:
Fixes https://github.com/ably/ably-js/issues/409

Bringing forward as we've had so many support tickets about this, and even though it's technically a breaking change, it's unlikely to affect anyone if we only apply it to people who don't use `recover`.


paddybyers:
LGTM, thanks

SimonWoolf:
Merged as 293aed1

PR #442, opened 2017-12-21
Move Webstorage to instance scope; implement in nodejs

paddybyers:
We need persistent storage for push of a deviceId and other credentials; currently this is only available for the browser. Our support push for nodejs will only be for testing purposes, but we do want it to be functional, and so we require storage.

This PR adds persistent storage for nodejs by persisting to a file whose name is derived from the `keyName` so that stored data survives between sessions, but realistic development use-cases do not give rise to conflicts between persisted data for different apps.

This in turn means that the `WebStorage` functionality needs to be initialised on a per-instance basis and cannot simply be static as before.

### DIFF
--- a/common/lib/transport/connectionmanager.js
+++ b/common/lib/transport/connectionmanager.js
@@ -1,6 +1,4 @@
 var ConnectionManager = (function() {
-	var haveWebStorage = !!(typeof(WebStorage) !== 'undefined' && WebStorage.get);
-	var haveSessionStorage = !!(typeof(WebStorage) !== 'undefined' && WebStorage.getSession);
 	var actions = ProtocolMessage.Action;
 	var PendingMessage = Protocol.PendingMessage;
 	var noop = function() {};
@@ -9,15 +7,6 @@ var ConnectionManager = (function() {
 	var transportPreferenceName = 'ably-transport-preference';
 
 	var sessionRecoveryName = 'ably-connection-recovery';
-	function getSessionRecoverData() {
-		return haveSessionStorage && WebStorage.getSession(sessionRecoveryName);
-	}
-	function setSessionRecoverData(value) {
-		return haveSessionStorage && WebStorage.setSession(sessionRecoveryName, value);
-	}
-	function clearSessionRecoverData() {
-		return haveSessionStorage && WebStorage.removeSession(sessionRecoveryName);
-	}
 
 	function betterTransportThan(a, b) {
 		return Utils.arrIndexOf(transportPreferenceOrder, a.shortName) >
@@ -169,6 +158,31 @@ var ConnectionManager = (function() {
 	}
 	Utils.inherits(ConnectionManager, EventEmitter);
 
+	/*******************************
+	 * session preference management
+	 *******************************/
+
+	ConnectionManager.prototype.getSessionRecoverData = function() {
+		var storage = this.realtime.storage;
+		if(storage.supportsSession) {
+			return storage.getSession(sessionRecoveryName);
+		}
+	};
+
+	ConnectionManager.prototype.setSessionRecoverData = function(value) {
+		var storage = this.realtime.storage;
+		if(storage.supportsSession) {
+			storage.setSession(sessionRecoveryName, value);
+		}
+	};
+
+	ConnectionManager.prototype.clearSessionRecoverData = function() {
+		var storage = this.realtime.storage;
+		if(storage.supportsSession) {
+			storage.removeSession(sessionRecoveryName);
+		}
+	};
+
 	/*********************
 	 * transport management
 	 *********************/
@@ -190,7 +204,7 @@ var ConnectionManager = (function() {
 			}
 
 			var recoverFn = self.options.recover,
-				lastSessionData = getSessionRecoverData();
+				lastSessionData = self.getSessionRecoverData();
 			if(lastSessionData && typeof(recoverFn) === 'function') {
 				Logger.logAction(Logger.LOG_MINOR, 'ConnectionManager.getTransportParams()', 'Calling clientOptions-provided recover function with last session data');
 				recoverFn(lastSessionData, function(shouldRecover) {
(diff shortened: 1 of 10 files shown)


nodejs/lib/util/webstorage.js
```diff
@@ -0,0 +1,109 @@
...
+	function valid(item) {
+		var expires = item.expires;
+		return !expires || (Date.now() < expires);
+	}
+
+	function WebStorage(client) {
+		this.client = client;
+		this.storagePath = null;
+		this.localStorage = null;
+		this.sessionStorage = new Map();
```
SimonWoolf: Note that `Map()` is node 4 or later. But looking now, ably-js hasn't worked on node pre-4 for a while, and no-one seems to have noticed. So I think this is fine, we should just update the readme to note that we support nodejs 4 or later, and remove 0.12 from .travis.yml.

nodejs/lib/util/webstorage.js
```diff
@@ -0,0 +1,109 @@
...
+
+	WebStorage.prototype.initPath = function() {
+		var storagePath = this.storagePath;
+		if(!storagePath) {
+			var options = this.client.options, appId;
+			if(options.key) {
+				appId = options.key.split('.')[0];
+			} else {
+				var authOptions = this.client.auth.authOptions,
+					token = authOptions.tokenDetails.token;
```
SimonWoolf: > `authOptions.tokenDetails.token`

This will crash if the client is using authurl/authcallback, as authOptions.tokenDetails won't be available yet.

You could delay Storage init for node until the first time the client successfully authorizes if we're quite sure that that the storage api won't be used by push until after that's happened. Pretty nasty though.

I haven't been following the push conversation, but what's the issue with having a device's device id used across multiple apps (i.e. if we used a fixed path here)? After all, it is ultimately the same device.

nodejs/lib/util/webstorage.js
```diff
@@ -0,0 +1,109 @@
...
+		}
+		return storagePath;
+	};
+
+	WebStorage.prototype.init = function() {
+		var localStorage = this.localStorage;
+		if(!localStorage) {
+			var storagePath = this.initPath();
+			mkdirs(storagePath);
+			console.log('**** storage path: ' + storagePath);
```
SimonWoolf: This should be a `Logger.logAction(Logger.LOG_MINOR, ...)` with a more informative message

nodejs/lib/util/webstorage.js
```diff
@@ -0,0 +1,109 @@
...
+		var storagePath = this.storagePath;
+		if(!storagePath) {
+			var options = this.client.options, appId;
+			if(options.key) {
+				appId = options.key.split('.')[0];
+			} else {
+				var authOptions = this.client.auth.authOptions,
+					token = authOptions.tokenDetails.token;
+				if(!token) {
+					throw new Error('Unable to initialise WebStorage; no key or token available');
```
SimonWoolf: Aside from comment above about getting the app ID, if it can't initialize storage for some reason, it should just set `supportsLocal` and `supportsSession` to `false` and return, not throw an exception and abort the whole client init.

nodejs/lib/util/webstorage.js
```diff
@@ -0,0 +1,109 @@
...
+				}
+				appId = token.split('.')[0];
+			}
+			storagePath = this.storagePath = path.resolve(os.homedir(), '.ably-js/localStorage', appId);
```
SimonWoolf: we should probably try a write & read, and if we get an error (e.g. because of filesystem permissions), again set `supportsLocal` and `supportsSession` to false

SimonWoolf (changes requested):

nodejs/lib/util/webstorage.js
paddybyers: Yes, I think we can be fairly aggressive in retiring support for older node versions (except where there are dependencies from things like Lambda where end-users have no control).

nodejs/lib/util/webstorage.js
paddybyers: > You could delay Storage init for node until the first time the client successfully authorizes if we're quite sure that that the storage api won't be used by push until after that's happened. Pretty nasty though.

But `init()` is only called when `set()` is called, so whta's here will work for connection persistence. We could have `init()` force authorisation if it's not already happened, or we can have the push implementation force authorisation ahead of attempting to set any persisted values.

> I haven't been following the push conversation, but what's the issue with having a device's device id used across multiple apps (i.e. if we used a fixed path here)? After all, it is ultimately the same device.

A `deviceId` belongs to an app, and the activation state, which is persisted, depends on the realtime state, which is persisted for the `appId/deviceId` combination.

Aside from that, we do still want to be able to run tests by creating a new app and know that the state of client and server are both clean at that point.

nodejs/lib/util/webstorage.js
paddybyers: > or better, keep the exception here, but in init wrap initPath() and the JSONStorage init in a try/catch, and set the supports vars to false in the catch

Yes

nodejs/lib/util/webstorage.js
paddybyers: yes, good point

nodejs/lib/util/webstorage.js
paddybyers: oops

nodejs/lib/util/webstorage.js
SimonWoolf: > But init() is only called when set() is called,

No it isn't? `this.storage = new WebStorage(this);` is called from the Rest client constructor

paddybyers: Yes, but the `WebStorage` constructor doesn't call `init()`.

SimonWoolf: Fair enough. I saw it was crashing the tests and assumed that it was because it was being initialized too early, but looking closer, it's actually it was because it's looking at `auth.authOptions.tokenDetails`; that should be `auth.tokenDetails` (the token is only an authOption if the library was initialized or authorize()d with a token literal, as opposed to having fetched one with authUrl/authCallback).

common/lib/transport/connectionmanager.js
```diff
@@ -1483,22 +1497,24 @@ var ConnectionManager = (function() {
...
 
 	ConnectionManager.prototype.persistTransportPreference = function(transport) {
 		if(Utils.arrIn(Defaults.upgradeTransports, transport.shortName)) {
 			this.transportPreference = transport.shortName;
-			if(haveWebStorage) {
-				WebStorage.set(transportPreferenceName, transport.shortName);
+			var storage = this.realtime.storage;
+			if(storage.supportsLocal) {
+				storage.set(transportPreferenceName, transport.shortName);
 			}
```
SimonWoolf: The result of this is that node clients persist their transport preference unnecessarily (they default to using websockets). I'd suggest either adding a conditional to persistTransportPreference for `connectionManager.transports.length > 1` (or maybe just a `persistTransportPreference` bool in the Platform object set to false for node)

PR #445, opened 2018-01-02
Fix sync connection serial in the event of a failed upgrade

SimonWoolf:
And rename various connectionSerial variables for clarity

(context from slack: "when it was first written, that code was correct, because at the time, we activated the new transport before syncing, so the first newConnectionSerial was passed to activeTransport and served its purpose. Later on in an unrelated change (https://github.com/ably/realtime/issues/522), to address a race condition, we switched the order of those and started syncing first, but I clearly failed to notice that the logic for when the upgrade failed needed to be updated too")

### DIFF
--- a/common/lib/transport/connectionmanager.js
+++ b/common/lib/transport/connectionmanager.js
@@ -332,7 +332,7 @@ var ConnectionManager = (function() {
 	 * @param transport, the transport instance
 	 * @param connectionKey
 	 */
-	ConnectionManager.prototype.scheduleTransportActivation = function(error, transport, connectionKey, connectionSerial, connectionId, connectionDetails) {
+	ConnectionManager.prototype.scheduleTransportActivation = function(error, transport, connectionKey, preUpgradeConnectionSerial, connectionId, connectionDetails) {
 		var self = this,
 			currentTransport = this.activeProtocol && this.activeProtocol.getTransport(),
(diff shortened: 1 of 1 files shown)


common/lib/transport/connectionmanager.js
```diff
@@ -387,14 +387,14 @@ var ConnectionManager = (function() {
 			* be a sync with the new connectionSerial (which will be -1). (And it
 			* needs to be set in the library, which is done by activateTransport). */
 			var connectionReset = connectionId !== self.connectionId,
-				newConnectionSerial = connectionReset ? connectionSerial : self.connectionSerial;
+				syncConnectionSerial = connectionReset ? preUpgradeConnectionSerial : self.connectionSerial;
```
paddybyers: `scheduleTransportActivation()` is passed the `connectionSerial` from the `CONNECTED` message when the upgrade transport connects. So how is this the `preUpgradeConnectionSerial` ?

SimonWoolf: In the sense that the sync as the atomic moment of the actual 'upgrade', so preUpgradeConnectionSerial is the serial of the upgrade transport before that (likely -1). Can change to preSyncConnectionSerial if you think that's clearer

SimonWoolf:
Merged as 7a63754

PR #451, opened 2018-01-22
Add example for browser testing

funkyboy:
@paddybyers @SimonWoolf ready for review.


SimonWoolf:
:-1: The client lib readmes are public and optimised for explaining to external contributors how they can help, contribute, get support etc. Hence why the tests run against sandbox by default, which is publicly accessible, so anyone can run them. This addition would be quite misleading to an external contributer, who may not realise that you're using "developing/testing locally" to mean running against a local `realtime`, and not just testing ably-js locally (as opposed to on ci or something), and would not be able to run a local `realtime` even if they wanted to.

funkyboy:
Fair enough. One more URL to bookmark for me :)

PR #452, opened 2018-01-23
Run Karma tests with BrowserStack and reconfigure Travis

### LINKED ISSUE #449: Rethink browser compatibility test approach
Our current automated cross-browser tests have been unreliable [for some time](https://ably-real-time.slack.com/archives/C030APSH3/p1454430886001336):

> Had a look at some runs of browsers-all. Failures can be divided into:

> - Random http timeouts. Saucelabs internet connection seems to be occasionally flakey, and any of the VMs (inc firefox and chrome) sometimes get random timeouts with comet requests.

> - Systematic saucelabs-only internet failures. Eg:
>  - for IE9 and 9, the first attempt to set up an app (which happens to be the realtime auth tests) pretty much always times out, resulting in those tests all failing. Subsequent attempts succeed.
> - IE10 is almost the opposite: after 50 or so successful tests, starting usually somewhere the early channel tests the VM seems to stop being able to connect to ably in any way. All subsequent tests timeout.

> (I've fired up local IE 9 and 10 virtual machines and double-checked that the test suite passes in both).

> I spent some time trying to debug the IE10 one back when I did IE10 support, without success.

> Not really sure what to suggest tbh. Reliable cross-browser testing is something we're going to have to have eventually

-------------

> I've just tried with testingbot, and got the same results, for both IE10 and IE8/9.
> - The IE10 problem (`network error 0x2ee4`) appears to be a bug in IE10 triggered by saucelabs/testingbot's use of a network proxy, which is why I'm not seeing it locally. Was reported to MS, closed as "fixed in latest version" (ie IE11) , which is unhelpful. (https://connect.microsoft.com/IE/feedback/details/852785/ie10-network-error-0x2ee4-https-proxy-ajax).
> - The IE8/9 problem seems to only happen the first time the page is loaded after the VM is created. On subsequent page refeshes (even if cache cleared, 'always refresh from server' ticked etc.) it works fine; which is why I couldn't recreate it in my local VM.

Matt and I spent some time trying to fix this a couple years ago (e.g. https://github.com/ably/ably-js/pull/229 ) without success. I fell back to using virtualbox running locally to test Windows browsers. This isn't sustainable; we need reliable automated cross-browser CI.

### LINKED ISSUE #460: Pick on CI to run unit and browser tests on
While add BrowserStack tests [here](https://github.com/ably/ably-js/pull/452) I am also taking Travis for a spin.

One drawback is that the log not always fits a web page, like [here](https://travis-ci.org/ably/ably-js/builds/337921934) and you are forced to read the [raw log](https://api.travis-ci.org/v3/job/337921936/log.txt) which isn't a lot readable.

### LINKED ISSUE #465: Add BrowserStack test on Android

funkyboy:
Adding BrowserStack for karma tests.
Fix #449 
Fix #460 
Fix #465 

- [x] add script for CI
- [x] add more combinations of os/browsers
- [x] inspect with IE hits timeouts and takes 4x than Firefox tests
- [x] remove local Firefox
- [x] add all browsers to CI script (in batches)
- [x] add Safari for Mac
- [x] add one Android device
- [x] Set up Travis to run 2 jobs (see https://github.com/ably/ably-js/issues/460#issuecomment-364444888)
- [x] Re-add local Firefox


funkyboy:
IE10 issues have been fixed here: https://github.com/ably/ably-js/pull/453

SimonWoolf:
> IE10 issues have been fixed here

To be clear, those appear to have been regressions introduced since a couple years ago (genuine bugs affecting IE10), rather than the IE10 problems we had with saucelabs/testingbot that couldn't be reproduced in local vms (and which caused every test after some initial number to time out with connectivity issues, rather than only the crypto tests). Looking forward to seeing how we fare now on saucelabs or browserstack.

test/bin/ci-travis-all.sh
```diff
@@ -1,3 +1,3 @@
 grunt $B0
-grunt test:karma --browsers bs_firefox_sierra,bs_chrome_sierra,bs_ie11_win81,bs_ie10_win81
+grunt test:karma --browsers $B1
 grunt test:karma --browsers bs_ie9_win7,bs_ie16_win10,bs_safari_10_3,bs_safari_11
```
funkyboy: This _would_ be nice but as soon as `$B0` (nodeunit) tests end, Travis starts running the script on line 3. But given that line 2 has not finished yet, BrowserStack will hang, because we are allowed at most 5 tests in parallel. :(

funkyboy: When running tests in a serial fashion, we hit travis (.org) limits of 50m per build. Sequence tested:
- nodeunit 
- bs_firefox_sierra,bs_chrome_sierra,bs_ie11_win81,bs_ie10_win81
- bs_ie9_win7,bs_ie16_win10,bs_safari_10_3,bs_safari_11

funkyboy:
We hit Travis 50m timeout also when removing iOS10 tests, thus running tests on 8 browsers (plus nodeunit).

test/bin/ci-travis-all.sh
```diff
@@ -0,0 +1,3 @@
+grunt test:nodeunit
```
SimonWoolf: No point duplicating the nodeunit tests in the browserstack run. (If you're only doing that get the concat task, you can just do `grunt concat` directly. have a look in the gruntfile to see what tasks do what)

karma.conf.js
```diff
@@ -3,86 +3,89 @@
...
-      version: '5.0'
+    bs_android_6: {
+      base: 'BrowserStack',
+      browser: 'android',
+      os: 'android',
+      os_version: '6.0',
+      device: 'Google Nexus 6',
+      real_mobile: true
     }
   };
```
SimonWoolf: No IE8? (I'm not against dropping IE8 if usage numbers are finally low enough, but that should be a deliberate decision, as long as we claim we support it we should be testing on it)

karma.conf.js
```diff
@@ -3,86 +3,89 @@
...
 
   config.set({
 
+    browserStack: {
+      username: process.env.BROWSERSTACK_USERNAME,
+      accessKey: process.env.BROWSERSTACK_ACCESSKEY,
```
SimonWoolf: Would be convenient if these were in the private repo in a script in `sourcable`

karma.conf.js
```diff
@@ -3,86 +3,89 @@
...
+      browser_version: '16.0',
+      os: 'Windows',
+      os_version: '10'
     },
-    sl_ie_10: {
-      base: 'SauceLabs',
-      browserName: 'internet explorer',
-      platform: 'Windows 7',
-      version: '10'
+    bs_safari_10_3: {
```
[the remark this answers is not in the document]
funkyboy: @SimonWoolf My bad. This is not referenced in the the travis file. Will substitute it with IE8.

karma.conf.js
```diff
@@ -166,24 +169,14 @@ module.exports = function(config) {
 
     // level of logging
     // possible values: config.LOG_DISABLE || config.LOG_ERROR || config.LOG_WARN || config.LOG_INFO || config.LOG_DEBUG
-    logLevel: config.LOG_INFO,
+    logLevel: process.env.ABLY_LOG_LEVEL,
```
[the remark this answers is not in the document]
funkyboy: @SimonWoolf That's why it doesn't work :) 
Goal was to keep the Travis log shorter. Will set to `WARN` to see what happens.

test/bin/ci-travis-all.sh
funkyboy: @SimonWoolf  True, but browser tests failed if not preceded by nodeunit tests because of https://github.com/ably/ably-js/issues/462

karma.conf.js
funkyboy: @SimonWoolf Fixing. See above.

karma.conf.js
funkyboy: @SimonWoolf Added

karma.conf.js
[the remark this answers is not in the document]
SimonWoolf: karma INFO-level logs add around 10 lines to the beginning of the test; the remaining ~40k lines are mostly from ably-js Logger. I'd focus on the 40k rather than the 10 lines to 5 ☺. Check out the `envPreprocessor` config line for exposing external env vars to karma tests

test/bin/ci-travis-all.sh
SimonWoolf: No, `grunt concat` doesn't run the minifier, so is unaffected by #462.

karma.conf.js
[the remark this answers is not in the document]
funkyboy: Oh so [this](https://github.com/ably/ably-js/blob/add-browserstack-testing/.travis.yml#L11) doesn't work because `ABLY_LOG_LEVEL` doesn't appear in the `envPreprocessor` "enumeration"?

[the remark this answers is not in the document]
funkyboy: Answering myself: yes :)

funkyboy:
@simonwoolf one last check before merging?

SimonWoolf:
:+1: . Git history's a bit messed up though, why is there a commit merging master into this branch? Get rid of that and rebase this on master, and regen package-lock.json. Let me know when done and I'll check it and merge

funkyboy:
@SimonWoolf that was the result of of using GH web UI to resolve a conflict :(

funkyboy:
@SimonWoolf here you go

SimonWoolf:
Thanks @funkyboy

PR #453, opened 2018-01-25
Fix IE10 crypto

funkyboy:
In particular this makes the following tests pass also on IE10:

```
2.generateRandomKey0 (2, 0, 2)
3.generateRandomKey1 (2, 0, 2)
4.getDefaultParams_wordArray_key (2, 0, 2)
5.getDefaultParams_base64_key (2, 0, 2)
6.getDefaultParams_check_keylength (2, 0, 2)
7.getDefaultParams_preserves_custom_algorithms (2, 0, 2)
15.single_send_128_with_web_socket_binary_transport (2, 0, 2)
16.single_send_128_with_web_socket_text_transport (2, 0, 2)
17.single_send_128_with_xhr_streaming_binary_transport (2, 0, 2)
18.single_send_128_with_xhr_streaming_text_transport (2, 0, 2)
19.single_send_128_with_xhr_polling_binary_transport (2, 0, 2)
20.single_send_128_with_xhr_polling_text_transport (2, 0, 2)
21.single_send_128_with_binary_transport (2, 0, 2)
22.single_send_128_with_text_transport (2, 0, 2)
23.single_send_256_with_web_socket_binary_transport (2, 0, 2)
24.single_send_256_with_web_socket_text_transport (2, 0, 2)
25.single_send_256_with_xhr_streaming_binary_transport (2, 0, 2)
26.single_send_256_with_xhr_streaming_text_transport (2, 0, 2)
27.single_send_256_with_xhr_polling_binary_transport (2, 0, 2)
28.single_send_256_with_xhr_polling_text_transport (2, 0, 2)
29.single_send_256_with_binary_transport (2, 0, 2)
30.single_send_256_with_text_transport (2, 0, 2)
31.multiple_send_binary_2_200 (2, 0, 2)
32.multiple_send_text_2_200 (2, 0, 2)
33.multiple_send_binary_20_100 (2, 0, 2)
```

### DIFF
--- a/browser/fragments/platform-browser.js
+++ b/browser/fragments/platform-browser.js
@@ -21,10 +21,12 @@ var Platform = {
 	addEventListener: window.addEventListener,
 	inspect: JSON.stringify,
 	getRandomValues: (function(crypto) {
+		if (crypto === undefined) {
+			return undefined;
+		}
 		return function(arr, callback) {
 			crypto.getRandomValues(arr);
 			callback(null);
 		};
 	})(window.crypto || window.msCrypto) // mscrypto for IE11
 };
-
--- a/browser/lib/util/crypto.js
+++ b/browser/lib/util/crypto.js
@@ -14,7 +14,7 @@ var Crypto = (function() {
 	 * @param callback
 	 */
 	var generateRandom;
-	if(typeof Uint32Array !== 'undefined' && Platform.getRandomValues && window.crypto) {
+	if(typeof Uint32Array !== 'undefined' && Platform.getRandomValues) {
 		var blockRandomArray = new Uint32Array(DEFAULT_BLOCKLENGTH_WORDS);
 		generateRandom = function(bytes, callback) {
 			var words = bytes / 4, nativeArray = (words == DEFAULT_BLOCKLENGTH_WORDS) ? blockRandomArray : new Uint32Array(words);
--- a/browser/lib/util/crypto.js
+++ b/browser/lib/util/crypto.js
@@ -14,7 +14,7 @@ var Crypto = (function() {
 	 * @param callback
 	 */
 	var generateRandom;
-	if(typeof Uint32Array !== 'undefined' && Platform.getRandomValues) {
+	if(typeof Uint32Array !== 'undefined' && Platform.getRandomValues && window.crypto) {
 		var blockRandomArray = new Uint32Array(DEFAULT_BLOCKLENGTH_WORDS);
 		generateRandom = function(bytes, callback) {
 			var words = bytes / 4, nativeArray = (words == DEFAULT_BLOCKLENGTH_WORDS) ? blockRandomArray : new Uint32Array(words);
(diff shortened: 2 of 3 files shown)


funkyboy:
screen shot 2018-01-25 at 09 48 40

browser/lib/util/crypto.js
```diff
@@ -14,7 +14,7 @@ var Crypto = (function() {
 	 * @param callback
 	 */
 	var generateRandom;
-	if(typeof Uint32Array !== 'undefined' && Platform.getRandomValues) {
+	if(typeof Uint32Array !== 'undefined' && Platform.getRandomValues && window.crypto) {
```
SimonWoolf: This is the sort of check that needs to be encapsulated in the Platform object. As it stands, the effect of this change will be to force weak random numbers for nativescript, react-native, and IE11 (which uses `window.mscrypto` rather than `window.crypto`), all of which do support strong random numbers via `Platform.getRandomValues`.

To put it another way, the code already checks that `Platform.getRandomValues` is defined; the bug is that it shouldn't be defined if in a browser context where you have neither `window.crypto` nor `window.mscrypto`.

spec/common/modules/testapp_manager.js
```diff
@@ -76,7 +76,7 @@ define(['globals', 'browser-base64', 'ably'], function(ablyGlobals, base64, ably
 				}).then(function (results) {
 					callback(null, results.content.toString());
 				})
-				.catch(function(err) {
+				['catch'](function(err) {
```
SimonWoolf: If the parsing issue is just that they don't like the newline before the method call with the dot syntax, would be clearer to just move it up to the previous line, avoid having to remember the rules of ASI

funkyboy: I tried using 

```
}).then(function (results) {
					callback(null, results.content.toString());
				}).catch(function(err) {
					callback(err);
				});
```
but no luck :(

funkyboy: @SimonWoolf Is this preferable?

```js
}).then(function (results) {
	callback(null, results.content.toString());
})['catch'](function(err) {
	callback(err);
});
```

SimonWoolf: ok, so they're confused by catch being a keyword. Yup, I'd prefer that latter one to the original, as then it makes it clearer that it's a property access rather than an array literal

browser/fragments/platform-browser.js
```diff
@@ -21,10 +21,12 @@ var Platform = {
 	addEventListener: window.addEventListener,
 	inspect: JSON.stringify,
 	getRandomValues: (function(crypto) {
+		if (crypto === undefined) {
```
funkyboy: @SimonWoolf did you mean something like this?

SimonWoolf: Sure, that should work. can you confirm it works as expected in IE10 etc.?

funkyboy: @SimonWoolf Yup. 
Tests keep on passing on Firefox and Chrome.
And with this also on IE10.
Attaching some screenshots for reference.

ie

[chrome.pdf](https://github.com/ably/ably-js/files/1666935/chrome.pdf)

[ff.pdf](https://github.com/ably/ably-js/files/1666936/ff.pdf)

SimonWoolf:
Thanks @funkyboy

PR #458, opened 2018-02-01
Remove old node version from Travis

funkyboy:


funkyboy:
@SimonWoolf Wondering: should we add tests also on newer versions of Node?

SimonWoolf:
> Wondering: should we add tests also on newer versions of Node?

Maybe. We should have a think about what we want the purpose of the travis ci builds to be going forward, given that we use jenkins as the main ci -- like, it seems odd to use travis for testing node versions and jenkins for testing browser versions, it just means you have two places you need to check instead of one. What do you think @funkyboy ?

funkyboy:
@SimonWoolf made an [issue](https://github.com/ably/ably-js/issues/460) about this. In the meantime, If you agree, I'd merge this PR. Let me know.

SimonWoolf:
:+1: (subsumed by https://github.com/ably/ably-js/pull/452/files)

PR #474, opened 2018-02-21
Add more versions of nodejs to Travis tests

### LINKED ISSUE #473: Run node unit tests on multiple versions on Node.js
- [x] node 4.8.7
- [x] node 6.13
- [x] node 8.9.4
- [x] node 9.5.0
- [x] Update `engines` field

funkyboy:
Fix #473


paddybyers:
Looking at https://nodejs.org/en/download/releases/:

6.x became LTS in 2016-10-18, which is 18 months ago, so I do think it is reasonable for us to continue to support 4.x. There continue to be 4.x releases from node (the most recent is nodejs 4.8.7 on 2017-12-07).

However, I don't see why we don't test against the most recent supported 4.x release (4.8.7 in this case). If someone is using 4.x still in a production system, they ought to be at or near the latest release in the 4.x series, not back on 4.2.

Also I think you can simply specify `6` to get the latest 6.x release but you'd need to try that out.

paddybyers:
> Ok?

SGTM

funkyboy:
I am gonna wait for @SimonWoolf before merging.
I am sure that if there's anything wrong or to be improved he can spot it :)

SimonWoolf:
We now have -short which does nodeunit and karma/firefox58, and -full which does the saucelabs tests. With this pr we run the -short tests on 4 different node versions, but that also runs the ff58 tests four times. So let's split those up so the nodeunit tests are run four times, one for each node version, and the firefox 58 tests are only run once.

funkyboy:
@SimonWoolf can Firefox58 tests be run right after one (and only one) of the nodeunit tests? E.g.

Task1: node4+FF58
Task2: node6
Task3: node8
Task4: nodelast
Task5: BrowserStack tests

SimonWoolf:
> can Firefox58 tests be run right after one (and only one) of the nodeunit tests? 

I guess it could, but I think it makes more sense to have it be its own task, so we can see at a glance from the result page whether the browser tests passed independently of the node tests

funkyboy:
@SimonWoolf re-ready for review.

SimonWoolf:
:+1: thanks

Issue #441, opened 2017-12-12
Add linter to enforce es version 


SimonWoolf:
See https://github.com/ably/ably-js/pull/435#issuecomment-350217410

PR #475, opened 2018-02-21
WIP: Update typescript types

funkyboy:
This is meant to support the TS sample [here](https://github.com/funkyboy/ts)


ably.d.ts
```diff
@@ -4,478 +4,480 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 declare namespace ablyLib {
-  namespace ChannelState {
```
mattheworiordan: I really dislike that you have changed indentation and made changes in code in the same commit.  As a result, I have no way to review what changed here as everything other than `declare namespace ablyLib` has now changed.  Please have separate commits to change whitespace / indentation from other commits.

common/lib/client/clientoptions.js
```diff
@@ -0,0 +1,9 @@
+var ClientOptions = (function() {
+
+  /* public constructor */
+	function ClientOptions() {
```
mattheworiordan: What's with the indentation here?  Don't you have a linter running?

spec/realtime/auth.test.js
```diff
@@ -71,8 +72,9 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
 			}
 
 			var authPath = echoServer + "/?type=json&body=" + encodeURIComponent(JSON.stringify(tokenDetails));
-
-			realtime = helper.AblyRealtime({ authUrl: authPath });
+			var clientOptions = new ClientOptions();
+			clientOptions.authUrl = authPath;
```
mattheworiordan: If you have introduced a new `ClientOptions` class into JS, should we not have a set of tests to ensure that the class has the default values?  

Also, if we are taking this approach now, are we not creating inconsistencies unless we provide constructors now for all the types, see https://docs.ably.io/realtime/types/?

ably.d.ts
mattheworiordan: P.S. I can review with `?w=1` to hide whitespace, but then I cannot add comments to commit lines in that same view

common/lib/client/clientoptions.js
funkyboy: @mattheworiordan Not yet. Related issue here: https://github.com/ably/ably-js/issues/441

ably.d.ts
funkyboy: @mattheworiordan This is all quite experimental. I don't know why I asked for a review. It's definitely not ready for review.

mattheworiordan:
Having reconsidered what is being done here, I am in fact a bit concerned by the flattened approach that we are taking for JS.

For example, a `Message` object is accessible currently as:

* `Ably::Models::Message` in Ruby
* `io.ably.lib.types.Message` in Java
* `ARTMessage` in iOS, however I believe this was done because namespaces are not supported in Objective-C so we were forced to this
* `IO.Ably.Message` in C#
* `Ably\Models\Message` in PHP
* `Message` apparently in Python, but I think the docs are perhaps incorrect.

Why are we flattening the types in TypeScript which may result in conflicts down the line for naming.  Surely we could instead introduce something like:

```typescript
import { Realtime, Rest, Types } from `ably`;
let options = new Types.ClientOptions({ foo: bar });
let client = new Realtime(options);
```

This is a lot cleaner and does not pollute the root namespace with an unknown number of types that will grow over time.

@funkyboy why do you want these types exposed in the root unlike almost all other client SDKs we have?  @paddybyers should we consider something in the spec to this effect?

funkyboy:
@mattheworiordan @paddybyers I opted for `Message`-as-root-object purely because it's a personal preference. I didn't consider the possibility that we want to have homogeneous namespaces across SDKs. If that's the case, let's talk about it.

Nevertheless this `... = new Types.ClientOptions(...)` is not idiomatic TypeScript, AFAICT.

mattheworiordan:
> Nevertheless this ... = new Types.ClientOptions(...) is not idiomatic TypeScript, AFAICT.

Why is that not idiomatic?   [TypeScript provides namespaces](https://www.typescriptlang.org/docs/handbook/namespaces-and-modules.html) for this very reason AFAICT

funkyboy:
Well the page you linked labels this

`let t = new shapes.Shapes.Triangle();` 

as "confusing and annoying for consumers of your module" :)

And shows this

`let t = new shapes.Triangle();`

as "revised example". 

If I were _exclusively_ using Ably TS I'd prefer `let t = new Triangle();`. 
If I were using more than one Ably SDK I'd probably appreciate a homogeneous organization of types across SDKs. 

My concern is that such an effort might probably stretch the way some programming languages are used.

mattheworiordan:
> `let t = new shapes.Shapes.Triangle();`
> as "confusing and annoying for consumers of your module" :)

No, you have misread why this annoying. This is because the module contains Shapes, so having a Shapes namespace in a Shapes module is unnecessary.

> `let t = new shapes.Triangle();`
> as "revised example".

How is that different from new `let clientOptions = new Types.ClientOptions();`?

> If I were exclusively using Ably TS I'd prefer let t = new Triangle();.

Sure, but that seriously limits our future ability to introduce types and classes with the same name in future.  Whilst we don't have this issue now, namespaces are designed to help with this.  So if we go with a flat solution now for JS, we're seriously restricting what we can do in the future, which means potentially unnecessary breaking changes for customers.

Additionally, if you want to import all Types and have `let clientOptions = new ClientOptions()` is it no possible in TyepScript to do something like :

```typescript
import { Realtime, Rest } from `ably`;
import { Types.* } from `ably`;

let clientOptions = new ClientOptions();
let realtime = new Realtime(clientOptions);
```

funkyboy:
@mattheworiordan 
> Additionally, if you want to import all Types and have let clientOptions = new ClientOptions() is it no possible in TyepScript to do something like

If that is a question, that's _exactly_ what I am trying to find out. So far: either I am not expert enough at TS (likely) or it's not possible :)

We use `Type`s throughout the spec, but wouldn't it be more appropriate to call it `Model`(s) in code (and packages and namespaces)?

funkyboy:
Restarting with a clean slate

browser/fragments/ably-epilogue.js
```diff
@@ -1,4 +1,8 @@
 	if(typeof Realtime !== 'undefined') {
+		var Types = {
```
mattheworiordan: If you're not using the `Types` var, then why not:

```javascript
var typesConstructor = function(obj) { return (obj); };
Ably.Types = {
  ClientOptions: typesConstructor,
  SomeOtherObject: typesConstructor
};
```

browser/fragments/ably-commonjs-epilogue.js
```diff
@@ -1,3 +1,8 @@
+var Types = {
+  ClientOptions: function(obj) { return (obj); }
```
mattheworiordan: You're now duplicating code between this file and ably-epilogue?  Can we avoid that please?

browser/static/ably-commonjs.js
```diff
@@ -11442,6 +11442,11 @@ var JSONPTransport = (function() {
 	return JSONPTransport;
 })();
 
+var Types = {
+  ClientOptions: function(obj) { return (obj); }
```
mattheworiordan: Can you please not commit static compiled files until this is ready for release?  Looking through this PR, only three files have actually changed, yet this PR has ten changed files, which I have to look through to make sure I'm not missing anything.

ably.d.ts
```diff
@@ -465,7 +465,7 @@ export declare class Rest {
 }
 
 export declare class Realtime {
-  constructor(options: ablyLib.ClientOptions | string);
+  constructor(options: Types.ClientOptions | string);
   static Crypto: ablyLib.Crypto;
   static Message: ablyLib.MessageStatic;
```
mattheworiordan: What are you going to do about this type?

ably.d.ts
```diff
@@ -479,3 +479,29 @@ export declare class Realtime {
...
+
+    /**
+     * When true, time is queried from the Ably servers. Example help text.
+     */
+    queryTime?: boolean;
+
+    constructor(options?: ClientOptions);
+  }
+
+  export class Message {
```
mattheworiordan: I've just realised you've duplicated the Message class, see https://github.com/ably/ably-js/blob/13566592e0c10730f79d797f51e4aaab5deedf91/ably.d.ts#L371-L383

browser/static/ably-commonjs.js
funkyboy: I committed those just to test the results of the concatenation. They won't be there when this is ready for review.

ably.d.ts
funkyboy: Yes, that's necessary if we do not want breaking changes. My plan was not to touch anything in the `ablyLib` namespace, if possible. That means we will have duplicate classes like:

- `ablyLib.Message`
- `Types.Message`

Happy to amend this, if we don't want it.

mattheworiordan:
See updated https://github.com/mattheworiordan/ts-example-with-definitions/commit/a136ccf7fbff25fdcd4a9bf34d6083e8a4d1167f

and:

VS code example

ably.d.ts
mattheworiordan: I really don't think we want that.  Is there any evidence that anyone is using explicitly defined types such as `ablyLib.Message` anywhere?  I appreciate this change may break that, but I think that a) the underlying JS will still work so it's not a breaking change, just a breaking typ check change , b) it's going to be very rare that anyone references those types, and if they do, they will simply get a type breaking change which is very easily resolved.

If everyone disagrees, why not set up an alias instead of duplicating types?  I believe:

```typescript
/* within ablyLib */
type Message = Types.Message;
```

would also work

funkyboy:
Superseded by #492.
Created a separate issue for TS project https://github.com/ably/ably-js/issues/500

PR #481, opened 2018-03-02
Add supported platform section to README

funkyboy:


README.md
```diff
@@ -2,9 +2,34 @@
...
+|          | Safari iOS 11      |
+|          | Chrome on Android 6|
+ 
+We maintain compatibility and explicitly support these platform versions, including performing CI testing on all library revisions.
+
+We do not explicitly maintain compatibility with older platform versions; we no longer perform CI testing on iOS7 as of version 1.0.13 (released on February 2nd 2018). Any known incompatibilities with older versions can be found [here](https://github.com/ably/ably-js/issues?q=is%3Aissue+is%3Aopen+label%3A%22compatibility%22).
+
+**Note**: the ably-js browser library contains fallback mechanisms in order to be able to support older browsers; specifically it supports comet-based connections for browsers that do not support websockets, and this includes JSONP for browsers that do not support cross-origin XHR.
+Each of these fallback transport mechanisms is supported and tested on all of the listed target browsers; even when those browsers do not themselves require those fallbacks.
+These mean that the library should be compatible with nearly any recent browser, on the majority of platforms.
```
mattheworiordan: > compatible with nearly any recent browser

Do we not mean nearly every browser being used?  Recent makes it sound like we're only supporting ones released in the last few months / year at best

paddybyers: Yes, just remove "recent"

README.md
```diff
@@ -2,9 +2,34 @@
...
+ This SDK is compatible with the following platforms:
+ 
+| Node.js  | Browsers           | 
+|----------|--------------------|
+|   4      | IE 8+              |
+|   6      | Chrome 64+         |
+|   8      | Firefox 58+        |
+|   9      | Safari macOS 11    |
+|          | Safari iOS 11      |
+|          | Chrome on Android 6|
```
SimonWoolf: This seems overly restrictive -- I think we're underselling ourselves.

E.g. it makes it sound like we don't support firefox <=57, which may be rather off-putting for someone on, say, the current extended-support release (52). Or safari on iOS 10, or Chrome on Android 8, or the default browser on a Samsung phone, etc. etc. etc.

In reality we we do support all of those, in the sense that if someone found a bug in one of them we would put effort into reproducing and fixing it. We just don't test them on ci. But that's just for practical reasons -- if we could test every one of the thousands of platforms we consider 'supported' in that sense we would; since we can't, we test a selection of common ones. But that doesn't mean the sdk isn't compatible with others.

ISTM what's relevant to the customer is what we support, in that sense (what we'll accept and fix bug reports on), not what we happen to test in ci (which we reserve the right to change at any point without notice). So I'd suggest making the list of what we support be that. WDYT?

funkyboy: @SimonWoolf I have a similar concern.
We still have to extra time on the CI (before we max out the allowed session length). So we could add some older browsers like: 

- Firefox 52
- Chrome TDB
- Safari iOS10
- Chrome on Android 4

SimonWoolf: sure, we can discuss that in the CI issue. my main point for the purposes of this PR is that the libraries we support is a (large) superset of the libraries we test in ci, and we shouldn't imply otherwise.

paddybyers: Isn't that addressed with my comments re fallbacks?

SimonWoolf: I don't think so. ISTM that having a big table at the top saying "we're only compatible with firefox 58+, chrome on Android 4.0, ..." just because those are the versions we currently test in ci is just misleading in a way that isn't really anything to do with whether we have transport fallbacks (it's not like firefox 57 doesn't support websockets). Rather it's to do with whether we'll fix bugs reported against firefox 57. Which we will, even through we don't have that specific version in ci. No?

mattheworiordan: I agree with @SimonWoolf's comments here.  A visitor is probably not going to read the detail in reality, they are going to scan, see a table, and assume we don't support Opera or old versions, when we do.

funkyboy:
Let's resume this conversation after we have [added more browser tests on Browserstack](https://github.com/ably/ably-js/issues/493)

mattheworiordan:
> Let's resume this conversation after we have 

Should we not resolve this now and add more browsers once #493 is fixed.  We should avoid coupling an update of the README to a task that will simply improve that README IMHO.

funkyboy:
@mattheworiordan @paddybyers @SimonWoolf ready for another look.

README.md
paddybyers: > just remove "recent"

We still need to do this

funkyboy:
@paddybyers Ooops. Fixed now.

SimonWoolf:
> ready for another look.

I don't think my and Matt's issues with this approach (https://github.com/ably/ably-js/pull/481#discussion_r172136135) have really been addressed.

Rather than going over those again, I'll do my own version as a counterpoint and push it up to a branch, see what you think.

SimonWoolf:
See https://github.com/ably/ably-js/pull/498 for my proposal for this ( https://github.com/ably/ably-js/blob/c658b5ae76c49ef65026cf21a48abb1762a25504/README.md#supported-platforms )

paddybyers:
> See #498 for my proposal for this

lgtm

PR #484, opened 2018-03-05
Add nvmrc file

### LINKED ISSUE #482: Add .nvmrc file
This will be ONLY for developers who wish to contribute to the SDK. It won't affect [supported platforms](https://github.com/ably/ably-js/issues/457).

funkyboy:
Fix #482


SimonWoolf:
One thing to watch is that node 8.9.4 ships with npm 5.6, which suffers from https://github.com/npm/npm/issues/18870 . So if you want to use node 8.9.4 you'll have to manually install (and make sure you use to publish packages) an older npm version than ships with it. Which is fine if you're happy with that, just something to be aware of.

(I've just tested npm 5.7 too, still has the same bug)

funkyboy:
@SimonWoolf We are in a kind of limbo :(
`6.10.2` has `npm 3.10.10`, good for the release but doesn't regenerate the lock file.

SimonWoolf:
> 6.10.2 has npm 3.10.10, good for the release but doesn't regenerate the lock file.

you can update npm independently of node. `npm install -g npm@5.1`

PR #490, opened 2018-03-19
Avoid xhr with local files on chrome 65+

SimonWoolf:
Fixes the problem @tomczoink reported with CORS errors when using a local file in chrome 65+. (See e.g. https://stackoverflow.com/questions/49256429/chrome-65-unable-to-make-post-requests-from-local-files-to-flask or https://stackoverflow.com/questions/49336294/read-write-error-with-cloud-firestore-the-access-control-allow-origin-header ).

The effect of this is currently to make the base transport jsonp. (Half-wondering about just getting rid of jsonp in this case and make websockets the only transport, any opinions on that?)

### DIFF
--- a/browser/fragments/platform-browser.js
+++ b/browser/fragments/platform-browser.js
@@ -2,13 +2,22 @@ if(typeof window !== 'object') {
 	console.log("Warning: this distribution of Ably is intended for browsers. On nodejs, please use the 'ably' package on npm");
 }
 
+function allowComet() {
+	/* xhr requests from local files are unreliable in some browsers, such as Chrome 65 and higher -- see eg
+	 * https://stackoverflow.com/questions/49256429/chrome-65-unable-to-make-post-requests-from-local-files-to-flask
+	 * So if websockets are supported, then just forget about comet transports and use that */
+	var loc = window.location;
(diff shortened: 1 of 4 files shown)


paddybyers:
I think we need a more generic check. For example, what about Blink? I think there's a reasonable likelihood that there are other browsers out there that support CORS but not from `file:////` because, after all, that's what the spec says they should do.

Personally I think it's cleaner to go with no XHR in the `file:///` case.

SimonWoolf:
> I think we need a more generic check. For example, what about Blink?

This works for that -- blink browsers seem to all have `Chrome/xx` with the base chrome version in their UA. E.g. opera/blink has `Chrome/47.0.2526.73 Safari/537.36 OPR/34.0.2036.42`; chromium has `Chromium/66.0.3359.26 Chrome/66.0.3359.26`, etc.

> Personally I think it's cleaner to go with no XHR in the file:/// case.

ok, sure. (note that that won't be all browsers with local files -- firefox in particular just has `null` in location.origin for local files -- but that's fine)

SimonWoolf:
74e88f5

SimonWoolf:
comments addressed: e1b8748

Issue #461, opened 2018-02-06
DefinitelyTyped


funkyboy:
Not clear if we should make a PR to https://github.com/DefinitelyTyped anymore or if they changed process. I replied to the PR here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/22860

funkyboy:
cc @SimonWoolf @mattheworiordan

funkyboy:
BTW a search for "ably" [here](https://microsoft.github.io/TypeSearch/) leads to https://www.npmjs.com/package/@types/ably

SimonWoolf:
> BTW a search for "ably" here leads to https://www.npmjs.com/package/@types/ably

yes, those are the types from definitelytyped. See https://github.com/DefinitelyTyped/DefinitelyTyped#npm

> Not clear if we should make a PR to https://github.com/DefinitelyTyped anymore or if they changed process.

As you pointed out in slack, they're pretty clear in the readme: `If you are the library author, or can make a pull request to the library, bundle types instead of publishing to DefinitelyTyped.` - https://github.com/DefinitelyTyped/DefinitelyTyped#create-a-new-package . So we don't need to duplicate our types there. Only action is for you to maybe try out ably in a sample typescript project and make sure the way we bundle types works for typescript users, per slack discussion.

funkyboy:
Update: see project by @withakay https://github.com/withakay/ably-ts-hack

withakay:
I should point out that that above linked project was borne out of a desire to be able to create a `Message` instance from TypeScript. To this end, a modified ably.d.ts type definition file has been created. Consider it an experiment, it probably has flaws and can almost certainly be improved upon.

mattheworiordan:
> Update: see project by @withakay https://github.com/withakay/ably-ts-hack

@funkyboy what's the context of that?

I don't really follow what the issue is to be honest.  As far as I can tell from what @SimonWoolf  has stated, the types will be inferred in the bundled package automatically.  Where are you up to at this point and how is @withakay's TS relevant?

funkyboy:
@mattheworiordan current [types in ably-js](https://github.com/ably/ably-js/blob/master/ably.d.ts) allow writing something like 
```
let message = new Ably.ablyLib.Message();
message.name = 'ciao';
channel.publish(message);
```

which:

- throws a runtime exception (still investigating [here](https://github.com/funkyboy/ts))
- doesn't feel "natural"

[...]

PR #492, opened 2018-03-23
Typescript namespace fixes

mattheworiordan:
@funkyboy I had a quick shot at fixing the type issues.

Seems like a very simple fix as far as I can tell.  WDYT?

Should supersede https://github.com/ably/ably-js/pull/475 and fix https://github.com/ably/ably-js/issues/461


mattheworiordan:
@funkyboy?  This shouldn't take more than a minute to review.

mattheworiordan:
@jdmnd as the TS expert, are you happy with this?

README.md
```diff
@@ -58,12 +58,18 @@ For React Native, do not use this package. Instead use the [ably-react-native](h
...
+let client = new Ably.Realtime(options); /* inferred type Ably.Realtime */
+let channel = client.channels.get('feed'); /* inferred type Ably.Types.RealtimeChannel */
 ```
 
-Additionally, the type definitions are registered with [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/ably).
+Intellisense in IDEs with TypeScript support is supported:
+
+![TypeScript suggestions](./resources/typescript-demo.gif)
+
+If your need to explicitly import the type definitions, see [ably.d.ts](./ably.d.ts).
```
jdmnd: Should say 'if you need'.

jdmnd:
Looks good to me. It's a little unusual for the types to all live in an embedded namespace, but it does make the autocomplete a bit more intuitive.

paddybyers:
> It's a little unusual for the types to all live in an embedded namespace, but it does make the autocomplete a bit more intuitive.

Any alternative suggestion?

ably.d.ts
```diff
@@ -3,7 +3,7 @@
 // Definitions by: Ably <https://github.com/ably/>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-declare namespace ablyLib {
+declare namespace Types {
   namespace ChannelState {
     type INITIALIZED = 'initialized';
     type ATTACHING = 'attaching';
```
jdmnd: This file uses double-quotes for all other strings.

jdmnd:
> Any alternative suggestion?

Ordinarily the types would be in the top-level namespace. i.e.: 
```
import { Realtime, ClientOptions } from 'ably'
let options: ably.ClientOptions = { key: 'foo' }
```
My guess is the way it's being done currently is so that when you type `ably.` the suggestions list is short – `Realtime` is likely what you want, not `HttpPaginatedResponse`.

I think I prefer the approach used.

PR #498, opened 2018-03-29
Supported platforms proposal

SimonWoolf:
Proposal re https://github.com/ably/ably-js/pull/481 , to solve https://github.com/ably/ably-js/pull/481#discussion_r172136135

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ We do not explicitly maintain compatibility with older platform versions; we no 
 
 **Note**: the ably-js browser library contains fallback mechanisms in order to be able to support older browsers; specifically it supports comet-based connections for browsers that do not support websockets, and this includes JSONP for browsers that do not support cross-origin XHR.
 Each of these fallback transport mechanisms is supported and tested on all of the listed target browsers; even when those browsers do not themselves require those fallbacks.
-These mean that the library should be compatible with nearly any recent browser, on the majority of platforms.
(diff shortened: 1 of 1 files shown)


Srushtika:
@mattheworiordan, this would done as part of the gradual docs restructure process that we (you, me and @tomczoink ) need to catchup on and finalise a plan.

However, broadly, here's what I thought:
On the [SDK downloads page](https://www.ably.io/download) on our docs, next every platform/framework available for download, we can add a small flag with 'See supported versions', which when clicked, would lead to the correct section in the relevant README file on GitHub. 

This would solve the maintenance issue of having to update the website everytime new info is available and also not duplicate already available data on GitHub. I think it's a good way to let interested users know the availability of this info via website docs, however, I'm yet to discuss this with @funkyboy and get his feedback and ideas.

mattheworiordan:
> next to every platform/framework available for download, we can add a small flag with 'See supported versions', 

Sure, sounds good, but might be tricky to actually fit in that page design wise though.

> This would solve the maintenance issue of having to update the website everytime new info is available and also not duplicate already available data on GitHu

Agreed.

>  I think it's a good way to let interested users know the availability of this info via website docs, however, I'm yet to discuss this with @funkyboy and get his feedback and ideas.

Ok, although I would like to see notes on this added to the usage pages as well which provide instructions for people to start using Ably, where platform support is paramount.

PR #499, opened 2018-04-01
Use valid React-Native timeout intervals

Gerharddc:
React-Native complains on Android if setTimeout receives anything larger than 60000ms.

### DIFF
--- a/browser/static/ably-reactnative.js
+++ b/browser/static/ably-reactnative.js
@@ -4043,9 +4043,9 @@ Defaults.TIMEOUTS = {
 	httpRequestTimeout         : 15000,
 	channelRetryTimeout        : 15000,
 	/* Not documented: */
-	connectionStateTtl         : 120000,
+	connectionStateTtl         : 60000,
 	realtimeRequestTimeout     : 10000,
-	recvTimeout                : 90000,
+	recvTimeout                : 60000,
 	preferenceConnectTimeout   : 6000,
 	parallelUpgradeDelay       : 6000
 };


paddybyers:
Thanks for this - we'll have a look.

SimonWoolf:
> Setting a timer for a long period of time, i.e. multiple minutes, is a performance and correctness issue on Android as it keeps the timer module awake, and timers can only be called when the app is in the foreground. See facebook/react-native#12981 for more info.

Interesting.

I don't think this PR is the correct solution to this, because (aside from only modifying the concatenated static file):

1. changing the comet recvTimeout to 60s will cause issues with the xhr streaming transport (for which the server refreshes the httprequest every 60s), and the default connectionStateTtl will just be overridden by the value in the connectiondetails on connection anyway. (Not to mention that the connectionStateTtl is still two minutes, and we shouldn't pretend otherwise just to get around a timer issue).

2. given the warning is saying the issue is that we're keeping the timer module awake, the thing is that there's always at least one timer running anyway -- when disconnected, aside from the connectionStateTtl timer, there's the disconnected/suspendedRetryTimer, and while connected there's the idleTimer. So the timer module will be kept awake anyway, so changing the times to appease the warning feels like a bit of a fig leaf -- if doing that we may as well just [suppress the warning](https://github.com/socketio/socket.io/issues/3054#issuecomment-365456894) for all the good it'd do.

I suggest we explore using https://github.com/ocetnik/react-native-background-timer , which replaces the js setTimeouts with wrappers around native equivalents, at least for android. (Does raise the question of why react-native doesn't do that itself, though -- https://github.com/facebook/react-native/issues/12981#issuecomment-377338655).

Gerharddc:
I agree that https://github.com/ocetnik/react-native-background-timer is the best options, just a lot of work...

SimonWoolf:
I'm closing this -- I've becoming more convinced that there is no correctness issue in the way ably-js is using timers, that making sure all the timers we use are under 60s would be completely pointless, that switching to using native timers (that would keep firing even when the activity is backgrounded and the websocket paused) would be a surprising and potentially problematic (wrt battery life) thing for a library like ably-js to do, and that it's perfectly fine to [ignore or suppress](https://stackoverflow.com/a/48778011) the warning. See https://github.com/ably/ably-js-react-native/issues/9#issuecomment-455680414 for a more detailed writeup.

Issue #501, opened 2018-04-05
Update TypeScript file also for browsers


funkyboy:
When writing `import * as Ably from 'ably';`

`'ably'` resolves to `/node_modules/ably/browser/static/ably` 

which wasn't updated in a [previous PR](https://github.com/ably/ably-js/pull/492/files)

PR #502, opened 2018-04-05
Update TypeScript file for browsers

funkyboy:
Fix https://github.com/ably/ably-js/issues/501


paddybyers:
Is this a cut and paste copy of a file that exists also for node?

funkyboy:
Correct. It’s technically a copy and paste, but made by `concat`, not by hand.

PR #504, opened 2018-04-13
Better to save package that is added

ndastur:


SimonWoolf:
Hi! Thanks for the contribution. This is actually the default behaviour if you use npm 5 or above, see https://github.com/npm/npm/issues/5108 . But sure, I guess it doesn't hurt to make it explicit, for people still on old versions of npm

funkyboy:
I agree with @SimonWoolf. Being clearer can't hurt :)
Thanks @ndastur!

PR #515, opened 2018-06-14
Switch from git protocol to https for msgpack dependency

SimonWoolf:
Apparently some environments block the git protocol port, see https://app.intercom.io/a/apps/ua39m1ld/inbox/inbox/conversation/16992208828

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "./nodejs/index.js": "./browser/static/ably-commonjs.js"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


SimonWoolf:
> Is it our plan to make this change generally?

in ably-js, msgpack-js is the only production dependency that's fetched from github. No idea about any other client libs

paddybyers:
> No idea about any other client libs

Right, that was my question, but we don't have other public repos that use npm so I guess it's moot.

PR #525, opened 2018-07-04
Update ws to v5

SimonWoolf:
cf https://ably-real-time.slack.com/archives/C030APSH3/p1530702930000192

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "dependencies": {
     "msgpack-js": "https://github.com/ably-forks/msgpack-js.git",
     "request": "^2.83.0",
-    "ws": "~4.0"
+    "ws": "^5.1"
   },
   "devDependencies": {
     "async": "git://github.com/ably-forks/async.git#requirejs",


SimonWoolf:
> Does this place any constraint on the node versions we support?

Was about to say no (I checked it still supported 4.x before putting the pr up), but on closer investigation, yes: it requires v4.5 (for Buffer.alloc), we currently support back to v4.2.

[Their justification is](https://github.com/websockets/ws/pull/1313) that older 4.x's have security holes so people on 4.x should upgrade to 4.8.7 or 4.9.1 anyway. 

WDYT?

paddybyers:
> Their justification is that older 4.x's have security holes so people on 4.x should upgrade to 4.8.7 or 4.9.1 anyway.
> 
> WDYT?

I think it's fine for us to exclude those older versions - we just ought to document that.

SimonWoolf:
Fixed + readme & engines updated in 99c967e5a4603a3c0f3393c59a00d633368b583b

PR #526, opened 2018-07-11
Give presence.subscribe callback arg the same behaviour as channel.subscribe

SimonWoolf:
writing a realtime presence test, was annoyed that the presence.subscribe callback arg was only called in the event of an attach error, wheras the channel.subscribe callback arg is called once attached. this makes them behave consistently

### DIFF
--- a/common/lib/client/realtimepresence.js
+++ b/common/lib/client/realtimepresence.js
@@ -106,7 +106,7 @@ var RealtimePresence = (function() {
 					break;
 				case 'initialized':
 				case 'detached':
-					channel.autonomousAttach();
+					channel.attach();
 				case 'attaching':
 					self.pendingPresence.push({
 						presence : presence,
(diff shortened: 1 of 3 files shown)


common/lib/client/realtimepresence.js
```diff
@@ -402,11 +403,20 @@ var RealtimePresence = (function() {
...
+		if(channel.state === 'failed') {
+			callback(ErrorInfo.fromValues(RealtimeChannel.invalidStateError('failed')));
+			return;
+		}
+
+		this.subscriptions.on(event, listener);
+		if(callback) {
+			channel.attach(callback);
+		} else {
+			channel.autonomousAttach();
```
paddybyers: Why this? (I realise this predated this PR.)

It's no different from attach except that it doesn't call the callback. Why not in `attach()` simply have `callback || noop` or `callback = callback || (err) => err && Logger......` ?

PR #529, opened 2018-07-23
Remove RealtimeChannel pendingEvents queue

SimonWoolf:
Not actually used any more due to transient publish functionality


paddybyers:
LGTM

PR #537, opened 2018-09-06
Use git protocol to install msgpack-js instead of https

bwobst:
`msgpack-js` does not get installed when using node 8.x (tested with 8.10.0). Updating the protocol to use `git://` instead of `https://` fixes this issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "./nodejs/index.js": "./browser/static/ably-commonjs.js"
   },
   "dependencies": {
-    "msgpack-js": "https://github.com/ably-forks/msgpack-js.git",
+    "msgpack-js": "git://github.com/ably-forks/msgpack-js.git",
     "request": "^2.87.0",
     "ws": "^5.1"
   },


SimonWoolf:
It used to use `git://`; it was changed because some environments block the git protocol port. It would be very unusual for 443 (https) to be blocked when git (9418) is allowed. It certainly isn't the case that https dependencies don't work with node 8 in general.

Could you tell me the output of the following commands run from the environment you're trying to install ably-js?
```
git ls-remote -h -t https://github.com/ably-forks/msgpack-js.git
```
```
git ls-remote -h -t git://github.com/ably-forks/msgpack-js.git
```

If https is for some reason blocked from your environment, you can use `git config --global url."git://".insteadOf https://` to configure the change globally. Though if that's the case then the library will also have problems making https requests to Ably.

bwobst:
I should also note that this is running in an [AWS SAM Local](https://github.com/awslabs/aws-sam-cli) serverless environment. The `git://` dependencies are installed properly but the single `https://` one is not.

SimonWoolf:
Ok, so getting that output means git can contact that repo fine over https. Which is reassuring. But puzzling that it doesn't work with npm then -- all npm does is shell out to the git cli. Some weirdly messed docker port forwarding, maybe.

Anyway, your issue should be solved by putting that `git config` line somewhere in your build script before the npm install. (or if that one doesn't work, maybe `url."https://github.com/".insteadOf git@github.com:`). I'm reluctant to switch the default back to git:// from https:// for everyone -- my impression is there are more environments that work with https but not with git than the reverse.

bwobst:
Confirmed running `git config --global url."git://".insteadOf https://` fixed the issue. Thanks!

PR #548, opened 2018-10-23
Add support for promises to public api

SimonWoolf:
Did most of this this a while ago when I was rewriting the realtime limits tests in async/await style -- I wanted to be able to write things like `await channel.publish(name, data)` and `await realtime.connection.once('connected')` -- just didn't get around to writing tests for it until now.

Currently only works on nodejs as it uses `util.promisify`. At some point when there's enough call for it I'll  rewrite to support browsers too.

### DIFF
--- a/common/lib/client/realtimechannel.js
+++ b/common/lib/client/realtimechannel.js
@@ -37,11 +37,15 @@ var RealtimeChannel = (function() {
 
 	RealtimeChannel.processListenerArgs = function(args) {
 		/* [event], listener, [callback] */
-		if(typeof(args[0]) == 'function')
-			return [null, args[0], args[1] || noop];
-		else
-			return [args[0], args[1], (args[2] || noop)];
-	}
+		args = Array.prototype.slice.call(args);
+		if(typeof args[0] === 'function') {
+			args.unshift(null);
+		}
+		if(args[args.length - 1] == undefined) {
+			args.pop();
+		}
+		return args;
+	};
 
 	RealtimeChannel.prototype.publish = function() {
 		var argCount = arguments.length,
@@ -49,6 +53,9 @@ var RealtimeChannel = (function() {
 			callback = arguments[argCount - 1];
 
 		if(typeof(callback) !== 'function') {
+			if(Platform.promisify) {
+				return Platform.promisify(this.publish).apply(this, arguments);
+			}
 			callback = noop;
 			++argCount;
 		}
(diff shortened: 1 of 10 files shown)


common/lib/client/realtimechannel.js
```diff
@@ -110,11 +117,16 @@ var RealtimeChannel = (function() {
 			callback = flags;
 			flags = null;
 		}
-		callback = callback || function(err) {
-			if(err) {
-				Logger.logAction(Logger.LOG_MAJOR, 'RealtimeChannel.attach()', 'Channel attach failed: ' + err.toString());
+		if(!callback) {
+			if(Platform.promisify) {
```
paddybyers: Doesn't this change the behaviour if `Platform.promisify` ? Ie there is no longer a log message?

SimonWoolf: If the user does something with the returned promise (i.e. calling `.catch()`, they'll get the error there, so the situation's analogous to if they passed a callback.

If they're not -- i.e. if they're just doing `channel.attach()` with no callback, and throwing away the return value -- and they don't have an `unhandledRejection` handler, then then they'll get the error in an `UnhandledPromiseRejectionWarning`.

(...although trying now, they actually get `Unhandled promise rejection: [object Object]`, because it uses `util.toString` which is non-overridable. Which isn't great. But, well, they get _something_, so that should be a hint to them start handling the promise properly)

(If it caused the channel state to go to `failed` (rather than just `detached` or `suspended`), that will trigger a log message elsewhere.)

SimonWoolf:
> I don't know if there is a way to avoid this, but it doesn't look very DRY.

Shrug -- you can't return from inside a function, so can't see any obvious way to dry it up *that* much. (And I have to use `call` with explicit arguments in some places rather than `apply(this, arguments)` since promisify puts the callback after the end of the args it's passed, so if there's any chance that the callback might be present but undefined you need to pass the args explicitly)

The alternative would be an approach like the one I took in realtime with [`Utils.addPromisifiedMethods`](https://github.com/ably/realtime/blob/develop/common/lib/util/utils.js#L1381-L1395) that adds new versions of everything (so rather than `publish()` returning a promise, you'd have a new `publishAsync` that returns a promise (which would also solve the UnhandledPromiseRejectionWarning thing from people who don't want promises). But the cost is a messier-looking API. Would you prefer that?

SimonWoolf:
(or could take a `promises` client option, and only return promises if that is true?)

SimonWoolf:
Closing in favour of https://github.com/ably/ably-js/pull/550

Issue #532, opened 2018-08-03
Add idempotent support


mattheworiordan:
See spec at https://github.com/ably/docs/pull/473

Ruby implementation at https://github.com/ably/ably-ruby/pull/165

Whilst this is technically 1.1 functionality, we can add this now so long as idempotent publishing is disabled by default.

@paddybyers who should do this?

SimonWoolf:
Implemented in https://github.com/ably/ably-js/pull/557

PR #557, opened 2018-12-07
Support clientoptions.idempotentRestPublishing, default false

SimonWoolf:
(plus some rearranging to allow a decent cross-platform randomString utility function)

Fixes https://github.com/ably/ably-js/issues/532

### DIFF
--- a/browser/fragments/platform-browser.js
+++ b/browser/fragments/platform-browser.js
@@ -36,7 +36,9 @@ var Platform = {
 		}
 		return function(arr, callback) {
 			crypto.getRandomValues(arr);
-			callback(null);
+			if(callback) {
+				callback(null);
+			}
 		};
 	})(window.crypto || window.msCrypto) // mscrypto for IE11
 };
--- a/browser/fragments/platform-nativescript.js
+++ b/browser/fragments/platform-nativescript.js
@@ -41,7 +41,9 @@ var Platform = {
 		for (var i = 0; i < arr.length; i++) {
 			arr[i] = bytes[i];
 		}
-		callback(null);
+		if(callback) {
+			callback(null);
+		}
 	}
 };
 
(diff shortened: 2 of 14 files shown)


SimonWoolf:
> Did you test https://docs.ably.io/client-lib-development-guide/features/#RSL1k4 ?

Nah. I did test that the id was being set; I figured checking that idempotency results from that is realtime's concertn, not ably-js's. (And that is in any case checked in the previous test, with manually-set ids). Also because "simulating an error response" needed some fiddling with http.js to be able to do in a uniform way between browsers and node. 

But shrug, as you think it's important I've added it: https://github.com/ably/ably-js/pull/557/commits/b5187194cb47422cda81029f74fe1d6bbe99944c

mattheworiordan:
>  I figured checking that idempotency results from that is realtime's concern, not ably-js's.

Sure, but that spec item is specifically testing that automatically assigned idempotent IDs are consistent across publish failures and retries.

> Also because "simulating an error response" needed some fiddling with http.js to be able to do in a uniform way between browsers and node.

Given JS is arguably our most utilised library, I am not sure skipping tests that may prevent future regressions on a pretty key feature is the right decision.

SimonWoolf:
> Sure, but that spec item is specifically testing that automatically assigned idempotent IDs are consistent across publish failures and retries.

With the js lib, retries and fallback hosts are done in the Http abstraction, by which time the body is encoded and the calling method is forgotten, so there's no mechanism by which the id could change at that stage. There's enough stuff to occupy us here that does have a serious chance of going wrong (eg behaviour of Uint8Array, crypto.getRandomValues, BufferUtils etc. across various browsers, nativescript, and react-native, including with noencryption builds of the library) without spending time worrying about stuff that doesn't.

Anyway, as mentioned I have now added the test.

Issue #534, opened 2018-08-17
Sync up the spreadsheet spec tracker with this lib


mattheworiordan:
We've done lots of work on adding 1.0. & 1.1 functionality to JS, yet we've not always updated the spec spreadsheet.

We should go through all changes and sync this up with the spreadsheet.

SimonWoolf:
Gone through spreadsheet. (as discussed, just from memory rather than linking tests, to save time & sanity)

Found a few unimplemented things:
- [ ] rsa4a -- log on init if no means to renew
- [ ] rsl1i -- message size calculation
- [ ] rtn14b -- retry immediately if token
- [ ] rtl6d1 -- bundling (size)
- [ ] rtl6d2 -- bundling (clientId)
- [ ] rtl15 -- channel.properties
- [ ] TS* -- stats not up to date
- [ ] TS12 -- intervalTime
- [ ] TI5 -- log href

mattheworiordan:
> as discussed, just from memory rather than linking tests, to save time & sanity

Deferred loss of sanity IMHO.  The less accuracy we have in the spreadsheet, makes it considerably harder for future maintainers.

mattheworiordan:
Thanks 👍

PR #561, opened 2018-12-14
Miscellaneous 1.1 stuff

SimonWoolf:
Fixes the remaining bits of https://github.com/ably/ably-js/issues/534#issuecomment-446325747 , plus adds PUT support


SimonWoolf:
@paddybyers poke

PR #564, opened 2018-12-26
Update correct import for webpack

hckhanh:
I faced this bug: `Realtime of undefined` until I change the way to import the module.


SimonWoolf:
Good point, since we don't set a default export -- thanks for the contribution :slightly_smiling_face:

Issue #440, opened 2017-12-12
Multiple tokens can be fetched simultaneously


SimonWoolf:
If several events (e.g. a rest request and a realtime connection, or two rest requests) are triggered concurrently with a library using token auth, then both trigger `_ensureValidAuthCredentials`, both see that `auth.token` is unset, and so both fetch a new token in parallel. E.g. https://jsbin.ably.io/ewelul/16/edit .

funkyboy:
@SimonWoolf @paddybyers what would be the intended result of this? Only one fires? And the other uses auth data from the first?

SimonWoolf:
yup, e.g. an attempt to get auth creds while another is ongoing could instead add an event listener to fire when the first has completed, something like that

funkyboy:
@SimonWoolf So there should be some global object (e.g. `window.ongoingauthobject` in the case of browsers) that tells you if there's an ongoing attempt to fetch an auth token.

SimonWoolf:
why would it need to be a global? it's library state

funkyboy:
@SimonWoolf you are right. Let's call it "global with respect to the library" tho. 
I mean with respect to the `Ably` object.

SimonWoolf:
it should be a property on `Auth`, which is a library singleton, yes

SimonWoolf:
This was fixed by https://github.com/ably/ably-js/pull/567

PR #567, opened 2019-01-16
Auth: Fix multiple concurrent token requests

SimonWoolf:
Fixes https://github.com/ably/ably-js/issues/440

Could have just made Auth an eventemitter, but eh, seemed a bit overkill to do that just for one event, the multicaster works


SimonWoolf:
(based this on 1.1 but on rethinking there's no reason for that, cherry-picked into master as 79e68746)

Issue #565, opened 2019-01-11
Occasional token request to https://null/keys/<key>/requestToken


SimonWoolf:
Customer reports that ably-js is (very occasionally) trying to make a token request to `https://null:443/keys/<key>/requestToken`. Lib is initialized with an environment and two (definitely-non-null) fallback hosts. Chrome 71 on windows 10.

Can't immediately see how this could happen.

PR #568, opened 2019-01-17
Http host when connected

SimonWoolf:
Fixes https://github.com/ably/ably-js/issues/565. Also fixes the anomaly where if you're connected, http requests don't use fallbacks -- being connected is not a guarantee that a datacenter has free capacity to service REST requests. (Considered just making it ignore the connected host and always go for Defaults.getHosts, but figured it's probably still worth it to go there first for efficient untilAttached history requests)

### DIFF
--- a/browser/lib/util/http.js
+++ b/browser/lib/util/http.js
@@ -13,6 +13,20 @@ var Http = (function() {
 			(statusCode >= 500 && statusCode <= 504);
 	}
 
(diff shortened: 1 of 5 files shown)


SimonWoolf:
test https://github.com/ably/ably-js/pull/568/commits/5f154bbd93102801fbe5850eb0114f466e3251ad

Issue #574, opened 2019-02-10
Typescript throwing errors on latest version and unable to install any previous versions


realfresh:
Hey, not sure if I'm doing something wrong but something weird is happening.

First, since updating to the latest stable version, 1.1.2, when I try to compile my typescript files, I'm getting the following errors below. One extremely strange thing is, I'm completely unable to roll back to any previous versions. If i install any other version through either yarn or npm, it always downloads version 1.1.2 files. Here are the typescript errors

Edit: Seems I was finally able to get Ably to install the right version by removing the package from package.json and then manually installing with "yarn add ably@1.0.23". Starting to feel like this whole problem is probably just something to do with my setup but gonna leave this up just in case.

```
node_modules/ably/browser/static/ably.d.ts:367:3 - error TS2416: Property 'auth' in type 'RestPromise' is not assignable to the same property in base type 'Rest'.
  Type 'AuthPromise' is not assignable to type 'Auth'.
    Types of property 'authorize' are incompatible.
      Type '(tokenParams?: TokenParams | undefined, authOptions?: AuthOptions | undefined) => Promise<TokenDetails>' is not assignable to type '(tokenParams?: TokenParams | tokenDetailsCallback | undefined, authOptions?: AuthOptions | tokenDetailsCallback | undefined, callback?: tokenDetailsCallback | undefined) => void'.
        Types of parameters 'tokenParams' and 'tokenParams' are incompatible.
          Type 'TokenParams | tokenDetailsCallback | undefined' is not assignable to type 'TokenParams | undefined'.
            Type 'tokenDetailsCallback' has no properties in common with type 'TokenParams'.

367     auth: Types.AuthPromise;
        ~~~~

node_modules/ably/browser/static/ably.d.ts:368:3 - error TS2416: Property 'channels' in type 'RestPromise' is not assignable to the same property in base type 'Rest'.
  Type 'Channels<ChannelPromise>' is not assignable to type 'Channels<Channel>'.
    Type 'ChannelPromise' is not assignable to type 'Channel'.
      Types of property 'history' are incompatible.
        Type '(params?: RestHistoryParams | undefined) => Promise<PaginatedResult<Message>>' is not assignable to type '(paramsOrCallback?: RestHistoryParams | paginatedResultCallback<Message> | undefined, callback?: paginatedResultCallback<Message> | undefined) => void'.
          Types of parameters 'params' and 'paramsOrCallback' are incompatible.
[...]

PR #575, opened 2019-02-11
Rewrite typings to pass `tsc --strict`

SimonWoolf:
Subtypes can't change the return value of their supertype in strict mode

Fixes https://github.com/ably/ably-js/issues/574

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -363,7 +363,27 @@ declare namespace Types {
 	}
 
 	// Classes
-	class RestPromise extends Rest {
+	class RestBase {
+		constructor(options: Types.ClientOptions | string);
+		static Crypto: Types.Crypto;
+		static Message: Types.MessageStatic;
+		static PresenceMessage: Types.PresenceMessageStatic;
+	}
+
+	class RestCallbacks extends RestBase {
+		static Promise: typeof Types.RestPromise;
+		static Callbacks: typeof Types.RestCallbacks;
+		auth: Types.AuthCallbacks;
+		channels: Types.Channels<Types.ChannelCallbacks>;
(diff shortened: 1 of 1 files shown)


ably.d.ts
```diff
@@ -363,7 +363,27 @@ declare namespace Types {
...
-	class RestPromise extends Rest {
+	class RestBase {
+		constructor(options: Types.ClientOptions | string);
+		static Crypto: Types.Crypto;
+		static Message: Types.MessageStatic;
+		static PresenceMessage: Types.PresenceMessageStatic;
+	}
+
+	class RestCallbacks extends RestBase {
+		static Promise: typeof Types.RestPromise;
```
paddybyers: why is this declared here?

SimonWoolf: Because if I declare it in `RestBase`, then when `RealtimeBase` extends `RestBase` and overrides it with `static Promise: typeof Types.RealtimePromise;` etc., tsc objects with `--strict`

The alternative would be something like 
```
	class RestBaseBase {
		constructor(options: Types.ClientOptions | string);
		static Crypto: Types.Crypto;
		static Message: Types.MessageStatic;
		static PresenceMessage: Types.PresenceMessageStatic;
	}

	class RestBase extends RestBaseBase {
		static Promise: typeof Types.RestPromise;
		static Callbacks: typeof Types.RestCallbacks;
	}

	class RestCallbacks extends RestBase {
		...
	}

	class RestPromise extends RestBase {
		...
	}

	class RealtimeBase extends RestBaseBase {
		...
	}
```
which seemed a bit messy. But shrug, can switch to that if you prefer it

paddybyers: No, stick with what you have

PR #576, opened 2019-02-11
React-Native: switch to using RNRandomBytes directly rather than going through react-native-randombytes

SimonWoolf:
Fixes https://github.com/ably/ably-js-react-native/issues/11

- Avoids the need to fork the library to avoid loading sjcl at runtime
- Avoids the unnecessary round-tripping of decoding base64 into a polyfilled buffer in r-n-rb, then extracting into a UInt32Array, then convert into a WordArray
- Fixes broken insufficient-randomness on react native

Haven't got a convenient react-native setup to actually test, but [apparently it works](https://github.com/ably/ably-js-react-native/issues/11#issuecomment-461894857)

### DIFF
--- a/browser/fragments/platform-reactnative.js
+++ b/browser/fragments/platform-reactnative.js
@@ -28,22 +28,11 @@ var Platform = {
 			str.length;
 	},
 	Promise: window.Promise,
-	getRandomValues: (function(randomBytes) {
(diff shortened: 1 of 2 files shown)


SimonWoolf:
Merged manually

PR #577, opened 2019-02-11
Fix auth options

richarddd:
We we need to re-authenticate using `realtime.auth.authorize` we need to be able to update clientId via the options. This is missing from the type defs


SimonWoolf:
Good spot -- thanks :slightly_smiling_face:

PR #578, opened 2019-02-11
Fix _uncheckedSetClientId

richarddd:
There is currently no way to set a new clientId if it has changed since `_tokenClientIdMismatch` will always return `true` for another id

### DIFF
--- a/browser/fragments/platform-reactnative.js
+++ b/browser/fragments/platform-reactnative.js
@@ -28,22 +28,11 @@ var Platform = {
 			str.length;
 	},
 	Promise: window.Promise,
-	getRandomValues: (function(randomBytes) {
-		return function(arr, callback) {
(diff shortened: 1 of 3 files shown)


SimonWoolf:
> There is currently no way to set a new clientId if it has changed since `_tokenClientIdMismatch` will always return `true` for another id

@richarddd That's correct, I'm afraid. Once a realtime connection is established, you can't change the clientId that's associated with it. (You can get rid of the client-side check but it won't do you much good as we'll reject the reauth on the server side). You should close() the connection and re-instantiate the library to start a fresh connection.

Technically if you're only using it as a REST client, not a Realtime client, then clientId changes would be ok as everything is stateless. But for consistency with the realtime client we just recommend that you re-instantiate the library in that case too.

PR #579, opened 2019-02-11
Updated README to include known limitations section 

Srushtika:
This change is per the discussion about the inclusion of unsupported features in the READMEs of various client libraries: https://github.com/ably/docs/issues/382


SimonWoolf:
> This client library is currently *not compatible* with the following Ably feature: **Feature:** Push Notifications target **Spec reference:** [RSH2](https://ably.io/documentation/client-lib-development-guide/features/#RSH2)

This feels a bit user-unfriendly at the moment. For a start, surely we shouldn't be linking to spec items in the client library development guide in primary user-facing docs like the readme.

What's wrong with "This library does not currently support being the target of a push notification (i.e. web push)"?

PR #580, opened 2019-02-25
Avoid xhrrequest relying on strict-mode parsing; tweak closure-compiler settings

SimonWoolf:


spec/rest/push.test.js
```diff
@@ -10,6 +10,7 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
 		defaultHeaders = Utils.defaultPostHeaders('msgpack'),
 		testDevice = {
 			id: 'testId',
+			clientId: 'testClientId',
```
paddybyers: Is this related to this PR?

SimonWoolf:
> I'm not sure about deleting the `minified` property without knowing why it's there

Why? Right now it's not being used, so whatever it was expected to be used for at the time it was added clearly never happened. If someday we we want to re-add it there's always git history

spec/rest/push.test.js
SimonWoolf: not really, was going to push it straight to master

paddybyers:
> Right now it's not being used

You know it's not used by the library; you don't know there's nothing external that depends on it. I don't know what that might be - but you can see from the history that at the time it was added it wasn't on the strength of an internal library dependency.

SimonWoolf:
Found commit b15a141d8c267cd427946b439c2d820f4c352285 from the day after `minified` was added uses it for the iframe transport, which is no longer used

paddybyers:
> Found commit [b15a141](https://github.com/ably/ably-js/commit/b15a141d8c267cd427946b439c2d820f4c352285) from the day after `minified` was added

OK, I remember now :)

PR #589, opened 2019-04-26
Fixes for calling authorize() before connected, and multiple times in succession

SimonWoolf:
Ensures that in all cases the token from most recent authorize() call takes precedence, and things waiting for a new token (including connectionmanager) will wait for that one to complete.

Also fixes a bug where if authorize() was called in the early stages of the connection (before a transport had been proposed), you'd get multiple concurrent connection sequences.

### DIFF
--- a/spec/realtime/auth.test.js
+++ b/spec/realtime/auth.test.js
@@ -1103,5 +1103,28 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
 		});
 	};
 
+	/* Check that only the last authorize matters */
+	exports.multiple_concurrent_authorize = function(test) {
+		test.expect(4);
+		var realtime = helper.AblyRealtime({log: {level: 4}, useTokenAuth: true, defaultTokenParams: { capability: {'wrong': ['*']} }});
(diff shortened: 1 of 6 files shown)


spec/realtime/auth.test.js
```diff
@@ -1103,5 +1103,28 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
...
+		test.expect(4);
+		var realtime = helper.AblyRealtime({log: {level: 4}, useTokenAuth: true, defaultTokenParams: { capability: {'wrong': ['*']} }});
+		realtime.connection.once('connected', function() {
+			realtime.auth.authorize({ capability: {'stillWrong': ['*']} }, function(err) {
+				test.ok(!err, 'Check first authorize cb was called');
+			});
+			realtime.auth.authorize({ capability: {'alsoNope': ['*']} }, function(err) {
+				test.ok(!err, 'Check second authorize cb was called');
+			});
+			realtime.auth.authorize({ capability: {'wtfAreYouThinking': ['*']} }, function(err) {
```
mattheworiordan: Ha ha

PR #590, opened 2019-05-14
Remove buggy callback argument from channel#unsubscribe()

SimonWoolf:
Was throwing an error in the case of unsubscribe() called with no callback on a failed channel.

I don't like the idea of a method taking a callback that is only ever called as a failure condition, and otherwise remains uncalled. And it's inconsistent with the convention of how callbacks are used in the rest
of the library, which is that if a method takes a callback it is (a) guaranteed to be called eventually (on either success or failure), and (b) the method returns a promise if no callback is supplied. Neither of which is true for unsubscribe().

And there's no reason to conceptualize unsubscribe() as failing if called on a failed channel anyway, we can still remove the listener perfectly well. The channel is not going to receive any more messages anyway, but then that's also true of a detached channel. And nothing in the spec requires unsubscribe to fail if called on a failed channel.

### DIFF
--- a/common/lib/client/realtimechannel.js
+++ b/common/lib/client/realtimechannel.js
@@ -234,8 +234,6 @@ var RealtimeChannel = (function() {
 		var event = args[0];
 		var listener = args[1];
 		var callback = args[2];
-		var subscriptions = this.subscriptions;
-		var events;
 
 		if(!callback) {
 			if(this.realtime.options.promises) {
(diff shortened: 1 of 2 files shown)


paddybyers:
What's the impact of this change on the promises API?

SimonWoolf:
> What's the impact of this change on the promises API?

unsubscribe() never returned a promise anyway. So no impact (other than to make it a bit more coherent, since there's no longer an abnormal method that takes an optional callback but doesn't return a promise..)

PR #591, opened 2019-06-04
Only deduplicate messages on the same channel

SimonWoolf:


SimonWoolf:
> How do you know that a duplicate message will arrive with no intervening message?

In general we don't. But that logic isn't trying to provide idempotence in general, it was a workaround for a [specific transport upgrade issue](https://github.com/ably/realtime/issues/674), for which that was always be the case.

PR #593, opened 2019-06-19
Web worker support 

CliftonH:
Replacing 'window' with 'self' to support running the realtime client in a worker context.

In node environments, the global scope can be referred to as `global` whereas in the browser it is `window`. However, `window` is undefined in a worker context and the global scope implements the `WorkerGlobalScope` interface. Both `window` and `WorkerGlobalScope` use the `self` property for referencing globally-scoped properties and variables, thus code like `self.Promise` may be executed in both the browser window and worker contexts, whereas `window.Promise` may only be executed in the browser window.


SimonWoolf:
Thanks for the PR! :slightly_smiling_face: 

My one concern is that since (unlike `window`) `self` is not unforgeable (protected from rebinding), if anyone has used `self` as something else in the global scope before the library is sourced, then this change would break the library. Perhaps unlikely, and it could be argued they deserve what they get if they do that, but still technically a breaking change.

To avoid any chance of backcompat issues, I'm thinking we could chuck something like
```
var global = (typeof window === 'object' && window) || (typeof self === 'object' && self);
```
in ably-prologue.js, and then use `global` whenever we need an explicit reference to the global object. What do you think? (@paddybyers any opinion?)

CliftonH:
That seems reasonable, I'll make that change now.

SimonWoolf:
> EDIT: Actually, I think this will throw as window is referenced in ably-epilogue.js and global won't be defined at that point as it's declared inside the IFFE. I could add something like global.global = global after the line you suggested above, but that feels wrong. I'm open to suggestions here.

Fair point. Looking at ably-epilogue.js now, can you think of any reason we couldn't move those inside the IIFE and reference exports/define as properties on global? i.e.
```
/* CommonJS support */
if (typeof global.exports === "object" && global.exports) {
	for (var obj in Ably) {
		if (Ably.hasOwnProperty(obj)) {
			global.exports[obj] = Ably[obj];
		}
	}
	/* SystemJS support for default exports to be added to the root of the module
	   https://github.com/frankwallis/plugin-typescript/issues/185 */
	global.exports.__esModule = true;
}

/* AMD support */
if (typeof global.define === "function" && global.define.amd) {
  global.define("ably", [], function() {
    return Ably;
  });
}

}).call({});
```

(cc @paddybyers for any input)

SimonWoolf:
lgtm. Inclined to merge subject to @paddybyers 's :+1: ?

browser/fragments/ably-prologue.js
```diff
@@ -1,6 +1,12 @@
 ;(function() {
-	"use strict";
-	var Ably = window.Ably = this;
+  "use strict";
+
+  /*
```
paddybyers: whitespace looks wrong

browser/fragments/ably-epilogue.js
```diff
@@ -13,23 +13,24 @@
...
-			exports[obj] = window.Ably[obj];
+	/* CommonJS support */
+	if (typeof global.exports === "object") {
+		for (var obj in global.Ably) {
+			if (global.Ably.hasOwnProperty(obj)) {
+				exports[obj] = global.Ably[obj];
+			}
 		}
+		/* SystemJS support for default exports to be added to the root of the module
+			 https://github.com/frankwallis/plugin-typescript/issues/185 */
```
paddybyers: whitespace looks wrong

CliftonH:
Thanks for the feedback, @paddybyers. I noticed that some of the spacing was a little weird to begin with and I'd originally opted to leave as much of it untouched as possible to minimize noise. If you look at `ably-prologue.js` and `ably-epilogue.js` in master, you'll see that they both use a mix of tabs and spaces for indention. I've since changed them to be consistent with the rest of the codebase.

SimonWoolf:
Thanks @CliftonH ! I'll merge now, and will cut a release once it passes jenkins. Cheers for the contribution :slightly_smiling_face:

SimonWoolf:
Included in 1.1.14

CliftonH:
Thanks, @SimonWoolf and @paddybyers. I appreciate your help.

ccoffey:
I think this issue needs to be re-opened, I see the following when using `v1.2.30`:

Screenshot 2023-09-01 at 13 43 00

I see this in `node_modules/ably/build/ably-commonjs.js`:

Screenshot 2023-09-01 at 13 53 20

Upgrading to `v1.2.43` (latest) does not fix the problem however downgrading to `v1.1.14` as mentioned in the above comment https://github.com/ably/ably-js/pull/593#issuecomment-505446157 does make the error go away.

PR #594, opened 2019-06-24
Log warning for potentially double-encoded tokenDetails

SimonWoolf:


common/lib/client/auth.js
```diff
@@ -447,6 +447,8 @@ var Auth = (function() {
 				} else if(tokenRequestOrDetails === 'undefined' || tokenRequestOrDetails === 'null') {
 					/* common failure mode with poorly-implemented authCallbacks */
 					callback(new ErrorInfo('Token string was literal null/undefined', 40170, 401));
+				} else if(tokenRequestOrDetails.indexOf('{' === 0) && !(contentType && contentType.indexOf('application/jwt') > -1)) {
```
paddybyers: tokenRequestOrDetails.indexOf('{' === 0)

this is insanely expensive for what you're trying to check. What about `tokenRequestOrDetails[0] === '{'` ?

SimonWoolf: Oddly enough a microbenchmark doesn't seem to bear that out, at least on node 10 -- the `[0]` method is actually about 5x slower:
```js
const ROUNDS = 1000*1000*1000;
let start;
const str = "arisetnoaorstenoariesntoaiersnts";

start = new Date();
for(let i=0; i<ROUNDS; i++) {
  str.indexOf('{') === 0;
}
console.log("indexOf", new Date() - start)

start = new Date();
for(let i=0; i<ROUNDS; i++) {
  str[0] === '{'
}
console.log("[0]", new Date() - start)
```
gives
```
indexOf 713
[0] 3631
```
It takes some very large strings (around 1M characters) before the indexOf method becomes slower than `[0]` due to the time to search the whole string.

(Even odder, `startsWith` is slower than either.)

On further experimentation this is only true for node 8 and 10 though; on node 6 and most web browsers, `[0]` is faster. Interesting.

I'll make the change :slightly_smiling_face:

paddybyers: Nobody noticed that the actual check `tokenRequestOrDetails.indexOf('{' === 0)` is completely wrong .....

SimonWoolf: Oh yeah. Oops :disappointed:

PR #598, opened 2019-07-01
Add workaround for possible mechanism for transport being disconnected immediately after connection

SimonWoolf:
This is re the customer connection that appears to disconnect immediately after connection.

Since we have no real logs from the client, this is a pure stab-in-the-dark guess, I have no real reason to think that any of these are the culprit.


SimonWoolf:
> LGTM, but how will we know if any of these happen in the wild?

We won't. (cf [slack comment](https://ably-real-time.slack.com/archives/CF7GGBF2N/p1562004280033400)). I guess we have to rely on someone noticing and asking us about it. Unless we want to do the sentry thing. (I think we balked last time at the size & invasiveness of the sentry lib and it having effects outside the library, but we could do our own light version that uses the normal ably-js infrastructure for making http requests)

paddybyers:
> I think we balked last time at the size & invasiveness of the sentry lib and it having effects outside the library, but we could do our own light version that uses the normal ably-js infrastructure for making http requests

I think we could consider this because we don't have to intercept exceptions or anything else; just POST something when we decide to.

PR #602, opened 2019-07-11
Adding delta stats category

codemerx:


common/lib/types/stats.js
```diff
@@ -8,6 +8,19 @@ var Stats = (function() {
...
+		if (values && values.category) {
+			this.category = { };
+			for (const [key, value] of Object.entries(values.category)) {
+				if (value) {
+					this.category[key] = new MessageCount(value);
+				}
+			}
+		}
+	}
+
```
SimonWoolf: ably-js is ES3, I'm afraid. No const, no for-of, no `Object.entries`, no array destructuring. (Tip: if you run `grunt` before making a PR, one of the tasks is the minfier, which enforces that the input is ES3 and will warn you about some things -- but not everything (only too-new syntax, not new methods))

common/lib/types/stats.js
```diff
@@ -29,9 +43,9 @@ var Stats = (function() {
 	}
 
 	function MessageTypes(values) {
-		this.messages = new MessageCount(values && values.messages);
-		this.presence = new MessageCount(values && values.presence);
-		this.all = new MessageCount(values && values.all);
+		this.messages = new MessageCategory(values && values.messages);
+		this.presence = new MessageCategory(values && values.presence);
+		this.all = new MessageCategory(values && values.all);
```
mattheworiordan: Why no test?

PR #604, opened 2019-07-24
Adding delta generation stats infrastructure

codemerx:


common/lib/types/stats.js
```diff
@@ -83,6 +83,25 @@ var Stats = (function() {
 		this.directPublishes = (values && values.directPublishes) || 0;
 	}
 
+	function ProcessedCount(values) {
+		this.succeeded = (values && values.succeeded) || 0;
```
mattheworiordan: I realise this is not relevant to this PR, but these attributes with zero value defaults should really be DRY'd up...

PR #608, opened 2019-09-10
Add missing `unsubscribe` in RealtimePresencePromise

kazk:


SimonWoolf:
Thanks @kazk :slightly_smiling_face:  You're right that the typings are missing unsubscribe, though the signature in this pr isn't quite right — you can't pass an array of actions to unsubscribe and it doesn't return a promise (it's a synchronous action). I've fixed in master by moving the RealtimePresenceCallbacks.unsubscribe definition to RealtimePresenceBase: 1624f93a10a29ef19399d10cf8be71a8ca2c76aa

kazk:
Thanks :)

> it doesn't return a promise

Yeah, I missed that only `subscribe` has `callbackWhenAttached?`.

> you can't pass an array of actions to unsubscribe

The documentation mentions it

image

https://www.ably.io/documentation/realtime/presence#unsubscribe

SimonWoolf:
@kazk re unsubscribe with an array: right you are, I'd forgotten I implemented that :slightly_smiling_face:. Fixed in 45e87814bb65b891b446bfa0fa8e9b4fc4e66314

PR #612, opened 2019-09-25
Adding support for user defined decoders

codemerx:
As per https://github.com/ably/wiki/issues/380#issuecomment-533647591 all the client libraries should accommodate user defined codecs. This is an initial attempt on implementing this infrastructure and using it for decoding `vcdiff` encoded messages.


common/lib/types/message.js
```diff
@@ -172,17 +186,35 @@ var Message = (function() {
...
 						default:
+							var codec = context.codecs[xform];
+							if(codec) {
+								try {
+									data = codec.decode(data, {
+										channelOptions: context.channelOptions,
+										encoding: match[0],
+										baseEncodedPreviousPayload: context.baseEncodedPreviousPayload
+									});
+									if(xform === 'vcdiff') {
```
mattheworiordan: Why is this only done for `vcdiff`?  If we introduced another codec, why would we not store the last payload?

spec/realtime/codecs.test.js
```diff
@@ -0,0 +1,65 @@
...
+			var realtime = helper.AblyRealtime({
+				codecs: {
+					'vcdiff': new AblyDeltaCodec.AblyVcdiffCodec()
+				}
+			});
+			var channel = realtime.channels.get('[?delta=vcdiff]' + testName);
+
+			channel.subscribe(function(message) {
+				var index = Number(message.name);
+				test.ok(equals(testData[index], message.data), 'Check message.data');
```
mattheworiordan: Am I missing something, how does this test that the delta is working?  If `var channel = realtime.channels.get('[?delta=vcdiff]' + testName);` was changed to `var channel = realtime.channels.get(testName);` wouldn't this test still pass?  If so, that's a flawed test as it's not testing that deltas are being applied.  I would argue we should have another test as well to make sure providing the `codecs` in the constructor `ClientOptions`, without passing `[?delta=vcdiff]` has no side-effect.

common/lib/types/message.js
codemerx: This is `vcdiff` specific sadly. We operate under the assumption that we need to store the payload right after we decoded `vcdiff` encoding if it was there and before we removed any other encoding. We also operate under the assumption that there might be only `base64` after `vcdiff`

spec/realtime/codecs.test.js
codemerx: <quote>
Am I missing something, how does this test that the delta is working? If var channel = realtime.channels.get('[?delta=vcdiff]' + testName); was changed to var channel = realtime.channels.get(testName); wouldn't this test still pass? If so, that's a flawed test as it's not testing that deltas are being applied.
</quote>

You are not missing anything. This is actually fixed here - [LINK](https://github.com/ably/ably-js/pull/615/files#diff-2bc3dbc1297758ec55d556915690486dR60)

<quote>
I would argue we should have another test as well to make sure providing the codecs in the constructor ClientOptions, without passing [?delta=vcdiff] has no side-effect.
</quote>

This is actually a very good idea. It is coming.

codemerx: Test added in https://github.com/ably/ably-js/pull/615/commits/cece42863925fab6a92967a6a75755ced1d73d97

PR #613, opened 2019-10-11
Initial implementation of channel parameters

codemerx:
Initial implementation of channel parameters as per https://github.com/ably/wiki/issues/380#issuecomment-539586784

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -248,7 +248,8 @@ declare namespace Types {
 
 	type ChannelParams = { [key: string]: string };
 
-	type ChannelModes = Array<string>;
+	type ChannelMode = 'PUBLISH' | 'SUBSCRIBE' | 'PRESENCE' | 'PRESENCE_SUBSCRIBE';
+	type ChannelModes = Array<ChannelMode>;
 
 	interface ChannelOptions {
 		cipher: any;
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -246,8 +246,14 @@ declare namespace Types {
 		ttl?: number;
 	}
 
+	type ChannelParams = Map<string, string>;
+
+	type ChannelModes = Array<string>;
+
 	interface ChannelOptions {
 		cipher: any;
+		params: ChannelParams;
+		modes: ChannelModes;
 	}
 
 	interface RestHistoryParams {
@@ -508,7 +514,8 @@ declare namespace Types {
 		name: string;
 		errorReason: ErrorInfo;
 		state: ChannelState;
-		setOptions: (options: any) => void;
+		params: ChannelParams;
+		modes: ChannelModes;
 		unsubscribe: (eventOrListener?: string | Array<string> | messageCallback<Message>, listener?: messageCallback<Message>) => void;
 	}
 
@@ -517,6 +524,7 @@ declare namespace Types {
 		attach: (callback?: standardCallback) => void;
 		detach: (callback?: standardCallback) => void;
 		history: (paramsOrCallback?: RealtimeHistoryParams | paginatedResultCallback<Message>, callback?: paginatedResultCallback<Message>) => void;
+		setOptions: (options: ChannelOptions, callback?: errorCallback) => void;
 		subscribe: (eventOrCallback: messageCallback<Message> | string | Array<string>, listener?: messageCallback<Message>, callbackWhenAttached?: standardCallback) => void;
 		publish: (messagesOrName: any, messageDataOrCallback?: errorCallback | any, callback?: errorCallback) => void;
 		whenState: (targetState: ChannelState, callback: channelEventCallback) => void;
@@ -527,6 +535,7 @@ declare namespace Types {
 		attach: () => Promise<void>;
 		detach: () => Promise<void>;
 		history: (params?: RealtimeHistoryParams) => Promise<PaginatedResult<Message>>;
+		setOptions: (options: ChannelOptions) => Promise<void>;
 		subscribe: (eventOrCallback: messageCallback<Message> | string | Array<string>, listener?: messageCallback<Message>) => Promise<void>;
 		publish: (messagesOrName: any, messageData?: any) => Promise<void>;
 		whenState: (targetState: ChannelState) => Promise<ChannelStateChange>;
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -246,7 +246,7 @@ declare namespace Types {
 		ttl?: number;
 	}
 
-	type ChannelParams = Map<string, string>;
+	type ChannelParams = { [key: string]: string };
 
 	type ChannelModes = Array<string>;
 
--- a/common/lib/client/realtimechannel.js
+++ b/common/lib/client/realtimechannel.js
@@ -189,6 +189,7 @@ var RealtimeChannel = (function() {
 			}
 		}
 		if(flags) {
+			Logger.deprecated('channel.attach() with flags', 'channel.setOptions() with channelOptions.params');
 			/* If flags requested, always do a re-attach. TODO only do this if
 			 * current mode differs from requested mode */
 			this._requestedFlags = flags;
--- a/common/lib/client/realtimechannel.js
+++ b/common/lib/client/realtimechannel.js
@@ -69,7 +69,7 @@ var RealtimeChannel = (function() {
 			return;
 		}
 		Channel.prototype.setOptions.call(this, options);
-		if(this.shouldReattachToSetOptions(options)) {
+		if(this._shouldReattachToSetOptions(options)) {
 			this._attach(true, callback);
 		} else {
 			callback();
@@ -93,7 +93,7 @@ var RealtimeChannel = (function() {
 		}
 	}
 
-	RealtimeChannel.prototype.shouldReattachToSetOptions = function(options) {
+	RealtimeChannel.prototype._shouldReattachToSetOptions = function(options) {
 		// TODO: Check if the new options are different than the old ones
 		return (this.state === 'attached' || this.state === 'attaching') && (options.params || options.modes);
 	};
--- a/common/lib/client/realtimechannel.js
+++ b/common/lib/client/realtimechannel.js
@@ -86,7 +86,7 @@ var RealtimeChannel = (function() {
 			}
 			for(var i = 0; i < options.modes.length; i++){
 				var currentMode = options.modes[i];
-				if(!currentMode || typeof currentMode !== 'string' || !Utils.arrIn(Utils.modes, String.prototype.toUpperCase.call(currentMode))){
+				if(!currentMode || typeof currentMode !== 'string' || !Utils.arrIn(ProtocolMessage.channelModes, String.prototype.toUpperCase.call(currentMode))){
 					return new ErrorInfo('Invalid channel mode: ' + currentMode, 40000, 400);
 				}
 			}
@@ -94,7 +94,6 @@ var RealtimeChannel = (function() {
 	}
 
 	RealtimeChannel.prototype._shouldReattachToSetOptions = function(options) {
-		// TODO: Check if the new options are different than the old ones
 		return (this.state === 'attached' || this.state === 'attaching') && (options.params || options.modes);
 	};
 
--- a/common/lib/client/realtimechannel.js
+++ b/common/lib/client/realtimechannel.js
@@ -23,6 +23,7 @@ var RealtimeChannel = (function() {
 		this._mode = null;
 		/* Temporary; only used for the checkChannelsOnResume option */
 		this._attachedMsgIndicator = false;
+		this._attachResume = false;
 	}
 	Utils.inherits(RealtimeChannel, Channel);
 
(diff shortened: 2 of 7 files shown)


common/lib/client/realtimechannel.js
```diff
@@ -136,50 +161,53 @@ var RealtimeChannel = (function() {
 			}
 		}
 		if(flags) {
```
SimonWoolf: Now we have a properly-spec'd api for setting flags (or at least, will do soon) we're going to want to deprecate calling attach() with an array of flags. I don't believe we ever told any customers about it, and certainly didn't document it -- it was mainly for use by realtime tests -- but still probably worth doing a release or two with it working but printing loud warnings to the console about contacting us, in case anyone discovered it on their own

common/lib/types/protocolmessage.js
```diff
@@ -71,6 +73,30 @@ var ProtocolMessage = (function() {
 		return this.flags && (this.flags & flags.MODE_ALL);
 	};
 
+	ProtocolMessage.prototype.encodeModesToFlags = function(modes) {
+		var self = this;
+		Utils.arrForEach(modes, function(mode) {
+			self.setFlag(mode);
+		});
+	};
```
SimonWoolf: Now we're exposing this as a public api we should do least some kind of error checking on the mode array the user passes in, to prevent typos (e.g. setFlag throws if given a string it doesn't recognise)

ably.d.ts
```diff
@@ -236,8 +236,14 @@ declare namespace Types {
 		ttl?: number;
 	}
 
+	type ChannelParams = Map<string, string>;
```
SimonWoolf: This should be `{ [key: string]: string }` not `Map<string, string>;`.

(I know I wrote `Map<string, string>` in the cross-language spec; that was meant as a 'Map' in the generic sense of a whatever the particular language idiom is for a map from keys to values. It's an unfortunate coincidence that  `Map<string, string>` is valid type specification for Typescript that isn't the one we want in this case, eg since ES6 Maps maps aren't json-serialisable. Apologies for misleading.)

codemerx: Sorry, my mistake - didn't realize that ES6 Map is not JSON serializable.

common/lib/types/protocolmessage.js
codemerx: Does this validation cover all scenarios you have in mind - https://github.com/ably/ably-js/pull/613/files#diff-ee8e6cae22c16e2972f2b341a26d2ed0R77-R96?

common/lib/client/realtimechannel.js
```diff
@@ -49,6 +49,57 @@ var RealtimeChannel = (function() {
...
+			var params = options.params;
+			if(!Utils.isObject(params)){
+				return new ErrorInfo('options.params must be an object', 40000, 400);
+			}
+			if('modes' in params && (typeof params.modes !== 'string' || !params.modes.length)) {
+				return new ErrorInfo('options.params.modes must be a non-empty string', 40000, 400);
+			}
+			if('delta' in params && (typeof params.delta !== 'string' || !params.delta.length)) {
+				return new ErrorInfo('options.params.delta must be a non-empty string', 40000, 400);
+			}
```
SimonWoolf: Picking two params that we happen to support now and validating them client-side doesn't really fit with how we decided to treat the params object (as an opaque bunch of key-value pairs that the client library shouldn't be looking into, just passing them on to realtime), I think we should just get rid of these

common/lib/types/protocolmessage.js
SimonWoolf: I'm afraid not, sorry. I made the comment on `ProtocolMessage.prototype.encodeModesToFlag` (and mentioned `setFlag`) for a reason: part of the point of having `options.modes` as a thing (as opposed to just using params.modes) is that you can get clientside validation. The client library has a list of all the modes it knows how to encode in the flags field, and it can reject a mode not in that list. You haven't attempted to do that -- you've added some validation that checks someone hasn't passed as an empty string on a number, but that's kinda missing the point -- we have the ability to statically check that someone hasn't typoed 'subscribe' as 'subscbe', and we should be doing that. (And can be a union type in the type definitions file)

common/lib/types/protocolmessage.js
```diff
@@ -71,6 +73,30 @@ var ProtocolMessage = (function() {
...
+
+	ProtocolMessage.prototype.decodeModesFromFlags = function() {
+		var modes = [],
+			self = this;
+		Utils.arrForEach([
```
mattheworiordan: This array of modes is duplicated.  Might be worth keeping in one place

common/lib/types/protocolmessage.js
```diff
@@ -128,7 +154,20 @@ var ProtocolMessage = (function() {
 			result += '; flags=' + Utils.arrFilter(flagNames, function(flag) {
 				return msg.hasFlag(flag);
 			}).join(',');
-
+		if(msg.params) {
+			var stringifiedParams = '';
+			for (var key in msg.params) {
```
mattheworiordan: Surprised this pattern of `for (var key in msg.params) {  if (Object.prototype.hasOwnProperty.call(msg.params, key) && msg.params[key])` doesn't have a Utils helper.  Seems to be used quite widely and thus adds to the overall library weight.

spec/realtime/channel.test.js
```diff
@@ -336,6 +336,359 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
...
+				var channelOptions = {
+					params: params
+				};
+				var channel = realtime.channels.get(testName, channelOptions);
+				channel.attach(function(err) {
+					if(err)
+						test.ok(false, 'Attach failed with error: ' + displayError(err));
+					test.deepEqual(channel.channelOptions, channelOptions, 'Check requested channel options');
+					test.deepEqual(channel.params, params, 'Check result params');
+					test.deepEqual(channel.modes, ['subscribe'], 'Check result modes');
```
mattheworiordan: Wouldn't it be better to have a test that checks the state from realtime in some way, as opposed to just checking the attributes on the channel are set.  Checking the channel attributes are set is not a reliable test to ensure the params/modes are sent to Realtime and are being received correctly.

spec/realtime/channel.test.js
```diff
@@ -336,6 +336,359 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
...
+					params: params
+				};
+				var channel = realtime.channels.get(testName);
+				channel.setOptions(channelOptions);
+				channel.attach(function(err) {
+					if(err)
+						test.ok(false, 'Attach failed with error: ' + displayError(err));
+					test.deepEqual(channel.channelOptions, channelOptions, 'Check requested channel options');
+					test.deepEqual(channel.params, params, 'Check result params');
+					test.deepEqual(channel.modes, ['subscribe'], 'Check result modes');
```
mattheworiordan: As above.  Also, should we consider testing what happens if you subscribe after call `setOptions`?

spec/realtime/channel.test.js
```diff
@@ -336,6 +336,359 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
...
+							test.ok(setOptionsReturned, 'setOptions should return immediately and call back after the reattach');
+							test.ok(channelReattached, 'Check reattach with modes occurred');
+							cb();
+						});
+						setOptionsReturned = true;
+					},
+					function(cb) {
+						var setOptionsReturned = false;
+						channel.setOptions({ }, function() {
+							test.ok(!setOptionsReturned, 'setOptions failed to call back immediately, when no reattach is required');
```
mattheworiordan: As above, can we test that apart from the attributes being set on the client lib SDK channel, that this data is actually making it to Realtime.  I will stop adding this comment for subsequent tests :)

spec/realtime/channel.test.js
codemerx: The attributes of the channel are set based on the `ATTACHED` response from the server, i.e. the assumption is Realtime received and processed these correctly provided it sent the correct response.

common/lib/client/realtimechannel.js
codemerx: Fixed already.

common/lib/types/protocolmessage.js
codemerx: Fixed already.

common/lib/client/realtimechannel.js
codemerx: Fixed already.

common/lib/client/realtimechannel.js
```diff
@@ -49,6 +49,57 @@ var RealtimeChannel = (function() {
 		return args;
 	};
 
+	RealtimeChannel.prototype.setOptions = function(options, callback) {
+		if(!callback) {
+			if (this.rest.options.promises) {
+				return Utils.promisify(this, 'setOptions', arguments);
+			}
+
+			callback = (err) => {
```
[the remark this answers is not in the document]
codemerx: Fixed already

common/lib/types/protocolmessage.js
mattheworiordan: @SimonWoolf wdyt?

spec/realtime/channel.test.js
mattheworiordan: >  i.e. the assumption is Realtime received and processed these correctly provided it sent the correct response.

Sure, but that's a flawed assumption, because someone else could refactor how Realtime sets these attributes, and then this test would pass mistakenly. That test is assuming how the internals of the library works, which is not how test should work IMHO.

common/lib/types/protocolmessage.js
codemerx: Fixed in https://github.com/ably/ably-js/pull/613/commits/b727b65b35ef4d05a66bc1902d886d1ab1286ce6

spec/realtime/channel.test.js
codemerx: Test added in https://github.com/ably/ably-js/pull/613/commits/b727b65b35ef4d05a66bc1902d886d1ab1286ce6

common/lib/types/protocolmessage.js
codemerx: Refactored in https://github.com/ably/ably-js/pull/613/commits/b727b65b35ef4d05a66bc1902d886d1ab1286ce6

spec/realtime/channel.test.js
codemerx: Fixed in https://github.com/ably/ably-js/pull/613/commits/804461e067147114f10062235374bdf9b95a0c08

spec/realtime/channel.test.js
codemerx: Fixed in https://github.com/ably/ably-js/pull/613/commits/804461e067147114f10062235374bdf9b95a0c08

QuintinWillison:
In case it's of help to other reviewers, until we merge the [1.2 integration branch](https://github.com/ably/docs/tree/master-1-2) in [docs](https://github.com/ably/docs), there's a [draft PR](https://github.com/ably/docs/pull/796) which contains a deployment link where you can view the rendered features spec.

ably.d.ts
```diff
@@ -498,7 +504,8 @@ declare namespace Types {
 		name: string;
 		errorReason: ErrorInfo;
 		state: ChannelState;
-		setOptions: (options: any) => void;
```
QuintinWillison: Is this a breaking change?

common/lib/client/realtimechannel.js
```diff
@@ -49,6 +49,53 @@ var RealtimeChannel = (function() {
...
+				var currentMode = options.modes[i];
+				if(!currentMode || typeof currentMode !== 'string' || !Utils.arrIn(Utils.modes, String.prototype.toUpperCase.call(currentMode))){
+					return new ErrorInfo('Invalid channel mode: ' + currentMode, 40000, 400);
+				}
+			}
+		}
+	}
+
+	RealtimeChannel.prototype._shouldReattachToSetOptions = function(options) {
+		// TODO: Check if the new options are different than the old ones
```
QuintinWillison: Does this need to be addressed?

common/lib/util/utils.js
```diff
@@ -3,6 +3,8 @@ var Utils = (function() {
 
 	function Utils() {}
 
+	Utils.modes = [ 'PRESENCE', 'PUBLISH', 'SUBSCRIBE', 'PRESENCE_SUBSCRIBE' ];
```
QuintinWillison: Could this be documented, have a more qualified name or even be located somewhere else but utils? It seems to relate to `TB2d` in the features spec, but that took me a while to work out.

common/lib/util/utils.js
```diff
@@ -253,6 +255,14 @@ var Utils = (function() {
 		return result;
 	};
 
+	Utils.forInOwnNonNullProps = function(ob, fn) {
```
QuintinWillison: Ideally I would want to see all methods in a utility library documented. Especially one like this where the usage isn't entirely obvious from the name (i.e. it supplies `fn` with property names, not their mapped values).

common/lib/client/realtimechannel.js
codemerx: It points to a possible performance optimization but we do not have a performance issue with this now.

common/lib/util/utils.js
codemerx: Uhm the name here actually describes precisely what the function does. It applies `fnq in a `for...in` on own, non null properties of `ob`. It is also a one liner almost so probably not the best candidate for documenting.

ably.d.ts
```diff
@@ -236,8 +236,14 @@ declare namespace Types {
 		ttl?: number;
 	}
 
+	type ChannelParams = { [key: string]: string };
+
+	type ChannelModes = Array<string>;
```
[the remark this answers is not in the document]
codemerx: Fixed in 52b2ccc756aef3f6a16b786887cdcfcd152bfdb0

Issue #371, opened 2017-01-09
Naming conventions for public/private methods?


SimonWoolf:
We've never had a consistent naming convention for methods that are vs aren't part of our actual api (except in cases where there's a name clash). We could prefix all methods that aren't part of our public API with an underscore, to give people who don't read the docs and just explore through tab-completion a heads-up about methods that might change at any point?

paddybyers:
I would support this

PR #615, opened 2019-10-18
Adding recovery on delta decode failure

codemerx:
Adding recovery mechanism for delta failures - previous message id mismatch and delta decode algorithm failure.

### DIFF
--- a/common/lib/client/realtimechannel.js
+++ b/common/lib/client/realtimechannel.js
@@ -28,9 +28,11 @@ var RealtimeChannel = (function() {
 			 * I.e. all encodings added by Ably Realtime have been decoded. */
 			baseEncodedPreviousPayload: undefined
 		};
-		this.lastPayloadMessageId = null;
-		this.lastPayloadProtocolMessageChannelSerial = null;
-		this.decodeFailureRecoveryInProgress = false;
+		this._lastPayload = {
+			messageId: null,
+			protocolMessageChannelSerial: null,
+			decodeFailureRecoveryInProgress: null
+		};
 	}
 	Utils.inherits(RealtimeChannel, Channel);
 
@@ -254,8 +256,8 @@ var RealtimeChannel = (function() {
 		} else if(this.channelOptions.modes) {
 			attachMsg.encodeModesToFlags(Utils.allToUpperCase(this.channelOptions.modes));
 		}
-		if(this.decodeFailureRecoveryInProgress) {
-			attachMsg.channelSerial = this.lastPayloadProtocolMessageChannelSerial;
+		if(this._lastPayload.decodeFailureRecoveryInProgress) {
+			attachMsg.channelSerial = this._lastPayload.protocolMessageChannelSerial;
 		}
 		this.sendMessage(attachMsg, noop);
 	};
@@ -443,9 +445,9 @@ var RealtimeChannel = (function() {
 				connectionId = message.connectionId,
 				timestamp = message.timestamp;
 
-			if(firstMessage.extras && firstMessage.extras.delta && firstMessage.extras.delta.from !== this.lastPayloadMessageId) {
+			if(firstMessage.extras && firstMessage.extras.delta && firstMessage.extras.delta.from !== this._lastPayload.messageId) {
 				Logger.logAction(Logger.LOG_MAJOR, 'RealtimeChannel.onMessage()', 'Delta message decode failure - previous message not available.');
-				this.startDecodeFailureRecovery();
+				this._startDecodeFailureRecovery();
 				break;
 			}
 
@@ -458,7 +460,7 @@ var RealtimeChannel = (function() {
 					Logger.logAction(Logger.LOG_MINOR, 'RealtimeChannel.onMessage()', e.toString());
 					if(e.code === 40018) {
 						Logger.logAction(Logger.LOG_MAJOR, 'RealtimeChannel.onMessage()', 'Message decode failed.');
-						this.startDecodeFailureRecovery();
+						this._startDecodeFailureRecovery();
 						return;
 					}
 				}
@@ -466,8 +468,8 @@ var RealtimeChannel = (function() {
 				if(!msg.timestamp) msg.timestamp = timestamp;
 				if(!msg.id) msg.id = id + ':' + i;
 			}
-			this.lastPayloadMessageId = lastMessage.id;
-			this.lastPayloadProtocolMessageChannelSerial = message.channelSerial;
+			this._lastPayload.messageId = lastMessage.id;
+			this._lastPayload.protocolMessageChannelSerial = message.channelSerial;
 			this.onEvent(messages);
 			break;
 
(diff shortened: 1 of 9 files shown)


common/lib/client/realtimechannel.js
```diff
@@ -28,6 +28,10 @@ var RealtimeChannel = (function() {
 			 * I.e. all encodings added by Ably Realtime have been decoded. */
 			baseEncodedPreviousPayload: undefined
 		};
+		this.lastPayloadMessageId = null;
+		this.lastPayloadProtocolMessageChannelSerial = null;
```
SimonWoolf: You're storing three different bits of the last message in three different properties in the `channel` (`lastPayloadMessageId`, `lastPayloadProtocolMessageChannelSerial`, and `baseEncodedPreviousPayload` as part of the `encodingDecodingContext` (without which you wouldn't need keep that context as a property of the channel). Wouldn't it be a lot simpler & neater to just store the `lastMessage` wholesale?

common/lib/client/realtimechannel.js
```diff
@@ -423,22 +438,49 @@ var RealtimeChannel = (function() {
...
 				timestamp = message.timestamp;
 
+			if(this.decodeFailureRecoveryInProgress) {
+				if(message.channelSerial === this.decodeFailureRecoveryForChannelSerial) {
+					this.decodeFailureRecoveryInProgress = false;
+					this.decodeFailureRecoveryForChannelSerial = null;
+				} else {
+					Logger.logAction(Logger.LOG_MAJOR, 'RealtimeChannel.onMessage()', 'Discarding message received during delta failure recovery process.');
+					break;
+				}
```
SimonWoolf: I don't think this is right -- the channel can't be stuck in `decodeFailureRecoveryInProgress=true` unless/until you get a message with the requested channelSerial and nothing else. E.g. what if the attach fails to recover back to that channelSerial and just gives you the current position? Or any other reason that you might not get a message with the expect channel serial (e.g. the library becomes disconnected for > 2 minutes during this process). With what you have here the channel would then be permanently refusing to process messages, even if it would be able to fine.

I think you have to keep trying to process all messages that arrive on the basis of what you have; the only effect that a recovery in progress should make is that you don't send a second attach while there's one in progress, and that state should cease to hold the moment you get any attached or detached (or error)

common/lib/types/errorinfo.js
```diff
@@ -1,10 +1,11 @@
 var ErrorInfo = (function() {
 
-	function ErrorInfo(message, code, statusCode, cause) {
+	function ErrorInfo(message, code, statusCode, cause, recoveryStrategy) {
```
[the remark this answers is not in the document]
paddybyers: Should we have a category of error codes that trigger reattach? It's nice to propagate it in the way you're doing it here, but polluting `ErrorInfo` just in order to be able to do that is a problem I think.

The other way to approach this possibly is that the thing that gets thrown by `decode()` doesn't have to be the `ErrorInfo` - it can be some other `Error` type that wraps an `ErrorInfo`, with the other `retryStrategy` information  implicitly or explicitly embedded in that type.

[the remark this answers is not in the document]
SimonWoolf: a category seems like overkill -- we get along fine currently with only one error code for all types of decode errors (40013), adding a second one for decode errors that trigger a reattach seems fine to me

common/lib/client/realtimechannel.js
codemerx: We discussed this shortly in a call with @paddybyers and reached the conclusion to leave it as is. `baseEncodedPreviousPayload` is decoded only to the point where `vcdiff` is removed or `base64` is removed, i.e. it is quite specific. Mutating  `lastMessage` to contain this kind of data is also not a good solution.

common/lib/client/realtimechannel.js
```diff
@@ -423,22 +437,37 @@ var RealtimeChannel = (function() {
...
+			}
+
 			for(var i = 0; i < messages.length; i++) {
 				var msg = messages[i];
 				try {
 					Message.decode(msg, this.encodingDecodingContext);
 				} catch (e) {
 					/* decrypt failed .. the most likely cause is that we have the wrong key */
 					Logger.logAction(Logger.LOG_MINOR, 'RealtimeChannel.onMessage()', e.toString());
+					if(e.code === 40018) {
```
codemerx: @paddybyers, should I add this to https://github.com/ably/ably-common/blob/master/protocol/errors.json?

common/lib/client/realtimechannel.js
```diff
@@ -28,6 +28,9 @@ var RealtimeChannel = (function() {
 			 * I.e. all encodings added by Ably Realtime have been decoded. */
 			baseEncodedPreviousPayload: undefined
 		};
+		this.lastPayloadMessageId = null;
```
mattheworiordan: Out of interest @SimonWoolf, under what conditions do we prefix attributes with `_`?  
@codemerx could we not reduce the attribute count on the channel object by doing something like:

```javascript
this.lastPayload = {
  messageId: null,
  protocolMessageChannelSerial: null, 
  decodeFailureRecoveryInProgress: false
}
```

common/lib/client/realtimechannel.js
mattheworiordan: Yes please 👍

common/lib/client/realtimechannel.js
SimonWoolf: > Out of interest @SimonWoolf, under what conditions do we prefix attributes with _?

Nominally private methods/attributes on public objects should be prefixed with underscores, see https://github.com/ably/ably-js/issues/371 . But until we have a major version release we're only doing that for new ones added as would be a breaking change if anyone was using them.

mattheworiordan: OK, so then we should be doing this for all new private methods right, and these additions fall into that category @codemerx.

common/lib/client/realtimechannel.js
codemerx: https://github.com/ably/ably-common/pull/48

common/lib/types/errorinfo.js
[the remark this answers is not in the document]
codemerx: Fixed in https://github.com/ably/ably-js/pull/615/commits/2ff1bc3306b914d70079848ba37fbd51f84deea1

common/lib/client/realtimechannel.js
codemerx: Fixed in https://github.com/ably/ably-js/pull/615/commits/b0cc789cbe06b3441aa5c2a0c6af8c4c514abb45

common/lib/client/realtimechannel.js
codemerx: Refactored in https://github.com/ably/ably-js/pull/615/commits/f514708ebc10edf3c9faa748bc0506b745a94caf

Issue #597, opened 2019-06-25
Investigate possible node 12 compatibility issue 


SimonWoolf:
node 12 build seems to be failing at node-gyp rebuild: https://travis-ci.org/ably/ably-js/jobs/549912367

PR #618, opened 2019-11-26
Update kexec for node 12 compatibility

SimonWoolf:
https://github.com/jprichardson/node-kexec/issues/36

Fixes https://github.com/ably/ably-js/issues/597 , by using [Meteor's fork of kexec](https://github.com/meteor/node-kexec/tree/update-for-node-12)

(only used for karma, so a devdependency, so shouldn't have affected customers)

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "karma-nodeunit": "~0.2",
     "karma-requirejs": "~0.2",
     "karma-story-reporter": "~0.3",
-    "kexec": "^2.0.2",
+    "kexec": "git://github.com/ably-forks/node-kexec.git#update-for-node-12",
     "nodeunit": "^0.11.3",
     "requirejs": "~2.1",
     "shelljs": "~0.3"
--- a/package-lock.json
+++ b/package-lock.json
@@ -348,8 +348,8 @@
       "dev": true
     },
     "async": {
-      "version": "1.5.2",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -39,7 +39,7 @@
     "karma-nodeunit": "~0.2",
     "karma-requirejs": "~0.2",
     "karma-story-reporter": "~0.3",
-    "kexec": "^2.0.2",
+    "kexec": "git@github.com:ably-forks/node-kexec.git#update-for-node-12",
```
paddybyers: AFAIUI we're not forking this because we want to make changes - why don't we just point to someone else's fork?

SimonWoolf: I kinda thought we had a policy of cloning to ably-forks whenever we want to reference someone's github branch as a client lib dependency, in case the original repo is deleted. (at least for things where the referenced repo is not unlikely to be deleted, ie if it isn't the original project repo etc)

paddybyers: @QuintinWillison I'd welcome your input on this.

TBH if the fork is owned by what appears to be a stable organisation then I don't think we need to create a fork for ourselves. Perhaps though we do need to decide case-by-case whether or not that's true. It also depends on how long we think we're likely to need to use a fork.

QuintinWillison: I would err on the side of defensive and never assume any third party is stable.

PR #619, opened 2019-12-02
Use "npm run ..." to ensure we use the local version of grunt-cli

SimonWoolf:
grunt-cli 1.3.2 changes how args are passed in a way that breaks
'grunt test:nodeunit --test ...'. This is fine as we specify 1.2.0. But
just calling `grunt` will use whatever grunt is in the user's $PATH,
which we have no control over. So switch to `npm run` commands that
explicitly reference the local repo grunt-cli.

(nb: for passing further arguments, need to pass `--` after
`npm run <script>` and before argument passed to grunt, or the
arguments are passed to `npm` rather than `grunt`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -456,17 +456,17 @@ client.time(function(err, time) { ... }); // time is in ms since epoch
 
 To run both the NodeUnit & Karma Browser tests, simply run the following command:
 
-    grunt test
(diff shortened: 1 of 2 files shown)


QuintinWillison:
Thanks!

PR #622, opened 2019-12-18
Delta plugin support

codemerx:
Introduces  https://github.com/ably-forks/vcdiff-decoder as plugin for `ably-js` as per https://github.com/ably/docs/pull/779


common/lib/types/message.js
```diff
@@ -137,21 +137,35 @@ var Message = (function() {
 
 	Message.serialize = Utils.encodeBody;
 
-	Message.decode = function(message, options) {
+	Message.decode = function(message, context) {
+		/* The second argument could be either EncodingDecodingContext that contains ChannelOptions or ChannelOptions */
+		if(!context.channelOptions) {
+			var channelOptions = context;
+			context = {
```
QuintinWillison: I don't get why it's valid to create a new context within this function and only ever store it as a local variable, given you set `baseEncodedPreviousPayload` on it just before the function returns, meaning that that value gets immediately discarded. Or am I missing something?

spec/realtime/delta.test.js
```diff
@@ -0,0 +1,65 @@
...
+				{ foo: 'bar', count: 2, status: 'active' },
+				{ foo: 'bar', count: 2, status: 'inactive' }
+			];
+
+			var realtime = helper.AblyRealtime({
+				plugins: {
+					'vcdiffDecoder': vcdiffDecoder
+				}
+			});
+			var channel = realtime.channels.get('[?delta=vcdiff]' + testName);
```
SimonWoolf: Doesn't matter much, but since this is based on feature-channel-params, would be good to test with that

common/lib/types/message.js
QuintinWillison: Perhaps something using overloads might be neater but I'm not precious about it. At a minimum, however, there could be some commentary explaining this as it's not obvious for maintainers in the future.

spec/realtime/delta.test.js
codemerx: The problem is channel params are not deployed on sandbox at this point of time since they are also a PR. Until they make it to sandbox this is the only way to have this test run.

common/lib/client/realtimechannel.js
```diff
@@ -21,6 +21,11 @@ var RealtimeChannel = (function() {
 		this.errorReason = null;
 		this._requestedFlags = null;
 		this._mode = null;
+		this._encodingDecodingContext = {
```
paddybyers: Isn't this now just a `decodingContext` ?

PR #625, opened 2020-01-23
CometTransport.sendRequest: fix interpretation of non-ProtocolMessage errors

SimonWoolf:
Bug introduced by me in 1f511f48a, which made xhrrequest emit both the full response body even in the case of an error response in order that rest.request() could populate the httpPaginatedResponse. The comettransport code assumed that at most one of err and body was set, and checked the latter first.

### DIFF
--- a/common/lib/transport/comettransport.js
+++ b/common/lib/transport/comettransport.js
@@ -96,6 +96,12 @@ var CometTransport = (function() {
 					err = err || new ErrorInfo('Request cancelled', 80003, 400);
 				}
 				self.recvRequest = null;
+				/* Connect request may complete without a emitting 'data' event since that is not
+				 * emitted for e.g. a non-streamed error response. Still implies preconnect. */
+				if(!preconnected) {
+					preconnected = true;
+					self.emit('preconnect');
+				}
 				self.onActivity();
 				if(err) {
 					if(err.code) {
--- a/common/lib/transport/comettransport.js
+++ b/common/lib/transport/comettransport.js
@@ -216,18 +216,25 @@ var CometTransport = (function() {
 			if(err) Logger.logAction(Logger.LOG_ERROR, 'CometTransport.sendItems()', 'on complete: err = ' + Utils.inspectError(err));
 			self.sendRequest = null;
 
-			/* the results of the request usually get handled as protocol responses instead of send errors */
+			/* the result of the request, even if a nack, is usually a protocol response
+			 * contained in the data. An err is anomolous, and indicates some issue with the
+			 * network,transport, or connection */
+			if(err) {
+				if(err.code) {
+					/* A protocol error received from realtime. TODO: once realtime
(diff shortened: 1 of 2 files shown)


common/lib/transport/comettransport.js
```diff
@@ -216,18 +216,25 @@ var CometTransport = (function() {
 			if(err) Logger.logAction(Logger.LOG_ERROR, 'CometTransport.sendItems()', 'on complete: err = ' + Utils.inspectError(err));
 			self.sendRequest = null;
 
-			/* the results of the request usually get handled as protocol responses instead of send errors */
+			/* the result of the request, even if a nack, is usually a protocol response
+			 * contained in the data. An err is anomolous, and indicates some issue with the
+			 * network,transport, or connection */
+			if(err) {
+				if(err.code) {
```
paddybyers: I can't think of a better way to check this, but it's not ideal doing a duck-typing check this way. I take it because JSONP we can't check for `X-Ably-xxx` headers?

Edit: I appreciate there's a TODO also here and I think there is an issue for that already.

common/lib/transport/comettransport.js
```diff
@@ -216,18 +216,25 @@ var CometTransport = (function() {
...
+					/* A protocol error received from realtime. TODO: once realtime
+					 * consistendly sends errors wrapped in protocol messages, should be
+					 * able to remove this */
+					self.onData(protocolMessageFromRawError(err));
+				} else {
+					/* A network/xhr error. Don't bother wrapping in a protocol message,
+					 * just disconnect the transport */
+					self.disconnect(err);
+				}
+				return;
```
paddybyers: This `return` is a difference in behaviour from before. Do we want this in the `protocol error received from realtime` case?

SimonWoolf: My reasoning was that since all paths in the err case lead to deciding that the transport should be disconnected -- and `/close`/`/disconnect` requests don't wait for `/send` requests, they're exec'd immediately -- there's no point sending anything further, we should just wait for the disconnect to happen and the transport to be disposed, and any pending attaches or messages will be resent on the new transport

common/lib/transport/comettransport.js
SimonWoolf: you're right, it is a bit hacky. I made a stab at a realtime PR to actually fix this (by always returning ProtocolMessage responses to comet requests, rather than sometimes those and sometimes REST-style error responses) back in March: https://github.com/ably/realtime/pull/2353/files . But it got parked, as it turned out to be more complex than I was expecting and other things took priority since, while hacky, the current way does mostly just work

SimonWoolf:
(this is failing a comet test, investigating)

SimonWoolf:
> (this is failing a comet test in ci, investigating)

https://github.com/ably/ably-js/pull/625/commits/f3a7a198a425b827ea1d809eaa8f99f59faec2c8

PR #631, opened 2020-02-26
Fix basic cases for resume and recover with timeserials

codemerx:
[Issue](https://github.com/ably/realtime/issues/2827)
This PR fixes the following cases:
- Resume channel with no messages since last connected
- Resume channel with messages since last connected
- Recover channel with no messages since last connected
- Recover channel with messages since last connected

### DIFF
--- a/common/lib/transport/connectionmanager.js
+++ b/common/lib/transport/connectionmanager.js
@@ -32,7 +32,7 @@ var ConnectionManager = (function() {
 		this.format = options.useBinaryProtocol ? 'msgpack' : 'json';
 
 		this.connectionSerial = undefined;
-		this.timeSerial = undefined;
+		this.connectionTimeserial = undefined;
 	}
 
 	TransportParams.prototype.getConnectParams = function(authParams) {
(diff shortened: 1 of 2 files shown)


SimonWoolf:
:+1: this looks correct given what realtime currently does.

Though before we do this: @paddybyers in the time since you wrote the realtime side (using `connectionTimeserial`), we've had a bit of pressure to reduce protocolmessage size over the wire: https://github.com/ably/wiki/issues/249 . Do we want to consider making the field name, say, `cts` rather than `connectionTimeserial` in the protocol, and save ourselves 17 bytes per protocolmessage?

paddybyers:
> Do we want to consider making the field name, say, `cts` rather than `connectionTimeserial` in the protocol, and save ourselves 17 bytes per protocolmessage?

I'm fine with optimising it. `cts` might be going a bit far.

SimonWoolf:
discussed over slack; we'll go with `ctSerial` for connection timeserial and `chSerial` for channel timeserial

Matt-Esch:
Closing this out pending https://github.com/ably/realtime/issues/3071

PR #638, opened 2020-04-07
Implement RTL17

codemerx:
Implement RTL17


QuintinWillison:
@simon, this is ready for review now if you have a few minutes. Thanks.

common/lib/client/realtimechannel.js
```diff
@@ -441,6 +441,13 @@ var RealtimeChannel = (function() {
 			break;
 
 		case actions.MESSAGE:
+
+			//RTL17
+			if(this.state !== 'attached') {
+				Logger.logAction(Logger.LOG_MAJOR, 'RealtimeChannel.onMessage()', 'Message skipped. Channel not in ATTACHED state.');
```
SimonWoolf, line 447: This message could be more helpful for debugging - say,
`Logger.logAction(Logger.LOG_MAJOR, 'RealtimeChannel.onMessage()', 'Message ' + message.id + ' skipped as channel state is ' + channel.state;`

spec/realtime/failure.test.js
```diff
@@ -381,5 +381,56 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
...
+			var messageReceived = false;
+			
+			sender_channel.subscribe(function(message) {
+
+				if(messageReceived) {
+					test.ok(false, 'Message received when channel not in ATTACHED state.');
+				}
+
+				messageReceived = true;
+				test.ok(JSON.stringify(testMessage) === JSON.stringify(message.data), 'Check first message received');
```
SimonWoolf, line 406: Equality assertion should use `test.equal(a, b, 'additional info')` rather than `test.ok(a === b, 'additional info')`, so that if it fails you get an `Expected 'foo' to equal 'bar'` rather than the somewhat less helpful `Expected false to be true`

SimonWoolf: (you can also use test.deepEqual to avoid the need to json-stringify before comparing)

common/lib/client/realtimechannel.js
QuintinWillison: https://github.com/ably/ably-js/pull/638/commits/f76fc914dc88b40a0575734a4f3fed5137416684

spec/realtime/failure.test.js
QuintinWillison: https://github.com/ably/ably-js/pull/638/commits/abfce6fa43db763a3b92b2013a96c58294d75569

PR #648, opened 2020-05-08
Push activation

paddybyers:
This PR replaces https://github.com/ably/ably-js/pull/435. The relevant commits from that branch are here and rebased onto the latest `master`.

This is not tested; I have just brought forward the relevant commits and resolved conflicts.


tcard:
In order to test this and make sure we're not missing anything, I suggest just [porting the Android test suite](https://github.com/ably/ably-java/blob/master/android/src/androidTest/java/io/ably/lib/test/android/AndroidPushTest.java).

We have there a TestActivation class that sets things up for testing: creates a REST instance that uses a TestActivationStateMachine that lets you intercept state transitions. In Java, it uses blocking queues; in JavaScript it could use an EventEmitter or promises.

TestActivation also implements helpers to move the state machine to particular states: `registerAndWait` and `moveToAfterRegistrationUpdateFailed`.

Then, it's just a matter of testing transitions. Transitions are often from a stable state to eventually other stable states, with maybe temporary states in between (e. g. `NotActivated` (stable) -> `WaitingForRegistrationSync` (temporary; a HTTP request is underway -> `WaitingForNewPushDeviceDetails` (stable)). The spec is very granular and deals with single transitions, but in order to be practical and test things a bit more end-to-end, our tests cover transitions from a stable state to other stable states, testing we go through the expected intermediate states given the right conditions. Tests are thus not that unit-y but provide more confidence things actually work.

So for example, the [test for RSH3a2a](https://github.com/ably/ably-java/blob/master/android/src/androidTest/java/io/ably/lib/test/android/AndroidPushTest.java#L458) covers a bunch of spec items in order to exhaustively check all that could happen after event `CalledActivate` while on `NotActivated` before we reach a stable state: either `AfterRegistrationSyncFailed` or `WaitingForNewPushDeviceDetails`. It does so in a table-driven fashion: provide some inputs, run some common logic, expect some results.

Well, that's all I can think of right now. Let me know if I can help some more!

SimonWoolf:
this is still failing the closurecompiler linting (const usage etc), see comments on the previous pr

jessica02200:
Yes I can't we wait until I get my fartbot going im so happy I shartted u know left a brown skid in your🧨draws

QuintinWillison:
This pull request is very old and the state of `main` branch has moved on significantly since it was last updated. The reality is that the work required to get this pull request and/or the underlying branch into a working state would most likely exceed the work required simply to restart from scratch, branched from the current state of `main`.

PR #649, opened 2020-05-12
Fix deprecated Buffer constructor usage

SimonWoolf:
This fixes the annoying deprecation warnings on recent node versions, but would raise the minimum node version to 5.10 (see eg [here](https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_string_encoding)), wdyt?

(if we have to we can abstract over them in platform-nodejs to get both, but bit of a faff)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,17 +24,17 @@ addons:
   firefox: "58.0"
 matrix:
   include:
-    - node_js: '4'
+    - node_js: '6'
       env: RUN_TESTS=./test/bin/ci-travis-firefox.sh
-    - node_js: '4'
-      env: RUN_TESTS=./test/bin/ci-travis-nodeunit.sh
     - node_js: '6'
       env: RUN_TESTS=./test/bin/ci-travis-nodeunit.sh
     - node_js: '8'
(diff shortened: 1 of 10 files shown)


QuintinWillison:
Version 5.10.0 of Node.js [was released](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V5.md#5.10.0) in March 2016 as far as I can tell.

Given that's over four years ago (ancient in npm terms, surely! 😀) and given the fact that you're targeting this fix for our forthcoming 1.2 release (not the current 1.1 line) then I think it's unlikely we'll upset too many people with that requirement.

Could you update the readme as part of this change? It currently states "version 4.5 or newer".

Oh, and `package.json` states `>=4.5.x`.

Thanks!

QuintinWillison:
@SimonWoolf if you're happy with my addition then please feel free to merge. Thanks.

SimonWoolf:
LGTM. Added https://github.com/ably/ably-js/pull/649/commits/1cf68417afb8c287c7a370e18f8fd55f489f19fc as well.

PR #652, opened 2020-05-19
Configure ruby tool version to match what's expected by infrastructure

QuintinWillison:
I had to do this in order to be able to deploy from my machine, given I use ASDF for managing Ruby runtimes and my global Ruby is 2.7.0.

I've done similar elsewhere: https://github.com/ably-forks/vcdiff-decoder/commit/822f1d20f0f51243c706d22b7672a8ff1ab771e1

However I was keen to run this by you, @SimonWoolf, because I'm now removing the exclusion of ASDF's `.tool-versions` from Git's field of vision. Is there a reason that's bad?


lmars:
@QuintinWillison we just upgraded the Ruby version in infrastructure to 2.6.6 yesterday, is there a reason to manage that version here too?

SimonWoolf:
> However I was keen to run this by you, @SimonWoolf, because I'm now removing the exclusion of ASDF's .tool-versions from Git's field of vision. Is there a reason that's bad?

Shrug, not a major one. I added it to .gitignore so I could have a local one that specified the node version I was using, but didn't want to press that particular version on anyone else. 

But if I might suggest an alternative solution: rather than having to manually keep the ruby version in sync with infrastructure in all the repos that invoke ably-env, I think you should be able to just make the grunt script change directory to the infrastructure repo and execute the command, so the normal version-resolving mechanisms will use the version specified by the infra repo. Eg `(cd ../infrastructure; bundle exec ./bin/ably-env foo)`. (parens to spawn a subshell, might not be necessary if the exec function is already doing that, don't remember offhand how it works)

QuintinWillison:
@lmars @SimonWoolf Thanks both. #653 is a cleaner approach.

PR #655, opened 2020-05-22
React Native support for Push Notifications

steelbrain:
WIP


steelbrain:
Hi Simon, thanks for your comment, I'll update the non-React Native code to conform to ES3 syntax but ReactNative itself has built-in ES2015+ support so things work out of the box there and we should use latest features when we can IMHO.

The reason I added SJCL is because there's two ways of using crypto APIs, sync and async. Previous patch fixed up the async API using RN Random Bytes, but the sync API (required by ULID) relied on SJCL which relied on cryto, which created a whole mess for end-consumers to deal with so I just mocked it away to fix ULID support.

QuintinWillison:
This pull request is very old and the state of `main` branch has moved on significantly since it was last updated. The reality is that the work required to get this pull request and/or the underlying branch into a working state would most likely exceed the work required simply to restart from scratch, branched from the current state of `main`.

See also #648.

PR #658, opened 2020-06-04
[SDK-38] fixes deltas on browsers in ably-js

SimonWoolf:
And fixes a few of other issues issues I noticed while doing it.

Also implements what we decided in https://github.com/ably/docs/issues/891 for what if you have no vcdiff plugin (or are using a browser too old to support deltas).

Related error code pr: https://github.com/ably/ably-common/pull/56

### DIFF
--- a/common/lib/client/realtimechannel.js
+++ b/common/lib/client/realtimechannel.js
@@ -469,9 +469,17 @@ var RealtimeChannel = (function() {
 				} catch (e) {
 					/* decrypt failed .. the most likely cause is that we have the wrong key */
 					Logger.logAction(Logger.LOG_ERROR, 'RealtimeChannel.onMessage()', e.toString());
-					if(e.code === 40018) {
-						this._startDecodeFailureRecovery(e);
-						return;
+					switch(e.code) {
+						case 40018:
+							/* decode failure */
+							this._startDecodeFailureRecovery(e);
+							return;
+						case 40019:
+							/* No vcdiff plugin passed in - no point recovering, give up */
+						case 40021:
+							/* Browser does not support deltas, similarly no point recovering */
+							this.notifyState('failed', e);
+							return;
 					}
 				}
 				if(!msg.connectionId) msg.connectionId = connectionId;
--- a/common/lib/client/realtimechannel.js
+++ b/common/lib/client/realtimechannel.js
@@ -34,6 +34,9 @@ var RealtimeChannel = (function() {
 			protocolMessageChannelSerial: null,
 			decodeFailureRecoveryInProgress: null
 		};
+		/* Only differences between this and the public event emitter is that this emits an
+		 * update event for all ATTACHEDs, whether resumed or not */
+		this._allChannelChanges = new EventEmitter();
 	}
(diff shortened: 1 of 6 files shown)


spec/realtime/delta.test.js
```diff
@@ -45,23 +45,33 @@ define(['shared_helper', 'vcdiff-decoder'], function(helper, vcdiffDecoder) {
...
-					test.equal(testVcdiffDecoder.numberOfCalls, testData.length - 1, 'Check number of delta messages');
+			channel.attach(function(err) {
+				if(err) {
+					test.ok(false, displayError(err));
 					closeAndFinish(test, realtime);
 				}
-			});
 
-			realtime.connection.on('connected', function() {
+				channel.on('attaching', function(stateChange) {
```
paddybyers, line 61: Wouldn't it be more straightforward just to test that you actually receive deltas (eg by checking extras) ?

Edit: I see you did that later

common/lib/client/realtimechannel.js
```diff
@@ -83,7 +86,24 @@ var RealtimeChannel = (function() {
 		if (this._decodingContext)
 			this._decodingContext.channelOptions = this.channelOptions;
 		if(this._shouldReattachToSetOptions(options)) {
-			this._attach(true, null, callback);
+			/* This does not just do _attach(true, null, callback) because that would put us
```
paddybyers, line 89: Does this need a spec clarification?

SimonWoolf: yeah, we should

common/lib/types/message.js
```diff
@@ -187,22 +187,30 @@ var Message = (function() {
...
+							try {
+								var deltaBase = context.baseEncodedPreviousPayload;
+								if(typeof deltaBase === 'string') {
+									deltaBase = BufferUtils.utf8Encode(deltaBase);
 								}
-								continue;
-							} else {
-								throw new Error('Missing Vcdiff decoder (https://github.com/ably-forks/vcdiff-decoder)');
+
+								/* vcdiff expects Uint8Arrays, can't copy with ArrayBuffers. (also, if we
```
QuintinWillison, line 202: This provided me food for thought. Nothing to stop us merging now but, FWIW: https://github.com/ably/ably-js/issues/659

common/lib/client/realtimechannel.js
QuintinWillison: I've created https://github.com/ably/docs/issues/894 

@SimonWoolf, my only concern on a skim read of this code (sorry, I've not looked in great depth) is to ensure that you've considered that the `_shouldReattachToSetOptions` 'method' implementation returns `true` for both the `attaching` as well as the `attached` local states, per [RTL16a](https://docs.ably.io/client-lib-development-guide/features/#RTL16a). Would this code running when the channel is `attaching` have a different outcome?

QuintinWillison: [moved](https://ably-real-time.slack.com/archives/C030C5YLY/p1591598178340200) to internal Slack channel

PR #672, opened 2020-07-02
Add state check to channels.release()

SimonWoolf:
Plus comment warning off anyone who's just browsing the source code and thinks that because it's there they should use it.

Plus noticed channels in the rest client were kept in an object called 'attached', which makes no sense since rest channels have no concept of attachment. Changed it to 'all', which matches the realtime client. This is technically a breaking change, but we've never documented or told anyone to enumerate channels using rest.channels.attached, so hopefully no-one is.

### DIFF
--- a/common/lib/client/realtime.js
+++ b/common/lib/client/realtime.js
@@ -132,11 +132,19 @@ var Realtime = (function() {
 		return channel;
 	};
 
+	/* Included to support certain niche use-cases; most users should ignore this.
+	 * Please do not use this unless you know what you're doing */
 	Channels.prototype.release = function(name) {
+		name = String(name);
 		var channel = this.all[name];
-		if(channel) {
-			delete this.all[name];
+		if(!channel) {
+			return;
 		}
+		var releaseErr = channel.getReleaseErr();
+		if(releaseErr) {
+			throw releaseErr;
+		}
+		delete this.all[name];
 	};
 
 	/* Records operations currently pending on a transport; used by connectionManager to decide when
--- a/common/lib/client/realtimechannel.js
+++ b/common/lib/client/realtimechannel.js
@@ -723,5 +723,14 @@ var RealtimeChannel = (function() {
 		return EventEmitter.prototype.whenState.call(this, state, this.state, listener);
 	}
 
+	/* @returns null (if can safely be released) | ErrorInfo (if cannot) */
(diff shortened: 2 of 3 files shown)


common/lib/client/realtime.js
```diff
@@ -132,11 +132,19 @@ var Realtime = (function() {
 		return channel;
 	};
 
+	/* Included to support certain niche use-cases; most users should ignore this.
+	 * Please do not use this unless you don't know what you're doing */
```
QuintinWillison, line 136:
```suggestion
	 * Please do not use this unless you know what you're doing */
```

Surely? 😉 

Same goes for the duplicate of this commentary in `rest.js`.

common/lib/client/realtime.js
```diff
@@ -132,11 +132,19 @@ var Realtime = (function() {
...
+		name = String(name);
 		var channel = this.all[name];
-		if(channel) {
-			delete this.all[name];
+		if(!channel) {
+			return;
+		}
+		var s = channel.state;
+		if(s !== 'initialized' && s !== 'detached' && s !== 'failed') {
+			throw new ErrorInfo('Can only release a channel in a terminal state (initialized, detached, or failed); was ' + s, 90001, 400);
```
QuintinWillison, line 145: The definitions of the words "terminal" and "initialized" don't marry up for me when used together in this context. I also can't find the word "terminal" anywhere in our [features specification](https://docs.ably.io/client-lib-development-guide/features/).

I think there could be more clarity here for the casual maintainer - perhaps even a function or 'method' that computes this conclusion (i.e. encapsulate this logic, enabling re-use and certainly providing it the formal location for documentary commentary to describe to maintainers of this codebase what it tells us and how that links to the spec.).

common/lib/client/realtime.js
SimonWoolf: lol. yes

common/lib/client/realtime.js
SimonWoolf: Fair enough. https://github.com/ably/ably-js/pull/672/commits/c86dd0109b8b3ce317e3f27a3e833c76bb891530 any better?

re spec item: if you're happy with this, I'll write a spec item for https://github.com/ably/docs/issues/437 , and can edit it into ably-js post-facto

PR #680, opened 2020-09-02
Fail connection immediately if authorize() called and 403 returned

SimonWoolf:
That is: interpret RSA4d as including requests made by a client lib to authenticate triggered by an explicit authorize() or an AUTH received from ably, not just to connect-sequence-triggered token fetches.

We might want to clarify the spec on this point.

### DIFF
--- a/browser/lib/transport/xhrrequest.js
+++ b/browser/lib/transport/xhrrequest.js
@@ -211,7 +211,7 @@ var XHRRequest = (function() {
 				return;
 			}
 
-			var err = responseBody.error;
+			var err = responseBody.error && ErrorInfo.fromValues(responseBody.error);
 			if(!err) {
 				err = new ErrorInfo('Error response received from server: ' + statusCode + ' body was: ' + Utils.inspect(responseBody), null, statusCode);
 			}
--- a/nodejs/lib/util/http.js
+++ b/nodejs/lib/util/http.js
@@ -34,11 +34,11 @@ this.Http = (function() {
 					case 'application/x-msgpack':
 						body = msgpack.decode(body);
 				}
-				var error = body.error || {
-					statusCode: statusCode,
-					code: headers['x-ably-errorcode'],
-					message: headers['x-ably-errormessage']
-				};
+				var error = body.error ? ErrorInfo.fromValues(body.error) : new ErrorInfo(
+					headers['x-ably-errormessage'] || 'Error response received from server: ' + statusCode + ' body was: ' + Utils.inspect(body),
+					headers['x-ably-errorcode'],
+					statusCode
+				);
 				callback(error, body, headers, true, statusCode);
 				return;
 			}
--- a/spec/realtime/auth.test.js
+++ b/spec/realtime/auth.test.js
@@ -517,6 +517,37 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
 		authUrl: echoServer + '/respondwith?status=403'
 	}, true); /* expectFailed: */
 
+	exports.authUrl_403_previously_active = function(test) {
+		var realtime, rest = helper.AblyRest();
+		rest.auth.requestToken(null, null, function(err, tokenDetails) {
+			if(err) {
+				test.ok(false, displayError(err));
+				closeAndFinish(test, realtime);
+				return;
+			}
+
+			var authPath = echoServer + "/?type=json&body=" + encodeURIComponent(JSON.stringify(tokenDetails));
+
+			realtime = helper.AblyRealtime({ authUrl: authPath });
+
+			realtime.connection.on('connected', function() {
+				test.ok(true, 'Connected to Ably using authUrl with TokenDetails JSON payload');
+
+				/* replace the authUrl and reauth */
+				realtime.auth.authorize(null, {authUrl: echoServer + '/respondwith?status=403'}, function(err, tokenDetails) {
+					test.equal(err && err.statusCode, 403, 'Check err statusCode');
+					test.equal(err && err.code, 40300, 'Check err code');
+					/* auth endpoints don't envelope, so this won't work with jsonp */
+					if(helper.bestTransport !== 'jsonp') {
+						test.equal(realtime.connection.state, 'failed', 'Check connection goes to the failed state');
+						test.equal(realtime.connection.errorReason && realtime.connection.errorReason.statusCode, 403, 'Check correct cause error code');
+					}
+					closeAndFinish(test, realtime);
+				});
+			});
+		});
+	};
+
 	/*
 	 * Check state change reason is propogated during a disconnect
 	 * (when connecting with a token that expires while connected)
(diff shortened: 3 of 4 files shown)


paddybyers (approved):
LGTM, thanks

spec/realtime/auth.test.js
```diff
@@ -517,6 +517,37 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
...
+
+			realtime.connection.on('connected', function() {
+				test.ok(true, 'Connected to Ably using authUrl with TokenDetails JSON payload');
+
+				/* replace the authUrl and reauth */
+				realtime.auth.authorize(null, {authUrl: echoServer + '/respondwith?status=403'}, function(err, tokenDetails) {
+					test.equal(err && err.statusCode, 403, 'Check err statusCode');
+					test.equal(err && err.code, 40300, 'Check err code');
+					/* auth endpoints don't envelope, so this won't work with jsonp */
+					if(helper.bestTransport !== 'jsonp') {
```
QuintinWillison, line 541: This comparison looks a little suspect to me but then I don't know the ins and outs of this test suite. It seems `bestTransport` is more used in tests to feed in to a next stage (common pattern being `realtime = helper.AblyRealtime({transports: [helper.bestTransport]})`, rather than for validation, so this just looks a bit odd. But, I don't know enough to be opinionated around whether it could or should be improved!

nodejs/lib/util/http.js
```diff
@@ -34,11 +34,11 @@ this.Http = (function() {
 					case 'application/x-msgpack':
 						body = msgpack.decode(body);
 				}
-				var error = body.error || {
-					statusCode: statusCode,
-					code: headers['x-ably-errorcode'],
-					message: headers['x-ably-errormessage']
-				};
+				var error = body.error ? ErrorInfo.fromValues(body.error) : new ErrorInfo(
+					headers['x-ably-errormessage'] || 'Error response received from server: ' + statusCode + ' body was: ' + Utils.inspect(body),
```
QuintinWillison, line 38: If I'm right that this is client-side message synthesis, in lieu of a server header, is that specified anywhere? I can only find [HP7](https://docs.ably.io/client-lib-development-guide/features/#HP7). Is it strictly required as part of this fix?

browser/lib/transport/xhrrequest.js
```diff
@@ -211,7 +211,7 @@ var XHRRequest = (function() {
 				return;
 			}
 
-			var err = responseBody.error;
+			var err = responseBody.error && ErrorInfo.fromValues(responseBody.error);
```
QuintinWillison, line 214: This confuses me. Looks like a boolean but later on is assigned to a new `ErrorInfo`. What's going on here? Might some commentary help in the code to explain? Or feel free to tell me I'm just being thick! 😁

QuintinWillison (approved):
Happy to approve as I know you need to get this moving on... but I did have a few queries out of curiosity.

nodejs/lib/util/http.js
SimonWoolf: No, not strictly related to the other fix - this is just improving the user experience a bit when showing an errorinfo made from a response received from a customer's auth server (which won't populate ably-specific headers). (I added it for node just cause I noticed we already did something like it for browsers, so was inconsistent that we don't in node)

It's not actually necessary, and not specified anywhere, just a bit of polish

browser/lib/transport/xhrrequest.js
SimonWoolf: the standard error response body from Ably looks like
```
{
	"error": {
		"message": "Could not find path: /nonexistent. (See https://help.ably.io/error/40400 for help.)",
		"code": 40400,
		"statusCode": 404,
		"href": "https://help.ably.io/error/40400",
		"serverId": "frontend.912a.2.eu-west-1-A.i-0a525ab4287758c86.108ykbPEgAoyLO"
	}
}
```
so this is just presenting that as an ErrorInfo rather than a plain object, which makes it display a bit nicer.

This format is documented in https://www.ably.io/documentation/rest-api#general, but not really in the spec anywhere that I can see. (kindof sortof [TI2](https://docs.ably.io/client-lib-development-guide/features/#TI2), but that's way too vague/handwavey to actually be useful)

QuintinWillison: Thanks, @SimonWoolf ... so just to make sure I understand right, `responseBody.error` is evaluated as a boolean (`null` == `false`) but the statement as a whole returns an object?

SimonWoolf: yes, exactly

spec/realtime/auth.test.js
```diff
@@ -517,6 +517,39 @@ define(['ably', 'shared_helper', 'async'], function(Ably, helper, async) {
...
+
+			realtime = helper.AblyRealtime({ authUrl: authPath });
+
+			realtime.connection.on('connected', function() {
+				test.ok(true, 'Connected to Ably using authUrl with TokenDetails JSON payload');
+
+				/* replace the authUrl and reauth */
+				realtime.auth.authorize(null, {authUrl: echoServer + '/respondwith?status=403'}, function(err, tokenDetails) {
+					test.equal(err && err.statusCode, 403, 'Check err statusCode');
+					test.equal(err && err.code, 40300, 'Check err code');
```
sacOO7, line 544: As per [RSA4d](https://docs.ably.io/client-lib-development-guide/features/#RSA4d), the code should be 80019 right 🤔

browser/lib/transport/xhrrequest.js
sacOO7: @QuintinWillison these are truthy and falsy values. But, we also need to consider edge cases while using them. A small read will be helpful https://www.sitepoint.com/javascript-truthy-falsy/

spec/realtime/auth.test.js
SimonWoolf: No. Look closer: in RSA4d, 80019 is specified as the code for the ErrorInfo that is "emitted with the state change and set as the connection errorReason". We check that a couple lines below, and if you add an assertion for the code there you'll find it is indeed 80019. This line is checking a different `err`: the one returned in the callback to `authorize()`.

Conceptually: 80019 is a connection error, meaning ["We can't connect because there was a problem with the client-configured authentication provider"](https://github.com/ably/ably-common/blob/main/protocol/errors.json#L153). That would not be appropriate to return in the callback for authorize(). There you want the original error the authorize actually encountered.

browser/lib/transport/xhrrequest.js
SimonWoolf: The library is assuming, as it's entitled to, that the server is returning response bodies in the form that it expects per [our public api](https://www.ably.io/documentation/rest-api#general), which defines that the `error` field if present will be an object with at least the core three fields. JS Objects are always truthy.

spec/realtime/auth.test.js
sacOO7: So, can we add an assertion for the same there? It's difficult to grasp intent at first glance. Maybe I can quickly push the change.

SimonWoolf: go for it

PR #682, opened 2020-09-12
Defaults: Generate environment fallbacks

lmars:
As per `RSC15g` introduced in https://github.com/ably/docs/pull/965.

### DIFF
--- a/common/lib/util/defaults.js
+++ b/common/lib/util/defaults.js
@@ -49,6 +49,17 @@ Defaults.getHttpScheme = function(options) {
 	return options.tls ? 'https://' : 'http://';
 };
 
+// construct environment fallback hosts as per RSC15i
+Defaults.environmentFallbackHosts = function(environment) {
+	return [
+		environment + '-a-fallback.ably-realtime.com',
+		environment + '-b-fallback.ably-realtime.com',
+		environment + '-c-fallback.ably-realtime.com',
+		environment + '-d-fallback.ably-realtime.com',
+		environment + '-e-fallback.ably-realtime.com',
+	];
+};
+
 Defaults.getFallbackHosts = function(options) {
 	var fallbackHosts = options.fallbackHosts,
 		httpMaxRetryCount = typeof(options.httpMaxRetryCount) !== 'undefined' ? options.httpMaxRetryCount : Defaults.httpMaxRetryCount;
@@ -116,16 +127,22 @@ Defaults.normaliseOptions = function(options) {
 	if(!('queueMessages' in options))
 		options.queueMessages = true;
 
-	var production = false;
 	if(options.restHost) {
 		options.realtimeHost = options.realtimeHost || options.restHost;
+		options.fallbackHosts = options.fallbackHostsUseDefault && !options.port && !options.tlsPort ? Defaults.FALLBACK_HOSTS : options.fallbackHosts;
 	} else {
 		var environment = (options.environment && String(options.environment).toLowerCase()) || Defaults.ENVIRONMENT;
-		production = !environment || (environment === 'production');
+		var production = !environment || (environment === 'production');
 		options.restHost = production ? Defaults.REST_HOST : environment + '-' + Defaults.REST_HOST;
 		options.realtimeHost = production ? Defaults.REALTIME_HOST : environment + '-' + Defaults.REALTIME_HOST;
+		if(!options.fallbackHosts && !options.port && !options.tlsPort) {
+			if(production || options.fallbackHostsUseDefault) {
+				options.fallbackHosts = Defaults.FALLBACK_HOSTS;
+			} else {
+				options.fallbackHosts = Defaults.environmentFallbackHosts(environment);
+			}
+		}
 	}
-	options.fallbackHosts = (production || options.fallbackHostsUseDefault) ? Defaults.FALLBACK_HOSTS : options.fallbackHosts;
 	Utils.arrForEach((options.fallbackHosts || []).concat(options.restHost, options.realtimeHost), checkHost);
 
 	options.port = options.port || Defaults.PORT;
--- a/spec/rest/defaults.test.js
+++ b/spec/rest/defaults.test.js
@@ -48,17 +48,18 @@ define(['ably', 'shared_helper'], function(Ably, helper) {
 
 	/* init with given environment */
 	exports.defaults_given_environment = function(test) {
-		test.expect(10);
+		test.expect(11);
 		var normalisedOptions = Defaults.normaliseOptions({environment: 'sandbox'});
 
 		test.equal(normalisedOptions.restHost, 'sandbox-rest.ably.io');
 		test.equal(normalisedOptions.realtimeHost, 'sandbox-realtime.ably.io');
 		test.equal(normalisedOptions.port, 80);
 		test.equal(normalisedOptions.tlsPort, 443);
-		test.equal(normalisedOptions.fallbackHosts, undefined);
+		test.deepEqual(normalisedOptions.fallbackHosts.sort(), Defaults.environmentFallbackHosts('sandbox').sort());
 		test.equal(normalisedOptions.tls, true);
 
-		test.deepEqual(Defaults.getHosts(normalisedOptions), [normalisedOptions.restHost]);
+		test.deepEqual(Defaults.getHosts(normalisedOptions).length, 4);
+		test.deepEqual(Defaults.getHosts(normalisedOptions)[0], normalisedOptions.restHost);
 		test.deepEqual(Defaults.getHost(normalisedOptions, 'sandbox-rest.ably.io', false), 'sandbox-rest.ably.io');
 		test.deepEqual(Defaults.getHost(normalisedOptions, 'sandbox-rest.ably.io', true), 'sandbox-realtime.ably.io');
 
@@ -160,7 +161,7 @@ define(['ably', 'shared_helper'], function(Ably, helper) {
 
 	/* init with no endpoint-related options and given default environment */
 	exports.defaults_set_default_environment = function(test) {
-		test.expect(10);
+		test.expect(11);
 		Defaults.ENVIRONMENT = 'sandbox';
 		var normalisedOptions = Defaults.normaliseOptions({});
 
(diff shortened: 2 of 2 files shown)


SimonWoolf (approved):

common/lib/util/defaults.js
```diff
@@ -116,16 +127,22 @@ Defaults.normaliseOptions = function(options) {
 	if(!('queueMessages' in options))
 		options.queueMessages = true;
 
-	var production = false;
 	if(options.restHost) {
 		options.realtimeHost = options.realtimeHost || options.restHost;
```
sacOO7, line 131: When options.restHost is specified, realtimeHost is mutated with restHost when realtimeHost is null. Is it possible to have same url for both resthost and realtimehost?

common/lib/util/defaults.js
```diff
@@ -116,16 +127,22 @@ Defaults.normaliseOptions = function(options) {
...
+		options.fallbackHosts = options.fallbackHostsUseDefault && !options.port && !options.tlsPort ? Defaults.FALLBACK_HOSTS : options.fallbackHosts;
 	} else {
 		var environment = (options.environment && String(options.environment).toLowerCase()) || Defaults.ENVIRONMENT;
-		production = !environment || (environment === 'production');
+		var production = !environment || (environment === 'production');
 		options.restHost = production ? Defaults.REST_HOST : environment + '-' + Defaults.REST_HOST;
 		options.realtimeHost = production ? Defaults.REALTIME_HOST : environment + '-' + Defaults.REALTIME_HOST;
```
sacOO7, line 137: options.realtimeHost is mutated without checking if it contains a custom value.

common/lib/util/defaults.js
```diff
@@ -116,16 +127,22 @@ Defaults.normaliseOptions = function(options) {
 	if(!('queueMessages' in options))
 		options.queueMessages = true;
 
-	var production = false;
 	if(options.restHost) {
 		options.realtimeHost = options.realtimeHost || options.restHost;
+		options.fallbackHosts = options.fallbackHostsUseDefault && !options.port && !options.tlsPort ? Defaults.FALLBACK_HOSTS : options.fallbackHosts;
```
sacOO7, line 132: returns hosts based on options.fallbackHostsUseDefault. Prioroty is given to options.fallbackHostsUseDefault to set fallback hosts

common/lib/util/defaults.js
```diff
@@ -116,16 +127,22 @@ Defaults.normaliseOptions = function(options) {
...
 	} else {
 		var environment = (options.environment && String(options.environment).toLowerCase()) || Defaults.ENVIRONMENT;
-		production = !environment || (environment === 'production');
+		var production = !environment || (environment === 'production');
 		options.restHost = production ? Defaults.REST_HOST : environment + '-' + Defaults.REST_HOST;
 		options.realtimeHost = production ? Defaults.REALTIME_HOST : environment + '-' + Defaults.REALTIME_HOST;
+		if(!options.fallbackHosts && !options.port && !options.tlsPort) {
```
sacOO7, line 138: Prioroty is given to options.fallbackHosts over options.fallbackHostsUseDefault

common/lib/util/defaults.js
sacOO7: RTN17a AND RTN17b

sacOO7:
The above comments were added as a part of https://github.com/ably/ably-java/pull/618 review. Observations are based on spec for implementing fallbacks (https://github.com/ably/docs/pull/965)

common/lib/util/defaults.js
lmars: > Is it possible to have same url for both resthost and realtimehost?

It is possible yes, and the split between `restHost` and `realtimeHost` is actually now obsolete because they always point at the same endpoint for a given environment. We'll be deprecating them in the spec in the near future.

common/lib/util/defaults.js
lmars: Yes that is a bug considering the current wording of the spec

common/lib/util/defaults.js
lmars: Yes you are correct this should prioritise `options.fallbackHosts`

common/lib/util/defaults.js
lmars: This is the correct behaviour, no?

sacOO7: Actually, I was not sure about the priorities above each other. I mentioned both, so either one of them is right.

sacOO7: The problem was logic inconsistency : )

common/lib/util/defaults.js
QuintinWillison: @lmars do you feel this requires a change to the spec, or to this library, please? From the discourse on this point so far (at least here) it feels like it could go either way. 😕

common/lib/util/defaults.js
QuintinWillison: @lmars to be clear, is this a bug that we need to fix in this library? If so, are you aware whether an issue or PR has been created or is WIP yet?

common/lib/util/defaults.js
QuintinWillison: @sacOO7 Can you resolve this conversation if this was not an issue, please?

common/lib/util/defaults.js
lmars: It's a bug in the code here, should be gated with `if(!options.realtimeHost)`

common/lib/util/defaults.js
lmars: After re-reading the spec, as per [`TO3k7`](https://www.ably.io/documentation/client-lib-development-guide/features#TO3k7) `fallbackHosts` and `fallbackHostsUseDefault` are mutually exclusive, so there should be no sense of prioritising one over the other, and there should be an explicit check that they are not both set (e.g. like we do in ably-ruby [here](https://github.com/ably/ably-ruby/blob/main/lib/ably/rest/client.rb#L177-L179)).

common/lib/util/defaults.js
lmars: As per https://github.com/ably/ably-js/pull/682#discussion_r502371464 there should be no question of priority, either `fallbackHosts` is set and they are the fallback hosts, or `fallbackHostsUseDefault` is set and the default fallbacks are used.

common/lib/util/defaults.js
QuintinWillison: Quite a good example of where a factory type approach for client options would iron these sorts of issues out without the need for runtime exceptions at the point of initialisation of the operational Ably instance! If we can find the opportunity to engineer ClientOptions-v2 then that's going to be my preferred approach, for sure. 🤔

common/lib/util/defaults.js
sacOO7: Can you mark this as resolved @lmars . I am not able to do it, it seems : (

common/lib/util/defaults.js
sacOO7: Yes, you can throw an error if both are set : )

common/lib/util/defaults.js
sacOO7: Are we assuming options.realtimeHost will be set?

lmars: @sacOO7 I'm not sure I understand your question, `options.realtimeHost` is an optional value that can be specified in development environments where the realtime host cannot be otherwise inferred by the library, and if it is set then it should not be overridden as is done here, it should be gated with an if statement as mentioned previously.

common/lib/util/defaults.js
```diff
@@ -116,16 +127,22 @@ Defaults.normaliseOptions = function(options) {
...
 
-	var production = false;
 	if(options.restHost) {
 		options.realtimeHost = options.realtimeHost || options.restHost;
+		options.fallbackHosts = options.fallbackHostsUseDefault && !options.port && !options.tlsPort ? Defaults.FALLBACK_HOSTS : options.fallbackHosts;
 	} else {
 		var environment = (options.environment && String(options.environment).toLowerCase()) || Defaults.ENVIRONMENT;
-		production = !environment || (environment === 'production');
+		var production = !environment || (environment === 'production');
 		options.restHost = production ? Defaults.REST_HOST : environment + '-' + Defaults.REST_HOST;
```
sacOO7, line 136: @lmars maybe I am thinking too critical... is it possible for users to specify both rest and environment in options?
According to the current implementation, it will just ignore env. option. Not sure, if we want to generate an error for the same... like we did in fallbackhosts

common/lib/util/defaults.js
```diff
@@ -116,16 +127,22 @@ Defaults.normaliseOptions = function(options) {
...
 		options.realtimeHost = production ? Defaults.REALTIME_HOST : environment + '-' + Defaults.REALTIME_HOST;
+		if(!options.fallbackHosts && !options.port && !options.tlsPort) {
+			if(production || options.fallbackHostsUseDefault) {
```
sacOO7, line 139: Just an ambiguous question ... user specifies environment (sandbox) with fallbackHostsUseDefault as true. After disconnection with the primary host (sandbox-ably.realtime.com).. subsequent redirects will be with the wrong hosts. I think although fallbackHostsUseDefault is deprecated, it's wrong for the library allowing to do that.

sacOO7: I mean, nothing serious. But, it's still something the library can manage

sacOO7: We can close this conversation if it's already well thought : )

SimonWoolf: > After disconnection with the primary host (sandbox-ably.realtime.com).. subsequent redirects will be with the wrong hosts.

Yes, if you tell the library to use the fallback hosts for the `production` cluster when your primary endpoint is a different cluster, then the library will be broken. The solution is to not do that - `fallbackHostsUseDefault` is for when using a primary endpoint that's just an alias for the production cluster. The library has no way of knowing when that is the case based on static analysis of the options only, we don't want the lib to have to make http requests to decide whether the options are valid. If you're giving fallback options the library is entitled to assume you know what you're doing. (Anyway, we are deprecating that option soon in favour of creating dns entries for the full set of autogenerated fallbacks)

common/lib/util/defaults.js
SimonWoolf: Just to put this in perspective: restHost and realtimeHost [are not actually documented client options](https://www.ably.io/documentation/realtime/usage#client-options). Customers should only be using `environment`, except for very rare exceptions like whitelabelled client libraries with enterprise customers, who we will give a set of client options to.

They're still part of the spec because having a standard way of pointing the library to a custom endpoint for testing is good for internal usability, but at the end of the day their usability is not very customer-facing.

common/lib/util/defaults.js
lmars: The identified bug has been addressed in https://github.com/ably/ably-js/pull/689.

common/lib/util/defaults.js
lmars: This is addressed in https://github.com/ably/ably-js/pull/689.

common/lib/util/defaults.js
lmars: This is addresses in https://github.com/ably/ably-js/pull/689.

common/lib/util/defaults.js
lmars: `restHost` and `environment` are mutually exclusive (see [`TO3k2`](https://www.ably.io/documentation/client-lib-development-guide/features#TO3k2)). I chose not to throw an error, as I think it's pretty clear to developers that `restHost` and `realtimeHost` take priority over `environment` anyway.

common/lib/util/defaults.js
lmars: As Simon said, the library has no way to infer whether these options are compatible or not, so it is up to the user to do the right thing, but I have at least added a warning to make it more apparent since there is no longer a need to set both anyway: https://github.com/ably/ably-js/pull/689

PR #683, opened 2020-09-30
add async examples to readme and create contributing.md

owenpearson:
This branch, if merged,  will add code snippets to readme.md to show example usage of the promise based api exported by ably-js. It will also move information about contributing and running tests to contributing.md, and resolve an issue with markdown syntax highlighting for a couple of existing code snippets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,6 @@ This SDK supports the following platforms:
 
 **TypeScript:** see [below](#typescript)
 
-
 We regression-test the library against a selection of those (which will change over time, but usually consists of the versions that are supported upstream, plus old versions of IE).
 
 However, we aim to be compatible with a much wider set of platforms and browsers than we can possibly test on. That means we'll happily support (and investigate reported problems with) any reasonably-widely-used browser. So if you find any compatibility issues, please do [raise an issue](https://github.com/ably/ably-js/issues) in this repository or [contact Ably customer support](https://support.ably.io) for advice.
@@ -31,25 +30,15 @@ However, we aim to be compatible with a much wider set of platforms and browsers
 Ably-js has fallback mechanisms in order to be able to support older browsers; specifically it supports comet-based connections for browsers that do not support websockets, and this includes JSONP for browsers that do not support cross-origin XHR. Each of these fallback transport mechanisms is supported and tested on all the browsers we test against, even when those browsers do not themselves require those fallbacks. These mean that the library should be compatible with nearly any browser on most platforms.
 Known browser incompatibilities will be documented as an issue in this repository using the ["compatibility" label](https://github.com/ably/ably-js/issues?q=is%3Aissue+is%3Aopen+label%3A%22compatibility%22).
 
-## Known Limitations
-
-This library currently does not support being the [target of a push notification](https://www.ably.io/documentation/general/push#activate) (i.e. web push)
-
-## Async API style
-
-This library exposes two API variants. Firstly, the original (and presently the default) callback-based API, which follows the usual node.js error-first callback style. Second, a promises-based API. With the promises variant, you can still pass a callback to methods and the callback will work as expected, but if you do not pass a callback, the method will return a promise. The API in use can be selected explicitly by requiring that specific variant when requiring/importing the library (or in the case of the browser version, when instantiating it). The usage instructions below make reference to both variants.
-
-For this library version, and for all future 1.x versions, the callback-based API will be the default, and the promises-based variant will need to be explicitly selected, to avoid breaking backwards compatibility. However, a move to the promises-based variant as the default is possible at the next major release. If you are not handling promises, and want a version of the library that will not start returning promises for calls where you don't pass a callback in future versions, you can explicitly require the callback variant.
-
 #### Version: 1.2.3
 
 The latest stable version of the Ably JavaScript client library is version: 1.2.3 .
 
 For complete API documentation, see the [Ably documentation](https://www.ably.io/documentation).
 
-## For node.js
+## Installation
 
-### Installation from npm
+### NodeJS
 
     npm install ably --save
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -497,83 +496,71 @@ client.stats(function(err, statsPage) {        // statsPage as PaginatedResult
...
+const ablyRealtimePromiseExample = async () => {
+  const channel = client.channels.get('myChannel');
 
-### Debugging the tests in your browser with NodeUnit test runner
+  // Attaching to a channel
+  await channel.attach();
 
-Simply open [spec/nodeunit.html](./spec/nodeunit.html) in your browser to run the test suite with a nice GUI.
+  client.connection.on('connected', async () => {
+    // Getting presence on a channel
```
SimonWoolf, line 515:
```
  // Attaching to a channel
  await channel.attach();

  client.connection.on('connected', async () => {
    // Getting presence on a channel
```

So, I know your intentions here are just listing a bunch of stuff that people can do, in no particular order. Trouble is, when people see a block of code in a readme, they tend to take it as a starting point, copy and paste the whole thing, and modify it. And if they do that with this, they'll get something broken (because it awaits the attach and then adds a connected listener once the channel is attached, which will then never fire because by the time the channel is attached it's already connected).

So while we can definitely do an example code block that lists a bunch of stuff you can do, I think that stuff has to be in an order that will actually work if run in that order.

(TBH I would get rid of the `on('connected'` listener from the basic example block. Unless there's something someone specifically wants to do every time the library reconnects, which is relatively rare, there's no actual need to do stuff within the connected listener, including it here might give a misleading impression that it's required)

owenpearson: Thanks, I can see that the `connection.on` was pointless so I've removed that and made the examples _pretty much_ independently "runnable". The only real caveat is that the realtime example requires that the clientId has been specified (otherwise they throw at `channel.presence.enter()`). I'm not sure if it's worth making that obvious somehow or assuming that developers will generally be able to figure that out for themselves...

SimonWoolf: 👍 much better

Srushtika (approved):
Looks great, thanks for adding the info in, @owenpearson

README.md
```diff
@@ -98,13 +95,13 @@ Intellisense in IDEs with TypeScript support is supported:
 
 If you need to explicitly import the type definitions, see [ably.d.ts](./ably.d.ts) (or `promises.d.ts` if you're requiring the library as `ably/promises`).
 
-### Using WebPack
+## Async API style
 
-(This applies to using webpack to compile for a browser; for node, see [Serverside usage with webpack](#serverside-usage-with-webpack))
+This library exposes two API variants. Firstly, the original (and presently the default) callback-based API, which follows the usual node.js error-first callback style. Second, a promises-based API. With the promises variant, you can still pass a callback to methods and the callback will work as expected, but if you do not pass a callback, the method will return a promise. The API in use can be selected explicitly by requiring that specific variant when requiring/importing the library (or in the case of the browser version, when instantiating it). The usage instructions below make reference to both variants.
```
QuintinWillison, line 100: Would prefer "Node.js" over "node.js", for correctness.

README.md
```diff
@@ -98,13 +95,13 @@ Intellisense in IDEs with TypeScript support is supported:
...
 
-WebPack will search your `node_modules` folder by default, so if you include `ably` in your `package.json` file, when running Webpack the following will allow you to `require('ably')` (or if using typescript or ES6 modules, `import * as Ably from 'ably';`). If your webpack target is set to 'browser', this will automatically use the browser commonjs distribution.
+For this library version, and for all future 1.x versions, the callback-based API will be the default, and the promises-based variant will need to be explicitly selected, to avoid breaking backwards compatibility. However, a move to the promises-based variant as the default is possible at the next major release. If you are not handling promises, and want a version of the library that will not start returning promises for calls where you don't pass a callback in future versions, you can explicitly require the callback variant.
```
QuintinWillison, line 102: The first sentence is a little long so do correct me if I've misinterpreted / misrepresented with my structural change suggestions.

I would also prefer to drop that final sentence (in my suggestion 👇) as we don't need to commit to that level of support at this stage. We may decide to make a clean break and only offer a promise-based API from 2.x, so I would prefer to leave that option on the table.

```suggestion
For this library version, and for all future 1.x versions, the callback-based API will be the default. This means that the promises-based variant will need to be explicitly selected, to avoid breaking backwards compatibility. A move to the promises-based variant as the default is likely at the next major release (i.e. 2.x onwards).
```

README.md
```diff
@@ -145,21 +142,21 @@ This readme gives some basic examples; for our full API documentation, please go
 
 All examples assume a client has been created as follows:
 
-```javascript
+```ts
```
QuintinWillison, line 145: It does bother me that you're needing to use `ts` highlighting for code that is definitely not TypeScript. My preference would be go go with `js` hightlighting, preferring correctness in the markdown at source level, rather than compensating for GitHub's rendering.

It looks like that may apply elsewhere in this PR, also.

Related, perhaps it's worth explicitly stating in the surrounding copy the ES level of the JavaScript being quoted?

README.md
```diff
@@ -497,83 +496,83 @@ client.stats(function(err, statsPage) {        // statsPage as PaginatedResult
...
-Open your browser to [http://localhost:3000](http://localhost:3000). If you are using a remote browser, refer to https://docs.saucelabs.com/reference/sauce-connect/ for instructions on setting up a local tunnel to your Nodeunit runner web server.
+  client.close();
+};
 
-### Debugging the tests in your browser with Karma
+ablyRealtimePromiseExample();
+```
 
-If you would like to run the tests through Karma, then:
+### Rest Example
```
QuintinWillison, line 536:
```suggestion
### REST Example
```

QuintinWillison (changes requested):
Some minor tweaks, please. 🤓

README.md
owenpearson: Unless I've made a mistake somewhere the code here should be TypeScript? It was certainly neither valid TypeScript nor valid JavaScript before. With regards to the ES level it might be best to just change the `var`s here to `const`s since no sane TypeScript developer would be using `var` but perhaps the same could be said for JavaScript code snippets both here and in the docs repo as well...

mattheworiordan:
@owenpearson are you planning on ensuring these documentation updates are reflected in the canonical documentation?  www.ably.io/documentation is our canonical source of docs, so it's important that is always up to date.  I don't think it currently makes any mention of these changes and should.

owenpearson:
@mattheworiordan I've created a docs issue for this [here](https://github.com/ably/docs/issues/968). I've not put any work into it yet because seems non-obvious to me where something like this should be added - the repo is designed in a language-generic way so I'm a little unsure where a language-specific detail like this should go. I'll push for input from others in that thread...

QuintinWillison (approved):

Issue #688, opened 2020-10-07
Lib currently does not allow developers to specify a custom realtimeHost without also specifying a custom restHost


owenpearson:
[Link to internal discussion](https://ably-real-time.slack.com/archives/C017C5EPYG0/p1602054291005000)

> so there's plenty of small things (eg the implicit assumption that no-one would ever provide a realtimeHost without also providing a restHost) that technically don't quite match the logic required by the spec, but stick around because they don't actually cause any problems (there's really no reason anyone would set a custom realtimeHost without also setting a custom restHost), so no-one ever bothers to fix them.

lmars:
Copying the comment I made in https://github.com/ably/ably-js/pull/682#discussion_r502061214:

> the split between restHost and realtimeHost is actually now obsolete because they always point at the same endpoint for a given environment. We'll be deprecating them in the spec in the near future.

I'd be tempted to close this as a won't fix.

sacOO7:
@owenpearson can you check if https://github.com/ably/ably-js/pull/689 fix it. I didn't get much time to look at the PR. But, I think it should, because it's spec-compliant.

PR #689, opened 2020-10-10
ClientOptions: Refactor host and fallback inference logic

lmars:
This refactors the way the library infers hosts and fallbacks, and fixes the handling of the deprecated `fallbackHostsUseDefault` option.

@sacOO7 this should address the concerns you raised in https://github.com/ably/ably-js/pull/682, in particularly:

- an error is thrown if `fallbackHosts` and `fallbackHostsUseDefault` are both set

- an error is thrown if `fallbackHostsUseDefault` and either `port` or `tlsPort` is set (since fallbacks don't apply when using custom ports as documented in https://github.com/ably/docs/pull/978)

- a warning is emitted when `fallbackHostsUseDefault` and `environment` are both set (these could be a valid combination so no need for an error, but there is no need to set them both anymore since each environment now always maps to its correct fallbacks)

- a warning is emitted when `realtimeHost` is set to `restHost` so the user knows that has happened and can choose to set `realtimeHost` explicitly if that is incorrect

- `realtimeHost` is only inferred if it isn't explicitly set

Fixes https://github.com/ably/ably-js/issues/688.

### DIFF
--- a/common/lib/util/defaults.js
+++ b/common/lib/util/defaults.js
@@ -153,21 +153,29 @@ Defaults.normaliseOptions = function(options) {
 	if(!('queueMessages' in options))
 		options.queueMessages = true;
 
-	if(options.restHost) {
-		options.realtimeHost = options.realtimeHost || options.restHost;
-	} else {
-		var environment = (options.environment && String(options.environment).toLowerCase()) || Defaults.ENVIRONMENT;
-		var production = !environment || (environment === 'production');
-		options.restHost = production ? Defaults.REST_HOST : environment + '-' + Defaults.REST_HOST;
-		options.realtimeHost = production ? Defaults.REALTIME_HOST : environment + '-' + Defaults.REALTIME_HOST;
-		if(!options.fallbackHosts && !options.port && !options.tlsPort) {
-			if(production) {
-				options.fallbackHosts = Defaults.FALLBACK_HOSTS;
-			} else {
-				options.fallbackHosts = Defaults.environmentFallbackHosts(environment);
-			}
+	/* infer hosts and fallbacks based on the configured environment */
+	var environment = (options.environment && String(options.environment).toLowerCase()) || Defaults.ENVIRONMENT;
+	var production = !environment || (environment === 'production');
+
+	if(!options.fallbackHosts && !options.restHost && !options.realtimeHost && !options.port && !options.tlsPort) {
+		options.fallbackHosts = production ? Defaults.FALLBACK_HOSTS : Defaults.environmentFallbackHosts(environment);
+	}
+
+	if(!options.realtimeHost) {
+		/* prefer setting realtimeHost to restHost as a custom restHost typically indicates
+		 * a development environment is being used that can't be inferred by the library */
+		if(options.restHost) {
+			Logger.logAction(Logger.LOG_WARN, 'Defaults.normaliseOptions', 'restHost is set to "' + options.restHost + '" but realtimeHost is not set, so setting realtimeHost to "' + options.restHost + '" too. If this is not what you want, please set realtimeHost explicitly.');
+			options.realtimeHost = options.restHost
+		} else {
+			options.realtimeHost = production ? Defaults.REALTIME_HOST : environment + '-' + Defaults.REALTIME_HOST;
 		}
 	}
+
+	if(!options.restHost) {
+		options.restHost = production ? Defaults.REST_HOST : environment + '-' + Defaults.REST_HOST;
+	}
+
 	Utils.arrForEach((options.fallbackHosts || []).concat(options.restHost, options.realtimeHost), checkHost);
 
 	options.port = options.port || Defaults.PORT;
--- a/common/lib/util/defaults.js
+++ b/common/lib/util/defaults.js
@@ -107,14 +107,14 @@ Defaults.normaliseOptions = function(options) {
 		if(options.fallbackHosts) {
 			var msg = 'fallbackHosts and fallbackHostsUseDefault cannot both be set';
 			Logger.logAction(Logger.LOG_ERROR, 'Defaults.normaliseOptions', msg);
-			throw new Error(msg);
+			throw new ErrorInfo(msg, 40000, 400);
 		}
 
 		/* default fallbacks can't be used with custom ports */
 		if(options.port || options.tlsPort) {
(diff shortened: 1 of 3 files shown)


common/lib/util/defaults.js
```diff
@@ -102,6 +102,32 @@ Defaults.normaliseOptions = function(options) {
 		options.queueMessages = options.queueEvents;
 	}
 
+	if(options.fallbackHostsUseDefault) {
+		/* fallbackHostsUseDefault and fallbackHosts are mutually exclusive as per TO3k7 */
+		if(options.fallbackHosts) {
+			var msg = 'fallbackHosts and fallbackHostsUseDefault cannot both be set';
+			Logger.logAction(Logger.LOG_ERROR, 'Defaults.normaliseOptions', msg);
+			throw new Error(msg);
```
SimonWoolf, line 110: These (and below) should be `throw new ErrorInfo(msg, 40000, 400);`

SimonWoolf (approved):
👍

common/lib/util/defaults.js
lmars: Fixed in ddb5a34a

common/lib/util/defaults.js
```diff
@@ -102,6 +102,32 @@ Defaults.normaliseOptions = function(options) {
...
+		/* default fallbacks can't be used with custom ports */
+		if(options.port || options.tlsPort) {
+			var msg = 'fallbackHostsUseDefault cannot be set when port or tlsPort are set';
+			Logger.logAction(Logger.LOG_ERROR, 'Defaults.normaliseOptions', msg);
+			throw new ErrorInfo(msg, 40000, 400);
+		}
+
+		/* emit an appropriate deprecation warning */
+		if(options.environment) {
+			Logger.deprecatedWithMsg('fallbackHostsUseDefault', 'There is no longer a need to set this when the environment option is also set since the library will now generate the correct fallback hosts using the environment option.');
```
sacOO7, line 122: @lmars strongly referring to this code for .net implementation.
 Not sure if this is a bug, but based on the warning, it's supposed to set env fallbacks if env is set right (even if fallbackHostsUseDefault) is used? currently, it won't be able to do that.

common/lib/util/defaults.js
```diff
@@ -102,6 +102,32 @@ Defaults.normaliseOptions = function(options) {
...
+		} else {
+			Logger.deprecated('fallbackHostsUseDefault', 'fallbackHosts: Ably.Defaults.FALLBACK_HOSTS');
+		}
+
+		/* use the default fallback hosts as requested */
+		options.fallbackHosts = Defaults.FALLBACK_HOSTS;
```
sacOO7, line 128: Maybe this should be moved inside else statement Line 124

common/lib/util/defaults.js
lmars: > it's supposed to set env fallbacks if env is set right (even if fallbackHostsUseDefault) is used?

No, as per [RSC15g4](https://www.ably.io/documentation/client-lib-development-guide/features#RSC15g4), `fallbackHostsUseDefault` takes precedence over `environment`.

To provide some context as to why, the intention of `fallbackHostsUseDefault` is that default fallbacks can be used by customers even when setting `environment` which would otherwise disable fallback behaviour for them unless they explicitly set `fallbackHosts`, but that is no longer needed as every valid `environment` always now has associated environment fallbacks that point at the correct place (e.g. some of them are just CNAMEs to the default fallbacks), so we will eventually remove `fallbackHostsUseDefault` but for now it is just deprecated whilst keeping it's meaning (i.e. use default fallbacks even when `environment` is set).

common/lib/util/defaults.js
lmars: I don't think so, no.

sacOO7: Yeah ofc

common/lib/util/defaults.js
sacOO7: What about the message? I am sure, it conveys otherwise 😕

sacOO7: Or maybe it's a little confusing for me

lmars: Here's what the message says when printed:

> 22:07:14.433 Ably: Deprecation warning - 'fallbackHostsUseDefault' is deprecated and will be removed from a future version. There is no longer a need to set this when the environment option is also set since the library will now generate the correct fallback hosts using the environment option.

If it's not clear, then let's make it clearer, what do you suggest?

sacOO7: Warning should mention that environment value will be overridden because of fallbackHostsUseDefault

sacOO7: or maybe a little implicit message

lmars: So let's think about the user experience here of a user that is currently setting `fallbackHostsUseDefault`.

Prior to the change that introduced environment fallbacks, let's assume the user was using the following options:

```js
var client = new Ably.Realtime({
  environment:             'acme',
  fallbackHostsUseDefault: true,
})
```

In this case, they had `fallbackHostsUseDefault` set because they wanted to use the default fallbacks. If they didn't set it, they wouldn't get any fallback behaviour at all, because the library didn't have the concept of environment fallbacks.

Now, with the recent changes that introduce environment fallbacks and deprecate `fallbackHostsUseDefault`, what should the behaviour be? Well, given we're just talking about the option being deprecated, the behaviour should be the same, they should continue to use the default fallbacks like they've been using so far, but they should be made aware that `fallbackHostsUseDefault` is deprecated and should no longer be set since now the library will transparently determine the fallback hosts from the environment.

So, to your point:

> Warning should mention that environment value will be overridden because of fallbackHostsUseDefault. To the point !!

I actually think that would be more confusing to the user, because what is being overridden? From their point of view, setting `environment` alone meant getting no fallback behaviour, so they were "overriding" that by setting `fallbackHostsUseDefault` to force the library to use some fallbacks. They no longer need to do that, and that's what I tried to convey in the deprecation message. Do you see what I mean?

I'll conclude by saying, this is really inconsequential, because hardly anyone uses the `fallbackHostsUseDefault` option, and when we remove it, whoever does use it will just start using the generated environment fallbacks and be none the wiser.

SimonWoolf: I agree with everything lewis said ^

PR #690, opened 2020-10-12
mention support for Web Worker context in README

owenpearson:


QuintinWillison (approved):

mattheworiordan:
I like the fact we mention you can run this as a web worker, but we've pushed a lot of complexity downstream now to the end-user, something we should avoid at all costs (Ably takes away complexity for developers).

For example, how do you manage the connection lifecycle when using a web worker?  How do you handle subscribe event handlers across multiple tabs etc.  

I think we should do more than this and include a trivial Javascript example which demonstrates how to share a connection across tabs.  Remember, no one generally asks us if we support web workers, instead they ask us if we can share a single connection across multiple tabs.  So this change hasn't helped our customers answer that question, and in fact our solution article explicitly states this is largely not possible and people should find another solution, see https://support.ably.com/a/solutions/articles/3000067035.

@owenpearson please can you open a new issue and address this problem more holistically, from the perspective of the customer with a problem (i.e. sharing a connection) and not the solution (i.e. web worker support).

PR #691, opened 2020-10-26
Replace nodeunit with Mocha

### LINKED ISSUE #645: Replace NodeUnit
It's been deprecated for nearly two years (at the time of writing).

Replacement options seem abundant and include [Jasmine](https://jasmine.github.io/), [AVA](https://github.com/avajs/ava), [Tape](https://github.com/substack/tape) and [Mocha](https://mochajs.org/) with possible outriders and related offerings like [Jest](https://jestjs.io/), [Puppeteer](https://github.com/puppeteer/puppeteer) and [QUnit](https://qunitjs.com/) also being worth considering as part of the mix.

owenpearson:
This change, if merged, will resolve #645. It replaces nodeunit with Mocha and Chai for tests running in Node, Karma, and the local test webserver script. It also updates travis to point towards the new tests.

One minor regression is that I've had to implement my own assertion counter and the implementation I've used is less strict in that it only checks assertions on calls to `counter.assert()` whereas nodeunit would do this automatically after each test. I'm open to suggestions on how to fix this...


spec/browser/connection.test.js
```diff
@@ -1,283 +1,294 @@
...
+		it('setup_realtime', function(done) {
+			counter.expect(1);
+			helper.setupApp(function(err) {
+				if(err) {
+					expect(false, helper.displayError(err));
+				} else {
+					expect(true, 'app set up');
+				}
+				counter.assert();
+				done();
```
lmars, line 47: Could we perhaps provide a wrapper around `it` that invokes the test function with a `test` object that has `test.expect` and `test.done` to avoid the need for both asserting the counter and calling done in every test?

spec/rest/defaults.test.js
```diff
@@ -1,258 +1,245 @@
...
+			expect(Defaults.getHost(normalisedOptions, 'sandbox-rest.ably.io', false)).to.deep.equal('sandbox-rest.ably.io');
+			expect(Defaults.getHost(normalisedOptions, 'sandbox-rest.ably.io', true)).to.deep.equal('sandbox-realtime.ably.io');
+
+			expect(Defaults.getPort(normalisedOptions)).to.equal(443);
+			counter.assert();
+			done();
+		});
+
+		/* init with given environment */
+		it('defaults_given_environment2', function(done) {
```
lmars, line 75: Let's call this `defaults_given_environment_and_fallbackHostsUseDefault` but keep the existing comment:

```
/* init with given environment and default fallbacks */
/* will emit a deprecation warning */
```

spec/rest/defaults.test.js
```diff
@@ -1,258 +1,245 @@
...
+			expect(normalisedOptions.restHost).to.equal('sandbox-rest.ably.io');
+			expect(normalisedOptions.realtimeHost).to.equal('sandbox-realtime.ably.io');
+			expect(normalisedOptions.port).to.equal(80);
+			expect(normalisedOptions.tlsPort).to.equal(443);
+			expect(normalisedOptions.fallbackHosts.sort()).to.deep.equal(Defaults.FALLBACK_HOSTS.sort());
+			expect(normalisedOptions.tls).to.equal(true);
+	
+			expect(Defaults.getHosts(normalisedOptions).length).to.deep.equal(4);
+			expect(Defaults.getHosts(normalisedOptions)[0]).to.deep.equal(normalisedOptions.restHost);
+			expect(Defaults.getHost(normalisedOptions).to.deep.equal('sandbox-rest.ably.io', false), 'sandbox-rest.ably.io');
```
lmars, line 88: This fails for me with:

```
  1) rest/defaults
       defaults_given_environment2:
     TypeError: Cannot read property 'deep' of undefined
      at Context.<anonymous> (spec/rest/defaults.test.js:88:49)
```

I think the assertions should be:

```js
expect(Defaults.getHost(normalisedOptions, 'sandbox-rest.ably.io', false)).to.deep.equal('sandbox-rest.ably.io');
expect(Defaults.getHost(normalisedOptions, 'sandbox-rest.ably.io', true)).to.deep.equal('sandbox-realtime.ably.io');
```

lmars (commented):
I've tested the updated test commands and they all work for me (i.e. they run some tests):

- `npm run test:mocha -- --test spec/realtime/auth.test.js` ✅ 
- `npm run test:webserver` plus visiting http://localhost:3000 ✅ 

Given the majority of changes are due to indentation, I've not thoroughly reviewed the code changes but if the tests that currently pass on `main` continue to pass then, other than the minor comments I've left, LGTM 👍

SimonWoolf (approved):

owenpearson:
OK so after realising a few mistakes in the original PR I've just force pushed a very different approach to this - rather than rewriting the tests individually I'm now injecting the Mocha API in through the `test` object which now serves as an adapter for using the Mocha API to run nodeunit tests. This means that the tests code needs only minor changes and should make this PR much more reviewable.

A few notes:
- `test.equal` now uses strict equality by default, abstract equality can still be tested using `test.ok(val1 == val2)` but most of the tests work fine with strict equality anyway.
- All of the 'setup' tests now run as 'before' hooks, the tests will still fail if an assertion inside the hook fails. This should pave the way for running individual tests (Mocha has a `--grep testName` option).
- Mocha will throw an `AssertionError` when an assertion fails where nodeunit would continue the test and report failed assertions after completion. This shouldn't cause any differences in the results of tests but might stop a failing test from exiting gracefully by leaving handlers active after the test is complete - this was the case for `no_connection_lifecycle` in `spec/realtime/failure.test.js` hence my changes to that test.

lmars:
@owenpearson 

> This means that the tests code needs only minor changes and should make this PR much more reviewable.

Nice! It's great when that happens 😄

lmars (approved):

PR #692, opened 2020-10-29
DevicePushDetails should contain error not errorReason

cruickshankpg:
The DevicePushDetails struct was initially made with an `errorReason` field that was never filled by the server. `error` is now being sent from the frontends.

### DIFF
--- a/common/lib/types/devicedetails.js
+++ b/common/lib/types/devicedetails.js
@@ -77,6 +77,7 @@ var DeviceDetails = (function() {
 	};
 
 	DeviceDetails.fromValues = function(values) {
+		values.error = values.error && ErrorInfo.fromValues(values.error); 
 		return Utils.mixin(new DeviceDetails(), values);
 	};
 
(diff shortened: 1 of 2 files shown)


common/lib/types/devicedetails.js
```diff
@@ -54,8 +54,8 @@ var DeviceDetails = (function() {
 			result += '; push.recipient=' + JSON.stringify(this.push.recipient);
 		if(this.push.state)
 			result += '; push.state=' + this.push.state;
-		if(this.push.errorReason)
-			result += '; push.errorReason=' + this.push.errorReason;
+		if(this.push.error)
+			result += '; push.error=' + JSON.stringify(this.push.error);
```
SimonWoolf, line 58: Ideally `this.push.error` should be parsed into an `ErrorInfo` object on receipt, ala https://github.com/ably/ably-js/blob/1aa8bfcb8f04b579c74303ebd6abbe8347cb35dd/browser/lib/transport/xhrrequest.js#L214 , so then ErrorInfo#toString() has nice output so you shouldn't need the json.stringify, and it's an instance of the right class when passed to the user.

(in practice the errorinfo class doesn't currently have much in the way of extra behaviour (eg other methods that could be called on it) other than nice toString output, so it doesn't matter much. But, y'know, one day it could)

SimonWoolf (approved):

owenpearson (approved):

Issue #694, opened 2020-11-03
Rewind examples don't compile with Typescript


jscheid:
In your docs you have examples such as [this one](https://github.com/ably/docs/blob/baa0a4666079abba3a3e19e82eb99ca8b8a735d0/content/realtime/channels.textile#setting-channel-options):

```js
const realtime = new Ably.Realtime('{{API_KEY}}');
const channelOptions = {
  params: {
    delta: 'vcdiff',
    rewind: '1'
  }
};
const channel = realtime.channels.get('{{RANDOM_CHANNEL_NAME}}', channelOptions);
```

But your Typescript types [require all of `params`, `cipher`, and `mode` to be set](https://github.com/ably/ably-js/blob/1aa8bfcb8f04b579c74303ebd6abbe8347cb35dd/ably.d.ts#L255-L259) and so you end up with an error like

```
Argument of type '{ params: { rewind: string; }; }' is not assignable to parameter of type 'ChannelOptions'.
  Type '{ params: { rewind: string; }; }' is missing the following properties from type 'ChannelOptions': cipher, modes [2345]
```

SimonWoolf:
Thanks for the bug report. Fixed by https://github.com/ably/ably-js/pull/695

owenpearson:
@jscheid We've released the fixed type definitions as [1.2.4](https://github.com/ably/ably-js/releases/tag/1.2.4)

PR #695, opened 2020-11-03
Typings: lots of type definition fixes

SimonWoolf:
Fixes https://github.com/ably/ably-js/issues/694

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -253,9 +253,9 @@ declare namespace Types {
 	type ChannelModes = Array<ChannelMode>;
 
 	interface ChannelOptions {
-		cipher: any;
-		params: ChannelParams;
-		modes: ChannelModes;
+		cipher?: {key: ArrayBuffer | Uint8Array | string} | CipherParams;
+		params?: ChannelParams;
+		modes?: ChannelModes;
 	}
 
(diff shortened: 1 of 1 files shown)


owenpearson (commented):
LGTM, but should we also change the key property on `CipherParams` to the same union type?

SimonWoolf:
> should we also change the key property on CipherParams to the same union type?

No, in an actual CipherParams instance (as you get by calling `getDefaultParams`) the `key` type is fixed, either a Buffer on node or a Wordarray in browsers. When creating a CipherParams (either using `getDefaultParams` or the short form cipher options) you can supply it as an arraybuffer, base64-encoded string etc. and it'll convert it as appropriate.

...Now I come to look for it, the typings are also missing `Crypto.getDefaultParams`. So that should probably be added too. And the typings for `Crypto.generateRandomKey` are wrong, it claims it'll return a string. Aaargh, I'll fix some of these. I should read through them start to finish at some point and see what other errors are in there..

SimonWoolf:
@owenpearson I went ahead and fixed a whole bunch of type issues I spotted while doing the initial change, so this is now a somewhat bigger pr, so new deeper review needed - thanks

owenpearson (approved):

PR #704, opened 2020-12-01
Use ES6 Modules

### LINKED ISSUE #477: Convert library to ES6 modules; use webpack for creating bundles
`import * as Ably from 'ably'` (I'm using TypeScript) yields the follow stats in my Webpack build:

> Size: 369.66kb
> Parsed size: 151.46kb
> Gzipped size: 40.88kb

This makes Ably the single largest dependency in our application. For comparison, the alternative I'm evaluating is roughly 35% of all those numbers.

If Ably exported an ES2015 modules target using the `module` field in package.json, then modern tooling like Webpack could try to only bundle the minimum my application requires. I'm not sure how effective this might be, but those size stats could use some help.

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-296) by [Unito](https://www.unito.io/learn-more)

owenpearson:
Resolves #477.

Most of the changes here are `import` and `export` statements, pretty much all of the important stuff is in `webpack.config.js` which makes heavy use of webpack's `resolve.alias` config option. This allows us to import from aliases and resolve them to different modules for separate builds. `null-loader` is also used when we want a module to return null. Closure compiler is no longer being used to minify code in favour of `Terser` (default webpack minifier) but is still included to check the build output's ES level and now runs as a GitHub Actions job.


SimonWoolf (approved):
Looks good to me 🙂

Couple minor comments:
- Your text editor or something is generating invalid text files -- text files [must end in a newline to be well-formed](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206), these don't (hence the red warning icons in github)
- Now that webpack is wrapping all the modules in IIFEs as part of transpilation, we no longer need to have IIFEs in the source files, so we can get rid of those and make them look a bit cleaner

tagging @paddybyers in case he wants to chip in as ex-ex-maintainer

SimonWoolf:
> use webpack v4 instead of v5 

if there's a particular reason we can't use the latest version, could we make sure that's written down somewhere (either in a github issue, or in the commit message on that commit, or both)

.eslintrc.json
```diff
@@ -0,0 +1,14 @@
+{
+  "env": {
+    "es6": true,
+    "node": true,
+    "browser": true
+  },
+  "rules": {
```
sacOO7, line 7: You can also add error for console.log statements ( if we are already using logger )

.eslintrc.json
```diff
@@ -0,0 +1,14 @@
+{
```
sacOO7, line 1: Not sure if this config. adds default eslint recommended rules (https://eslint.org/docs/rules/)
You might like to add 
  ```
"extends": [
    "eslint:recommended"
  ]
```

File: package.json [hunk 3]
```diff
@@ -83,6 +88,8 @@
     "test:webserver": "grunt test:webserver",
     "concat": "grunt concat",
     "build": "grunt build",
-    "requirejs": "grunt requirejs"
+    "requirejs": "grunt requirejs",
+    "lint": "eslint nodejs/**/*.js common/**/*.js browser/lib/**/*.js",
```
sacOO7, line 92: You can also add one more command for ```lint:fix``` using --fix flag for fixing possible linting errors.

.eslintrc.json
owenpearson: Hi Sachin, I definitely intend to add this in the near future, but I'm leaving it out of this PR - The reason being we intend to migrate to typescript and ES6 syntax soon and that will make it a lot easier to write code that passes a ruleset like `eslint:recommended` (or even `airbnb-base`).

sacOO7: Yeah, makes sense ( I would also suggest airbnb styling for default linting rules)

AlexBHarley:
So excited this is landing! Any numbers on bundle size reduction?

Could you publish a beta or release candidate for this @owenpearson @SimonWoolf? I want to get using it ASAP.

SimonWoolf:
> Could you publish a beta or release candidate for this @owenpearson @SimonWoolf? I want to get using it ASAP.

Beta release is available at `npm install ably@1.2.5-beta`, feedback welcome 🙂

AlexBHarley:
I'm seeing a size increase from ~96kB (1.2.4) to ~105kB (1.2.5-beta.1). My import statement looks like this:

```
    const ably = await import(
      // @ts-ignore
      'ably/browser/static/ably-commonjs'
    );
```

Is that the right package to be importing?

owenpearson:
> I'm seeing a size increase from ~96kB (1.2.4) to ~105kB (1.2.5-beta.1). My import statement looks like this:
> 
> ```
>     const ably = await import(
>       // @ts-ignore
>       'ably/browser/static/ably-commonjs'
>     );
> ```
> 
> Is that the right package to be importing?

Yeah that import is fine. It may have increased slightly in size, my best guess is that this is down to Webpack adding some new polyfills but I'll have a look and see if I can shed that extra 10kb if possible. It's worth noting that if you're not using encryption we have a smaller 'noencryption' build ('ably/browser/static/ably-commonjs.noencryption') that should be about 20kb lighter.

Unfortunately bundle size isn't going to change much in this release, and since pretty much all of the code is necessary to call the realtime constructor I don't think we'll be able to reduce bundle size without releasing a 'lightweight' version with some features omitted.

AlexBHarley:
Gotcha OK thanks for clarifying. Found this PR after coming from the bundle size issue. I guess this is just a precursor to work  reducing that.

PR #719, opened 2021-03-09
Comet: Raise preconnect event if the server responds with a protocol message

cruickshankpg:
Even if the message is an error.

There was a realtime CI regression on upgrading to ably-js 1.2.5 (from 1.2.1). Rate enforcement errors for comet were being swallowed by the client and a no hosts error being raised in response.  On realtime sending a response of:
```
ProtocolMessage; action=DISCONNECTED; error=[ErrorInfo: Maximum account-wide instantaneous connections rate exceeded; permitted rate = 1; metric = connections.maxRate; statusCode=429; code=80021; see https://help.ably.io/error/80021 ]
```
We expected the connection to be disconnected with error:
```
[ErrorInfo: Maximum account-wide instantaneous connections rate exceeded; permitted rate = 1; metric = connections.maxRate; statusCode=429; code=80021;]
```
but instead saw:
```
[ErrorInfo: Unable to connect (and no more fallback hosts to try); statusCode=404; code=80003]
```

This is because the `preconnect` event was no longer being raised in error cases as a result of https://github.com/ably/ably-js/pull/698. That PR was a sensible fix for the case where the host is unreachable but if the host does respond then the `preconnect` event should be raised. I've used `code` being set in the `ErrorInfo` to discern this as it seems to be used elsewhere but it's a bit subtle for my taste so I'm happy for you to suggest more elaborate changes.

Ably js logs of failing test:
```
20:26:06.279 Ably: Transport.onProtocolMessage(): received on comet: [ProtocolMessage; action=DISCONNECTED; error=[ErrorInfo: Maximum account-wide instantaneous connections rate exceeded; permitted rate = 1; metric = connections.maxRate; statusCode=429; code=80021; see https://help.ably.io/error/80021 ]]; connectionId = undefined
20:26:06.279 Ably: Transport.onDisconnect(): err = [ErrorInfo: Maximum account-wide instantaneous connections rate exceeded; permitted rate = 1; metric = connections.maxRate; statusCode=429; code=80021; see https://help.ably.io/error/80021 ]
20:26:06.279 Ably: ConnectionManager.tryATransport(): transport comet disconnected, err: [ErrorInfo: Maximum account-wide instantaneous connections rate exceeded; permitted rate = 1; metric = connections.maxRate; statusCode=429; code=80021; see https://help.ably.io/error/80021 ]
20:26:06.279 Ably: ConnectionManager.notifyState(): new state: disconnected
20:26:06.279 Ably: ConnectionManager.cancelTransitionTimer():
20:26:06.280 Ably: Connection state: disconnected; reason: [ErrorInfo: Unable to connect (and no more fallback hosts to try); statusCode=404; code=80003]
20:26:06.280 Ably: ConnectionManager.enactStateChange: setting new state: disconnected; reason = Unable to connect (and no more fallback hosts to try)
```

Now Passing:
```
10:01:15.958 Ably: Transport.onProtocolMessage(): received on comet: [ProtocolMessage; action=DISCONNECTED; error=[ErrorInfo: Maximum account-wide instantaneous connections rate excee
ded; permitted rate = 1; metric = connections.maxRate; statusCode=429; code=80021; see https://help.ably.io/error/80021 ]]; connectionId = undefined
10:01:15.958 Ably: Transport.onDisconnect(): err = [ErrorInfo: Maximum account-wide instantaneous connections rate exceeded; permitted rate = 1; metric = connections.maxRate; statusCo
de=429; code=80021; see https://help.ably.io/error/80021 ]
10:01:15.958 Ably: ConnectionManager.deactivateTransport(): transport = NodeCometTransport; uri=https://local-rest.ably.io:8081/comet/; isConnected=false; format=undefined; stream=tru
e
10:01:15.958 Ably: ConnectionManager.deactivateTransport(): state = disconnected; was pending
10:01:15.958 Ably: ConnectionManager.deactivateTransport(): reason =  Maximum account-wide instantaneous connections rate exceeded; permitted rate = 1; metric = connections.maxRate
10:01:15.958 Ably: ConnectionManager.notifyState(): new state: disconnected
10:01:15.958 Ably: ConnectionManager.cancelTransitionTimer(): 
10:01:15.958 Ably: Connection state: disconnected; reason: [ErrorInfo: Maximum account-wide instantaneous connections rate exceeded; permitted rate = 1; metric = connections.maxRate; 
statusCode=429; code=80021; see https://help.ably.io/error/80021 ]
10:01:15.958 Ably: ConnectionManager.enactStateChange: setting new state: disconnected; reason = Maximum account-wide instantaneous connections rate exceeded; permitted rate = 1; metr
ic = connections.maxRate
10:01:15.958 Ably: CometTransport.dispose():
10:01:15.958 Ably: ConnectionManager.disconnectAllTransports(): Disconnecting all transports
```

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -14,5 +14,5 @@ jobs:
         with:
           node-version: 14.x
       - run: npm ci
-      - run: npm run lint
+      - run: npm run build
       - run: npm run check-closure-compiler
--- a/common/lib/transport/connectionerror.js
+++ b/common/lib/transport/connectionerror.js
@@ -42,12 +42,13 @@ ConnectionError.isRetriable = function(err) {
 	if (!err.statusCode || !err.code || err.statusCode >= 500) {
 		return true;
 	}
-	for (const [{code}] of ConnectionError.values()) {
-		if (code == err.code) {
-			return true;
+	var retriable = false;
+	Utils.valuesArray(ConnectionError).forEach(function(connErr) {
+		if (connErr.code && connErr.code == err.code) {
+			retriable = true;
 		}
-	}
-	return false;
+	});
+	return retriable;
 };
 
 export default ConnectionError;
--- a/common/lib/transport/connectionerror.js
+++ b/common/lib/transport/connectionerror.js
@@ -38,4 +38,16 @@ var ConnectionError = {
 	})
 };
 
+ConnectionError.isRetriable = function(err) {
+	if (!err.statusCode || !err.code || err.statusCode >= 500) {
+		return true;
+	}
+	for (const [{code}] of ConnectionError.values()) {
+		if (code == err.code) {
+			return true;
+		}
+	}
+	return false;
+};
+
 export default ConnectionError;
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -59,8 +59,8 @@ module.exports = function (grunt) {
 				/* By default, the compiler assumes you're using es6 and transpiles to
 				 * es3, adding various (unnecessary and undesired) polyfills. Specify
 				 * both in and out to es3 to avoid transpilation */
-				language_in: 'ECMASCRIPT3',
-				language_out: 'ECMASCRIPT3',
+				language_in: 'ECMASCRIPT5',
+				language_out: 'ECMASCRIPT5',
 				strict_mode_input: true,
 				checks_only: true,
 				warning_level: 'QUIET'
@@ -92,6 +92,7 @@ module.exports = function (grunt) {
 	]);
 
 	grunt.registerTask('check-closure-compiler', [
+		'build',
 		'closureCompiler:ably.js'
 	]);
 
--- a/common/lib/transport/connectionmanager.js
+++ b/common/lib/transport/connectionmanager.js
@@ -328,7 +328,7 @@ var ConnectionManager = (function() {
 					self.notifyState({state: 'failed', error: wrappedErr.error});
 					callback(true);
 				} else if(wrappedErr.event === 'disconnected') {
-					if(wrappedErr.error.code && wrappedErr.error.statusCode > 400 && wrappedErr.error.statusCode < 500) {
+					if(!ConnectionError.isRetriable(wrappedErr)) {
 						/* Error received from the server that does not call for trying a fallback host, eg a rate limit */
 						self.notifyState({state: self.states.connecting.failState, error: wrappedErr.error});
 						callback(true);
--- a/common/lib/transport/connectionmanager.js
+++ b/common/lib/transport/connectionmanager.js
@@ -330,7 +330,7 @@ var ConnectionManager = (function() {
 				} else if(wrappedErr.event === 'disconnected') {
 					if(wrappedErr.error.code && wrappedErr.error.statusCode < 500) {
 						/* Error received from the server that does not call for trying a fallback host, eg a rate limit */
-						self.notifyState({state: 'disconnected', error: wrappedErr.error});
+						self.notifyState({state: self.states.connecting.failState, error: wrappedErr.error});
 						callback(true);
 					} else {
 						/* Error with that transport only; continue trying other fallback hosts */
--- a/common/lib/transport/connectionmanager.js
+++ b/common/lib/transport/connectionmanager.js
@@ -328,7 +328,7 @@ var ConnectionManager = (function() {
 					self.notifyState({state: 'failed', error: wrappedErr.error});
 					callback(true);
 				} else if(wrappedErr.event === 'disconnected') {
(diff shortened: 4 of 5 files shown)


SimonWoolf (changes requested):
> but it's a bit subtle for my taste so I'm happy for you to suggest more elaborate changes.

So my worry with this as-is is that for the case where you get a proper errorinfo response with a 5xxxx error code (which the router occasionally does eg if it can't reach realtime). In that case we want it to be treated as a broken endpoint, the same like a network error (per [rtn17d](https://docs.ably.io/client-lib-development-guide/features/#RTN17d)) - that is, so it immediately tries a fallback endpoint to get another region. But with this change it won't do that any more.

You could make the check `(!err || (err.code && err.statusCode < 500))`, but would like to avoid putting even more logic into the decision of whether to emit a `preconnect` event unless absolutely necessary. Plausibly a more natural home for that logic is in connectionmanager in tryATransport, expanding the wrappedErr-disconnected clause into eg something like
```
} else if(wrappedErr.event === 'disconnected') {
	if(wrappedErr.error.code && wrappedErr.error.statusCode < 500) {
		/* Error received from the server that does not call for trying a fallback host, eg a rate limit */
		self.notifyState({state: 'disconnected', error: wrappedErr.error});
		callback(true);
	} else {
		/* Error with that transport only; continue trying other fallback hosts */
		callback(false);
	}
}
```
Which is a bit clearer and more explicit.

So with that you can get rid of the other change, as the no-preconnect path now explicitly allows for errors which are not `failed` but which still don't necessitate using fallback hosts.

(Also: I'd like to generally apologise for this code, it was 4 years ago, I'd not done much javascript before, it had to work on IE8, and the status quo ante was already quite convoluted, and the result is definitely not the easiest-to-read code in the world, so sorry about that)

SimonWoolf:
(btw, quick additional comment: not a big deal but generally we don't have compiled artifact changes in the branch you push for PR, just cause that makes the PR harder to review as it's cluttered up with autogenerated files. cf https://github.com/ably/ably-js/blob/main/CONTRIBUTING.md#contributing )

cruickshankpg:
Ok I've tried the change you've suggested and the comet limit test is passing

SimonWoolf (approved):

owenpearson (approved):
LGTM, thanks

cruickshankpg:
@SimonWoolf I've had to check whether it's a 400 error to handle all the local errors raised by the client which isn't great

SimonWoolf:
yeah, not ideal - in particular if we do get a genuine 400 from the server that shouldn't trigger the fallback logic.

Looks like the comet transports use the convention that network errors have a statusCode of 400 and no `code`, so you know that if you have a code then that was actually sent by the server. This is also semantically not great as far as the '400' goes, but I can't think of a good answer to that; I guess 400 was chosen to avoid implying we got a 503 or something from the server if it's actually just a client network issue. So one answer might be to change the `ErrorInfo(msg, 80003, 400)` that the websocket transport spits out for network errors for a `ErrorInfo(msg, null, 400)`, to be consistent with the convention used by comet. wdyt?

SimonWoolf (changes requested):
👍 Fine with the approach.

Though it won't pass the linter as-is due to all the ES6 features, as you'll see if you run `grunt`, so you'll need to tweak those. (I believe there's a long-term plan to switch to typescript and transpile rather than just have it written in ES3, but we're not there yet).

.github/workflows/check.yml
```diff
@@ -14,5 +14,6 @@ jobs:
         with:
           node-version: 14.x
       - run: npm ci
+      - run: npm run build
```
SimonWoolf, line 17: Adding this here doesn't really solve the problem, someone running `grunt check-closure-compiler` locally would still have the problem. Need to make `check-closure-compiler` run `build` first -- add it as the first element of the array here https://github.com/ably/ably-js/blob/main/Gruntfile.js#L95

common/lib/transport/connectionerror.js
```diff
@@ -38,4 +38,17 @@ var ConnectionError = {
 	})
 };
 
+ConnectionError.isRetriable = function(err) {
+	if (!err.statusCode || !err.code || err.statusCode >= 500) {
+		return true;
+	}
+	var retriable = false;
+	ConnectionError.keys().forEach(function(connErr) {
```
SimonWoolf, line 46: > `ConnectionError.keys().forEach(function(connErr) {...`

That won't work for a couple reasons (plain objects don't have a `keys()` function, and anyway you want the values not the keys). You want something like `Utils.valuesArray(ConnectionError).forEach(function(connErr) { ...`

(I think IE9 probably has Array#forEach? double check, if it doesn't use `Utils.arrForEach`)

SimonWoolf (changes requested):

PR #721, opened 2021-03-12
Ignore anything in the browser folder when searching within VS Code

QuintinWillison:
Until we fix the build not to require this folder at all, this tweak makes the developer UX easier when using VS Code (which most of us do) to search for code in this repositroy.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "files.exclude": {
+        "/browser/**": true
+    }
+}


QuintinWillison:
@owenpearson will any of your recent work obsolete this pull request?
if not, perhaps you would be happy to approve it?

owenpearson:
@QuintinWillison Yeah the PR for removing these files from Git will invalidate this (and effectively do what this PR would have done but for every other text editor too).

PR #725, opened 2021-03-24
Remove build output from git

owenpearson:
Removes build output files from the git repository (these will still be packaged in NPM releases)


package.json
```diff
@@ -7,7 +7,7 @@
     "url": "https://github.com/ably/ably-js/issues",
     "email": "support@ably.io"
   },
-  "main": "./nodejs/index.js",
+  "main": "./browser/static/ably-node.js",
```
SimonWoolf, line 10: Why this change?
Also this will presumably break the react-native and browser builds below unless you change the main-script alias for them to match this change

SimonWoolf (approved):
👍 with comment

package.json
owenpearson: The reason for this to change was so that I could exclude the `nodejs` directory from the package entirely. It makes sense to me to have a clear distinction between 'source code' and 'build output' so this change means the entire `nodejs` directory is now just source code and paves the way for us to do something like a top level `src` and `dist` folder structure in future. You're right about me needing to change the aliases for rn and browser so I've force pushed that change now.

owenpearson:
FYI @SimonWoolf I've added on another commit with a `prepare` script which just runs the build script. Seems to work fine even without any of the source code in the package.json `files` so people can still install from GH if they need and we still get a slimmer NPM bundle.

SimonWoolf:
> Seems to work fine even without any of the source code in the package.json files so people can still install from GH if they need and we still get a slimmer NPM bundle.

Nice 👍

My assumption is that `prepare` won't run if you install the package from npm. Might be worth testing that assumption by doing a beta release to npm and checking you don't get any errors? If that assumption is wrong then can always make the script conditional, eg `"prepare": "[ -e browser/static/ably.js ] && npm run build"` or something

owenpearson:
> Might be worth testing that assumption by doing a beta release to npm, then npm installing it somewhere and checking you don't get any errors?

Have tested this on NPM with `1.2.7-beta.1` and it works fine :+1:

package.json
```diff
@@ -91,6 +89,7 @@
     "requirejs": "grunt requirejs",
     "lint": "eslint nodejs/**/*.js common/**/*.js browser/lib/**/*.js",
     "lint:fix": "eslint --fix nodejs/**/*.js common/**/*.js browser/lib/**/*.js",
-    "check-closure-compiler": "grunt check-closure-compiler"
+    "check-closure-compiler": "grunt check-closure-compiler",
+	"prepare": "npm run build"
```
QuintinWillison, line 93: Spacing appears off for this line.

QuintinWillison (approved):
Just one minor niggle. Otherwise, looks good and a great leap forward! 🚀

package.json
owenpearson: Fixed: ba77f09.

PR #726, opened 2021-03-24
Fix faulty import of JSONP transport to React Native and NativeScript

owenpearson:
Resolves an issue which would cause the library to throw an error on import in React Native and NativeScript.

To summarise, the issue was that each platform was using an array of 'Transports' and explicitly initialising each one as soon as the library is loaded, but it was also using `null-loader` to omit the JSONP platform from React Native and NativeScript builds so it was essentially attempting to initialise an empty module as if it were a Transport. This PR adds a separate array of transports without JSONP to fix this error.

This PR would also allow React Native users to use the `ably` NPM module directly since the recent work around ES6 modules displaces the need for the `ably-js-react-native` wrapper.

### DIFF
--- a/browser/lib/transport/withoutjsonp.js
+++ b/browser/lib/transport/withoutjsonp.js
@@ -1,3 +1,6 @@
+/**
+ * This file exists for React Native and Nativescript in order to exclude the unsupported JSONP transport from these platforms.
+ */
 import XHRPollingTransport from './xhrpollingtransport';
 import XHRStreamingTransport from './xhrstreamingtransport';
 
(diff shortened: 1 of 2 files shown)


SimonWoolf (approved):

browser/lib/transport/withoutjsonp.js
```diff
@@ -0,0 +1,7 @@
+import XHRPollingTransport from './xhrpollingtransport';
```
QuintinWillison, line 1: Could you consider adding some commentary to the top of this file to explain why it's here and what it does for the wider project as a whole? It's a bit opaque right now.

QuintinWillison (changes requested):

browser/lib/transport/withoutjsonp.js
owenpearson: Good point, I've added a header in abeda9a.

QuintinWillison: Thanks. Perfect.

QuintinWillison (approved):

PR #731, opened 2021-03-26
Fix request typings

### LINKED ISSUE #728: REST client: Type Error in Types.HttpPaginatedResponse.items
When using the REST client to get channel occupancy metrics as documented here:
https://ably.com/documentation/rest/channel-status#metadata-rest

```ts
restClient
    .request('GET', `/channels/${channel}`, { params: { occupancy: 'metrics' } })
    .then((channelData) => {
        // @ts-expect-error incorrect type
        return channelData.items[0]?.status.occupancy.metrics.connections as number;
    });
```

`channelData.items` is typed as `Types.HttpPaginatedResponse.items: string[]` when the api is actually returning an object with this data

https://ably.com/documentation/realtime/channel-metadata#channel-details
```json
{
  "channelId": "foo",
  "status": {
    "isActive": true,
    "occupancy": {
      "metrics": {
        "connections": 1,
        "publishers": 1,
        "subscribers": 1,
        "presenceConnections": 1,
        "presenceMembers": 0,
        "presenceSubscribers": 1
      }
    }
  }
}
```

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-675) by [Unito](https://www.unito.io/learn-more)

owenpearson:
Resolves #728.

Also adds a generic type parameter for `Rest.request` to declare the response type for convenience.

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -652,7 +652,6 @@ declare namespace Types {
 	}
 
 	class HttpPaginatedResponse extends PaginatedResult<any> {
-		items: string[];
 		statusCode: number;
 		success: boolean;
 		errorCode: number;
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -401,7 +401,7 @@ declare namespace Types {
 		static Callbacks: typeof Types.RestCallbacks;
 		auth: Types.AuthCallbacks;
 		channels: Types.Channels<Types.ChannelCallbacks>;
-		request: (method: string, path: string, params?: any, body?: any[] | any, headers?: any, callback?: Types.StandardCallback<Types.HttpPaginatedResponse>) => void;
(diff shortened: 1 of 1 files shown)


ably.d.ts
```diff
@@ -651,8 +651,7 @@ declare namespace Types {
 		isLast: () => boolean;
 	}
 
-	class HttpPaginatedResponse extends PaginatedResult<any> {
-		items: string[];
+	class HttpPaginatedResponse<T = any> extends PaginatedResult<T> {
 		statusCode: number;
```
SimonWoolf, line 655: You accidentally deleted the `items` line instead of changing it to `items: T[];`..?

owenpearson: `PaginatedResult` actually has it's own `items` property which is already typed as `T[]` so deleting this line has the same effect but without duplication

SimonWoolf: ah 👍

QuintinWillison:
I've not got enough TypeScript knowledge to make an informed approval on this pull request, so I think just ensure that @SimonWoolf is happy. Thanks!

PR #734, opened 2021-03-31
Fix bugs in disconnection error filtering

cruickshankpg:
I must not have retested against realtime after making the last set of markups on my previous PR as I have found the code had a couple of bugs.


owenpearson (approved):
LGTM

PR #738, opened 2021-04-06
Add Playwright tests

owenpearson:
Based off #739 for compatibility.

Adds support for running the browser tests in Playwright. 

`playwrightSetup.js` sets up Mocha with a custom reporter which dispatches `CustomEvent`s at test lifecycle hooks.

`web_server` has been changed to send `playwright.html` instead of `mocha.html` when the `PLAYWRIGHT_TEST` env var is truthy. `playwright.html` is almost the same as `mocha.html` but includes `playwrightSetup.js` to set up Mocha. the reason for this separation is that the browser compatibility of `CustomEvent` is poor so `playwrightSetup.js` should only be included for playwright tests and not, for example, when testing on IE. It would also work to use some javascript which tests whether it's running inside a Playwright-launched browser and decide whether to use the custom reporter based on the result but I couldn't find a reliable way to do so.

`runPlaywrightTests` Uses `child_process` to launch the `test_webserver` (with the `PLAYWRIGHT_TEST` env var) and navigates to it with Playwright. It listens to the events emitted by the custom event reporter in `playwrightSetup.js` and writes test results to the NodeJS console. It also exits with code 1 if the any tests have failed. It also reads to the `PLAYWRIGHT_BROWSER` env var and tests on the specified browser if set.


spec/web_server
```diff
@@ -15,14 +15,18 @@ var express = require('express'),
 var server = express();
 
 server.use(function(req, res, next) {
-  console.log('%s %s %s', req.method, req.url, req.path);
+  if (!process.env.PLAYWRIGHT_TEST) console.log('%s %s %s', req.method, req.url, req.path);
```
QuintinWillison, line 18: I notice that you read this `env` variable twice in this file and I wonder if you couldn't perhaps raise it up as something like `const isPlaywright = !!process.env.PLAYWRIGHT_TEST` earlier on, maybe?

Also, is this environment variable name documented anywhere? It's not clear to me whether this is a name that you came up with or perhaps one that we're inheriting from a dependency. 🤔

spec/support/runPlaywrightTests.js
```diff
@@ -0,0 +1,56 @@
...
+				onTestResult({ type, detail });
+			});
+		});
+	});
+};
+
+(async () => {
+	try {
+		if (process.env.PLAYWRIGHT_BROWSER) {
+			await runTests(playwright[process.env.PLAYWRIGHT_BROWSER]);
```
QuintinWillison, line 47: Similar to my other comment in respect of the `PLAYWRIGHT_TEST` environment variable, is this documented anywhere?

QuintinWillison (changes requested):
I think overall there could be more commentary, at a bare minimum at class level, to provide higher level explanation as to what the moving parts are here. For me it's not entirely obvious without a lot of scrolling up and down the files in this pull request, which suggests it would be even less obvious when trying to work it out from looking at the entire codebase without the context of seeing just this change in isolation.

spec/support/runPlaywrightTests.js
```diff
@@ -0,0 +1,56 @@
+const playwright = require('playwright');
```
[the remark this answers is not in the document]
owenpearson: Thanks, sorted in 8a86795

spec/web_server
owenpearson: i've documented the env vars a bit better in 9e5c68e

QuintinWillison (approved):

PR #739, opened 2021-04-07
Improve Mocha tests

owenpearson:


QuintinWillison (approved):
Happy to land this though I'm a little troubled by https://github.com/ably/ably-js/commit/63e0d5fd4dc2b54e9ca21ff5bad92bcbf4eb8bb6 as it seems like hammer to crack a nut and not immediately obviously related to the scope of this pull request.

Also worth noting that the package lock file version has increased meaning a lot of noise in the diff there, which could perhaps have impacted the `npm ci` vs `npm install` thing. 🤷‍♂️ 

I've not spent long looking at https://github.com/ably/ably-js/commit/106a117e1f42679eefeb307dae3f6bd04765584d as it is, indeed, massive! I'll trust you that the tests that were running before this change are still running.

PR #740, opened 2021-04-07
Implement Ably-Agent connection param for ably-js and NodeJS versions

### LINKED ISSUE #737: Implement RSC7d (Ably-Agent header)
see [PR](https://github.com/ably/docs/pull/1034) and [spec point](https://ably-docs.herokuapp.com/client-lib-development-guide/features/#RSC7d) for context.

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-680) by [Unito](https://www.unito.io)

owenpearson:
Resolves #737

Adds the `Ably-Agent` header for Rest requests and the `agent` param for realtime requests. At the moment this will only include the version of `ably-js` being used and, if applicable, the version of NodeJS from `process.version`. It might be worth discussing in future whether there are appropriate product versions to add for browser, react native, and nativescript.


common/lib/transport/connectionmanager.js
```diff
@@ -101,6 +101,7 @@ var ConnectionManager = (function() {
 		}
 		params.v = Defaults.apiVersion;
 		params.lib = Defaults.libstring;
+		params.agent = Defaults.agent;
```
SimonWoolf, line 104: Per RSC7b this is supposed to replace the lib param & header, surely we shouldn't send both at the same time?

common/lib/util/defaults.js
```diff
@@ -39,6 +39,13 @@ Defaults.version          = '1.2.8';
 Defaults.libstring        = Platform.libver + '-' + Defaults.version;
 Defaults.apiVersion       = '1.2';
 
+var agent = 'ably-js/' + Defaults.version;
```
SimonWoolf, line 42: Given that this is supposed to replace lib, seems very odd we'd include less information than that. From the libstring I can tell whether the library is browser, react-native, nativescript etc., this drops that information. If this is to help us decide whether to drop support for things, that's moving in the wrong direction, surely.

Is this why you left the lib param in? If so I'm unconvinced - having two platfrom version strings in different formats where the information you need is spread across both of them is not great, if we're going to switch formats we should actually make the new one a replacement for the old, if we're not going to do that we shouldn't switch formats

owenpearson: I've now updated the agent param to make use of the existing libstring (2b44cdf), and removed the lib param completely (e5c8527).

SimonWoolf (approved):

nodejs/lib/util/defaults.js
```diff
@@ -9,7 +9,8 @@ var Defaults = {
 	baseTransportOrder: ['comet', 'web_socket'],
 	transportPreferenceOrder: ['comet', 'web_socket'],
 	upgradeTransports: ['web_socket'],
-	restAgentOptions: {maxSockets: 40, keepAlive: true}
+	restAgentOptions: {maxSockets: 40, keepAlive: true},
+	agent: 'nodejs/' + process.version
```
QuintinWillison, line 13: We need to document these agent identifiers as we add them. I've started a document (internal for now) so we can pull this information together, which you can find [here](https://docs.google.com/document/d/1duUdRknI1QNkanASgkTYo19LGXQIeT8oFlXS7TCU-Ek/edit?usp=sharing).

Also, looking at the docs for [process.version](https://nodejs.org/api/process.html#process_process_version) it appears that you'll be emitting a string that has a 'v' prefix. That would be superfluous and probably technically wrong if we decide (as I think could be the case) that this second component of each product added to the agent should conform to something sensible like [SemVer](https://semver.org/). This needs stronger specification and I would like to at least lay that down in the document I referenced earlier in this comment.

common/lib/util/defaults.js
QuintinWillison: I would like to explore this further as I think the updates that @owenpearson has made in response to @SimonWoolf's comments may be too simplistic. We only get one shot at this and I want to aim for maximum flexibility - so, as such, it would appear to me that a discrete agent product component purely to indicate the `ably-js` source codebase version may be more helpful, where it's then augmented by an additional agent product component(s) telling us about the runtime environment and/or wrapper.

I'm going to accumulate thoughts on this in [this internal document](https://docs.google.com/document/d/1duUdRknI1QNkanASgkTYo19LGXQIeT8oFlXS7TCU-Ek/edit?usp=sharing) for now.

QuintinWillison (changes requested):
I also think that we should have tests that verify that the agent string includes certain keywords based on the runtime environment we know that the test is running under.

common/lib/util/defaults.js
QuintinWillison: Looking at Node's [os module](https://nodejs.org/api/os.html#os_os), we could consider adding [platform()](https://nodejs.org/api/os.html#os_os_platform)`/`[version()](https://nodejs.org/api/os.html#os_os_version), for example.

nodejs/lib/util/defaults.js
owenpearson: I've resolved this now by using `platform.versions.node` which is valid SemVer.

owenpearson:
After looking in to the browser test failures I discovered that we needed to update our CORS policy to support the new header. It will take a couple of weeks for the new policy to be pushed out to all environments so this PR is blocked until then.

QuintinWillison:
@owenpearson When you pick this pull request up again, can you please visit the existing open conversations and see if they're still valid and/or perhaps can be resolved. The spec has moved on, in particular with the addition of the `agents` client option in https://github.com/ably/docs/pull/1155 - which I am sure this client library does need, at a minimum for use by [ably-asset-tracking-js](https://github.com/ably/ably-asset-tracking-js).

PR #754, opened 2021-05-14
Platform types

owenpearson:
Creates a common interface for `Platform` config files, based off crypto-js types PR.

I didn't convert the NativeScript platform config file, because it would be a lot of effort to annotate all of the weird NativeScript globals and is ultimately not too important.

### DIFF
--- a/browser/fragments/platform-browser.ts
+++ b/browser/fragments/platform-browser.ts
@@ -1,4 +1,8 @@
 import msgpack from '../lib/util/msgpack';
+import IPlatform, { TypedArray } from '../../common/types/IPlatform';
+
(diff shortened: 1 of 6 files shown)


QuintinWillison (approved):
Bit surprising that we have `boolean` properties of `IPlatform` that are optional, but I'm assuming you've done that because it reflects what you've found in the implementation.

There could be an opportunity to document these interfaces now, while they're being created, but I can understand if you feel that would slow you down too much.

owenpearson:
@QuintinWillison  Yeah I think that's a valid point, but it's a tradeoff because it's important to avoid bloating the browser library and being more explicit about these properties would be a few extra bytes each with no change in behaviour. I might revisit this later but I'm gonna leave it for now.

PR #756, opened 2021-05-14
Webworker support

owenpearson:
Webworker support is currently broken, a regression from the move to webpack. This PR adds a new webpack bundle which specifically supports running in a webworker context.


ben-xD (approved):

ben-xD:
Oh, but 1 question: would this prevent a regression in the future? I made an environment [here](https://github.com/ben-xD/ably-webworkers) which crashed with the `window` error. Im not sure how we can run tests in a webworker environment though. Otherwise it might be easy to reintroduce this issue again

owenpearson:
@ben-xD Yeah this should be fairly regression-proof - I can't think of an obvious way to break it, although the webpack 'webworker' target isn't well documented so it's hard to say for certain. I actually tested against your ably-webworker repo when I was developing this so it should work, have you checked that you're using the `browser/static/ably-webworker.min.js` bundle?

ben-xD:
Okay cool. Im not facing any issues anymore because I've switched to Firebase functions which runs in a nodeJS environment. 

There was the other issue that I faced previously in Cloudflare Workers that still lingers is: 
screenshot_2021-05-14_at_11 10 58

I think you know more about this now: it seemed to us that `.Request` isn't actually implemented/ defined in the file at all, so it makes sense this error is thrown. However, why it **only** happens in Cloudflare Workers (and not Web Workers in general), I don't know.

owenpearson:
Thanks, I'm aware of the typings issue and need to investigate what the current best approach is. At the moment we have a separate type definitions file for each module you can import (see `callbacks.d.ts`, `promises.d.ts`) but that obviously doesn't scale nicely. Can you give me instructions on how to reproduce the undefined tokens behaviour?

And yes, the `.Request` thing is unrelated to this AFAIK so should be a separate issue. I'm happy for you to open the issue if you don't mind :) please include some basic instructions on how to reproduce. FYI, the reason it looks like `.Request` is not implemented is because the module is actually mutated elsewhere if you search around you should see there are a couple of places that set the `.Request` property on `Http` (obviously this isn't ideal and should no longer be the case once we've converted http stuff to typescript).

ben-xD:
To reproduce,
- `git clone https://github.com/ben-xD/ably-webworkers`. If you cloned in the past, ensure you're up to date
- To ensure this cloned repo uses your branch of Ably, In ably-js directory, run `npx link` and in `ably-webworkers` directory, run `npx link ably`. I've already updated the code to import `ably-webworker.min`, see https://github.com/ben-xD/ably-webworkers/blob/e6ae8d2a736da6e2ad36548c7d9c50919d4cc9e7/src/auth.ts#L3
- Follow the steps in 'getting started' in the readme (npm i, update api key, and npm start)
- Open the browser, notice in the browser console that the token is undefined. 
- Click webworkers button in the website, and notice the token is also undefined there.


Cool, I will make a github issue for the other issue

jcapogna:
I'm currently adding Ably a Cloudflare worker and running into think is this issue: 

```
Error: Something went wrong! Status: 400 Bad Request, Details {
  "result": null,
  "success": false,
  "errors": [
    {
      "code": 10021,
      "message": "Uncaught ReferenceError: window is not defined\n  at line 34\n  at line 34\n  at line 1 in n\n  at line 22\n  at line 1 in n\n  at line 22\n  at line 1 in n\n  at line 22\n  at line 1 in n\n  at line 22\n"
    }
  ],
  "messages": []
}
```

Is there an old version of Ably that does work in Cloudflare Workers?

ben-xD:
Hey @jcapogna, 
- You can use `1.2.3` which I've tested, and it works. 
- Or you can wait until this fix is fully developed and merged
- Probably not a good idea for production: 
  - You can clone the ably master and then in the ably-js directory: run `npm i` then `npm link` 
  - In your repo, use `npm link ably` to load the locally linked library
  - Unfortunately to get typescript support you need to uncomment line 2 and comment line 3 in https://github.com/ben-xD/ably-cloudflare-workers/blob/a4855a2bf46c7d8be0fbd1c45c105ae0ec561d97/src/auth.ts#L2 and to compile it, you need to comment line 2 and uncomment line 3.
  - Example code that works here: https://github.com/ben-xD/ably-cloudflare-workers

@owenpearson will know more

owenpearson:
Hi @jcapogna, thanks for reporting this to us. As Ben said above `1.2.3` should work fine - we have a couple of issues to resolve with this PR but we should have a new version which works with Cloudflare Workers released soon.

jcapogna:
Thank you, downgrading to 1.2.3 helped. I'm now having what I believe to be a different problem w/ Ably & Cloudflare so I opened #766

QuintinWillison (approved):

ben-xD:
Were we going to add typescript support for this one? There is no autocomplete when using the webworker import, so all the usages everything is a red squiggly line. When I was using it, I commented out the webworker import and used the normal import during development/ coding, and uncommented to build it. 

Having said that, if typescript support is coming through other means it's probably not a big deal to wait for that.

owenpearson:
@ben-xD This should be fixed by 406a689, have you tried it out since that commit?

PR #757, opened 2021-05-14
Use setImmediate if available in browser

### LINKED ISSUE #768: Certain Chrome versions throttling timers in hidden tabs
Customer has provided us with two scenarios where throttling in chrome v86 affects behaviour of the library

### Example 1

Switch tab before the timer goes off, will attach to channel twice
```
setTimeout(() => {
    var realtime = new Ably.Realtime({
        log: { level: 2 },
        key: "REDACTED",
        clientId: "" + Math.random(),
    });
    var channel = realtime.channels.get("channel_name");
    channel.subscribe();
}, 1000);
```

### Example 2

Switch tab before the timer goes off, second auth callback will not trigger
```
setTimeout(() => {
    var realtime = new Ably.Realtime.Promise({
      key: "REDACTED",
      authCallback: (_, callback) => {
        realtime.auth.createTokenRequest(null, null, function(__, tokenRequest) {
          console.log("Calling callback...")
          callback(undefined, tokenRequest)
        })
      },
    });
    realtime.auth.authorize().then((tokenDetails) => {
      console.log("Callback 1")
      realtime.auth.authorize().then((tokenDetails) => {
        console.log("Callback 2")
      })
    })
}, 1000);
```

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-897) by [Unito](https://www.unito.io)

owenpearson:
Resolves #768 

Opening as a draft for now just to run this through CI. Resolves an issue with recent versions of chrome throttling chained timers in hidden tabs.

### DIFF
--- a/browser/fragments/platform-browser.js
+++ b/browser/fragments/platform-browser.js
@@ -34,7 +34,7 @@ var Platform = {
 	preferBinary: false,
 	ArrayBuffer: global.ArrayBuffer,
 	atob: global.atob,
-	nextTick: function(f) { setTimeout(f, 0); },
+	nextTick: typeof setImmediate !== 'undefined' ? setImmediate : function(f) { setTimeout(f, 0); },
 	addEventListener: global.addEventListener,
 	inspect: JSON.stringify,
 	stringByteSize: function(str) {


owenpearson:
For visibility: the reason this works is that webpack v4 has a polyfill for NodeJS setImmediate. This won't actually use the builtin `setImmediate` from any browser implementations.

ben-xD (approved):

PR #762, opened 2021-05-17
Convert library code to TypeScript

### LINKED ISSUE #686: Migrate code base to typescript
Using typescript with strict type-checking will allow us to enforce better type safety in the project and will allow us to emit type definition files as part of the compile process rather than having to maintain them ourselves.

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-319) by [Unito](https://www.unito.io)

owenpearson:
Resolves #686

- Converts the library code to TypeScript. I've also opted to rewrite the code using ES6+ syntax which is compiled down to ES3 using the TypeScript compiler.
- I've used a few generally undesirable TypeScript features here in the interest of being able to get this change landed sooner (in particular I've used the `any` type quite widely and there are quite a few explicit type casts). We don't need to keep a note of these since they can easily be identified using ESLint and I do intend to fix them in the near future.
- The bundle size for `ably.min.js` (the minified browser bundle) has increased somewhat (from 193.5kb to 222kb). I don't think there's too much we can do about this right now without a lot of effort/risk, and I think there are better things we can do to reduce bundle size for lib consumers in future (like making the library more compatible with tree shaking).

Note for reviewers: All of the changes in this branch have been reviewed independently so don't feel it necessary to read through all of the changes included here; I'm really just looking for more general feedback on the overall structure of what's being changed here (although feel free to comment on specific library code changes if you want).


browser/fragments/platform-browser.ts
```diff
@@ -8,36 +12,37 @@ function allowComet() {
...
+	// TODO: type this properly
+	msgpack: msgpack as unknown as typeof import('@ably/msgpack-js'),
 	supportsBinary: !!global.TextDecoder,
 	preferBinary: false,
 	ArrayBuffer: global.ArrayBuffer,
 	atob: global.atob,
-	nextTick: typeof global.setImmediate !== 'undefined' ? global.setImmediate.bind(global) : function(f) { setTimeout(f, 0); },
-	addEventListener: global.addEventListener,
+	nextTick: typeof global.setImmediate !== 'undefined' ? global.setImmediate.bind(global) : function(f: () => void) { setTimeout(f, 0); },
+	addEventListener: global.addEventListener as unknown as null, // This is not ideal but TypeScript doesn't seem to allow anything else here
```
SimonWoolf, line 43: > `global.addEventListener as unknown as null, // This is not ideal but TypeScript doesn't seem to allow anything else here`

removing `as unknown as null` it seems to compile fine for me?

common/lib/client/auth.ts
```diff
@@ -1,112 +1,135 @@
...
-		toBase64 = Base64.encode;
-		hmac = function(text, key) {
-			return stringifyBase64(HmacSHA256(text, key));
-		};
+import { createHmac } from 'crypto';
+import { ErrnoException, RequestCallback, RequestParams } from '../../types/http';
+import * as API from '../../../ably';
+import { StandardCallback } from '../../types/utils';
+
+// TODO: replace these with the real types once these classes are in TypeScript
```
SimonWoolf, line 16: Lots of TODOs like this that look like they were intended to be dealt with before doing the final typescript release?

common/lib/client/realtimechannel.ts
```diff
@@ -1,27 +1,84 @@
...
+import ConnectionErrors from '../transport/connectionerrors';
+import * as API from '../../../ably';
+import ConnectionManager from '../transport/connectionmanager';
+import ConnectionStateChange from './connectionstatechange';
+import { ErrCallback, PaginatedResultCallback } from '../../types/utils';
+
+// TODO: Replace this with the real type when Realtime is in TypeScript
+type Realtime = any;
+
+export type RealtimeChannelState = 'initialized' | 'attaching' | 'attached' | 'suspended' | 'detaching' | 'detached' | 'failed';
```
SimonWoolf, line 20: There seems to now be duplication in a lot of places now between types defined in ably.d.ts and in the code like this one (which looks like it should be equivalent to `API.Types.ChannelState`). This makes it quite fragile against future changes - would be better defined in one place and referenced in the other. (Doesn't strictly matter which, but might be nicer for future maintainability to have it be defined in the code itself and just imported by ably.d.ts?)

common/types/ClientOptions.ts
```diff
@@ -0,0 +1,29 @@
+import { Modify } from "./utils";
+import * as API from '../../ably';
+
+export default interface ClientOptions extends API.Types.ClientOptions {
+  restAgentOptions?: { keepAlive: boolean, maxSockets: number };
+  pushFullWait?: boolean;
+  checkChannelsOnResume?: boolean;
+  plugins?: Record<string, unknown>;
```
SimonWoolf, line 8: - Since there's only one accepted plugin -- and it's very easy to type the wrong name -- why not have it typed, at least the key -- `plugins?: { vcdiff?: any }`
- Why is plugins in the 'non-public' part of the client options (in here instaed of in ably.d.ts)?

SimonWoolf (commented):
Added a few comments based on scrolling to a few random places in the diff, not a comprehensive review

Generally this is a big step forward, great work 👍

browser/fragments/platform-browser.ts
owenpearson: Yeah you're right, I've removed it now

common/lib/client/auth.ts
owenpearson: mb, addressed them all now

common/lib/client/realtimechannel.ts
owenpearson: We started using API.Types while this change was still in review so it slipped through, I've fixed it now though

common/types/ClientOptions.ts
owenpearson: Good point, I've added both the suggestions

browser/lib/transport/xhrrequest.ts
```diff
@@ -0,0 +1,348 @@
...
+import Http from 'platform-http';
+import Logger from '../../../common/lib/util/logger';
+import Defaults from '../../../common/lib/util/defaults';
+import * as BufferUtils from 'platform-bufferutils';
+import HttpMethods from '../../../common/constants/HttpMethods';
+import IXHRRequest from '../../../common/types/IXHRRequest';
+import { ErrnoException, RequestCallback, RequestParams } from '../../../common/types/http';
+import XHRStates from '../../../common/constants/XHRStates';
+
+// TODO replace this with the real type when Rest is in TypeScript
```
QuintinWillison, line 14: Can this comment go now?

browser/lib/util/http.ts
```diff
@@ -0,0 +1,176 @@
+import * as Utils from '../../../common/lib/util/utils';
+import Defaults from '../../../common/lib/util/defaults';
+import ErrorInfo from '../../../common/lib/types/errorinfo';
+import { ErrnoException, IHttp, RequestCallback, RequestParams } from '../../../common/types/http';
+import HttpMethods from '../../../common/constants/HttpMethods';
+
+// TODO: replace these with the real types once these classes are in TypeScript
```
QuintinWillison, line 7: Can this comment be removed now?

common/lib/client/rest.ts
```diff
@@ -0,0 +1,224 @@
...
+
+		(new PaginatedResource(this, '/stats', headers, envelope, function(body: unknown, headers: Record<string, string>, unpacked?: boolean) {
+			const statsValues = (unpacked ? body : JSON.parse(body as string));
+			for(let i = 0; i < statsValues.length; i++) statsValues[i] = Stats.fromValues(statsValues[i]);
+			return statsValues;
+		})).get(params as Record<string, string>, callback);
+	}
+
+	time(params: RequestParams | StandardCallback<number>, callback?: StandardCallback<number>): Promise<number> | void {
+		/* params and callback are optional; see if params contains the callback */
```
QuintinWillison, line 107: Intriguing. Have we got an issue in the backlog to improve this API? Well, assuming you think it _does_ need improving, however seeing the preamble required in this method's implementation I am thinking that if an interface like this can't be cleanly expressed in TypeScript then it needs a refactor. 🤔

QuintinWillison (changes requested):
I've pretty much skim read thru, given we know that all of this work has already been reviewed once in upstream PRs.

Just a few things that I think could still have a tidy, or I've only just spotted...

ably.d.ts
```diff
@@ -265,7 +266,7 @@ declare namespace Types {
 
 	type ChannelParams = { [key: string]: string };
 
-	type ChannelMode = 'PUBLISH' | 'SUBSCRIBE' | 'PRESENCE' | 'PRESENCE_SUBSCRIBE';
+	type ChannelMode = 'PUBLISH' | 'SUBSCRIBE' | 'PRESENCE' | 'PRESENCE_SUBSCRIBE' | 'ATTACH_RESUME';
```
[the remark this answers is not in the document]
owenpearson: I don't think this is part of the public API that we would expect customers to ever use so it's probably not worth an announcement, this is more just for the purpose of consistency between external and internal typings

common/lib/client/rest.ts
owenpearson: So this is really a problem caused by us using the old NodeJS callback-style async API. The fact that the callback is always the last argument (rather than in a fixed position) means that most of the public API methods in the library have to do the same argument normalisation dance. The only way we can really solve this is through a huge breaking API change so it's not really an option in v1.x versions of the library. Once we can release v2 then as part of that I would like for us to move to a modern ES6 promise-based async API which would make all of this arg normalisation stuff redundant and allow us to remove a lot of lib code.

browser/lib/util/http.ts
owenpearson: [ce064d0](https://github.com/ably/ably-js/pull/762/commits/ce064d0ddb1ca40d21d618c425ec9f9eb42de62b)

browser/lib/transport/xhrrequest.ts
owenpearson: [ce064d0](https://github.com/ably/ably-js/pull/762/commits/ce064d0ddb1ca40d21d618c425ec9f9eb42de62b)

browser/fragments/platform-reactnative.ts
```diff
@@ -1,27 +1,29 @@
...
-	xhrSupported: XMLHttpRequest,
+	xhrSupported: true,
 	allowComet: true,
 	jsonpSupported: false,
 	streamingSupported: true,
 	useProtocolHeartbeats: true,
 	createHmac: null,
-	msgpack: msgpack,
-	supportsBinary: (typeof TextDecoder !== 'undefined') && TextDecoder,
+	// TODO: type this properly
```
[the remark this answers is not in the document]
owenpearson: There isn't an issue but it was a quick fix 🙂  [4cc0ef6](https://github.com/ably/ably-js/pull/762/commits/4cc0ef63b7a1e15ad686c9fecf6a937fc8fa1b60)

Peter-Maguire (approved):

lmars (approved):
Epic effort! LGTM

common/lib/client/rest.ts
QuintinWillison: Thanks. Great explanation. Appreciated. 😁

QuintinWillison (approved):

PR #765, opened 2021-05-18
Add missing types from ClientOptions

owenpearson:


QuintinWillison (approved):
Happy for this to land as is, given surrounding context...

Though I would suggest that when you are in context of updating these interface definitions that you also consider adding API commentary. In particular I find myself wondering the units of `Duration` and `Timeout` properties.

PR #774, opened 2021-07-20
Bind setImmediate to global object in browsers

owenpearson:
This fixes an error in IE9-11 where calls to `setImmediate` would fail with `Invalid calling object`.


QuintinWillison (approved):
See [internal Slack thread](https://ably-real-time.slack.com/archives/C8SPU4589/p1626766573218600).

PR #777, opened 2021-07-26
fix channel names as object prototype keys

Matt-Esch:
When an object key is free-form text, we should not use vanilla js objects, as this can cause key collisions with built-in prototype keys, such as `__proto__` or `toString`, etc. I have offered a quick solution for this issue, but it would be better tackled using an ES6 `Map`. I am happy for someone to replace these structures with Map instances if they would prefer. I'm not familiar with which versions of js we support and whether we would want a `Map` polyfill if things are being back-ported (babel?) for < ES6 compatibility.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,20 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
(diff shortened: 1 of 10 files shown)


owenpearson (approved):
LGTM, thanks for fixing the presence test as well 👍

Matt-Esch:
I think I have satisfied myself that these tests are unaffected by the changes I've made so please go ahead and merge this for me please @owenpearson - I have fixed a few of the tests, I got to the point of debugging firefox tests and that's a bit too deep in the weeds at this point. I have at least seen passing tests for node 10/12/14 so if you're happy with this let's get it merged.

PR #779, opened 2021-07-27
EventEmitter: add EventEmitter#once(event[], ...)

SimonWoolf:
Had occasion to use `once(['enter', 'update'])` in a realtime test, discovered I'd never gotten around to implementing it in ably-js


owenpearson (approved):
LGTM, thanks

PR #784, opened 2021-08-03
Implement RTL5b and RTL5j

owenpearson:
Spec alignment issue reported internally; see slack thread: https://ably-real-time.slack.com/archives/C01K3K2V5HB/p1627940634005500

Adds/modifies switch cases for detaching from channels in the `suspended` and `failed` states, along with associated tests.

### DIFF
--- a/common/lib/client/realtimechannel.js
+++ b/common/lib/client/realtimechannel.js
@@ -308,6 +308,10 @@ var RealtimeChannel = (function() {
 			return;
 		}
 		switch(this.state) {
+                        case 'suspended':
+                                this.requestState('detached');
+                                callback();
+                                break;
 			case 'detached':
 			case 'failed':
 				callback();
--- a/common/lib/client/realtimechannel.js
+++ b/common/lib/client/realtimechannel.js
@@ -309,7 +309,7 @@ var RealtimeChannel = (function() {
 		}
 		switch(this.state) {
                         case 'suspended':
-                                this.requestState('detached');
+                                this.notifyState('detached');
                                 callback();
                                 break;
 			case 'detached':
--- a/common/lib/client/realtimechannel.js
+++ b/common/lib/client/realtimechannel.js
@@ -313,8 +313,10 @@ var RealtimeChannel = (function() {
                                 callback();
                                 break;
 			case 'detached':
-			case 'failed':
 				callback();
+				break;
+			case 'failed':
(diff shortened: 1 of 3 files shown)


spec/realtime/channel.test.js
```diff
@@ -1506,5 +1506,35 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
...
 			);
 		});
+
+                // RTL5j
+                it('detaching from suspended channel transitions channel to detached state', function (done) {
+                    var realtime = helper.AblyRealtime({ transports: [helper.bestTransport] });
+                    var channelName = 'detach_from_suspended';
+                    var channel = realtime.channels.get(channelName);
+
+                    channel.state = 'suspended';
```
QuintinWillison, line 1516: Just to ensure my understanding: I'm assuming this isn't public API - i.e. the test is mutating an internal object?

QuintinWillison (approved):

spec/realtime/channel.test.js
owenpearson: It certainly is part of the public API and is documented in the typescript interface definition, but I think it should be clear to developers that the intention is for this to be readonly and I would expect no one would actually try mutating it unless they're trying to mess with the lib internals.

SimonWoolf: typescript has a `readonly` attribute modifier now, we can add that to the typings on things like this to make the intention even clearer

SimonWoolf (approved):
lgtm with comment.

I'm now thinking it probably makes sense to modify `(RTL5k) If the channel receives an ATTACHED message while in the DETACHING state, it should send a new DETACH message and remain in the DETACHING state` to also apply to the `detached` state. Imagine: client lib is in `detached` after an explicit detach() while in `suspended` after a failed attach, and then the server finally catches up and sends an ATTACHED; we want to enforce the user's stated preference and send a DETACH rather than just acquiesce and end up in the `attached` state after the user called detach(), wdyt?

common/lib/client/realtimechannel.js
```diff
@@ -308,10 +308,16 @@ var RealtimeChannel = (function() {
 			return;
 		}
 		switch(this.state) {
+                        case 'suspended':
+                                this.requestState('detached');
```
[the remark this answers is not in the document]
owenpearson: thanks, resolved in e6cfd1b

Issue #685, opened 2020-10-05
Replace JSHint with ESLint


owenpearson:
Currently JSHint is being used for linting. ESLint generally has a wider range of linting rules and plugins (including a Prettier plugin which AFAIK JSHint does not have). ESLint also seems to have a better ecosystem with out-of-the-box configurations like airbnb-base which we could leverage. There are ESLint rules which don't exist in JSHint which could identify issues, for example, #679 could be identified with the eslint `no-param-reassign`. ESLint also allows linting of typescript using the @typescript-eslint parser which is, AFAIK, not possible using JSHint.

### Acceptance criteria
- ESLint set up as an NPM script to check all source code (currently `browser`, `common` and `nodejs` directories)
- CI check set up to run ESLint on PR and push to main.
- Use the base `eslint:recommended` ruleset and get the CI check passing.

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-318) by [Unito](https://www.unito.io)

Issue #684, opened 2020-10-05
Use Prettier for code formatting


owenpearson:
Currently this project does not use a formatter. Adding [Prettier](https://prettier.io/) would allow us to enforce consistent styling in all JavaScript, TypeScript, JSON, and markdown files. Prettier also [integrates](https://prettier.io/docs/en/integrating-with-linters.html) nicely with linting tools such as ESLint.

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-317) by [Unito](https://www.unito.io/learn-more)

PR #794, opened 2021-09-21
Convert lib/types dir to TypeScript

owenpearson:
Resolves #790


common/lib/util/forInOwnNonNullProps.ts
```diff
@@ -0,0 +1,7 @@
+export default function(ob: Record<string, unknown>, fn: Function) {
```
QuintinWillison, line 1: For an implementation that literally uses a fully qualified / expanded name like [hasOwnProperty](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty), I'm confused why we're abbreviating the word 'properties'.

While I recognise that there are frameworks like React which have popularised `Props` as a shortening, if there's no strong reason why we need it shorter then I think it would be clearer for those reading the code if this file were called `forInOwnNullProperties` and it were imported as such where needed.

common/lib/util/logger.ts
```diff
@@ -0,0 +1,94 @@
...
+
+enum LogLevels {
+    None = 0,
+    Error = 1,
+    Major = 2,
+    Minor = 3,
+    Micro = 4,
+}
+
+function pad(str: unknown, three?: number) {
```
QuintinWillison, line 11: The abbreviation `str` is not helpful, nor is the name I assume is the expansion ('string'), especially with the type being `unknown`.

common/lib/types/stats.ts
```diff
@@ -0,0 +1,308 @@
...
+	xchgConsumer?: XchgMessages;
+	push?: PushStats;
+	processed?: ProcessedMessages;
+	inProgress?: never;
+	unit?: never;
+	intervalId?: never;
+
+	constructor(values?: StatsValues) {
+		super(values as MessageDirectionsValues);
+		this.persisted     = new MessageTypes(values && values.persisted);
```
QuintinWillison, line 289: Variable spacing within statements to form columns like this is generally not a good idea as it will cause bigger than necessary diff noise when lines get added or removed from this 'bag of setters'.

QuintinWillison (changes requested):
In an ideal world I would have liked to see new API commentary alongside at least some of the modules, types and methods/functions. This could be an opportunity to inject some of that, though I recognise it may not have been a priority for you.

I'm also noticing a lot of variance in terms of indentation. e.g. `common/lib/types/stats.ts` vs `common/lib/util/dataSizeBytes.ts`. I've not pulled the code in locally to inspect further but the impression is that some files are using tabs and others using spaces. I assume, therefore, that we're not running any tools over the code in CI that would spot for formatting inconsistencies? I think we should be and these tools should form foundational underpinnings to this work - otherwise we're only going to need to go back over all the files again some time in the future to conform them.

common/types/platform-bufferutils.d.ts
```diff
@@ -0,0 +1,16 @@
+declare module 'platform-bufferutils' {
```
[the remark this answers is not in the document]
owenpearson: I'm a bit cautious about using `data` because `buffer` is more precise, but I understand the preference to avoid abbreviations so I've changed all buffer operations to use `buffer` in 3891c09.

common/lib/util/forInOwnNonNullProps.ts
owenpearson: Agreed, 33a6451.

common/lib/util/logger.ts
owenpearson: I've rewritten this function, should be a lot more readable now.

common/lib/types/stats.ts
owenpearson: Yeah I'm a bit unsure about how to proceed with this (and the other styling differences you mentioned above). I actually don't think it's necessarily a bad thing for us to conform these files again in future using `prettier`. The alternative would be to reformat every file as I change them which would massively increase the burden on the reviewer to understand what's changed in each PR. I absolutely agree that this TypeScript migration is the right time to introduce prettier to the repo but my intention at the moment is to do so at the end.

owenpearson:
@QuintinWillison The tabs vs spaces stuff should be fixed now.

As for the API commentary that would go on the external type interface (`ably.d.ts`) and would be unrelated to this work. Maybe one day we can automatically generate the external type interface from the source code but it would be hard to do that without exposing a bunch of internal typings that we wouldn't want users of the library to see.

common/lib/types/stats.ts
QuintinWillison: Fair enough and thanks for explaining. I can see we have #684 in the backlog, supported by #685. In terms of code quality, implementing both of those will raise the game massively and will also reduce the need for code review comments like the one I made to start this conversation even exist. 😄

QuintinWillison (approved):

PR #795, opened 2021-09-22
Convert multicaster to typescript

owenpearson:
TypeScript doesn't really support class instances with call signatures so had to switch to a static factory method which returns a plain object.. Should still be fairly self explanatory.


common/lib/util/multicaster.ts
```diff
@@ -0,0 +1,43 @@
...
+	push(...args: Array<Function>): void {
+		this.members.push(...args);
+	}
+
+	static create(members?: Array<Function>) {
+		const instance = new Multicaster(members);
+		return Object.assign(
+      (...args: unknown[]) => instance.call(args),
+      {
+        push: (fn: Function) => instance.push(fn)
```
sacOO7, line 37: push can be updated as `push:instance.push` , same goes for `call` ig

owenpearson: Nice spot, changed both in ed3209cb

common/lib/util/multicaster.ts
```diff
@@ -0,0 +1,43 @@
...
+class Multicaster {
+	members: Array<Function>;
+
+  // Private constructor; use static Multicaster.create instead
+	private constructor(members?: Array<Function>) {
+		this.members = members || [];
+	}
+
+	call(...args: unknown[]): void {
+		for (let i = 0; i < this.members.length; i++) {
```
sacOO7, line 12: `for (const member in members)`

owenpearson: Thanks, I've changed it in 7e03f30

QuintinWillison:
You have [this failure](https://github.com/ably/ably-js/pull/795/checks?check_run_id=3688595907#step:5:5424) for the `test-node` workflow's `10.x` run:

> Error: The operation was canceled

I wonder if that's a timeout (perhaps increase `timeout-minutes` at [step](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepstimeout-minutes) or [job](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes) level?).

QuintinWillison:
@owenpearson's `test-node` workflow fix, addressing [his comment](https://github.com/ably/ably-js/pull/795#issuecomment-928075486), is https://github.com/ably/ably-js/pull/802

common/lib/util/multicaster.ts
```diff
@@ -0,0 +1,49 @@
...
+  // Private constructor; use static Multicaster.create instead
+	private constructor(members?: Array<AnyFunction | undefined>) {
+		this.members = members as Array<AnyFunction> || [];
+	}
+
+	call(...args: unknown[]): void {
+		for (const member of this.members) {
+			if (member) {
+				try {
+          member(...args);
```
QuintinWillison, line 22: I'm trying to ignore the indentation issue here as I know you plan to fix this later with [Prettier](https://github.com/ably/ably-js/issues/684) (/ [ESLint](https://github.com/ably/ably-js/issues/685)), but I couldn't simply pass by without commenting as it does make it harder to review when things like this stand out (cognitive load on reviewers as they parse the code for themselves).

sacOO7 (approved):

QuintinWillison (approved):
While I acknowledge that `args` is a commonly understand programmers' abbreviation for `arguments`, I'm not sure I can see a reason why it couldn't be left expanded for clarity (though I can see that nobody is that consistent on this - e.g. `arguments` [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments), yet `theArgs` [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters)).

owenpearson:
@QuintinWillison `arguments` is a JavaScript keyword. It was used to make a function take a variable number of arguments before ES6 spread syntax. It's idiomatic to use `args` when using spread syntax to make it clear that you're not using the old `arguments` keyword.

QuintinWillison:
@owenpearson Thanks for explaining re `arguments` vs `args`. Of course I hadn't actually tried to make what I suggested work in some real code. 😆 Every day's a school day. 🚀

PR #799, opened 2021-09-23
Convert eventemitter to typescript

owenpearson:


common/lib/util/eventemitter.ts
```diff
@@ -86,7 +88,7 @@ var EventEmitter = (function() {
 	 * @param listener (optional) the listener to remove. If not
 	 *        supplied, all listeners are removed.
 	 */
-	EventEmitter.prototype.off = function(event, listener) {
+	off(event: string | null, listener: Function) {
```
QuintinWillison, line 91: I'm not sure this method signature matches the implementation, or there's something subtle going on that I don't understand. In the parlance of other languages, it looks like `off` is overloaded to allow just `listener` to be provided with no `event` name. I'm talking about the block starting with this statement, further down in the implementation of this method:

```js
if(typeof(event) == 'function') {
```

common/lib/util/eventemitter.ts
```diff
@@ -109,18 +111,17 @@ var EventEmitter = (function() {
...
 		}
 
 		/* "normal" case where event is an actual event */
 		if(listener) {
-			removeListener([this.events, this.eventsOnce], listener, event);
+			removeListener([this.events, this.eventsOnce], listener, event as string);
 		} else {
-			delete this.events[event];
-			delete this.eventsOnce[event];
+			delete this.events[event as string];
```
QuintinWillison, line 123: The use of [type assertion](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions) feels like a red flag here. Are you not able to provide TypeScript with enough information by this point not to need this?

common/lib/util/eventemitter.ts
```diff
@@ -176,19 +176,21 @@ var EventEmitter = (function() {
 	 * @param event the name of the event to listen to
 	 * @param listener the listener to be called
 	 */
-	EventEmitter.prototype.once = function(event, listener) {
-		var argCount = arguments.length, self = this;
+	once(event: string, listener: Function) {
+		const argCount = arguments.length;
```
QuintinWillison, line 180: I would prefer `argumentCount`. 😄

common/lib/util/eventemitter.ts
```diff
@@ -212,17 +214,15 @@ var EventEmitter = (function() {
 	 * @param currentState the name of the current state of this object
 	 * @param listener the listener to be called
 	 */
-	EventEmitter.prototype.whenState = function(targetState, currentState, listener /* ...listenerArgs */) {
-		var eventThis = {event:targetState},
-			self = this,
-			listenerArgs = Array.prototype.slice.call(arguments, 3);
+	private whenState(targetState: string, currentState: string, listener: Function, ...listenerArgs: unknown[] /* ...listenerArgs */) {
```
QuintinWillison, line 217: Couple of things here:

- would prefer `listenerArguments` 😝 
- does the `/* ...listenerArgs */` comment need to remain?

QuintinWillison (changes requested):
I think there's enough things I felt the need to comment on in this pull request that I would like to see those comments addressed before I approve, please.

common/lib/util/eventemitter.ts
```diff
@@ -2,81 +2,83 @@ import * as Utils from './utils';
...
+
+/**
+ * Remove listeners that match listener
+ * @param targetListeners is an array of listener arrays or event objects with arrays of listeners
+ * @param listener the listener callback to remove
+ * @param eventFilter (optional) event name instructing the function to only remove listeners for the specified event
+ */
+function removeListener(targetListeners: any, listener: Function, eventFilter?: string) {
+	let listeners: Record<string, unknown>;
+	let idx;
```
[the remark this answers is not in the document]
owenpearson: 52f1d6b1

common/lib/util/eventemitter.ts
```diff
@@ -2,81 +2,83 @@ import * as Utils from './utils';
...
 		if(arguments.length == 1 && typeof(event) == 'function') {
 			this.any.push(event);
 		} else if(Utils.isEmptyArg(event)) {
 			this.any.push(listener);
 		} else if(Utils.isArray(event)) {
-			var self = this;
-			Utils.arrForEach(event, function(ev) {
-				self.on(ev, listener);
-			});
+			event.forEach((ev) => {
```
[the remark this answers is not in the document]
owenpearson: ee321214

common/lib/util/eventemitter.ts
```diff
@@ -2,81 +2,83 @@ import * as Utils from './utils';
 import Logger from './logger';
 import Platform from 'platform';
 
-var hasOwnProperty = Object.prototype.hasOwnProperty;
+/* Call the listener, catch any exceptions and log, but continue operation*/
+function callListener(eventThis: { event: string }, listener: Function, args: unknown[]) {
```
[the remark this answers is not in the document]
owenpearson: See [my response](https://github.com/ably/ably-js/pull/795#issuecomment-933525518)

common/lib/util/eventemitter.ts
```diff
@@ -2,81 +2,83 @@ import * as Utils from './utils';
...
-	}
-
 	/**
 	 * Add an event listener
 	 * @param event (optional) the name of the event to listen to
 	 *        if not supplied, all events trigger a call to the listener
 	 * @param listener the listener to be called
 	 */
-	EventEmitter.prototype.on = function(event, listener) {
+	on (event: string, listener: Function) {
```
SimonWoolf, line 68: This is restricting to a single signature, (even, listener), but on() is overloaded, eg you can call with a listener but no event. With this that call will fail typechecking.

With typescript the idiomatic way of doing this is multiple function signatures, where the last one is the 'active' one which is the union of all the others, and the others are basically just documentation -- eg:

```
on(listener: Function): void;
on(event: null | string | string[], listener: Function): void;
on(event: null | string | string[] | Function, listener?: Function): void {
```

common/lib/util/eventemitter.ts
SimonWoolf: Similar comment applies to previous, as Quintin correctly points out off is overloaded, so eg
```
off(listener?: Function): void;
off(event: null | string | string[], listener?: Function): void;
off(event?: null | string | string[] | Function, listener?: Function) {
```

common/lib/util/eventemitter.ts
SimonWoolf: I agree with Quintin, a better way to convince ts that if we get this far then event must be a string is just adding a runtime check before the 'event is string' section starts, eg
```
if(typeof event !== 'string') {
	throw new Error('EventListener.off: invalid function arguments ' + Utils.inspect(arguments));
}
```
that'll let you remove the assertion

common/lib/util/eventemitter.ts
```diff
@@ -176,19 +176,21 @@ var EventEmitter = (function() {
 	 * @param event the name of the event to listen to
 	 * @param listener the listener to be called
 	 */
-	EventEmitter.prototype.once = function(event, listener) {
-		var argCount = arguments.length, self = this;
+	once(event: string, listener: Function) {
```
SimonWoolf, line 179: similarly overloaded

SimonWoolf (changes requested):
Other than specific comments, my main feedback is on the structure of the commits. I recommend you split it up, with one commit just to rename the file from .js to .ts, and a second commit to actually make the changes. E.g. see a recent realtime pr https://github.com/ably/realtime/pull/3708/commits

The reason for this is how git tracks file history. There's no explicit rename tracking, so it uses heuristics to tell when a file has been renamed (for the purposes of walking back through the history of a file, seeing commit messages with blame, etc.). When the rename is one commit which doesn't change the contents of the file, the heuristics always succeed, they don't have to do any work. When a single commit changes both name and contents, they... often work. But sometimes don't. And they start working less as repos get longer (there's a configurable limit to the total amount of guesswork it'll do to avoid git operations taking too long). So git best practice is to isolate renames in their own commit.

(It also makes code reviewing easier -- github can sometimes get confused and show a typescript conversion as a delete and new file, because again heuristics; if it's split you can review just the commit that actually makes the contents changes, which works properly. Compare the diffs in lrumap.ts [here](https://github.com/ably/realtime/pull/3708/files) vs [here](https://github.com/ably/realtime/pull/3708/commits/c39bf2d6f7b021b88d82f516cdb8021af667ded5), which is easier to review?)

You may be interested in https://ably.atlassian.net/wiki/spaces/ENG/pages/312345055/Typescript-Conversion-Guide , which has that and more advice.

owenpearson:
@SimonWoolf Thanks for pointing that out, I do see that it's valuable to keep git history intact and is much easier to review when git is aware that the file has moved. I was actually trying to follow the guidance that each commit should compile/pass tests hence the all-in-one "convert x to typescript" commits but I'll change the file extensions first in future PRs.

common/lib/util/eventemitter.ts
owenpearson: I'd rather keep it as `listenerArgs` given my comment about `args` being more idiomatic. As for the comment, 6f0aea0.

common/lib/util/eventemitter.ts
owenpearson: 32f8d04

common/lib/util/eventemitter.ts
owenpearson: 32f8d04

common/lib/util/eventemitter.ts
owenpearson: I've refactored this to use `args` in 32f8d04 so I'd prefer to keep this one as `argCount`.

SimonWoolf (approved):

Issue #701, opened 2020-11-16
NPM module 'request' is deprecated


owenpearson:
See deprecation announcement [here](https://github.com/request/request/issues/3142) and a discussion around recommended alternatives [here](https://github.com/request/request/issues/3143).

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-600) by [Unito](https://www.unito.io)

sacOO7:
Go for axios : )

owenpearson:
Commenting here so I don't forget; it looks like [got](https://www.npmjs.com/package/got) is currently the only widely used high level HTTP client for NodeJS with support for HTTP/2.

If axios/axios#1175 or node-fetch/node-fetch#372 become resolved then they might be worth looking in to as well.

jimmywarting:
I think you should go with node-fetch cuz then your frontend wouldn't need any extra large lib, beside Deno also has built in fetch

jimmywarting:
besides axios don't work in web workers, axios is built around xhr (not fetch)
so by using axios then it would not work in web workers

PR #800, opened 2021-09-24
Convert HTTP clients to TypeScript

owenpearson:


jamienewcomb:
@owenpearson is this ready to have some reviewers assigned?

browser/lib/util/http.ts
```diff
@@ -0,0 +1,176 @@
+import * as Utils from '../../../common/lib/util/utils';
+import Defaults from '../../../common/lib/util/defaults';
+import ErrorInfo from '../../../common/lib/types/errorinfo';
+import { IHttp, RequestCallback, RequestParams } from '../../../common/types/http';
+import HttpMethods from '../../../common/constants/HttpMethods';
+
+type Rest = any;
```
QuintinWillison, line 7: I'm guessing that this and `Realtime` below it, both repeated in the `nodejs` variant, represent some kind of temporary placeholder - i.e. they'll be replaced at some point later, presumably when these types get formal TypeScript-aware definitions?

If that is the case then I think best practice here would be to add comments to these type definitions, individually:

- stating that they are temporary
- explicitly mentioning `TODO` as a marker in the code
- including a link to the GitHub issue which you will work on to fix them

browser/lib/util/http.ts
```diff
@@ -0,0 +1,176 @@
...
+type Realtime = any;
+
+const noop = function() {};
+
+function shouldFallback(err: ErrorInfo) {
```
QuintinWillison, line 12: It's me again, on the top of abbreviations. 😁 

For me, `err` is a confusing abbreviation to use here. I can see from the type information you've added that it's clearly not a JS-built-in [Error](https://developer.mozilla.org/en-US/docs/web/javascript/reference/global_objects/error), but that would have been my default assumption. I would expand fully to `errorInfo` (and, yes, I accept that this would sub-optimally be `errorInfo` rather than `errorInformation` - however I accept that this is the state of play for this type as `ErrorInfo` is [baked into the spec.](https://docs.ably.com/client-lib-development-guide/features/#TI1)).

package.json
```diff
@@ -33,6 +33,7 @@
     "@ably/vcdiff-decoder": "1.0.4",
     "@types/crypto-js": "^4.0.1",
     "@types/node": "^15.0.0",
+    "@types/request": "^2.48.7",
```
QuintinWillison, line 36: I was curious as to what this dependency does, so I had a quick look at [the npm page for it](https://www.npmjs.com/package/@types/request). That led me to click [the GitHub link in the first sentence](https://github.com/request/request). I see you've created https://github.com/ably/ably-js/issues/701 to handle that deprecation in our backlog, however I think we should perhaps have a conversation (outside of the scope of this pull request) around how we label GitHub issues that represent architectural debt like this. 🤔

common/types/http.d.ts
```diff
@@ -0,0 +1,39 @@
...
+	static put(rest: Rest | null, path: PathParameter, headers: Record<string, string> | null, body: unknown, params: RequestParams, callback: RequestCallback): void;
+	static putUri(rest: Rest | null, uri: string, headers: Record<string, string> | null, body: unknown, params: RequestParams, callback: RequestCallback): void;
+	static patch(rest: Rest | null, path: PathParameter, headers: Record<string, string> | null, body: unknown, params: RequestParams, callback: RequestCallback): void;
+	static patchUri(rest: Rest | null, uri: string, headers: Record<string, string> | null, body: unknown, params: RequestParams, callback: RequestCallback): void;
+	static checkConnectivity?: (callback: (err?: ErrorInfo | null, connected?: boolean) => void) => void;
+	static Request?: (method: HttpMethods, rest: Rest | null, uri: string, headers: Record<string, string> | null, params: RequestParams, body: unknown, callback: RequestCallback) => void;
+	static _getHosts: (client: Rest | Realtime) => string[];
+	static supportsAuthHeaders: boolean;
```
QuintinWillison, line 28: Could this be an opportunity to expand this abbreviation - i.e. define as `supportsAuthorizationHeaders`?

common/types/http.d.ts
```diff
@@ -0,0 +1,39 @@
...
+	static supportsLinkHeaders: boolean;
+}
+
+export interface ErrnoException extends Error {
```
QuintinWillison, line 32: `Errno` 🤯  - has this horse bolted or could we name this more sensibly?

nodejs/lib/util/http.ts
```diff
@@ -0,0 +1,236 @@
...
+import ErrorInfo from '../../../common/lib/types/errorinfo';
+import request, { Options as RequestOptions, RequestResponse } from 'request';
+import { ErrnoException, IHttp, PathParameter, RequestCallback, RequestParams } from '../../../common/types/http';
+import HttpMethods from '../../../common/constants/HttpMethods';
+
+type Rest = any;
+type Realtime = any;
+
+const msgpack = Platform.msgpack;
+const noop = function() {};
```
QuintinWillison, line 13: While I recognise most programmers are familiar with the concept of a 'no-op', it's also earnt itself a [disambiguation page on Wikipedia](https://en.wikipedia.org/wiki/Noop_(disambiguation)). Also, for what this variable is actually doing, I think a name like `doNothing` would be more explicit and perhaps more self-explanatory at the sites where it is used.

nodejs/lib/util/http.ts
```diff
@@ -0,0 +1,236 @@
...
+
+/***************************************************
+ *
+ * These Http ops are used for REST operations
```
QuintinWillison, line 17:
```suggestion
 * These Http operations are used for REST operations
```

nodejs/lib/util/http.ts
```diff
@@ -0,0 +1,236 @@
...
+		}
+		if(params)
+			doOptions.qs = params;
+
+		doOptions.uri = uri;
+		doOptions.timeout = (rest && rest.options.timeouts || Defaults.TIMEOUTS).httpRequestTimeout;
+		request[method](doOptions, handler(uri, params, callback));
+	}
+
+	/** Http.get, Http.post, Http.put, ...
```
QuintinWillison, line 159: This whole comment structure looks somewhat irregular... is this some kind of multi-method commentary syntax that I'm unaware of? 😕

nodejs/lib/util/http.ts
```diff
@@ -0,0 +1,236 @@
...
+			} else {
+				/* Fallback expired; remove it and fallthrough to normal sequence */
+				rest._currentFallback = null;
+			}
+		}
+	
+		const hosts = getHosts(rest);
+	
+		/* see if we have one or more than one host */
+		if(hosts.length == 1) {
```
QuintinWillison, line 117:
```suggestion
		if(hosts.length === 1) {
```

browser/lib/util/http.ts
```diff
@@ -0,0 +1,176 @@
...
+	const statusCode = err.statusCode as number;
+	/* 400 + no code = a generic xhr onerror. Browser doesn't give us enough
+		* detail to know whether it's fallback-fixable, but it may be (eg if a
+		* network issue), so try just in case */
+	return (statusCode === 408 && !err.code) ||
```
QuintinWillison, line 17: If HTTP status codes aren't available from one of our dependencies then perhaps we could pull them out to our own module somewhere?

browser/lib/util/http.ts
```diff
@@ -0,0 +1,176 @@
...
+			} else {
+				/* Fallback expired; remove it and fallthrough to normal sequence */
+				rest._currentFallback = null;
+			}
+		}
+
+		const hosts = getHosts(rest);
+
+		/* if there is only one host do it */
+		if(hosts.length == 1) {
```
QuintinWillison, line 73:
```suggestion
		if(hosts.length === 1) {
```

nodejs/lib/util/http.ts
```diff
@@ -0,0 +1,236 @@
...
+}
+
+const Http: typeof IHttp = class {
+	static methods = [HttpMethods.Get, HttpMethods.Delete, HttpMethods.Post, HttpMethods.Put, HttpMethods.Patch];
+	static methodsWithoutBody = [HttpMethods.Get, HttpMethods.Delete];
+	static methodsWithBody = [HttpMethods.Post, HttpMethods.Put, HttpMethods.Patch];
+
+	/* Unlike for doUri, the 'rest' param here is mandatory, as it's used to generate the hosts */
+	static do(method: HttpMethods, rest: Rest, path: PathParameter, headers: Record<string, string> | null, body: unknown, params: RequestParams, callback: RequestCallback): void {
+		const uriFromHost = (typeof(path) == 'function') ? path : function(host: string) { return rest.baseUri(host) + path; };
```
QuintinWillison, line 92:
```suggestion
		const uriFromHost = (typeof(path) === 'function') ? path : function(host: string) { return rest.baseUri(host) + path; };
```

Also, I'm curious as to why it's `typeof(path)` (function-like) here but `typeof` (operator-like) elsewhere - e.g. `typeof responseText` later on in this file. 🤔 This may well just be a gap in my JS knowledge (something obvious I'm not seeing), but I think it would be nice to be consistent.

QuintinWillison (changes requested):

nodejs/lib/util/http.ts
SimonWoolf: (no gap in your knowledge, just style differences - I (and Owen) have generally gone for the operator-like style (to make it clear that it isn't actually a function), but Paddy I think liked the function-like style)

SimonWoolf (commented):
(comment [here](https://github.com/ably/ably-js/pull/799#pullrequestreview-770576835) re commit structure applies here too)

nodejs/lib/util/http.ts
owenpearson: I think it's novel but not sure if there's a problem with it. Consumers of this class will get the typings from `http.d.ts` because this class isn't directly imported anywhere in source code, so it's only gonna be visible to people looking in this file and I think it's fairly obvious what the intended meaning is.

common/types/http.d.ts
QuintinWillison: ok, thanks for explaining.

browser/lib/util/http.ts
QuintinWillison: Owen's resolution (addition of comments) looks acceptable to me, given [his explanation](https://github.com/ably/ably-js/pull/800#discussion_r723320445), however I would rather leave it to @SimonWoolf to resolve this conversation in order to get his acknowledgement that this is satisfactory for now.

PR #802, opened 2021-09-27
`fail-fast` for test-node workflow

owenpearson:
Stops node test workflows from being cancelled as soon as one of them fails.

### DIFF
--- a/.github/workflows/test-node.yml
+++ b/.github/workflows/test-node.yml
@@ -7,8 +7,8 @@ on:
 jobs:
   test-node:
     runs-on: ubuntu-latest
-    continue-on-error: true
     strategy:
+      fail-fast: false
       matrix:
         node-version: [10.x, 12.x, 14.x]
     steps:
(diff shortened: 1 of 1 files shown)


QuintinWillison (changes requested):
I'm not sure I like the look at this. Are you sure this will do what you were seeking it to?

My fear is that this will make this workflow always look like it's passed, even when it's failed, perhaps even on a step that doesn't have any `npm` commands. 🤔 

So:

1. I'm wondering if `continue-on-error` is _ever_ appropriate in a workflow
2. I'm wondering if, even if it is appropriate and does do what you want it to, whether it should be [at step level](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) rather than [at job level](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error) like you have at the moment

owenpearson:
@QuintinWillison I've replaced `continue-on-error` with `fail-fast` as suggested

QuintinWillison (approved):

PR #803, opened 2021-09-30
Use package.json version directly in library

owenpearson:
This works because webpack can automatically resolve JSON modules as plain JS objects. We will need to add the `resolveJsonModule` compiler option when these files are migrated to TypeScript.

This PR also removes a section of the README which advertises the current library version. IMO this is an improvement because people will know where to look for the version if that's what they're looking for (probably in the releases/tags section or over at npmjs.com) so if someone has scrolled that far down the README then they're gonna be looking for something else.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -87,7 +87,6 @@ module.exports = function (grunt) {
 	grunt.initConfig(gruntConfig);
 
 	grunt.registerTask('build', [
-		'set-library-version',
 		'webpack'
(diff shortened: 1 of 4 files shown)


QuintinWillison (approved):
Nice. And I completely agree that there is no need to explicitly mention the version, in the way that it has been, in the README. We still do in some other SDKs but the reasons behind it vary, and some are probably not very strong reasons. 🤔

PR #804, opened 2021-09-30
Check for alternative casing in x-ably-errorcode header

owenpearson:
Some parts of the realtime app use X-Ably-ErrorCode instead of x-ably-errorcode for some reason which was causing problems in some of the browser tests.

### DIFF
--- a/browser/lib/transport/xhrrequest.js
+++ b/browser/lib/transport/xhrrequest.js
@@ -221,7 +221,7 @@ var XHRRequest = (function() {
 				return;
 			}
 
-			var err = headers['x-ably-errorcode'] && responseBody.error && ErrorInfo.fromValues(responseBody.error);
+			var err = (headers['x-ably-errorcode'] | headers['X-Ably-ErrorCode']) && responseBody.error && ErrorInfo.fromValues(responseBody.error);
 			if(!err) {
 				err = new ErrorInfo('Error response received from server: ' + statusCode + ' body was: ' + Utils.inspect(responseBody), null, statusCode);
 			}


browser/lib/transport/xhrrequest.js
```diff
@@ -8,6 +8,14 @@ import Defaults from '../../../common/lib/util/defaults';
 import BufferUtils from 'platform-bufferutils';
 import DomEvent from '../util/domevent';
 
+function getAblyError(responseBody, headers) {
+	if (headers['x-ably-errorcode'] | headers['X-Ably-ErrorCode']) {
```
QuintinWillison, line 12: I think HTTP headers are case insensitive. Can't you just do a case insensitive compare to catch all variants?

QuintinWillison (changes requested):

browser/lib/transport/xhrrequest.js
SimonWoolf: ^ what quintin said

browser/lib/transport/xhrrequest.js
```diff
@@ -8,6 +8,14 @@ import Defaults from '../../../common/lib/util/defaults';
...
+		return responseBody.error && ErrorInfo.fromValues(responseBody.error);
+	}
+	// Partial errors don't have an X-Ably-ErrorCode header so check if the response has that specific error code
+	return responseBody.error && responseBody.error.code === 40020 && ErrorInfo.fromValues(responseBody.error);
```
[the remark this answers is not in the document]
owenpearson: I've opened an issue, and removed the workaround in 13d70f8

PR #817, opened 2021-10-12
Rest request emits error properly, if our server answers with empty or unparsable body

### LINKED ISSUE #676: Rest#request can't cope with an empty response body
Not really a problem at the moment -- every endpoint that currently exists that we might use it for gives a response -- but still, a weird restriction and a spec violation (per HP1 it should be able to copy with an empty response)

vodolaz095:
nodejs http client implementation can now work if server provides error page in a way it is either empty or cannot be parsed by JSON.parse or `msgpack.decode`. 

Closes #676


owenpearson:
@vodolaz095 We have an echo server which allows us to test for arbitrary responses like this, for example: a GET to `echo.ably.io/?body=${jsonBody}&status=200&type=json` will allow you to write a test which does just that (make sure you remember to URL encode the json body). Would you be able to have a go at writing that test?

owenpearson:
@vodolaz095 Sorry I didn't spot this before but your new tests are failing in web browsers. I think we need the same changes in `browser/lib/util/https.js`.

vodolaz095:
@owenpearson i cannot say i totally understand this code, but i think `try...catch` block resulting in similiar error emited is implemented.

this is `try..catch` logic similiar to one i implemented
https://github.com/ably/ably-js/blob/13d70f8b4f4723f6b92b60e2c31f96d518d630c1/browser/lib/transport/xhrrequest.js#L185-L219

so, i think similiar behaviour is already implemented in frontend code.

Do you think we need to implement same error names to be emitted?



This error is returned, when client side javascript client fails to parse body:
https://github.com/ably/ably-js/blob/13d70f8b4f4723f6b92b60e2c31f96d518d630c1/browser/lib/transport/xhrrequest.js#L217

vodolaz095:
i think i'm stuck...

in browser:


on serverside


i mean, i terminate browser request on 204 status code, but there is still somehow appearing its unenveloping by
https://github.com/ably/ably-js/blob/d0d0bca390254e8f28a4c4b98f08ccf79aafdf86/common/lib/client/resource.js#L31

PR #830, opened 2021-10-14
Convert XHRRequest to TypeScript

owenpearson:


QuintinWillison (approved):
If I ignore the wildly varying indentation (🙈) then all good. 😉 

(formatting, presumably, to be addressed under #684 and/or #685? ... and, in which case, perhaps we should consider moving that work up the priority list as it would clean up reviews generally across the board in this repository 🤔)

owenpearson:
@QuintinWillison I've been trying to not change indentation styles in these PRs and as far as I'm aware this PR is clean in that sense. I can see that `bufferutils.ts` has a lot of indentation changes in the diff but that's because I unwrapped everything from the top level `module` block, so that would have been the case even with `prettier`.

PR #831, opened 2021-10-14
Update NodeJS versions in CI to current LTS versions

owenpearson:


vodolaz095:
so, i agree, deprecating 10x version makes sense, funny things there is still somebody using nodejs of 8.x branch... i hope he/she doesn't have few million lines of code system...

PR #832, opened 2021-10-14
Convert CometTransport to TypeScript

owenpearson:


common/lib/transport/comettransport.ts
```diff
@@ -1,165 +1,177 @@
...
-			/* Likely a network or transport error of some kind. Certainly not fatal to the connection */
-			return false;
-		}
+import IXHRRequest from '../../types/IXHRRequest';
+import TokenDetails from '../../types/TokenDetails';
+
+type ConnectionManager = any;
+type TransportParams = any;
+
+const REQ_SEND = 0,
```
QuintinWillison, line 15: I'm yet to learn TypeScript fully myself, but this looks like [an `enum`](https://www.typescriptlang.org/docs/handbook/enums.html). 🤔

common/lib/transport/comettransport.ts
```diff
@@ -1,165 +1,177 @@
...
 	}
-	Utils.inherits(CometTransport, Transport);
 
-	CometTransport.REQ_SEND = REQ_SEND;
-	CometTransport.REQ_RECV = REQ_RECV;
-	CometTransport.REQ_RECV_POLL = REQ_RECV_POLL;
-	CometTransport.REQ_RECV_STREAM = REQ_RECV_STREAM;
+	abstract createRequest(uri: string, headers: Record<string, string> | null, params?: Record<string, unknown> | null, body?: unknown, requestMode?: number): IXHRRequest;
+
+	static REQ_SEND = REQ_SEND;
```
QuintinWillison, line 70: Curious here around use of `static`, relating to [previous comment](https://github.com/ably/ably-js/pull/834#discussion_r736296323).

common/lib/transport/comettransport.ts
owenpearson: Good point, I've added an XHRStates enum in e9d0e80.

common/lib/transport/comettransport.ts
owenpearson: As in `protocol.js`, this is an artefact of the way the library was modularised before webpack (basically you could only export one thing from a module so adding a static property was a hacky way to add a named export to the module alongside the default export). Fortunately, by creating the XHRStates enum in e9d0e80 this has already been taken care of.

PR #850, opened 2021-11-05
Convert PaginatedResource to TypeScript

owenpearson:


common/lib/client/paginatedresource.ts
```diff
@@ -0,0 +1,195 @@
...
+import { PaginatedResultCallback } from '../../types/utils';
+
+// TODO: Replace this once rest.js is converted to TypeScript
+type Rest = any;
+
+export type BodyHandler = (body: unknown, headers: Record<string, string>, packed?: boolean) => any;
+
+function getRelParams(linkUrl: string) {
+	const urlMatch = linkUrl.match(/^\.\/(\w+)\?(.*)$/);
+	return urlMatch && Utils.parseQueryString(urlMatch[2]);
```
ikbalkaya, line 14: Should you be also checking `urlMatch.size > 2` ?

common/lib/client/paginatedresource.ts
```diff
@@ -0,0 +1,195 @@
...
+	const relParams: Record<string, Record<string, string>> = {};
+	for(let i = 0; i < linkHeader.length; i++) {
+		const linkMatch = linkHeader[i].match(/^\s*<(.+)>;\s*rel="(\w+)"$/);
+		if(linkMatch) {
+			const params = getRelParams(linkMatch[1]);
+			if(params)
+				relParams[linkMatch[2]] = params;
+		}
+	}
+	return relParams;
```
ikbalkaya, line 30: There are a couple of unsafe array accesses on this function. I am not sure if those functions have their corresponding unit tests but I think it would be a great way to discover and fix those edge cases

common/lib/client/paginatedresource.ts
```diff
@@ -0,0 +1,195 @@
...
+			Logger.logAction(Logger.LOG_ERROR, 'PaginatedResource.handlePage()', 'Unexpected error getting resource: err = ' + Utils.inspectError(err));
+			callback(err);
+			return;
+		}
+		let items, linkHeader, relParams;
+		try {
+			items = this.bodyHandler(body, headers, unpacked);
+		} catch(e) {
+			/* If we got an error, the failure to parse the body is almost certainly
+			 * due to that, so cb with that in preference to the parse error */
```
ikbalkaya, line 105: "so cb with that in preference to the parse error" I think there is a typo here

common/lib/client/paginatedresource.ts
```diff
@@ -0,0 +1,195 @@
...
+	hasNext?: () => boolean;
+	isLast?: () => boolean;
+
+	constructor(resource: PaginatedResource, items: T[], relParams?: Record<string, any>) {
+		this.resource = resource;
+		this.items = items;
+
+		if(relParams) {
+			if('first' in relParams) {
+				this.first = (cb: (result?: ErrorInfo | null) => void) => {
```
ikbalkaya, line 137: Ah. Now I can get what 'cb' is. Can this be named `callback`?

ikbalkaya (dismissed):
Looks good overall. However there is a pattern I recognise where arrays or parsed content is used without safety check. I'm not sure if there is a plan in pipeline but I think writing some unit tests that covers all cases would make those a bit more obvious.

common/lib/client/paginatedresource.ts
QuintinWillison: I can't disagree with this. 😄 Though, perhaps, worth acknowledging that this variable name was inherited from the original JS implementation.

QuintinWillison (approved):

common/lib/client/paginatedresource.ts
owenpearson: 2e1fd3b1

common/lib/client/paginatedresource.ts
owenpearson: cb is short for callback but I agree that it's still not worded great so I improved it in cdfa6f7

common/lib/client/paginatedresource.ts
owenpearson: 500ea75

common/lib/client/paginatedresource.ts
owenpearson: You're right, but I don't want to change this now. It's always worked up until this point so I would assume it's safe, even if it would take a lot of work to prove it.

common/lib/client/paginatedresource.ts
QuintinWillison: Resolving this conversation on @ikbalkaya's behalf.

common/lib/client/paginatedresource.ts
QuintinWillison: Resolving this conversation on @ikbalkaya's behalf.

common/lib/client/paginatedresource.ts
QuintinWillison: Resolving this conversation on @ikbalkaya's behalf.

common/lib/client/paginatedresource.ts
QuintinWillison: Resolving this conversation on @ikbalkaya's behalf.

PR #853, opened 2021-11-05
Convert Channel to TypeScript

owenpearson:


common/lib/client/channel.ts
```diff
@@ -0,0 +1,178 @@
...
+function normaliseChannelOptions(options?: ChannelOptions) {
+	const channelOptions = options || {};
+	if(channelOptions.cipher) {
+		if(!Crypto) throw new Error('Encryption not enabled; use ably.encryption.js instead');
+		const cipher = Crypto.getCipher(channelOptions.cipher);
+		channelOptions.cipher = cipher.cipherParams;
+		channelOptions.channelCipher = cipher.cipher;
+	} else if('cipher' in channelOptions) {
+		/* Don't deactivate an existing cipher unless options
+			* has a 'cipher' key that's falsey */
```
ikbalkaya, line 43: Hmm. I wasn't able to associate this comment and the lines below 🤔

common/lib/client/channel.ts
```diff
@@ -0,0 +1,178 @@
...
+
+	constructor(rest: Rest | Realtime, name: string, channelOptions?: ChannelOptions) {
+		super();
+		Logger.logAction(Logger.LOG_MINOR, 'Channel()', 'started; name = ' + name);
+		this.rest = rest;
+		this.name = name;
+		this.basePath = '/channels/' + encodeURIComponent(name);
+		this.presence = new Presence(this);
+		this.channelOptions = {};
+		this.channelOptions = normaliseChannelOptions(channelOptions);
```
ikbalkaya, line 65: Will this second assignment of channelOptions not override the first one ?

common/lib/client/channel.ts
```diff
@@ -0,0 +1,178 @@
...
+		const options = this.channelOptions;
+		(new PaginatedResource(rest, this.basePath + '/messages', headers, envelope, function(body: any, headers: Record<string, string>, unpacked?: boolean) {
+			return Message.fromResponseBody(body, options, unpacked ? undefined : format);
+		})).get(params as Record<string, unknown>, callback);
+	}
+
+	publish(): void | Promise<void> {
+		const argCount = arguments.length,
+			first = arguments[0],
+			second = arguments[1];
```
ikbalkaya, line 108: Is it safe to directly access arguments[N]?

common/lib/client/channel.ts
```diff
@@ -0,0 +1,178 @@
...
+
+		if(typeof(callback) !== 'function') {
+			if(this.rest.options.promises) {
+				return Utils.promisify(this, 'publish', arguments);
+			}
+			callback = noop;
+		}
+
+		if(typeof first === 'string' || first === null) {
+			/* (name, data, ...) */
```
ikbalkaya, line 121: This comment may not be needed 🤔

common/lib/client/channel.ts
```diff
@@ -0,0 +1,178 @@
...
+			messages = [Message.fromValues({name: first, data: second})];
+			params = arguments[2];
```
ikbalkaya, line 123: Direct index access here and some more places below

ikbalkaya (dismissed):

common/types/channel.d.ts
```diff
@@ -10,5 +10,5 @@ export interface ChannelOptions {
     algorithm: string;
     encrypt: Function;
     decrypt: Function;
-  };
+  } | null;
```
QuintinWillison, line 13: What does this do?

common/lib/client/channel.ts
owenpearson: So the first branch of the if statement is `if (channelOptions.cipher)` which checks if `channelOptions.cipher` is truthy. The `else if` branch merely checks if `channelOptions.cipher` exists at all, so it will only trigger if the `cipher` is falsey. So essentially this branch will 'deactivate' the cipher if `channelOptions.cipher` is falsey. Perhaps it would be equivalent to say 'Deactivate the existing cipher if the 'cipher' key is falsey'.

common/lib/client/channel.ts
owenpearson: Yup, if you don't provide an argument here then it will return `undefined` and we handle all those cases below

common/lib/client/channel.ts
owenpearson: I think it's helpful, it's just saying that if we reach this branch the first argument is the event name and the second argument is the data to publish

common/lib/client/channel.ts
owenpearson: Will return undefined if there's no third argument and that's handled below so it's safe :)

common/types/channel.d.ts
owenpearson: Just types it so that `channelCipher` can be null. There's a place in `channel.ts` which sets options.channelCipher to null so this had to be nullable.

common/lib/client/channel.ts
QuintinWillison: Resolving this conversation on @ikbalkaya's behalf.

common/lib/client/channel.ts
QuintinWillison: Resolving this conversation on @ikbalkaya's behalf.

common/lib/client/channel.ts
QuintinWillison: Resolving this conversation on @ikbalkaya's behalf.

common/lib/client/channel.ts
QuintinWillison: Resolving this conversation on @ikbalkaya's behalf.

common/lib/client/channel.ts
QuintinWillison: Resolving this conversation on @ikbalkaya's behalf.

QuintinWillison (approved):

PR #854, opened 2021-11-05
Add typings for ClientOptions

owenpearson:


common/types/ClientOptions.ts
```diff
@@ -0,0 +1,98 @@
...
+  transports?: Transport[];
+
+  httpMaxRetryCount?: number;
+  restAgentOptions?: { keepAlive: boolean, maxSockets: number };
+  pushFullWait?: boolean;
+  checkChannelsOnResume?: boolean;
+  plugins?: Record<string, unknown>;
+  agents?: string[];
+}
+
```
ikbalkaya, line 80: Here too,  there are tons of optional properties. I am not sure whether the usage of optionals are similar in JS but I would use them sparingly just for the values that might be null.  I would definitely consider rescanning those and make most of those non-optionals

common/types/ClientOptions.ts
```diff
@@ -0,0 +1,98 @@
...
+export type DeprecatedClientOptions = Modify<ClientOptions, {
+  host?: string;
+  wsHost?: string;
+  queueEvents?: boolean;
+  promises?: boolean;
+  headers?: Record<string, string>;
+  maxMessageSize?: number;
+  timeouts?: Record<string, number>;
+}>
```
ikbalkaya, line 89: Lots of optionals :)

ikbalkaya (dismissed):
I realised lots of usage of optional properties in this PR. I think they make things flexible but also make code a bit more vulnerable. (correct me if this is not the case in Javascript) my understanding of optional usage is that they should be used with a purpose and choosing them is part of design of the app. For example, by design the autoconnect property is intended to be true by default, but making it optional you declare it so that it may not even exist. I hope that is helpful.

common/types/utils.d.ts
```diff
@@ -1,3 +1,4 @@
 export type StandardCallback<T> = (err?: ErrorInfo | null, result?: T) => void;
 export type ErrCallback = (err?: ErrorInfo | null) => void;
 export type PaginatedResultCallback<T> = StandardCallback<PaginatedResult<T>>;
+export type Modify<T, R> = Omit<T, keyof R> & R;
```
QuintinWillison, line 4: If you're adding a brand new type here, which I think you are, then this would be a great opportunity to document what it does / conveys. It's quite opaque.

QuintinWillison (changes requested):
I acknowledge that it's perhaps unrealistic to try to add API commentary to every type as part of the JS to TS conversion, however I think new types should be easy to document now.

common/types/AuthOptions.ts
```diff
@@ -0,0 +1,33 @@
...
+  /**
+   * A URL that the library may use to obtain a token string (in plain text format), or a signed TokenRequest or TokenDetails (in JSON format).
+   **/
+  authUrl?: string;
+  key?: string;
+  queryTime?: boolean;
+  token?: TokenDetails | string;
+  tokenDetails?: TokenDetails;
+  useTokenAuth?: boolean;
+
```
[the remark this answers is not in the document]
owenpearson: These are just options that might be passed in by a user of the library so they are all optional. For example, someone might authenticate with a key, whereas someone else might authenticate with an authUrl.

common/types/ClientOptions.ts
owenpearson: As above, these are options passed in by library users so they're all optional.

common/types/ClientOptions.ts
owenpearson: Above applies here too.

common/types/AuthOptions.ts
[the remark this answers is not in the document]
QuintinWillison: Resolving this conversation on @ikbalkaya's behalf.

common/types/ClientOptions.ts
QuintinWillison: Resolving this conversation on @ikbalkaya's behalf.

common/types/ClientOptions.ts
QuintinWillison: Resolving this conversation on @ikbalkaya's behalf.

QuintinWillison (approved):

PR #855, opened 2021-11-05
Convert Rest to TypeScript

owenpearson:
When reviewing bear in mind this PR contains a breaking change; creating an `Ably.Rest` instance without using the `new` keyword will no longer work after this change. This change was made because the existing logic to support this wouldn't work nicely with TypeScript but tbh I think we shouldn't support it anyway.


common/lib/client/rest.ts
```diff
@@ -0,0 +1,225 @@
...
+		host: string,
+		validUntil: number,
+	};
+	serverTimeOffset: number | null;
+	auth: Auth;
+	channels: Channels;
+	push: Push;
+
+	constructor(options: ClientOptions | string) {
+		/* normalise options */
```
ikbalkaya, line 35: I think this comment is not linked with code provided in the next line

common/lib/client/rest.ts
```diff
@@ -0,0 +1,225 @@
...
+		/* process options */
+		if(normalOptions.key) {
+			const keyMatch = normalOptions.key.match(/^([^:\s]+):([^:.\s]+)$/);
+			if(!keyMatch) {
+				const msg = 'invalid key parameter';
+				Logger.logAction(Logger.LOG_ERROR, 'Rest()', msg);
+				throw new Error(msg);
+			}
+			normalOptions.keyName = keyMatch[1];
+			normalOptions.keySecret = keyMatch[2];
```
ikbalkaya, line 59: Access to array elements without check ☝️

common/lib/client/rest.ts
```diff
@@ -0,0 +1,225 @@
...
+		if(this.options.headers)
+			Utils.mixin(headers, this.options.headers);
+		const timeUri = (host: string) => { return this.authority(host) + '/time' };
+		Http.get(this, timeUri, headers, params as RequestParams, (err?: ErrorInfo | ErrnoException | null, res?: unknown, headers?: Record<string, string>, unpacked?: boolean) => {
+			if(err) {
+				_callback(err);
+				return;
+			}
+			if(!unpacked) res = JSON.parse(res as string);
+			const time = (res as number[])[0];
```
ikbalkaya, line 132: Unsafe access to array (and potentially unsafe cast?) ☝️

ikbalkaya (changes requested):

QuintinWillison (approved):
We have to balance up whether we feel there is 'risk' of trying to deal with some of the (historically present) issues that @ikbalkaya is spotting... perhaps it's better to defer them to later by creating issues, so that the scope of these PRs can focus on pure JS to TS conversion.

In respect of the breaking API change, I assume @owenpearson is raising this because customers could encounter this. In which case, we need a mechanism for ensuring this makes its way into the change log - and, perhaps even, is reflected in the SemVer bump we make to this library (see [internal ADR 34](https://ably.atlassian.net/wiki/spaces/ENG/pages/1505886246/ADR34+SDK+versioning+policy), if we get that over the line soon).

owenpearson:
@QuintinWillison With regards to the breaking change, this particular feature (being able to initialise clients without the `new` keyword) is an undocumented feature and I'd be surprised if anyone was actually using it so it doesn't feel like a significant change. I'm pretty much certain that there's no way to do this in TypeScript without getting a compiler error so it feels necessary. Good point about making sure it makes it in to the changelog though, not sure how to ensure that.

common/lib/client/rest.ts
owenpearson: Yeah it's definitely not helpful, I've removed it in 73294d4

common/lib/client/rest.ts
owenpearson: the REST API will always return an array with a single number here so I think it's safe enough

common/lib/client/rest.ts
owenpearson: Inclined to agree but this is a pretty fundamental part of the library and hasn't caused a problem so far so I say we keep it for now

QuintinWillison: Resolving this conversation on @ikbalkaya's behalf.

common/lib/client/rest.ts
QuintinWillison: Resolving this conversation on @ikbalkaya's behalf.

(though I would note that I do have sympathy with the opening comment... an API boundary is a place where we _should_ be more diligent in checking things / not making assumptions of correctness of external systems)

Issue #805, opened 2021-10-05
Continuous Delivery via GitHub workflow


QuintinWillison:
I think we should model something on what we do in `ably-asset-tracking-android`'s [publish workflow](https://github.com/ably/ably-asset-tracking-android/blob/main/.github/workflows/publish.yml).

The only possible reason we may wish not do do it that way is that there's a chance the publish workflow will fail, and in which case we then end up with a Git tag that's been pushed optimistically to indicate that we've done a release when there may actually be subsequent commits required to the `main` branch in order to fix it up into a state where it's _actually, really_ able to do a release (e.g. [to npm](https://www.npmjs.com/package/ably)).

Therefore, perhaps we should be refactoring our general release process (across all SDKs) so that the tag is not pushed to GitHub until we know it definitely has (or can, in the case of [ably-go](https://github.com/ably/ably-go) where the release _is_ the tag) be released. 

Relates to https://github.com/ably/ably-js/issues/551.

Also see GitHub's blog post (23 September 2021): [Announcing npm’s new access token format](https://github.blog/2021-09-23-announcing-npms-new-access-token-format/)

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1263) by [Unito](https://www.unito.io)

jamienewcomb:
@QuintinWillison you mentioned in the stand up we should reach descriptions before closing tickets. We have gone through a few people now who don't actually understand what is being asked off here. after reading the description I am not sure either. 

Could you add an acceptance criteria please so it is clear for who ever picks this up

owenpearson:
Closing in favour of #893

PR #863, opened 2021-11-22
Add release PR stage to release process

owenpearson:
Modifies the release process slightly so that a release PR can be created and reviewed before the release. Previously the whole release process happened directly on the `main` branch and didn't allow for a release PR to be created.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,14 +16,18 @@
 ## Release Process
 
 1. Make sure the tests are passing in CI for the branch you're building
-2. Update the CHANGELOG.md with any customer-affecting changes since the last release and add this to the git index
-3. Run `npm run grunt -- release:patch` (or: "major", "minor", "patch", "prepatch") - creates commit and local tag
-4. Ensure that the infrastructure repository is in the same directory as `ably-js` and is up to date with the remote main branch
-5. Run `npm run grunt -- release:deploy` (requires the SDKDeveloper AWS role) - pushes commit and tag, then publishes to the Ably CDN
-6. Run `npm publish .` (should require OTP) - publishes to NPM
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -16,14 +16,18 @@
...
-6. Run `npm publish .` (should require OTP) - publishes to NPM
-7. Visit https://github.com/ably/ably-js/tags and add release notes to the release (generally you can just copy the notes you added to the CHANGELOG)
-8. For nontrivial releases: update the ably-js submodule ref in the realtime repo
-9. Update the Ably changelog with these changes (again, you can just copy the notes you added to the CHANGELOG)
+2. Create a new branch for the release, for example `release/1.2.3`
+3. Update the CHANGELOG.md with any customer-affecting changes since the last release and add this to the git index
+4. Run `npm version <VERSION_NUMBER> --no-git-tag-version` with the new version and add the changes to the git index
+5. Create a PR for the release branch
+6. Once the release PR is landed to the `main` branch, checkout the `main` branch locally (remember to pull the remote changes) and run `npm run build`
+7. Run `git tag <VERSION_NUMBER>` with the new version and push the tag to git
```
QuintinWillison, line 24: I would ideally only push the tag _after_ it's been published (Ably CDN _and_ NPM)... is that possible?

QuintinWillison (changes requested):

CONTRIBUTING.md
owenpearson: Not yet. The infrastructure script `ably-env deploy javascript` uses the tag to publish to the CDN. I agree that we should ideally push the tag after publishing but I think we deal with that when we move publishing into a CD workflow.

QuintinWillison: Right, got you. Thanks for explaining. Will be addressed under https://github.com/ably/ably-js/issues/805.

QuintinWillison (approved):

PR #873, opened 2022-01-13
Remove mostRecentMsg when detaching from channel

owenpearson:
Resolves an issue where users could attach to a channel with rewind, detach from the channel and resubscribe, and not receive the rewind message the second time.

### DIFF
--- a/spec/realtime/channel.test.js
+++ b/spec/realtime/channel.test.js
@@ -1536,5 +1536,32 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
                       done();
                     });
                 });
+
+                it('rewind works on channel after reattaching', function (done) {
+                    var realtime = helper.AblyRealtime({ transports: [helper.bestTransport] });
+                    var channelName = 'rewind_after_detach';
+                    var channel = realtime.channels.get(channelName);
+                    var channelOpts = { params: { rewind: '1' } };
+                    channel.setOptions(channelOpts);
+
+                    var subscriber = function(message) {
+                      expect(message.data).to.equal('message');
+                      channel.unsubscribe(subscriber);
+                      channel.detach(function(err) {
+                        if (err) {
+                          closeAndFinish(done, realtime, err);
+                          return;
+                        }
+                        channel.subscribe(function(message) {
+                          expect(message.data).to.equal('message');
+                          closeAndFinish(done, realtime);
+                        });
+                      });
+                    }
+
+                    channel.publish('event', 'message');
+
+                    channel.subscribe(subscriber);
+                });
 	});
 });


QuintinWillison (approved):
The fix looks plausible, though I can't say that my approval carries that much weight as I've not looked at the rest of the codebase to see where these variables are used and/or otherwise mutated.

Also, it would be nice to see a test for this behaviour. Is that something you considered, or were you planning to add that in a subsequent PR?

QuintinWillison:
Oh, and if other client libraries could be making the same mistake, does that mean this needs a spec update?

owenpearson:
@QuintinWillison Adding a test is a good idea. I'll add that to this PR soon.

It won't require a spec update, the `mostRecentMsg` thing is an implementation detail of ably-js that isn't in the spec (and AFAIK isn't in the other libraries). I will check the other libs to make sure they behave in the same way though...

SimonWoolf (approved):
The logic seems reasonable. Reaching into connectionmanager to change one of its internal members is maybe a little messy, if this was realtime I'd suggest doing `ConnectionManager#clearDedupCacheIfChannel(this.name)` or something, but it'd be far from the worst messy thing to happen in ably-js so shrug.

(The indentation is messed up, surprised the linter didn't catch that now we have one -- can we chuck a `"indent": ["error", "tab", {"SwitchCase": 1}]` in the .eslintrc.json?)

SimonWoolf:
> there is inconsistent indentation all over the place in ably-js so this would turn in to a pretty messy PR.

Grepping the main codebase there are very few places that use spaces for indentation, and they're all from recent PRs.
```
~/ably/ably-js [main] $ gr "^    " common/ nodejs/ browser/
browser/fragments/platform-browser.js:        agent: 'browser',
browser/fragments/platform-nativescript.js:        agent: 'nativescript',
browser/fragments/platform-reactnative.js:        agent: 'reactnative',
common/lib/client/realtimechannel.js:                        case 'suspended':
common/lib/client/realtimechannel.js:                                this.notifyState('detached');
common/lib/client/realtimechannel.js:                                callback();
common/lib/client/realtimechannel.js:                                break;
common/lib/client/realtimechannel.js:                                if (this.connectionManager.mostRecentMsg && this.connectionManager.mostRecentMsg.channel === this.name) {
common/lib/client/realtimechannel.js:                                  this.connectionManager.mostRecentMsg = null;
common/lib/client/realtimechannel.js:                                }
common/lib/util/defaults.js:        if(options.agents) {
common/lib/util/defaults.js:          for(var key in options.agents) {
common/lib/util/defaults.js:            Defaults.agent += ' ' + key + '/' + options.agents[key];
common/lib/util/defaults.js:          }
common/lib/util/defaults.js:        }
```

I wasn't suggesting you should add linting in this pr specifically and run it over the whole codebase - clearly that'd be a separate PR. But I do think we shouldn't be deliberately merging PRs with indentation that makes the code really quite difficult to read for someone who doesn't have tabstop=8. I mean, look at it:



There is a `.editorconfig` file in this repo (and most other ably repos) that defines the style, configure your editor to respect that then it'll automatically adopt the right formatting settings when editing files in every repo.

Issue #882, opened 2022-02-06
TypeError: Cannot read properties of undefined (reading 'bind') in browser


tencircles:
When using the ably in the browser the following error occurs when the library is imported. I am using nextjs and deploying on vercel, but I don't believe that should cause this issue.

`TypeError: Cannot read properties of undefined (reading 'bind')`

Is the library not meant for use on the client side? I'm a bit confused.

owenpearson:
Hi @tencircles, thanks for reporting this. The library should generally work fine on the client side. To help us understand what's causing this issue we would need some more information to help with our investigation. In particular we would like to know:
- What version of NextJS and Ably are you using?
- Which web browser are you seeing the error in?
- Does the error message come with a stack trace? If so could you post it here in the comments?

tencircles:
Hey there @owenpearson .

I have tried Ably `^1.2.18-beta.0` and `^1.2.17` both produce the same error.
I am using next version `12.0.8`.
I am seeing the error in chrome, safari, and FF.

Below is a stack trace. 

```
index.js:430 TypeError: Cannot read properties of undefined (reading 'bind')
    at Object.<anonymous> (ably-commonjs.js:829:1)
    at Object.<anonymous> (ably-commonjs.js:858:85)
    at c (ably-commonjs.js:42:1)
    at Module.<anonymous> (ably-commonjs.js:12377:1)
    at c (ably-commonjs.js:42:1)
    at ably-commonjs.js:106:11
    at ably-commonjs.js:107:1
    at ably-commonjs.js:20:28
    at Object.8351 (ably-commonjs.js:22:1)
    at g (bootstrap:22:1)
```

It seems to be pointing at this line

`nextTick: typeof setImmediate !== 'undefined' ? global.setImmediate.bind(global) : function(f) { setTimeout(f, 0); },`

Here the value of `setImmediate.bind` is `undefined` causing the above error.
I can only think of two causes.

1. Something in nextjs or webpack is modifying Function.prototype (very unlikely).
2. The value of `global` in the browser is referencing something other than `window`.

Any input is appreciated.

tencircles:
Potential third cause would be something overwriting `window.setImmediate` with something other than a function, but just a quick check in the console reveals this not to be the case.

owenpearson:
[...]

PR #883, opened 2022-02-07
fix: refer to global.setImmediate in nextTick check

tencircles:
Fixes a strange edge case that occurs wherein sometimes global may not refer to the actual global object in a given environment.


QuintinWillison:
Thanks. I see @tencircles has already been conversing with @owenpearson on this, and this PR comes out of [that](https://github.com/ably/ably-js/issues/882#issuecomment-1031890278).

Peter-Maguire (approved):

owenpearson (approved):
LGTM, thanks

PR #884, opened 2022-02-09
fix spelling/grammar errors and JSDoc inaccuracies

Peter-Maguire:
A (hopefully) small PR to fix some inaccuracies and spelling/grammar mistakes in the JSDoc for typescript functions

### DIFF
--- a/common/lib/client/auth.ts
+++ b/common/lib/client/auth.ts
@@ -826,7 +826,7 @@ class Auth {
 				return;
 			}
 			/* RSA4b1 -- if we have a server time offset set already, we can
-			 * auto remove expired tokens. Else just use the cached token. If it is
+			 * automatically remove expired tokens. Else just use the cached token. If it is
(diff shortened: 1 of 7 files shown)


common/lib/client/auth.ts
```diff
@@ -825,7 +826,7 @@ class Auth {
 				return;
 			}
 			/* RSA4b1 -- if we have a server time offset set already, we can
-			 * autoremove expired tokens. Else just use the cached token. If it is
+			 * auto remove expired tokens. Else just use the cached token. If it is
```
owenpearson, line 829: I don't think this is correct either. How about "automatically remove"?

owenpearson (changes requested):

common/lib/client/auth.ts
Peter-Maguire: I think grammatically it would be 'auto-remove' but 'automatically remove' is definitely clearer.

PR #889, opened 2022-02-14
Publish builds and sourcemaps to CDN from GitHub Actions

### LINKED ISSUE #191: Source maps should be generated and published to our CDN and NPM
Source maps can be generated on `npm run build` by setting the webpack [devtool configuration option](https://webpack.js.org/configuration/devtool/).

The source code should be published as well so that users can jump to the source of an error from a stack trace.

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-276) by [Unito](https://www.unito.io)

Peter-Maguire:
Resolves #191 

This PR creates a tool and CI script for publishing library files and their associated source maps to the CDN.

The script takes an input tag (e.g 1.2.3), checks it out from git and then uploads the build folders contents to an S3 bucket, adding the major, minor and patch versions into the file names and uploading. e.g:
```
ably-1.2.3.js
ably-1.2.js
ably-1.js
```

This is also coupled with a github actions script to check out the tag, then run a grunt script which builds and runs this script.


owenpearson (changes requested):
Looks good on the whole. Couple of general comments:

1. It would be cool to see how this is gonna work with GitHub actions, are you planning on saving that step for a separate PR?
2. The formatting of your JS is a bit inconsistent and sometimes unorthodox. Things in particular that stick out to me are
- Missing semicolons
- Random double line breaks
- Lack of spacing around `=>` for arrow functions (this one's maybe more of a preference but I think it makes a lot of the iterator chaining quite jarring to read)

Once we resolve #684 this won't be so much of a problem but for now I think you should address these things.

scripts/README.md
```diff
@@ -0,0 +1,10 @@
+These are scripts used by the CI for deployment, etc.
+
+
+### cdn_deploy.js
+
+Deploys code and source maps to the CDN.
+Requires aws-sdk, nodegit and minimist:
+```
+# npm install --no-save aws-sdk nodegit minimist
```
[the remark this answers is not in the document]
Peter-Maguire: I was hesitant to add them into the package since 99% of the time this script won't be used, but if it's okay I'm happy to add them in to dev dependencies

scripts/README.md
```diff
@@ -0,0 +1,10 @@
+These are scripts used by the CI for deployment, etc.
+
```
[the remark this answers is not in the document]
Peter-Maguire: Fixed, the readme was a placeholder so I've replaced it with something more descriptive.

scripts/cdn_deploy.js
```diff
@@ -0,0 +1,124 @@
...
+			if(stat.isDirectory()) return scanDir(excludes)(filePath);
+			return filePath;
+		})
+	}
+}
+
+
+function fatal(message){
+	console.error(message);
+	process.exit(1);
```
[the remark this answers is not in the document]
Peter-Maguire: Good point, I've replaced those calls.

scripts/cdn_deploy.js
```diff
@@ -0,0 +1,124 @@
...
+			}
+		}
+		console.log("Success!");
+	}catch(e){
+		fatal(e);
+	}
+}
+
+function upload(s3, upload){
+	return new Promise((fulfill, reject)=>{
```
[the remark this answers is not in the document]
Peter-Maguire: Changed it. I've always used fulfill for whatever reason, but I will stick with the standard.

scripts/README.md
```diff
@@ -0,0 +1,20 @@
...
+**--root**: The base directory inside the bucket to deploy to, defaults to `lib`.
+**--s3Key**: S3 Access Key. Can also be set with AWS_ACCESS_KEY env variable.
+**--s3Secret**: S3 Secret Access Key. Can also be set with AWS_SECRET_ACCESS_KEY env variable.
+**--path**: The local path to retrieve source files from. Defaults to `.`.
+**--includeDirs**: A comma separated list of directories to include. Defaults to `.`.
+**--excludeDirs**: A comma separated list of directories to exclude. Defaults to `node_modules,.git`.
+**--fileRegex**: A regular expression to test file names against for upload. Defaults to `^(?!\.).*\.(map|js|html)$`.
+**--endpoint**: Optional. The S3 endpoint to deploy to.
+**--skipCheckout**: Optional. Skip checking out the branch before running.
+
```
owenpearson, line 19: Double newline here

owenpearson (changes requested):
Looks great 👍

Can we update the release process in `CONTRIBUTING.md`?(sorry, should have asked about this when I reviewed last time)
 
I've also opened #893 which I think should be our next step here.

CONTRIBUTING.md
```diff
@@ -22,12 +22,11 @@
...
 7. Run `git tag <VERSION_NUMBER>` with the new version and push the tag to git
-8. Ensure that the infrastructure repository is in the same directory as `ably-js` and is up to date with the remote main branch
-9. Run `npm run grunt -- release:deploy` (requires the SDKDeveloper AWS role) - pushes commit and tag, then publishes to the Ably CDN
-10. Run `npm publish .` (should require OTP) - publishes to NPM
-11. Visit https://github.com/ably/ably-js/tags and add release notes to the release (generally you can just copy the notes you added to the CHANGELOG)
-12. For nontrivial releases: update the ably-js submodule ref in the realtime repo
-13. Update the [Ably Changelog](https://changelog.ably.com/) (via [headwayapp](https://headwayapp.co/)) with these changes (again, you can just copy the notes you added to the CHANGELOG)
+8. Run `npm publish .` (should require OTP) - publishes to NPM
+9. Visit https://github.com/ably/ably-js/tags and add release notes to the release (generally you can just copy the notes you added to the CHANGELOG)
+10. Run the GitHub action "Publish to CDN" with the latest version number
```
owenpearson, line 27: Let's swap 9 and 10 around - we don't want to prematurely announce a release until we know the publish workflow has worked correctly.

owenpearson (changes requested):

CONTRIBUTING.md
Peter-Maguire: Sure, although this is only temporary anyway until we move the npm publishing into the CI too so I don't want to dwell on the wording too much

owenpearson: Yeah that's fair, but I would have thought that particular step would stay even after the NPM CD stuff is done

owenpearson (approved):

CONTRIBUTING.md
```diff
@@ -22,12 +22,11 @@
...
 6. Once the release PR is landed to the `main` branch, checkout the `main` branch locally (remember to pull the remote changes) and run `npm run build`
 7. Run `git tag <VERSION_NUMBER>` with the new version and push the tag to git
-8. Ensure that the infrastructure repository is in the same directory as `ably-js` and is up to date with the remote main branch
-9. Run `npm run grunt -- release:deploy` (requires the SDKDeveloper AWS role) - pushes commit and tag, then publishes to the Ably CDN
-10. Run `npm publish .` (should require OTP) - publishes to NPM
-11. Visit https://github.com/ably/ably-js/tags and add release notes to the release (generally you can just copy the notes you added to the CHANGELOG)
-12. For nontrivial releases: update the ably-js submodule ref in the realtime repo
-13. Update the [Ably Changelog](https://changelog.ably.com/) (via [headwayapp](https://headwayapp.co/)) with these changes (again, you can just copy the notes you added to the CHANGELOG)
+8. Run `npm publish .` (should require OTP) - publishes to NPM
+9. Run the GitHub action "Publish to CDN" with the new tag name
```
QuintinWillison, line 26: "with the new tag name" is incorrect. I would avoid being too specific here, as the `description` on the `version` input to the 'Publish to CDN' workflow already says enough. Perhaps take inspiration from how the 'Run the publish workflows' step is written [in `ably-asset-tracking-android`](https://github.com/ably/ably-asset-tracking-android/blob/main/CONTRIBUTING.md#release-process)?

scripts/README.md
```diff
@@ -0,0 +1,18 @@
...
+### cdn_deploy.js
+
+Deploys correctly versioned code and source maps to the CDN.
+
+Arguments:
+
+**--bucket**: The S3 bucket name to deploy to, defaults to `cdn.ably.io`.
+**--root**: The base directory inside the bucket to deploy to, defaults to `lib`.
+**--s3Key**: S3 Access Key. Can also be set with AWS_ACCESS_KEY env variable.
```
QuintinWillison, line 11:
```suggestion
**--s3Key**: S3 Access Key. Can also be set with `AWS_ACCESS_KEY` env variable.
```

scripts/README.md
```diff
@@ -0,0 +1,18 @@
...
+**--s3Secret**: S3 Secret Access Key. Can also be set with AWS_SECRET_ACCESS_KEY env variable.
```
QuintinWillison, line 12:
```suggestion
**--s3Secret**: S3 Secret Access Key. Can also be set with `AWS_SECRET_ACCESS_KEY` env variable.
```

scripts/cdn_deploy.js
```diff
@@ -0,0 +1,137 @@
...
+const fs = require('fs');
+const path = require('path');
+const {exec} = require('child_process');
+const argv = require('minimist')(process.argv.slice(2));
+const S3_DEFAULT_BUCKET = "cdn.ably.io";
+const S3_DEFAULT_ROOT = "lib";
+
+async function run() {
+
+	let config = {
```
QuintinWillison, line 13: Is this tab indented? Would prefer 2 space if possible.

.github/workflows/publish-cdn.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.inputs.version }}
+      - name: Use Node.js 14.x
+        env:
+          AWS_ACCESS_KEY: ${{ secrets.AWS_ACCESS_KEY }}
```
QuintinWillison, line 20: I think both the `AWS_` secret keys could have a more qualified name, perhaps expand the prefix to `CDN_PUBLISH_AWS_`?

(as our use of secrets expands for this repository, over time, then we may have other AWS secrets come in that could clash... and 'renaming secrets' isn't a thing in GitHub so may as well not set ourselves up for that pain)

.github/workflows/publish-cdn.yml
```diff
@@ -0,0 +1,27 @@
...
+  publish:
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.inputs.version }}
+      - name: Use Node.js 14.x
+        env:
```
QuintinWillison, line 19: Why are you injecting secrets into the environment of this step? First, it's a third-party action so that bears risk. But, also, I'm assuming you need the secrets in the final step that runs the `node` script? 🤔

QuintinWillison (changes requested):
A few things I've spotted.

Plus, as a final thought, I note that `cdn_deploy.js` is a new piece of code. Could you consider adding JSDoc annotations to functions in that file from the outset? I think it's worth capturing that stuff as early as possible. A good example is the `upload` function - which is a bit opaque to look at initially (being a callback to `Promise` converter).

CONTRIBUTING.md
owenpearson: This was actually my suggestion and it's not immediately obvious to me why it's incorrect?

scripts/cdn_deploy.js
owenpearson: The whole repo is generall tab indented (see `.editorconfig`) I would love it if we could use 2 space indents but if we're enforcing a rule I'd like for it to be enforced across the entire codebase and it's just not worth overwriting all the git blame history for something like this 🙁

CONTRIBUTING.md
QuintinWillison: ah, sorry... ok, so in `ably-asset-tracking-android` we have a `v` prefix on release tag names. Therefore, actually, this comment should have been targeting [the `description`](https://github.com/ably/ably-js/blob/1edb0d258f86f9313bfbf14e5e7d39ac6ec6da3b/.github/workflows/publish-cdn.yml#L6) as incorrect!

Perhaps that description should be 'Release tag', not 'Version, without v prefix'. 🤔 

(FWIW, I like the `v` prefix as it makes it a _bit_ more difficult for the publish script to checkout something that's not a release tag 🤷)

owenpearson: Ah, that makes sense, I can see that we need to change that description 👍 

I like the `v` prefix too but it's a bit too late to change that for ably-js 😄

scripts/cdn_deploy.js
QuintinWillison: Fair point, though this is new code and if we know we will be adopting 2 space in near future then this is a file that will then not need to be touched at that point.

Peter-Maguire: If we are going to change the indentation, I think it would make more sense to do it all in one go whilst changing the .editorconfig and any linting rules at the same time to keep it all consistent

CONTRIBUTING.md
Peter-Maguire: I've updated the description of the gh action

.github/workflows/publish-cdn.yml
Peter-Maguire: You're right, I'm not too familiar with github actions so I wasn't sure how this actually worked, should be right now.

CONTRIBUTING.md
QuintinWillison: https://github.com/ably/ably-js/pull/889/commits/45484ca41c7b4e07bbe2bab6111ae81f25e87c38

.github/workflows/publish-cdn.yml
QuintinWillison: Fixed in https://github.com/ably/ably-js/pull/889/commits/04ff346c729c68e371eeada6a4f41aeaa13c8882

.github/workflows/publish-cdn.yml
QuintinWillison: Thanks 😁  https://github.com/ably/ably-js/pull/889/commits/04ff346c729c68e371eeada6a4f41aeaa13c8882

.github/workflows/publish-cdn.yml
```diff
@@ -0,0 +1,27 @@
...
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Release tag'
+        required: true
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    permissions:
```
QuintinWillison, line 12: Sorry, I should have noticed this in my first-pass review but I don't think you need any elevated permissions for this workflow. You're not pushing to GitHub Packages, which is what `packages` is needed for.

QuintinWillison (changes requested):

.github/workflows/publish-cdn.yml
Peter-Maguire: Good spot, removed it.

PR #890, opened 2022-02-15
Tie internal and external typings together

### LINKED ISSUE #857: TypeScript: Tie together internal typings with `ably.d.ts`
Currently in the TypeScript integration branch the internal types don't import from `ably.d.ts`. We should ensure that these types are used so that TypeScript won't compile if the library diverges from this type interface. Ideally this should be done in a way so that it's obvious when public API typings are used so that library developers can easily tell when they are making breaking API changes.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1358) by [Unito](https://www.unito.io)

Peter-Maguire:
Resolves #857 

Internal types now extend public types where appropriate, and use the public type if there is no difference.


Peter-Maguire:
There's some bits here where I had to change `null` to `undefined` to comply with the typings, I don't think they will cause problems, but it might be worth taking a look at them to make sure I'm not missing any reason why they might.

common/lib/client/auth.ts
```diff
@@ -10,10 +10,11 @@ import HmacSHA256 from 'crypto-js/build/hmac-sha256';
 import { stringify as stringifyBase64 } from 'crypto-js/build/enc-base64';
 import { createHmac } from 'crypto';
 import { ErrnoException, RequestCallback, RequestParams } from '../../types/http';
-import TokenDetails from '../../types/TokenDetails';
-import TokenParams from '../../types/TokenParams';
+import {Types} from '../../../ably';
 import { StandardCallback } from '../../types/utils';
 
+
```
owenpearson, line 16: I guess you didn't mean to commit this

ably.d.ts
```diff
@@ -122,7 +122,7 @@ declare namespace Types {
 		 * Can be used to explicitly recover a connection.
 		 * See https://www.ably.com/documentation/realtime/connection#connection-state-recovery
 		 */
-		recover?: string | ((lastConnectionDetails: {
+		recover?: boolean | string | ((lastConnectionDetails: {
```
owenpearson, line 125: Setting `options.recover` as a boolean is deprecated [see here](https://github.com/ably/ably-js/blob/main/common/lib/util/defaults.js#L144). I'm not sure why this change was needed here but you should be able to make a similar change to `DeprecatedClientOptions` and get the same effect.

common/lib/client/auth.ts
```diff
@@ -10,10 +10,11 @@ import HmacSHA256 from 'crypto-js/build/hmac-sha256';
 import { stringify as stringifyBase64 } from 'crypto-js/build/enc-base64';
 import { createHmac } from 'crypto';
 import { ErrnoException, RequestCallback, RequestParams } from '../../types/http';
-import TokenDetails from '../../types/TokenDetails';
-import TokenParams from '../../types/TokenParams';
+import {Types} from '../../../ably';
```
owenpearson, line 13: We should try and keep the styling of these consistent with the existing imports

common/lib/client/auth.ts
```diff
@@ -863,7 +864,7 @@ class Auth {
 
 
 	/* User-set: check types, '*' is disallowed, throw any errors */
-	_userSetClientId(clientId: string | null) {
+	_userSetClientId(clientId: string | undefined) {
```
owenpearson, line 867: What's the purpose behind this change? I'm sure it was necessary but it does concern me that two lines below we have "clientId must be either a string or null"

common/types/channel.d.ts
```diff
@@ -1,11 +1,7 @@
-type ChannelParams = { [key: string]: string };
-export type ChannelMode = 'PUBLISH' | 'SUBSCRIBE' | 'PRESENCE' | 'PRESENCE_SUBSCRIBE';
-type ChannelModes = Array<ChannelMode>;
```
owenpearson, line 1: You've got a newline at the start of this file

common/lib/client/auth.ts
Peter-Maguire: `_userSetClientId` takes `authOptions.clientId` which is an optional string, typescript treats it as if it were typed as `string | undefined` rather than `string | null`. The only option here was to either change the signature of this method, or change the public typings.

ably.d.ts
Peter-Maguire: Changing the signature of recover on `DeprecatedClientOptions` causes `ClientOptions` to no longer be compatible with it, meaning functions that take `ClientOptions` instead of `DeprecatedClientOptions` throw an error instead: 

```
Argument of type 'DeprecatedClientOptions' is not assignable to parameter of type 'ClientOptions'.   
Types of property 'recover' are incompatible.
```

common/lib/client/auth.ts
Peter-Maguire: Fixed.

ably.d.ts
owenpearson: Ah that's annoying. Unfortunately it's not really an option to make this change to the public API since setting `options.recover` to a boolean is deprecated. Could you look in to alternatives? I'm happy for you to use a workaround like `any` just to get this over the line and then we can deal with it later, just need to make sure we're not changing the public type interface.

Peter-Maguire: Got it, there's only one place where this is checked, so [I've modified that](https://github.com/ably/ably-js/pull/890/files#diff-915197b352a9482f9518c50eb5909c5820d931511a1aeb8a21836e32c04b1d40R178) to convert to `any` first

common/lib/client/auth.ts
Peter-Maguire: Changed to `import * as API` as we talked about

owenpearson (approved):
LGTM

PR #899, opened 2022-02-24
fix retry timer setting wrong variable to null (Fixes #898)

### LINKED ISSUE #898: Retry timer is improperly cleared, causing it to fail to start the second time
cancelRetryTimer clears the timeout but incorrectly sets suspendTimer to null instead of retryTimer, so the second time that the retry timer is called (if the connection is lost, regained, then lost again) it will not start, meaning that the channel will not try to re-attach

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1631) by [Unito](https://www.unito.io)

Peter-Maguire:


owenpearson (approved):
LGTM, thanks

PR #906, opened 2022-02-25
Refactor HTTP clients

### LINKED ISSUE #905: Platform HTTP classes should be used as instances
- Currently the platform specific HTTP APIs are accessed as static classes.
- This means that if you create two clients in the same javascript environment they will share a single HTTP class and any changes to HTTP configuration in either client will affect the other client.
- Using class instances will make the http code a bit cleaner and will resolve the aforementioned bug.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1638) by [Unito](https://www.unito.io)

owenpearson:
Resolves #905

- Changes the HTTP clients (node and browser implementations) to use non-static methods so that multiple library instances have their own unique http clients.
- Converts JSONPTransport to typescript.
- Refactors a lot of the http related code.


browser/lib/transport/jsonptransport.ts
```diff
@@ -0,0 +1,232 @@
...
+
+	toString() {
+		return 'JSONPTransport; uri=' + this.baseUri + '; isConnected=' + this.isConnected;
+	}
+
+	createRequest(uri: string, headers: Record<string, string> | null, params?: Record<string, string>, body?: unknown, requestMode?: number, timeouts?: Record<string, number>, method?: HttpMethods) {
+		/* JSONP requests are used either with the context being a realtime
+		 * transport, or with timeouts passed in (for when used by a rest client),
+		 * or completely standalone.  Use the appropriate timeouts in each case */
+		timeouts = (this && this.timeouts) || timeouts || Defaults.TIMEOUTS;
```
Peter-Maguire, line 87: Can be simplified to this
```suggestion
		timeouts = this?.timeouts || timeouts || Defaults.TIMEOUTS;
```

browser/lib/transport/jsonptransport.ts
```diff
@@ -0,0 +1,232 @@
...
+					this.complete(err);
+				}
+			} else {
+				/* Handle as non-enveloped -- as will be eg from a customer's authUrl server */
+				this.complete(null, message);
+			}
+		};
+
+		const timeout = (this.requestMode == XHRStates.REQ_SEND) ? this.timeouts.httpRequestTimeout : this.timeouts.recvTimeout;
+		this.timer = setTimeout(() => { this.abort(); }, timeout);
```
Peter-Maguire, line 189: sorry I know this is a direct port from the js version but isn't this functionally the same as this:
```suggestion
		this.timer = setTimeout(this.abort.bind(this), timeout);
```

Peter-Maguire (changes requested):

Peter-Maguire (approved):

QuintinWillison (approved):
Tricky to review line by line, but on a skim read looks ok.

PR #910, opened 2022-02-28
Replace Base64.js with CryptoJS in browser

### LINKED ISSUE #372: remove Base64
https://ably-real-time.slack.com/archives/client-libraries/p1484263977000583

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-559) by [Unito](https://www.unito.io)

Peter-Maguire:
Fixes #372. Removed base64.js and replaced with CryptoJS. Node library already uses Buffer instead, so this only affects the browser.

### DIFF
--- a/common/lib/client/auth.ts
+++ b/common/lib/client/auth.ts
@@ -7,6 +7,7 @@ import * as BufferUtils from 'platform-bufferutils';
 import ErrorInfo from '../types/errorinfo';
 import HmacSHA256 from 'crypto-js/build/hmac-sha256';
 import { stringify as stringifyBase64 } from 'crypto-js/build/enc-base64';
+import { parse as parseUtf8 } from 'crypto-js/build/enc-utf8';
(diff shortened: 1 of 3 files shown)


Peter-Maguire:
playwright tests seem to be unrelated/random so are most likely due to the sandbox environment being flaky, I have ran the tests locally multiple times and they all pass.

owenpearson (approved):
LGTM, thanks

QuintinWillison (approved):

SimonWoolf:
This currently removes all references to base64.js, but leaves the actual file still there https://github.com/ably/ably-js/blob/main/browser/lib/util/base64.js , any particular reason to leave it there?

Peter-Maguire:
@SimonWoolf good spot, I must have reverted the commit where I deleted it at some point and not caught it. I'll make a new MR to delete the file completely.

PR #936, opened 2022-03-21
Refactor Resource.ts typings

Peter-Maguire:
This PR clears up the typings for Resouce.ts by specifying the typing for the callback function for resource requests.


common/lib/client/resource.ts
```diff
@@ -51,16 +45,19 @@ function unenvelope(callback: Function, format: Utils.Format | null) {
       }
     }
 
-    if (body.statusCode === undefined) {
+    if (!body) {
+      callback(new ErrorInfo('Body is missing', null));
```
owenpearson, line 49: Would be good to add something to this error message to indicate where it's coming from, eg "unenvelope(): Body is missing"

common/lib/client/resource.ts
```diff
@@ -216,11 +226,11 @@ class Resource {
             /* token has expired, so get a new one */
             rest.auth.authorize(null, null, function (err: Error) {
               if (err) {
-                callback(err);
+                callback(ErrorInfo.fromValues(err));
```
owenpearson, line 229: Looks like `rest.auth.authorize` will only invoke with an `ErrorInfo` (sorry it should be typed better). You should be able to just change the callback arg type to `ErrorInfo` and not have to use this.

owenpearson (changes requested):

common/lib/client/resource.ts
Peter-Maguire: Switched it to err: ErrorInfo

common/lib/client/resource.ts
```diff
@@ -105,7 +107,7 @@ function logResponseHandler(callback: Function, method: HttpMethods, path: strin
         'Received; ' +
           urlFromPathAndParams(path, params) +
           '; Headers: ' +
-          paramString(headers) +
+          (headers ? paramString(headers) : 'None') +
```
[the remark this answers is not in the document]
Peter-Maguire: Done

common/lib/client/resource.ts
Peter-Maguire: Done

owenpearson (approved):
LGTM, thanks

PR #938, opened 2022-03-28
fix: take into account that some of the transportParams are numbers -…

incognos:
… heartbeatInterval & remainPresentFor


Peter-Maguire (approved):
:+1:  Nice catch, we will include this change in the next release.

PR #947, opened 2022-04-04
Ensure that Ably-Agent is always URI encoded

owenpearson:
This fixes a bug in the react-native library where the websocket constructor would throw an error because the Ably-Agent query param was not being URI encoded.

### DIFF
--- a/common/lib/transport/connectionmanager.ts
+++ b/common/lib/transport/connectionmanager.ts
@@ -154,7 +154,7 @@ export class TransportParams {
       params.heartbeats = this.heartbeats;
     }
     params.v = Defaults.apiVersion;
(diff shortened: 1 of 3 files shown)


paddybyers:
What issue is this aiming to fix?

owenpearson:
@paddybyers this fixes a regression in ably-js 1.2.18 where the websocket constructor would throw an error. I'm still looking for a better solution hence the draft state.

Peter-Maguire (approved):

lmars:
Is this URI encoding the header value too? We should check that still parses correctly.

owenpearson:
@lmars yeah it is. I did the release earlier today and I've just checked Grafana, looks like everything is working correctly.

PR #954, opened 2022-04-07
Remove support for MozWebSocket

owenpearson:
This was only included to support websockets on ancient versions of chrome and firefox. Would be stunned if it's ever being used anymore.


Peter-Maguire:
Does this not count as a breaking change?

paddybyers:
I don't see why any external code would be depending on this.

Peter-Maguire (approved):

PR #960, opened 2022-04-19
Update publish action to use GitHub OIDC for AWS access

lmars:
This updates the publish GitHub action to use GitHub OIDC to obtain short-lived AWS credentials in place of the current long lived stored in the secrets.

The AWS IAM role (`github-actions-sdk`) is managed by the internal infrastructure team.


surminus (approved):
:tada:

PR #982, opened 2022-05-12
Platform Entrypoints

Peter-Maguire:
Currently, Ably JS uses module aliases inside webpack to switch platform-specific configuration/dependencies. The PR changes the webpack configuration to use platform-specific entrypoints. This will allow us to better unit test the library in the future.


README.md
```diff
@@ -47,7 +47,7 @@ For complete API documentation, see the [Ably documentation](https://www.ably.co
 and require as:
 
 ```javascript
-var Ably = require('ably');
+var Ably = require('src/ably');
```
owenpearson: I'm guessing the changes in this file are from an automatic find and replace, obviously we don't want to make a breaking change like this.

package.json
```diff
@@ -16,12 +16,12 @@
...
   },
   "files": [
-    "ably.d.ts",
+    "src/ably.d.ts",
     "browser/static/**",
-    "callbacks.d.ts",
-    "callbacks.js",
-    "promises.d.ts",
-    "promises.js",
+    "src/callbacks.d.ts",
```
owenpearson: Yeah the changes here are breaking changes, I would suggest using webpack copy plugin to these files into the dist folder and change the release process to use `npm publish dist`

src/callbacks.js
```diff
@@ -1,6 +1,6 @@
 "use strict";
 /* When returning promises becomes the default in some future version, this
  * file will start modifying the exports as promises.js does today.
- * Please note that the file imported below is only generated after running 
+ * Please note that the file imported below is only generated after running
  * the build task. */
-module.exports = require('./browser/static/ably-node');
+module.exports = require('../browser/static/ably-node');
```
owenpearson: I'm guessing this import won't work anymore since ably-node has moved to the `dist` folder

src/common/types/http.d.ts
```diff
@@ -1,7 +1,7 @@
 import HttpMethods from '../constants/HttpMethods';
 import Rest from '../lib/client/rest';
 import ErrorInfo from '../lib/types/errorinfo';
-import http from 'http';
+import http from 'src/common/types/http';
```
owenpearson, line 4: This should be importing `http` from the node standard library so this change should be reverted

webpack.config.js
```diff
@@ -218,8 +218,8 @@ const webworkerConfig = {
     new CopyPlugin({
       patterns: [
         {
-          from: path.resolve(browserPath, 'fragments', 'ably.d.ts'),
-          to: path.resolve(browserPath, 'static', 'ably-webworker.min.d.ts'),
+          from: path.resolve(browserPath,  'fragments', 'ably.d.ts'),
+          to: path.resolve(browserPath, 'build', 'ably-webworker.min.d.ts'),
```
owenpearson: I would guess that `build` is top level directory, rather than a child of browserPath

webpack.config.js
```diff
@@ -53,11 +53,11 @@ const nodeConfig = {
     ...baseConfig.resolve,
     alias: {
       platform: path.resolve(nodePath, 'platform'),
-      'platform-http': path.resolve(nodePath, 'lib', 'util', 'http'),
-      'platform-bufferutils': path.resolve(nodePath, 'lib', 'util', 'bufferutils'),
-      'platform-defaults': path.resolve(nodePath, 'lib', 'util', 'defaults'),
-      'platform-crypto': path.resolve(nodePath, 'lib', 'util', 'crypto'),
-      'platform-transports': path.resolve(nodePath, 'lib', 'transport'),
+      'platform-http': path.resolve(nodePath,  'lib', 'util', 'http'),
```
owenpearson: You added some extraneous whitespace here and on line 161

owenpearson (changes requested):
few comments:

ikbalkaya (approved):
Warning! very light review - LGTM

webpack.config.js
owenpearson: Happy for you to add webpack.config.js to our prettier targets as part of this PR

QuintinWillison:
I am not going to be able to invest enough time into reviewing this PR in any way that will bring real value, given its nature, so am removing my request for review. I also note, as I look at it right now, that it has conflicts with the target branch - so I'm going to move it back to a Draft state until those have been addressed.

README.md
Peter-Maguire: Already fixed [here](https://github.com/ably/ably-js/pull/982/commits/f042f1cb73ddd4c122b807c8f712f875cdd26277)

src/callbacks.js
Peter-Maguire: Fixed [here](https://github.com/ably/ably-js/pull/982/commits/f042f1cb73ddd4c122b807c8f712f875cdd26277)

webpack.config.js
Peter-Maguire: The webpack config was mostly entirely rewritten in [another commit](https://github.com/ably/ably-js/pull/982/files#diff-1fb26bc12ac780c7ad7325730ed09fc4c2c3d757c276c3dacc44bfe20faf166f)

Gruntfile.js
```diff
@@ -11,12 +11,12 @@ module.exports = function (grunt) {
...
-    fragments: 'browser/fragments',
-    static: 'browser/static',
-    dest: 'browser/static',
-    compat: 'browser/compat',
+    common: 'src/common',
+    browser: 'src/platform/web',
+    fragments: 'src/platform/web/fragments',
+    static: 'build',
+    dest: 'build',
+    compat: 'src/platform/web/compat', // still used?
```
owenpearson, line 19: Not still used - pls delete
```suggestion
```

README.md
```diff
@@ -84,7 +84,8 @@ The TypeScript typings are included in the package and so all you have to do is:
 
 ```typescript
 import * as Ably from 'ably';
-let options: Ably.Types.ClientOptions = { key: 'foo' };
+
+let options: Ably.Types.ClientOptions = {key: 'foo'};
```
owenpearson, line 88: You've changed the formatting slightly here
```suggestion
let options: Ably.Types.ClientOptions = { key: 'foo' };
```

README.md
```diff
@@ -95,7 +96,7 @@ Intellisense in IDEs with TypeScript support is supported:
 
 ![TypeScript suggestions](./resources/typescript-demo.gif)
 
-If you need to explicitly import the type definitions, see [ably.d.ts](./ably.d.ts) (or `promises.d.ts` if you're requiring the library as `ably/promises`).
+If you need to explicitly import the type definitions, see [ably.d.ts](src/ably.d.ts) (or `promises.d.ts` if you're requiring the library as `ably/promises`).
```
owenpearson, line 99: We moved `ably.d.ts` back to the root so this should be reverted
```suggestion
If you need to explicitly import the type definitions, see [ably.d.ts](./ably.d.ts) (or `promises.d.ts` if you're requiring the library as `ably/promises`).
```

package.json
```diff
@@ -7,17 +7,17 @@
...
   "react-native": {
-    "./browser/static/ably-node.js": "./browser/static/ably-reactnative.js"
+    "./build/ably-node.js": "./build/ably-reactnative.js"
   },
   "browser": {
-    "./browser/static/ably-node.js": "./browser/static/ably-commonjs.js"
+    "./build/ably-node.js": "./build/ably-commonjs.js"
   },
   "files": [
+    "src/**",
```
owenpearson, line 19: The `files` field specifies which files should be included in the npm package so we need to keep the build output in here.
```suggestion
    "build/**",
```

src/common/lib/util/utils.ts
```diff
@@ -337,7 +338,7 @@ export function allSame(arr: Array<Record<string, unknown>>, prop: string): bool
   });
 }
 
-export const nextTick = Platform.nextTick;
+//export const nextTick = Platform.Config.nextTick;
```
owenpearson, line 341: Can be removed

src/common/lib/util/utils.ts
```diff
@@ -397,34 +398,36 @@ export const now =
     return new Date().getTime();
   };
 
-export const inspect = Platform.inspect;
+//export const inspect = Platform.Config.inspect;
+// TODO: remove
```
owenpearson, line 402:
```suggestion
```

owenpearson (changes requested):

README.md
Peter-Maguire: I really need to get my editor config setup to be inline with the prettier config

owenpearson (approved):
LGTM 👍

PR #985, opened 2022-05-16
Rest lifecycle status RSL8

### LINKED ISSUE #981: Add support to get channel lifecycle status
Add support to get channel lifecycle status 

Channel lifecycle status is described at [Channel status / REST / [Docs]](https://ably.com/docs/rest/channel-status#metadata-rest)

It is currently possible to retrieve channel lifecycle status in ably-go by using REST.Request and marshalling the result into a slice of empty interface.

**Example:**

```go
result, _ := restClient.Request("get", "/channels/test").Pages(context.Background())

result.Next(ctx)

var items []interface{}
if err := result.Items(&items); err != nil {
	fmt.Println(err)
}

fmt.Printf("%+v\n", items)
// [map[channelId:test name:test status:map[isActive:true occupancy:map[metrics:map[connections:1 presenceConnections:1 presenceMembers:1 presenceSubscribers:1 publishers:1 subscribers:1]]]]]
```

**Out of Scope**
Realtime subscription to Channel Lifecycle Events. This ticket is only for supporting REST client ‘Get’.

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-1853) by [Unito](https://www.unito.io)

owenpearson:
Resolves #981

The descriptions in the README examples are taken from https://ably.com/docs/api/rest-sdk/channel-status.


ikbalkaya (approved):
LGTM 👍

README.md
```diff
@@ -459,6 +459,16 @@ client.stats(function(err, statsPage) {        // statsPage as PaginatedResult
 client.time(function(err, time) { ... }); // time is in ms since epoch
 ```
 
+### Getting the status of a channel
+
+```javascript
+channel.status(function(err, channelDetails) {
```
sacOO7, line 465: This example can be moved right next to other `Channel Examples`. Currently, it's under `Ably Time`, which kinda feels not related to channel context.

common/lib/client/channel.ts
```diff
@@ -189,6 +190,17 @@ class Channel extends EventEmitter {
...
     Resource.post(this.rest, this.basePath + '/messages', requestBody, headers, params, null, callback);
   }
+
+  status(callback?: StandardCallback<API.Types.ChannelDetails>): void | Promise<API.Types.ChannelDetails> {
+    if (typeof callback !== 'function' && this.rest.options.promises) {
+      return Utils.promisify(this, 'status', []);
+    }
+
+    const format = this.rest.options.useBinaryProtocol ? Utils.Format.msgpack : Utils.Format.json;
+    const headers = Utils.defaultPostHeaders(format);
```
sacOO7, line 200: Do we need to pass any headers? I think it works fine without passing any headers -> 
https://github.com/ably/ably-go/pull/521

test/rest/status.test.js
```diff
@@ -0,0 +1,71 @@
+'use strict';
+
+define(['shared_helper', 'chai'], function (helper, chai) {
+  var rest;
+  var utils = helper.Utils;
+  var expect = chai.expect;
+
+  // RSL8
+  describe('rest/status', function () {
```
sacOO7, line 9: Optional - maybe test can be updated to check for channel status values

owenpearson: I think it would take a lot of effort for us to test this on the ably-js side and really it's something the realtime team should be testing themselves so I'm gonna opt to not do this one for now

common/lib/client/channel.ts
owenpearson: We definitely need to be sending the `Ably-Agent` and `X-Ably-Version` headers with REST API requests in all of our SDKs. Requests will still work fine without them but it would stop us from being able to track those requests in our metrics.

README.md
owenpearson: Good point - will do 👍

Peter-Maguire (approved):

PR #986, opened 2022-05-17
CDN workflow fixes

owenpearson:
Resolves #975

I also decided to run prettier on the cdn script so probably best to review this commit-wise.

### DIFF
--- a/scripts/cdn_deploy.js
+++ b/scripts/cdn_deploy.js
@@ -65,7 +65,8 @@ async function run() {
           config.includeDirs.find((d) => file.startsWith(d)),
           file
         );
-        const newPath = `${relativePath.split('.js')[0]}-${version}.js`;
+        const split = relativePath.split('.js');
+        const newPath = `${split[0]}-${version}.js${split[1]}`;
         let fileData = fs.readFileSync(file).toString();
         let ContentType;
         if (newPath.endsWith('.min.js')) {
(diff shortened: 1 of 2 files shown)


scripts/cdn_deploy.js
```diff
@@ -3,124 +3,132 @@
...
+      console.log(`Uploading ${file}...`);
+      for (let version of versions) {
+        const relativePath = path.relative(
+          config.includeDirs.find((d) => file.startsWith(d)),
+          file
+        );
+        const split = relativePath.split('.js');
+        const newPath = `${split[0]}-${version}.js${split[1]}`;
+        let fileData = fs.readFileSync(file).toString();
+        let ContentType;
```
QuintinWillison, line 71: I'm not sure why this is cased this way - why not `contentType`? 😕

QuintinWillison (approved):
Only one minor observation. Otherwise, looks good.

scripts/cdn_deploy.js
owenpearson: I did that so that I can use the shorthand object property assignment on line 83 of that file:
```ts
        await upload(s3, {
          Body: fileData,
          Key: path.join(config.root, newPath),
          Bucket: config.bucket,
          ContentType,
        });
```
As for why AWS decided to name their properties like that I have no clue 🤷

QuintinWillison: Interesting... breaking a naming convention to take advantage of a language feature. 😆 

Not sure I'm with you there, but fair enough, thanks for explaining. If you prefer this 'look' then keep it as is. 😝

PR #1003, opened 2022-06-10
Message interactions

Peter-Maguire:


src/common/lib/client/realtimechannel.ts
```diff
@@ -430,15 +441,92 @@ class RealtimeChannel extends Channel {
...
+      return [];
+    }
+    const realListenerMap = this.filteredSubscriptions.get(realListener) as Map<MessageFilter, MessageListener[]>;
+    // If no filter is specified return all listeners using that function
+    if (!filter) {
+      // array.flat is not available unless we support es2019 or higher
+      return Array.from(realListenerMap.values()).reduce((prev, cur) => prev.concat(...cur), []);
+    }
+
+    // No subscriptions matching this filter
```
owenpearson, line 513: You could remove this logic by just returning `realListenerMap.get(filter) || [];`

ably.d.ts
```diff
@@ -700,6 +700,13 @@ declare namespace Types {
...
   };
 
+  type MessageFilter = {
+    name?: string;
+    refId?: string;
+    refType?: string;
+    hasRef?: boolean;
+  };
+
   class RealtimeChannelCallbacks extends RealtimeChannelBase {
```
owenpearson, line 710: We also need to update the `RealtimeChannel.unsubscribe` signatures.

test/realtime/message.test.js
```diff
@@ -917,6 +917,75 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
       );
     });
 
+    it('subscribe_with_filter_object', function (done) {
+      const realtime = helper.AblyRealtime();
+      const channel = realtime.channels.get('subscribe_with_filter_array');
```
owenpearson, line 922:
```suggestion
      const channel = realtime.channels.get('subscribe_with_filter_object');
```

test/realtime/message.test.js
```diff
@@ -917,6 +917,75 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
...
+            refType: 'com.ably.test',
+            refId: '0123456789',
+          },
+          (m) => {
+            try {
+              expect(m.name).to.be.equal('correct', 'Correct message received');
+            } catch (e) {
+              return cb(e);
+            }
+            config.nextTick(cb);
```
owenpearson, line 974: I'd like to see more comments in this fixture, in particular the purpose of this line is non-obvious; what is it that we're expecting to happen before the next tick?

test/realtime/message.test.js
```diff
@@ -917,6 +917,75 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
       );
     });
 
+    it('subscribe_with_filter_object', function (done) {
```
owenpearson, line 920: We should also have a test for unsubscribing with a filter object

owenpearson (changes requested):

ably.d.ts
Peter-Maguire: RealtimeChannelCallbacks doesn't have an unsubscribe signature, and RealtimeChannelBase has been updated

Jmgr (commented):
LGTM, but I'm neither an expert in JS nor the in ably-js internals so can't really approve. I really like the comments though :+1:

src/common/lib/client/realtimechannel.ts
```diff
@@ -430,15 +441,92 @@ class RealtimeChannel extends Channel {
...
+      }
+      listener(m);
+    };
+    this._addFilteredSubscription(filter, listener, filteredListener);
+    this.subscriptions.on(filteredListener);
+  }
+
+  // Adds a new filtered subscription
+  _addFilteredSubscription(filter: MessageFilter, realListener: MessageListener, filteredListener: MessageListener) {
+    if (this.filteredSubscriptions.has(realListener)) {
```
cruickshankpg, line 478: should this throw an error instead of quietly overwriting the old filter?

Peter-Maguire: Overwriting shouldn't happen here, if the filter and listener are both identical to a previous subscription then the new listener is added to the array (`realListenerMap.set(filter, realListenerMap.get(filter).concat(filteredListener)`)

test/realtime/message.test.js
Peter-Maguire: [Done](https://github.com/ably/ably-js/pull/1003/commits/3192a0bdede12e59e37edb4eda726dcf8e708f75)

test/realtime/message.test.js
Peter-Maguire: [Done](https://github.com/ably/ably-js/pull/1003/commits/3192a0bdede12e59e37edb4eda726dcf8e708f75)

ably.d.ts
```diff
@@ -700,6 +700,13 @@ declare namespace Types {
     quickAck?: boolean;
   };
 
+  type MessageFilter = {
+    name?: string;
+    refId?: string;
```
cruickshankpg, line 705:
```suggestion
    refTimeserial?: string;
```

cruickshankpg (changes requested):

cruickshankpg (approved):
LGTM

test/realtime/message.test.js
```diff
@@ -917,6 +917,119 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
...
+            'Listener should no longer be present after unsubscribing'
+          ).to.be.false;
+          config.nextTick(cb);
+        } catch (e) {
+          cb(e);
+        }
+      }
+
+      async.series(
+        [(cb) => realtime.connection.once('connected', () => cb()), (cb) => channel.attach(cb), unsubscribe, send],
```
owenpearson, line 1028: These test fixtures should run on IE so avoid arrow functions (and any other es6+ syntax). We should probably add an eslint config for the test directory tbh.

owenpearson (changes requested):

test/realtime/message.test.js
Peter-Maguire: [Done](https://github.com/ably/ably-js/pull/1003/commits/19a22ec2d1d673c99156dd215699a1de6a307865)

src/common/lib/client/realtimechannel.ts
```diff
@@ -83,6 +87,10 @@ class RealtimeChannel extends Channel {
     this.connectionManager = realtime.connection.connectionManager;
     this.state = 'initialized';
     this.subscriptions = new EventEmitter();
+    this.filteredSubscriptions = new Map<
```
owenpearson, line 90: This will affect platform support, we should initialise the map only when `_subscribeFilter` is called for the first time

owenpearson (changes requested):

owenpearson (approved):
LGTM

PR #1008, opened 2022-06-13
Implement incremental backoff and jitter (RTB1)

### LINKED ISSUE #997: Implement incremental backoff and jitter
Incremental backoff and jitter was added to the features spec in https://github.com/ably/docs/pull/1445

To implement this we need to change the retry behaviour for:
- When the client connection is in the `DISCONNECTED` state
- When a `RealtimeChannel` is in the `SUSPENDED` state

The new retry behaviour is specified in RTB1 of the features spec.

### RTB1a

The backoff coefficient described in `RTB1a` can be calculated in javascript like so:
```ts
function getBackoffCoefficient(n) {
  return Math.min((n + 2) / 3, 2);
}
```

### RTB1b

The jitter coefficient described in `RTB1b` can be calculated in javascript like so:
```ts
function getJitterCoefficient() {
  return 1 - Math.random() * 0.2;
}
```

### RTB1

The overall retry time calculation should look like:
```ts
function getRetryTime(initValue, n) {
  return initValue * getBackoffCoefficient(n) * getJitterCoefficient();
}
```

The following code can be used to test it:
```ts
[1, 2, 3, 4, 5].forEach((n) => {
  console.log(getRetryTime(initValue, n));
});
```
Which, with an `initValue` of 15 (seconds) should print something like:
```
13.917470451245713
18.415226855678757
20.444851032898747
26.650729887092275
27.803382948778786
```

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1989) by [Unito](https://www.unito.io)

owenpearson:
Resolves #997


Peter-Maguire (approved):

src/common/lib/client/realtimechannel.ts
```diff
@@ -699,6 +700,10 @@ class RealtimeChannel extends Channel {
       state + (reason ? '; reason: ' + reason : '')
     );
 
+    if (state !== 'attaching' && state !== 'suspended') {
+      this.retryCount = 0;
```
sacOO7, line 704: Is this supposed to be `if (state !== 'attached' && state !== 'suspended')` @owenpearson ?

owenpearson: Hey, so this was intentional and will work correctly. In the channel retry sequence the channel state alternates between `'attaching'` and `'suspended'` so this condition is met whenever the channel leaves the retry sequence for whatever reason.

It would probably also work to have `if (state === 'attached' || state === 'failed')` (note that it's `'failed'` here rather than `'suspended'`) although I'm not 100% sure whether it's possible for the channel to transition to `'detaching'` or `'detached'` during the retry sequence so it may be necessary to include one or both of those in that condition too.

sacOO7: I don't think it tries to re-attach on the channel `failed`. I have tried the same myself, when the channel doesn't have enough capability available, it will go into a `failed` state and will not re-attach.

owenpearson: Yeah you're right, I'm not saying that it tried to re-attach on failed. This condition is supposed to be for when the channel *leaves* the retry sequence so that it can reset the retry counter whenever that happens. Checking that the channel state is not `attaching` or `suspended` is IMO the easiest way for us to know when the retry sequence has ended.

sacOO7: I think it would work just fine for `if (state === 'attached' || state === 'failed')`

owenpearson: I just checked and for ably-js if the user calls `RealtimeChannel.detach` while the client is `suspended` the channel becomes `detached, and if you call `detach` while the client is `attaching` the channel becomes `detaching` so it seems possible for the client to enter the retry sequence again without ever becoming `attached` or `failed`. It might work differently in ably-dotnet though. Also worth considering what would happen if the client receives a `DETACHED` protocol message during the retry sequence.

sacOO7: Although, we can say before `detaching` state would always be `attached` which resets `retryCount` to zero.

owenpearson: > Although, we can say before detaching state would always be attached which resets retryCount to zero.

This isn't necessarily true for ably-js, see my comment above

sacOO7: yep .. that was previous assumption as per old code.

owenpearson: > So, as of now, it can become a bug if the channel state engine follows a different pattern for the retry sequence.

IIRC the spec is pretty explicit that during retries the channel state alternates between `attaching` and `suspended` so I would be surprised if the condition I used here didn't work for the other SDKs.

sacOO7: Well, we do know as of now that during suspended it can go into other states too ... It does alternate between these two states ... Though for resetting retry thats not the right condition. The right condition should be the state where it shouldn't retry. That would be `attached` and `failed`

test/realtime/failure.test.js
```diff
@@ -328,6 +358,58 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
...
+                  return;
+                }
+                var elapsedSinceSuspneded = performance.now() - lastSuspended;
+                lastSuspended = performance.now();
+                try {
+                  // JS timers don't work precisely and realtimeRequestTimeout can't be 0 so add 5ms to the max timeout length each retry
+                  expect(elapsedSinceSuspneded).to.be.below(
+                    channelRetryTimeout + Math.min(retryCount, 3) * 50 + 5 * (retryCount + 1)
+                  );
+                  expect(elapsedSinceSuspneded).to.be.above(0.8 * (channelRetryTimeout + Math.min(retryCount, 3) * 50));
```
sacOO7, line 399: This generalized calculated lower bound breaks for any value other than 150. 
If `channelRetryTimeout` is 50, for first retry, calculated value is `8`.
While actual value generated using `getRetryTime` is `4.7`

sacOO7: resolved as a part of #1344

PR #1009, opened 2022-06-21
Test observability

owenpearson:
TODO:
- [ ] Figure out how to get command line output so that we can still get useful workflow run logs while using the junit reporter (would likely only be possible by running mocha programmatically and writing a wrapper for `mocha-junit-reporter`)

### DIFF
--- a/.github/workflows/test-node.yml
+++ b/.github/workflows/test-node.yml
@@ -24,6 +24,12 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
       - run: npm ci
-      - run: npm run test:node
+      - run: npm run test:node:junit
         env:
           CI: true
+      - name: Upload test results
+        if: always()
+        uses: ably-labs/test-observability-action@main
+        with:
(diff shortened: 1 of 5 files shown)


.github/workflows/test-node.yml
```diff
@@ -24,6 +24,12 @@ jobs:
...
-      - run: npm run test:node
+      - run: npm run test:node:junit
         env:
           CI: true
+      - name: Upload test results
+        if: always()
+        uses: ably-labs/test-observability-action@main
+        with:
+          server-url: 'https://test-observability.herokuapp.com'
+          server-auth: ${{ secrets.TEST_OBSERVABILITY_SERVER_AUTH_KEY }}
```
lawrence-forooghian, line 35: Just a note that this now exists as an org-wide secret (see https://github.com/ably/test-observability-action/pull/9), so if it's a repo-level secret here it can now be removed.

lawrence-forooghian:
Replaced by #1551.

PR #1031, opened 2022-07-18
Update TypeScript to 4.6.4 and TypeDoc to 0.23.8

lawrence-forooghian:
Updates TypeDoc to 0.23.8 (so I can use [declaration references](https://typedoc.org/guides/declaration-references/) in an upcoming PR), which requires that I update TypeScript to at least 4.6, so I do that too. It fixes a couple of TypeScript errors that I guess result from stricter compiler checks. It also fixes one error that TypeDoc starts to throw.

(Note that the target here is `integration/docs`, not `main`.)

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -1232,6 +1232,12 @@ declare namespace Types {
   /**
    * Not yet documented.
    *
+   * @param shouldRecover - Not yet documented.
+   */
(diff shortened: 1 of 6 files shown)


lawrence-forooghian:
Oh my, CI is failing horribly. Will look at that.

lawrence-forooghian:
Alrighty, ready for review now 👍

owenpearson (approved):
LGTM, thanks

PR #1035, opened 2022-07-27
Transports: implement timeouts in tryConnect()

SimonWoolf:
In tryConnect rather than in connectionManager#tryATransport so that it can properly dispose of the transport if it times out.

NB: the test requires a realtime that has https://github.com/ably/realtime/pull/4231 running.

### DIFF
--- a/src/common/lib/transport/websockettransport.ts
+++ b/src/common/lib/transport/websockettransport.ts
@@ -48,7 +48,7 @@ class WebSocketTransport extends Transport {
 
     const realtimeRequestTimeout = connectionManager.options.timeouts.realtimeRequestTimeout;
     transportAttemptTimer = setTimeout(() => {
-      transport.off(['wsopen', 'disconnected', 'failed']);
+      transport.off(['preconnect', 'disconnected', 'failed']);
       transport.dispose();
       errorCb.call(
         { event: 'disconnected' },
@@ -57,7 +57,7 @@ class WebSocketTransport extends Transport {
     }, realtimeRequestTimeout);
 
     transport.on(['failed', 'disconnected'], errorCb);
-    transport.on('wsopen', function () {
+    transport.on('preconnect', function () {
       Logger.logAction(Logger.LOG_MINOR, 'WebSocketTransport.tryConnect()', 'viable transport ' + transport);
       clearTimeout(transportAttemptTimer);
       transport.off(['failed', 'disconnected'], errorCb);
@@ -169,7 +169,7 @@ class WebSocketTransport extends Transport {
 
   onWsOpen() {
     Logger.logAction(Logger.LOG_MINOR, 'WebSocketTransport.onWsOpen()', 'opened WebSocket');
-    this.emit('wsopen');
+    this.emit('preconnect');
   }
 
   onWsClose(ev: number | CloseEvent) {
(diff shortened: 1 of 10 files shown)


src/common/lib/transport/websockettransport.ts
```diff
@@ -39,12 +39,27 @@ class WebSocketTransport extends Transport {
...
   ) {
     const transport = new WebSocketTransport(connectionManager, auth, params);
+
+    let transportAttemptTimer: NodeJS.Timeout | number;
     const errorCb = function (this: { event: string }, err: ErrorInfo) {
+      clearTimeout(transportAttemptTimer);
       callback({ event: this.event, error: err });
     };
+
+    const realtimeRequestTimeout = connectionManager.options.timeouts.realtimeRequestTimeout;
```
owenpearson, line 49: I would have thought you could remove a lot of the duplication by adding a couple of methods to the base transport class (`startTryConnectTimer` and `clearTryConnectTimer` or whatever).

test/browser/connection.test.js
```diff
@@ -401,6 +401,38 @@ define(['shared_helper', 'chai'], function (helper, chai) {
         }
         closeAndFinish(done, realtime);
       });
+
+      /*
+       * Check behaviour when going through a proxy which doesn't send any xhr streaming
+       * responses until the stream closes (simulated by a transportparam interpreted by
+       * realtime)
+       */
+      it('connection behaviour with a proxy through which streaming is broken', function (done) {
```
owenpearson, line 410: Also, I'm not comfortable adding a test which will break CI until that realtime change is on sandbox. I think the best option I can think of is to add this as an `it.skip` for now with a TODO/gh issue

owenpearson (changes requested):

src/common/lib/transport/websockettransport.ts
SimonWoolf: istm a halfway-house of sharing just the timer helper functions would be a bit messy -- you'd have to pass in errorCb, the transport instance, and the timeout, the list of event names to deregister varies a bit, etc. If we want to dedup, better would be to go the whole hog and deduplicate the whole of tryConnect -- I'll have a quick go at that

SimonWoolf: https://github.com/ably/ably-js/pull/1035/commits/92e4671c3871c9f43f23784d43367e0f12ddf2fa, https://github.com/ably/ably-js/pull/1035/commits/b2aeea8fe997777494342ffbaf99dcde8fb52ffb -- that went smoother than expected, turns out they were pretty much all identical

test/browser/connection.test.js
SimonWoolf: I've merged https://github.com/ably/realtime/pull/4231, will automatically deploy to sandbox in the next hour or so

src/common/lib/transport/websockettransport.ts
owenpearson: nice, that's a big improvement thanks 👍

owenpearson:
I just realised the node tests are broken here. I assumed before that this was because of the new test depending on a realtime patch but I just realised the new tests is only for browsers so something must have broken. I'm looking into it now.

PR #1043, opened 2022-08-08
[CAP-48] remove connectionTimeserial spike stuff

andydunstall:
removes unused `timeSerial` added in https://github.com/ably/ably-js/commit/3779562813ce589083cd23a2093716082d15dd27#diff-4950999f00a83f9a58097a13788e54b21306f3b895c03c8a5b85ae06bf30447e (so we can add changes for CAP-44 later)

### DIFF
--- a/src/common/lib/transport/connectionmanager.ts
+++ b/src/common/lib/transport/connectionmanager.ts
@@ -90,7 +90,6 @@ export class TransportParams {
   format?: Utils.Format;
   connectionKey?: string;
   connectionSerial?: number;
-  timeSerial?: string;
   stream?: any;
   heartbeats?: boolean;
 
@@ -102,7 +101,6 @@ export class TransportParams {
     this.format = options.useBinaryProtocol ? Utils.Format.msgpack : Utils.Format.json;
 
     this.connectionSerial = undefined;
-    this.timeSerial = undefined;
   }
 
   getConnectParams(authParams: Record<string, unknown>): Record<string, string> {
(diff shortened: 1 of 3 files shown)


SimonWoolf:
@owenpearson poke (this has been open for a month)

src/common/lib/transport/connectionmanager.ts
```diff
@@ -832,7 +816,7 @@ class ConnectionManager extends EventEmitter {
    * @param transport the transport instance
    * @param connectionId the id of the new active connection
    * @param connectionDetails the details of the new active connection
-   * @param connectionPosition the position at the point activation; either {connectionSerial: <serial>} or {timeSerial: <serial>}
+   * @param connectionPosition the position at the point activation, {connectionSerial: <serial>}
```
owenpearson, line 819: I think this comment is wrong (and was wrong prior to these changes). The type of `connectionPosition` is `ConnectionManager` for some reason, although it should be a `Message` (not a serial).

owenpearson (changes requested):
Changes look good to me, just one thing I'd like to clean up before we merge

src/common/lib/transport/connectionmanager.ts
SimonWoolf: It's a subtype of both `ConnectionManager` and `Message`, but really the type should have been exactly what the comment says: `{connectionSerial: number} | {timeSerial: string}`

but it only existed to be that union type, now that no longer exists we can just make the param just plain ol' `connectionSerial: number` , no?

owenpearson: Oh I just saw this after I already did it a different way, but yeah that makes a lot more sense

owenpearson (approved):

PR #1045, opened 2022-08-19
[WIP] [CAP-49] no connection serials

andydunstall:
Add support for no connection serials as described in https://github.com/ably/specification/pull/88.

(**Note** I've updated the original PR to try to make it clearer which commits implement which spec items - so should be easier to walk though commit by commit if thats useful when implementing the other SDKs - and I've moved the ably-js specific upgrade stuff into the last two commits so those can be ignored)

**Changes**
* updates the protocol version to `2.0` (*RSC7a*)
  * *testing*: covered already by `initbase0`
* fix channel serial type
  * `message.channelSerial` is a string not a number (eg `'e02Tut8QwBFsQM05921682:5'`)
* tracks channel serial of last message/presence (*RTL15b*)
  * note unlike with connection serials we arn't checking for duplicates by comparing serials - this was only needed since the old upgrade scheme could potentially lead to duplicates which is no longer the case, and we want to keep the `channelSerial` opaque rather than having to parse it to compare
  * *testing*: covered already by `resume_active`
* includes `channelSerial` in `ATTACH` (*RTL4c1*)
  * *testing*: covered already by `resume_active`
* remove `connectionSerial` from resume request (*RTN15b2*)
  * *testing*: covered already by `resume_active`
* reattaches channels once connected
  * reattaches `attached` channels once connected (*RTN15c6*/*RTN15c7*)
  * `attaching` and `suspended` states aleady go though the attach sequence on connected (*RTN15c6*/*RTN15c7*)
  * no longer need to reattach all `attached` channels in `ConnectionManager` as this is now always done when we get a new connection, regardless of whether the connection id changed (*RTN15c3*)
  * `msgSerial` is already reset and `errorReason` is already set (*RTN15c7*)
  * already processes queued messages upon connection (*RTN15c6*/*RTN15c7*)
* clears `channelSerial` when detached/suspended/failed (*RTP5a1*)
  * *testing*: covered by channel `rewind works on channel after reattaching` test (if we don't clear the rewind will fail)
* adds recovery with channel serials
  * *RTN16f* is already implemented - this just updates to use the new serialised recovery key 
  * loads the recovery channel serials, applies to any existing channels (which are not yet attached) and stores for later (*RTN16j*)
    * note currently this doesn't clear the recovery channel serials, so if the developer releases the channel then reattaches it uses the recovered channel serial again (not defined in the spec)
  * *RTN16k* is already implemented - this just updates to use the new serialised recovery key  
  * encodes the recovery key as JSON (*RTN16g*) 
  * *RTN16d* is implemented already since in those states the `connectionKey` is `undefined`
  * channel recovery keys are only loaded in `connect`, since the `recover` param may be a user defined callback it can't be loaded in the constructor, and by the time we get to `connect` we know theres no attached channels but `recover` has been resolved into a string (by `getTransportParams`)
  * *testing*: added a test for recovering multiple channels
* presence changes
  * sends the member ID on re-entry (`data` and `channelId` already sent) (*RTP17g*)
  * updates presencemessage to include the id when encoding
  * re-enter whenever we become `ATTACHED` (*RTP17f*) (regardless of if `HAS_PRESENCE` has been set (*RTP17c2*))
  * no longer re-enter when we complete sync (*RTP17c1*)
  * when re-entering sends all members with our `connectionId`, regardless of whether its in the main presence map, and no longer removes from the 'my members' presence map (*RTP17g*/*RTP17d*)
  * *testing*: already tested by `presence_auth_reenter`
* removes connection serial (*RTN10*/*RTN15b2*)
  * this breaks upgrade - which will be fixed later - though prefered to group all the spec changes together (then add the ably-js specific upgrade stuff later)
  * removes `duplicateConnectionId` test as now redundant
* removes `checkChannelsOnResume` options
  * given we're already reattaching each channel on resume, reattaching again after 30s is redundant
* channel doesn't wait for `channelSerial` if empty (*RTP17h*)
* presence index mymembers by client ID (*RTP17h*)

**Upgrade Changes**
(ably-js only - from https://github.com/ably/realtime/issues/4216#issuecomment-1221407649)
* replaces upgrade `SYNC` with `ACTIVATE`
* doesn't wait for no channels pending before upgrading
* removes `mostRecentMessage` as this is no longer needed in the new upgrade scheme (and causes a bug - see [thread](https://ably-real-time.slack.com/archives/C02P7F81DPS/p1663751041394809))
  * removes `duplicateMsgId` test as now redundant

### DIFF
--- a/src/common/lib/client/connection.ts
+++ b/src/common/lib/client/connection.ts
@@ -16,7 +16,6 @@ class Connection extends EventEmitter {
   state: string;
   key?: string;
   id?: string;
-  serial: undefined;
   errorReason: ErrorInfo | null;
 
   constructor(ably: Realtime, options: NormalisedClientOptions) {
@@ -26,7 +25,6 @@ class Connection extends EventEmitter {
     this.state = this.connectionManager.state.state;
     this.key = undefined;
     this.id = undefined;
-    this.serial = undefined;
     this.errorReason = null;
 
     this.connectionManager.on('connectionstate', (stateChange: ConnectionStateChange) => {
--- a/src/common/lib/client/connection.ts
+++ b/src/common/lib/client/connection.ts
@@ -17,7 +17,6 @@ class Connection extends EventEmitter {
   key?: string;
   id?: string;
   serial: undefined;
-  recoveryKey?: string | null;
   errorReason: ErrorInfo | null;
 
   constructor(ably: Realtime, options: NormalisedClientOptions) {
@@ -28,7 +27,6 @@ class Connection extends EventEmitter {
     this.key = undefined;
     this.id = undefined;
     this.serial = undefined;
-    this.recoveryKey = undefined;
     this.errorReason = null;
 
     this.connectionManager.on('connectionstate', (stateChange: ConnectionStateChange) => {
@@ -73,6 +71,10 @@ class Connection extends EventEmitter {
   close(): void {
     Logger.logAction(Logger.LOG_MINOR, 'Connection.close()', 'connectionKey = ' + this.key);
     this.connectionManager.requestState({ state: 'closing' });
+  }
+
+  getRecoveryKey(): string | null {
+    return this.connectionManager.getRecoveryKey();
   }
 }
 
--- a/src/common/lib/client/connection.ts
+++ b/src/common/lib/client/connection.ts
@@ -71,6 +71,10 @@ class Connection extends EventEmitter {
     this.connectionManager.requestState({ state: 'closing' });
   }
 
+  get recoveryKey(): string | null {
+    return this.createRecoveryKey();
+  }
+
   createRecoveryKey(): string | null {
     return this.connectionManager.createRecoveryKey();
   }
--- a/src/common/lib/client/connection.ts
+++ b/src/common/lib/client/connection.ts
@@ -71,8 +71,8 @@ class Connection extends EventEmitter {
     this.connectionManager.requestState({ state: 'closing' });
   }
 
-  getRecoveryKey(): string | null {
-    return this.connectionManager.getRecoveryKey();
+  createRecoveryKey(): string | null {
+    return this.connectionManager.createRecoveryKey();
   }
 }
 
--- a/src/common/lib/client/realtime.ts
+++ b/src/common/lib/client/realtime.ts
@@ -125,30 +125,6 @@ class Channels extends EventEmitter {
     }
   }
 
-  resetAttachedMsgIndicators() {
-    for (const channelId in this.all) {
-      const channel = this.all[channelId];
-      if (channel.state === 'attached') {
-        channel._attachedMsgIndicator = false;
-      }
-    }
-  }
-
-  checkAttachedMsgIndicators(connectionId: string) {
-    for (const channelId in this.all) {
-      const channel = this.all[channelId];
-      if (channel.state === 'attached' && channel._attachedMsgIndicator === false) {
-        const msg =
-          '30s after a resume, found channel which has not received an attached; channelId = ' +
-          channelId +
-          '; connectionId = ' +
-          connectionId;
-        Logger.logAction(Logger.LOG_ERROR, 'Channels.checkAttachedMsgIndicators()', msg);
-        channel.requestState('attaching');
-      }
-    }
-  }
-
   /* Connection interruptions (ie when the connection will no longer queue
    * events) imply connection state changes for any channel which is either
    * attached, pending, or will attempt to become attached in the future */
--- a/src/common/lib/client/realtime.ts
+++ b/src/common/lib/client/realtime.ts
@@ -8,7 +8,6 @@ import Defaults from '../util/defaults';
 import ErrorInfo from '../types/errorinfo';
 import ProtocolMessage from '../types/protocolmessage';
 import { ChannelOptions } from '../../types/channel';
-import { ErrCallback } from '../../types/utils';
 import ClientOptions, { DeprecatedClientOptions } from '../../types/ClientOptions';
 import * as API from '../../../../ably';
 import ConnectionManager from '../transport/connectionmanager';
@@ -55,13 +54,11 @@ class Realtime extends Rest {
 class Channels extends EventEmitter {
   realtime: Realtime;
   all: Record<string, RealtimeChannel>;
-  inProgress: Record<string, RealtimeChannel>;
 
   constructor(realtime: Realtime) {
     super();
     this.realtime = realtime;
     this.all = Object.create(null);
-    this.inProgress = Object.create(null);
     realtime.connection.connectionManager.on('transport.active', () => {
       this.onTransportActive();
     });
@@ -107,9 +104,7 @@ class Channels extends EventEmitter {
   }
 
   /* called when a transport becomes connected; reattempt attach/detach
-   * for channels that are attaching or detaching.
-   * Note that this does not use inProgress as inProgress is only channels which have already made
-   * at least one attempt to attach/detach */
+   * for channels that are attaching or detaching. */
   onTransportActive() {
     for (const channelName in this.all) {
       const channel = this.all[channelName];
@@ -177,37 +172,6 @@ class Channels extends EventEmitter {
       throw releaseErr;
     }
     delete this.all[name];
-    delete this.inProgress[name];
-  }
-
-  /* Records operations currently pending on a transport; used by connectionManager to decide when
-   * it's safe to upgrade. Note that a channel might be in the attaching state without any pending
-   * operations (eg if attached while the connection state is connecting) - such a channel must not
-   * hold up an upgrade, so is not considered inProgress.
-   * Operation is currently one of either 'statechange' or 'sync' */
-  setInProgress(channel: RealtimeChannel, operation: string, inProgress: boolean) {
-    this.inProgress[channel.name] = this.inProgress[channel.name] || {};
-    (this.inProgress[channel.name] as any)[operation] = inProgress;
-    if (!inProgress && this.hasNopending()) {
-      this.emit('nopending');
-    }
-  }
-
-  onceNopending(listener: ErrCallback) {
-    if (this.hasNopending()) {
-      listener();
-      return;
-    }
-    this.once('nopending', listener);
-  }
-
-  hasNopending() {
-    return Utils.arrEvery(
-      Utils.valuesArray(this.inProgress, true) as any,
-      function (operations: Record<string, unknown>) {
-        return !Utils.containsValue(operations, true);
-      }
-    );
   }
 }
 
--- a/src/common/lib/client/realtime.ts
+++ b/src/common/lib/client/realtime.ts
@@ -67,6 +67,23 @@ class Channels extends EventEmitter {
     });
   }
 
+  channelSerials(): { [ name: string ]: string } {
+    let serials: { [ name: string ]: string } = {};
+    for (const [name, channel] of Object.entries(this.all)) {
+      if (channel.channelSerial) {
+        serials[name] = channel.channelSerial;
+      }
+    }
+    return serials;
+  }
+
+  setRecoveryChannelSerials(channelSerials: { [ name: string ]: string }) {
+    for (const [name, serial] of Object.entries(channelSerials)) {
+      const channel = this.get(name);
+      channel.channelSerial = serial;
+    }
+  }
+
   onChannelMessage(msg: ProtocolMessage) {
     const channelName = msg.channel;
     if (channelName === undefined) {
@@ -153,7 +170,7 @@ class Channels extends EventEmitter {
     }
   }
 
-  get(name: string, channelOptions: ChannelOptions) {
+  get(name: string, channelOptions?: ChannelOptions) {
     name = String(name);
     let channel = this.all[name];
     if (!channel) {
--- a/src/common/lib/client/realtime.ts
+++ b/src/common/lib/client/realtime.ts
@@ -67,8 +67,8 @@ class Channels extends EventEmitter {
   channelSerials(): { [name: string]: string } {
     let serials: { [name: string]: string } = {};
     for (const [name, channel] of Object.entries(this.all)) {
-      if (channel.channelSerial) {
-        serials[name] = channel.channelSerial;
+      if (channel.properties.channelSerial) {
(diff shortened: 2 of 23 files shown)


src/common/lib/client/realtimechannel.ts
```diff
@@ -596,11 +587,18 @@ class RealtimeChannel extends Channel {
   }
 
   onMessage(message: ProtocolMessage): void {
+    const isNotControlMsg = message.action === actions.MESSAGE || message.action === actions.PRESENCE;
+    if (isNotControlMsg || message.action === actions.ATTACHED) {
+      const suppressed = this.setChannelSerial(message);
```
sacOO7, line 592: this always returns `false`, do we even need `suppress`?

src/common/lib/transport/connectionmanager.ts
```diff
@@ -114,22 +109,11 @@ export class TransportParams {
...
-          params.recover = match[0];
-          const recoverSerial = match[1];
-          if (isNaN(Number(recoverSerial))) {
-            params.timeSerial = recoverSerial;
-          } else {
-            params.connectionSerial = recoverSerial;
-          }
+        const recoveryContext = RecoveryContext.decode(options.recover);
+        if (recoveryContext) {
+          params.recover = recoveryContext.connectionKey;
```
sacOO7, line 116: Why are we assigning `connectionKey` to `recover`

sacOO7: It seems we are using `recover` as a string at several places.

src/common/lib/types/protocolmessage.ts
```diff
@@ -23,6 +23,7 @@ const actions = {
   MESSAGE: 15,
   SYNC: 16,
   AUTH: 17,
+  ACTIVATE: 18,
```
sacOO7, line 26: This is not documented in the draft

src/common/lib/types/protocolmessage.ts
```diff
@@ -70,13 +71,12 @@ class ProtocolMessage {
   connectionKey?: string;
   connectionSerial?: number;
```
sacOO7, line 72: `ConnectionSerial` needs to be removed as per spec

src/common/lib/client/realtimechannel.ts
andydunstall: nope - that was added before we decided duplicates we're not possible with the new upgrade scheme - thanks for spotting will remove

src/common/lib/types/protocolmessage.ts
andydunstall: This is only used for upgrade (no other SDKs will need this) - we haven't agreed what the message will be called so I just added this for the draft

src/common/lib/types/protocolmessage.ts
andydunstall: thanks will remove

src/common/lib/transport/connectionmanager.ts
andydunstall: Realtime expects the params `?<mode>=<connectionKey>` - this is the same behaviour as before

Yeah it decodes it in a few places to match the existing behaviour - can look at refactoring later

sacOO7: Yeah, I mean I am not getting why `connectionKey` is being assigned to `recover`

andydunstall: Since realtime expects `?recover=<connectionKey>` - where `recover` is the mode

This is the same behaviour as before

SimonWoolf: (I was about to comment linking the relevant spec item that requires this, but I... actually can't find it? Possibly this was never properly specified. I'll add something to my docs pr...)

sacOO7: That would be great!

SimonWoolf: made the spec a lot more explicit: https://github.com/ably/docs/commit/2a932d4c9e19a9855060994bc882cc27cf23b9a7

sacOO7: Thanks!!

test/realtime/resume.test.js
```diff
@@ -629,5 +622,121 @@ define(['shared_helper', 'async', 'chai'], function (helper, async, chai) {
...
+                    recoveredCount++;
+                  });
+                  setTimeout(function () {
+                    try {
+                      expect(recoveredCount).to.equal(MSG_COUNT);
+                      cb();
+                    } catch (err) {
+                      cb(err);
+                    }
+                  }, 2000);
```
sacOO7, line 710: Tests will become flaky because of the timeout, we can use [waitforexpect](https://www.npmjs.com/package/wait-for-expect) instead. It will move forward as soon as expectation is matched against given assertion.

sacOO7:
```
waitForExpect(() => {expect(numberToChange).toEqual(100)}).then(cb);
```

test/realtime/resume.test.js
```diff
@@ -629,5 +622,121 @@ define(['shared_helper', 'async', 'chai'], function (helper, async, chai) {
...
+                }), cb);
+              },
+              function (cb) {
+                // Subscribe to the next 5 channels after becoming connected
+                // (so the recovery key should be used to lookup channel serials
+                // when getting a channel).
+                async.parallel(rxRecoverChannels.slice(5).map((channel) => function (cb) {
+                  let recoveredCount = 0;
+                  channel.subscribe('sentWhileDisconnected', function(msg) {
```
sacOO7, line 719: Is this supposed to be `channel.subscribe('sentWhileConnected', function(msg) `

test/realtime/resume.test.js
sacOO7: We were using it in the `ably-echo` library -> https://github.com/ably-forks/echo/blob/afdf9c63ea81f2a8f54811d1a3e926577ea7dd2f/tests/ably/ably-channel.test.ts#L177

test/realtime/resume.test.js
```diff
@@ -629,5 +622,121 @@ define(['shared_helper', 'async', 'chai'], function (helper, async, chai) {
...
+              function (cb) {
+                rxRealtimeRecover = helper.AblyRealtime({ recover: recoveryKey });
+                rxRecoverChannels = channelNames.map((name) => rxRealtimeRecover.channels.get(name));
+                cb();
+              },
+              function (cb) {
+                // Subscribe to the first channels before becoming connected
+                // (so when the transport is activated the channels serials
+                // should be set).
+                async.parallel(rxRecoverChannels.slice(0, 5).map((channel) => function (cb) {
```
sacOO7, line 698: How do we make sure the connection is not yet in a `connected` state here?

sacOO7: Comment says to subscribe to channels before becoming `connected`, the `attach` cannot happen without connection being `connected`

test/realtime/resume.test.js
```diff
@@ -629,5 +622,121 @@ define(['shared_helper', 'async', 'chai'], function (helper, async, chai) {
...
+                    const tx = txRest.channels.get(name);
+                    tx.publish('sentWhileDisconnected', null, cb);
+                  }), cb);
+                }
+                // Publish another MSG_COUNT messages on each channel (which are no
+                // longer received by rxRealtime).
+                async.timesSeries(MSG_COUNT, publishAll, cb);
+              },
+              function (cb) {
+                rxRealtimeRecover = helper.AblyRealtime({ recover: recoveryKey });
```
sacOO7, line 690: Can we add a negative test for a scenario where `recoveryKey` is not passed as a clientOption after connection drops

src/common/lib/client/realtime.ts
```diff
@@ -55,18 +54,39 @@ class Realtime extends Rest {
...
+    }
+    return serials;
+  }
+
+  setRecoveryChannelSerials(channelSerials: { [ name: string ]: string }) {
+    for (const [name, serial] of Object.entries(channelSerials)) {
+      if (name in this.all) {
+        this.all[name].channelSerial = serial;
+      }
+      this.recoverySerials[name] = serial;
```
sacOO7, line 86: We don't need `recoverySerials` instead the simplified code can be 
```
    for (const [name, serial] of Object.entries(channelSerials)) {
        this.get(name, null).channelSerial = serial;
      }
```

src/common/lib/client/realtime.ts
```diff
@@ -165,6 +152,9 @@ class Channels extends EventEmitter {
     let channel = this.all[name];
     if (!channel) {
       channel = this.all[name] = new RealtimeChannel(this.realtime, name, channelOptions);
+      if (name in this.recoverySerials) {
```
sacOO7, line 155: This chunk of code can be removed once modified with above code.

src/common/lib/client/realtime.ts
sacOO7: `recoverySerials` var is introducing extra complexity and reduced code readability

test/realtime/resume.test.js
```diff
@@ -629,5 +624,63 @@ define(['shared_helper', 'async', 'chai'], function (helper, async, chai) {
...
+      // Attach recovered channels and check we receive the expected messages.
+      await Promise.all(rxRecoverChannels.map(async (channel) => {
+        await channel.attach();
+
+        let recoveredCount = 0;
+        channel.subscribe((msg) => {
+          recoveredCount++;
+        });
+        // Use a timeout to check we don't receive more messages than we expect.
+        await new Promise((resolve) => setTimeout(resolve, 2000));
```
sacOO7, line 677: Can we try avoid using timeouts, I had a chat with @SimonWoolf, seems we are not using timeouts anywhere else in tests

sacOO7:
Thanks for adding new test, it's more readable now : )

test/realtime/resume.test.js
andydunstall: That was just copying the existing tests in `resume.test.js` - though i agree kinda messy - have updated

test/realtime/resume.test.js
```diff
@@ -629,5 +624,67 @@ define(['shared_helper', 'async', 'chai'], function (helper, async, chai) {
...
+      }
+
+      const rxRealtimeRecover = helper.AblyRealtime({ recover: recoveryKey });
+      const rxRecoverChannels = channelNames.map((name) => rxRealtimeRecover.channels.get(name));
+
+      // Attach recovered channels and check we receive the expected messages.
+      await Promise.all(rxRecoverChannels.map(async (channel) => {
+        await channel.attach();
+
+        await new Promise((resolve) => {
```
sacOO7, line 672: I see two issues with this approach
1. Risk of test waiting for an indefinite amount of time to match the assertion, so it either passes or goes into wait state 
( rather than failing )
2. Where are we passing exception thrown by `expect` statement. As far as I know, thing work differently when you put `expect` statements inside a method. 
So, we need a separate method that does that, the `new Promise` would look like

```
const safeAssert = async (assertions: Function) => {
await new Promise((resolve, reject) => {
    try {
        assertions();
        resolve();
    } catch (err) {
        reject(err)
    }
});
});
await safeAssert(()=> {
          let recoveredCount = 0;
          channel.subscribe((msg) => {
            // Check we don't receive unexpected messages.
            expect(msg.name).to.equal('sentWhileDisconnected');

            recoveredCount++;
            if (recoveredCount === NUM_MSGS) {
              resolve();
            }
          });
})

```
This fix still doesn't cover first problem.

sacOO7:
Can we add spec id as a comment next to code onwards ... it's really hard to refer to the code and relevant spec item when looking at the PR

andydunstall:
> Can we add spec id as a comment next to code onwards ... it's really hard to refer to the code and relevant spec item when looking at the PR

Each commit is labeled with the spec ID - so probably best to walk though commit by commit

src/common/lib/client/realtimechannel.ts
```diff
@@ -79,6 +76,7 @@ class RealtimeChannel extends Channel {
   stateTimer?: number | NodeJS.Timeout | null;
   retryTimer?: number | NodeJS.Timeout | null;
   retryCount: number = 0;
+  channelSerial?: string | null;
```
sacOO7, line 79: This is supposed to go inside properties ->
```
  properties: { attachSerial: string | null | undefined, channelSerial?: string | null };
```

src/common/lib/client/realtime.ts
andydunstall: updated thanks!

src/common/lib/client/realtimechannel.ts
```diff
@@ -596,11 +586,21 @@ class RealtimeChannel extends Channel {
...
   onMessage(message: ProtocolMessage): void {
+    // Note setting channel serial here rather than the switch below to be
+    // explicit (as with fall throughs its easy to update from the wrong
+    // message).
+    if (
+      message.action === actions.ATTACHED ||
+      message.action === actions.MESSAGE ||
+      message.action === actions.PRESENCE
+    ) {
+      this.setChannelSerial(message.channelSerial);
```
sacOO7, line 597: Check if `channelSerial` is not empty, then only set it

andydunstall: Thats handled by `setChannelSerial`

sacOO7: As per RTL15b, use only if its set

andydunstall: Yep handled by `setChannelSerial`:
```javascript
    if (channelSerial) {
      this.properties.channelSerial = channelSerial;
    }
```

sacOO7:
We will be merging this PR to `integration-2.0`. rather than main right?

src/common/lib/client/connection.ts
```diff
@@ -71,6 +71,10 @@ class Connection extends EventEmitter {
     this.connectionManager.requestState({ state: 'closing' });
   }
 
+  get recoveryKey(): string | null {
+    return this.createRecoveryKey();
+  }
```
SimonWoolf, line 76: 😌 beautiful

andydunstall:
Given this PR is now old and the commit history quite messy I've replaced with https://github.com/ably/ably-js/pull/1110

PR #1046, opened 2022-08-26
eventemitter: fix once with array using promise

andydunstall:
when looking at upgrading the realtime test suite to use the latest SDK found `await conn.once(['foo', 'bar'])` no longer works using promises, throwing `Invalid arguments` (though callbacks is fine ie `conn.once(['foo', 'bar'], () => {})`

looks like a check was added in https://github.com/ably/ably-js/commit/32f8d04363df2e5ae40e2bae91c0248614111e48#diff-21811b221b0e30b57b9d5c084d2fd83f1709eb97a129e25732474b92b8fb63a6R237 - though I can't see what this is checking given its just calling itself so will be checked later (or if its just so typescript doesn't complain)?


owenpearson (approved):
LGTM, thanks

PR #1051, opened 2022-09-13
Configurable internetUpUrl

owenpearson:
Implementation is based on the one described by Lewis [here](https://github.com/ably/specification/issues/36). The only difference is that this implementation allows any success status code.

Includes some very basic URL parsing so that provided `internetAvailabilityUrl` can optionally contain a url scheme and/or query params.

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -210,6 +210,15 @@ declare namespace Types {
      */
     useBinaryProtocol?: boolean;
 
+    /**
+     * Override the URL used by the realtime client to check if the internet is available.
+     *
+     * In the event of a failure to connect to the primary endpoint, the client will send a
+     * GET request to this URL to check if the internet is available. If this request returns
+     * a success response the client will attempt to connect to a fallback host.
+     */
+    connectivityCheckUrl?: string;
+
     disconnectedRetryTimeout?: number;
     suspendedRetryTimeout?: number;
     closeOnUnload?: boolean;
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -219,6 +219,12 @@ declare namespace Types {
      */
     connectivityCheckUrl?: string;
 
+    /**
+     * Disable the check used by the realtime client to check if the internet
+     * is available before connecting to a fallback host.
+     */
+    disableConnectivityCheck?: boolean;
+
     disconnectedRetryTimeout?: number;
     suspendedRetryTimeout?: number;
     closeOnUnload?: boolean;
--- a/src/common/lib/util/defaults.ts
+++ b/src/common/lib/util/defaults.ts
@@ -295,6 +295,17 @@ export function normaliseOptions(options: DeprecatedClientOptions): NormalisedCl
     }
   }
 
+  let connectivityCheckParams = null;
+  let connectivityCheckUrl = options.connectivityCheckUrl;
+  if (options.connectivityCheckUrl) {
+    let [uri, qs] = options.connectivityCheckUrl.split('?');
+    connectivityCheckParams = qs ? Utils.parseQueryString(qs) : {};
+    if (uri.indexOf('://') === -1) {
+      uri = 'https://' + uri;
+    }
+    connectivityCheckUrl = uri;
+  }
+
   return {
     ...options,
     useBinaryProtocol:
@@ -305,6 +316,8 @@ export function normaliseOptions(options: DeprecatedClientOptions): NormalisedCl
     restHost,
     maxMessageSize: options.maxMessageSize || Defaults.maxMessageSize,
     timeouts,
+    connectivityCheckParams,
+    connectivityCheckUrl,
   };
 }
 
--- a/src/platform/nodejs/lib/util/http.ts
+++ b/src/platform/nodejs/lib/util/http.ts
@@ -226,6 +226,10 @@ const Http: typeof IHttp = class {
   }
 
   checkConnectivity = (callback: (errorInfo: ErrorInfo | null, connected?: boolean) => void): void => {
+    if (this.options.disableConnectivityCheck) {
+      callback(null, true);
+      return;
+    }
     const connectivityCheckUrl = this.options.connectivityCheckUrl || Defaults.connectivityCheckUrl;
     const connectivityCheckParams = this.options.connectivityCheckParams;
     const connectivityUrlIsDefault = !this.options.connectivityCheckUrl;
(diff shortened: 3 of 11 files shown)


Peter-Maguire (approved):

ably.d.ts
```diff
@@ -210,6 +210,14 @@ declare namespace Types {
      */
     useBinaryProtocol?: boolean;
 
+    /**
+     * Override the url used by the realtime client to check if the internet is available
+     * for private networks where access to the public internet not available.
```
lmars, line 215: I don't think this comment is very clear, since if a client is in a private network with no access to the public Internet, then why would the user configure `internetAvailabilityUrl` when they know the Internet is not reachable?

There are two use cases here:

1. a user who wants to use an "is the Internet reachable" URL under their own domain (e.g. `https://internet-up.acme.com`)
2. a user who knows their clients can't reach the Internet, so they want to disable the check completely

Perhaps these should be catered for by different client options?

src/platform/web/lib/util/http.ts
```diff
@@ -71,17 +80,27 @@ const Http: typeof IHttp = class {
...
+            headers?: any,
+            packed?: boolean,
+            statusCode?: number
+          ) {
+            let result = false;
+            if (!internetUpUrlIsDefault) {
+              result = !err && isSuccessCode(statusCode as number);
+            } else {
+              result = !err && (responseText as string)?.replace(/\n/, '') == 'yes';
+            }
```
lmars, line 103: I think this logic is unnecessary, if we get _any_ HTTP response, then the check is surely complete?

owenpearson: Yeah that's a fair point. I don't know the history of why we check the response body every time so I decided to err on the side of caution and preserve the existing behaviour.

ably.d.ts
owenpearson: OK I'll look into adding an option for disabling the check completely.

As for the comment being unclear, I'm not entirely sure what you're objecting to sorry, if I changed it to "check if the 
_network_ is available" would that make it clear?

lmars: > As for the comment being unclear, I'm not entirely sure what you're objecting to sorry, if I changed it to "check if the
network is available" would that make it clear?

I think just remove the "for private networks where access to the public internet not available.", since I don't believe this option relates to that use case.

src/platform/web/lib/util/http.ts
lmars: I suspect the history is that the original implementor thought, well if there's just one URL we control, and we know what we expect the response to be, perhaps we just assert what that response is, but "checking the Internet is available" doesn't require that the URL returns "yes", nor that it returns a success response, simply establishing a TCP connection is enough.

cc @paddybyers in case you have further insight into why we might want to this check to continue to be this specific.

paddybyers: There's always the possibility that there's something wrong with the client's DNS resolver, say, where the request resolves to something that responds with an error message, but isn't indicative of there being general connectivity.

That issue would be addressed by insisting that it's an HTTPS URL.

ably.d.ts
owenpearson: I've removed that part of the comment and added a `disableConnectivityCheck` client option to skip the check entirely.

lmars: Thanks 👍

ably.d.ts
```diff
@@ -210,6 +210,19 @@ declare namespace Types {
...
+     *
+     * Should return a success response when available.
+     */
+    internetAvailabilityUrl?: string;
+
+    /**
+     * Disable the check used by the realtime client to check if the internet
+     * is available before connecting to a fallback host.
+     */
+    disableConnectivityCheck?: boolean;
```
lmars, line 224: Perhaps match the naming with the URL option, and call it `disableInternetAvailabilityCheck`?

src/platform/nodejs/lib/util/http.ts
```diff
@@ -219,15 +226,32 @@ const Http: typeof IHttp = class {
   }
 
   checkConnectivity = (callback: (errorInfo: ErrorInfo | null, connected?: boolean) => void): void => {
-    // TODO: fix this
+    if (this.options.disableConnectivityCheck) {
+      callback(null, true);
+      return;
+    }
+    const internetUpUrl = this.options.internetAvailabilityUrl || Defaults.internetUpUrl;
```
lmars, line 233: Should we be consistent and call the default value `internetAvailabilityUrl` too?

src/common/lib/util/defaults.ts
```diff
@@ -295,6 +295,17 @@ export function normaliseOptions(options: DeprecatedClientOptions): NormalisedCl
...
   }
 
+  let internetAvailabilityParams = null;
+  if (options.internetAvailabilityUrl) {
+    let [uri, qs] = options.internetAvailabilityUrl.split('?');
+    const params = qs ? Utils.parseQueryString(qs) : {};
+    if (uri.indexOf('://') === -1) {
+      uri = (options.tls ? 'https://' : 'http://') + uri;
+    }
+    options.internetAvailabilityUrl = uri;
```
lmars, line 305: What's the use case for mutating this URL? Should it not just be opaque? What if the user wants to use TLS with Ably, but for whatever reason, their internet availability URL is non-TLS?

src/common/lib/util/defaults.ts
```diff
@@ -295,6 +295,17 @@ export function normaliseOptions(options: DeprecatedClientOptions): NormalisedCl
...
+    internetAvailabilityParams = params;
```
lmars, line 306: Is it necessary to split out the query string and pass it around separately, only to join it again when making the request?

ably.d.ts
lmars: As long as we're consistent, I'm indifferent 🙂

mattheworiordan (approved):

src/common/lib/util/defaults.ts
owenpearson: I've changed it a bit now so you can use plain HTTP if you pass in a scheme but otherwise it defaults to TLS.

As for mutating the URL, it's just easier to normalise it once on init I guess.

ably.d.ts
owenpearson: done

lmars (approved):

QuintinWillison:
I've added @m-hulbert as a reviewer, given there is API commentary added.

QuintinWillison (approved):
I think there could perhaps be more attention given to capitalisation within the API commentaries that have been added (e.g. `url` -> `URL`) but I will defer to DevEd/Docs team for input on that.

owenpearson:
@QuintinWillison I've updated the API doc for that option, fixing the mistake you mentioned and also adding a better description.

paddybyers (approved):
LGTM, thanks

m-hulbert (approved):
LGTM

PR #1058, opened 2022-10-04
RealtimeChannel: only set channelSerial if present

SimonWoolf:
(also properties.channelSerial rather than channel.channelSerial. honestly I'm not really sure what the point of putting some properties in a "properties" object, but shrug, I'm following the existing style)

### DIFF
--- a/src/common/lib/client/realtimechannel.ts
+++ b/src/common/lib/client/realtimechannel.ts
@@ -1,4 +1,8 @@
-import ProtocolMessage from '../types/protocolmessage';
+import ProtocolMessage, {
+  MessageProtocolMessage,
+  PresenceProtocolMessage,
+  SyncProtocolMessage,
+} from '../types/protocolmessage';
 import EventEmitter from '../util/eventemitter';
 import * as Utils from '../util/utils';
 import Channel from './channel';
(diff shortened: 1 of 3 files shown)


src/common/lib/client/realtimechannel.ts
```diff
@@ -650,6 +644,7 @@ class RealtimeChannel extends Channel {
         if (!message.presence) break;
       // eslint-disable-next-line no-fallthrough
       case actions.PRESENCE: {
+        this.setChannelSerial(message.channelSerial, false); // RTL15b
```
andydunstall, line 647: Will this include `channelSerial` from the `SYNC` - as it falls though - which is used for the sync cursor? (which is why I initially put it all at the top as this switch statement is confusing)

SimonWoolf: oooh yes. very good point.

SimonWoolf: https://github.com/ably/ably-js/pull/1058/commits/442f404e56969d1dfe2b18f435a32fa1bfd2264a WDYT? refactors out the fallthrough in the switch statement, makes it a bit more readable

andydunstall: Nice looks good - in which case can we split up the commits (to keep refactoring separate from behaviour changes - and remove the original commit)

(sorry I also pushed a change to my branch so this will need a rebase)

SimonWoolf: https://github.com/ably/ably-js/pull/1058/commits/4375b1a4630591064bb3c1a9139d742aa43505fd

SimonWoolf: (with your commits there's no behaviour changes left, all refactoring)

andydunstall (approved):
nice looks good

PR #1064, opened 2022-10-18
Add missing/undocumented docstrings

Peter-Maguire:
This PR adds the remaining docstrings that are missing from https://github.com/ably/ably-js/pull/1028


ably.d.ts
```diff
@@ -2024,11 +1986,9 @@ declare namespace Types {
      */
     get(params?: RestPresenceParams, callback?: paginatedResultCallback<PresenceMessage>): void;
     /**
-     * BEGIN LEGACY DOCSTRING
-     * Get the current presence member set for this channel. In the REST client library this method directly queries [Ably’s REST presence API](https://ably.com/documentation/rest-api#presence).
-     * END LEGACY DOCSTRING
+     * Retrieves the current members present on the channel and the metadata for each member, such as their [PresenceAction]{@link PresenceAction} and ID.
      *
-     * @param callback - Not yet documented.
+     * @param callback - A function which returns a [PaginatedResult]{@link PaginatedResult} object, containing an array of [PresenceMessage]{@link PresenceMessage} objects.
```
owenpearson, line 1991: The function here doesn't return a `PaginatedResult`, it's invoked with one when the operation is complete.

Following the convention used elsewhere in this file this should be something like:
>  A function which, upon success, will be called with a [PaginatedResult]{@link PaginatedResult} object, containing an array of [PresenceMessage]{@link PresenceMessage} objects.

owenpearson (changes requested):
Looks good! just one thing that needs changing

ably.d.ts
Peter-Maguire: https://github.com/ably/ably-js/pull/1064/commits/c2fc24ebbad5ccc7650b8f35ee44aad973ca8fb6

owenpearson (changes requested):
Looks great, nice work 🙂 

Just one thing I think we need to address: it is not clear from the descriptions for `EventEmitter.whenState` how it differs from `EventEmitter.once` (the difference is that whenState resolves immediately if the emitter is already in the supplied state).

owenpearson (approved):

PR #1068, opened 2022-11-03
fix: stop mutating `Defaults.agent` on client init

### LINKED ISSUE #1067: `normaliseOptions` mutates `Defaults` object
The `Defaults` object should be static and immutable once created, however when setting the `agent` string in `normaliseOptions` we mutate the `Defaults` object. This means that creating multiple clients in the same JS environment will lead to duplicate identifiers in the agent string.

owenpearson:
Resolves #1067

### DIFF
--- a/src/common/lib/client/auth.ts
+++ b/src/common/lib/client/auth.ts
@@ -609,7 +609,7 @@ class Auth {
           return client.baseUri(host) + path;
         };
 
-      const requestHeaders = Utils.defaultPostHeaders();
+      const requestHeaders = Utils.defaultPostHeaders(this.client.options);
       if (authOptions.requestHeaders) Utils.mixin(requestHeaders, authOptions.requestHeaders);
       Logger.logAction(
         Logger.LOG_MICRO,
--- a/src/common/lib/client/channel.ts
+++ b/src/common/lib/client/channel.ts
@@ -92,7 +92,7 @@ class Channel extends EventEmitter {
     const rest = this.rest,
       format = rest.options.useBinaryProtocol ? Utils.Format.msgpack : Utils.Format.json,
       envelope = this.rest.http.supportsLinkHeaders ? undefined : format,
-      headers = Utils.defaultGetHeaders(format);
+      headers = Utils.defaultGetHeaders(rest.options, format);
 
     if (rest.options.headers) Utils.mixin(headers, rest.options.headers);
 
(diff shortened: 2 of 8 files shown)


sacOO7:
Can we add tests for this, recreating multiple realtime instances scenario?
Also, we need one more test for checking `agent-param` being sent along when websocket connection is first made.
Currently, we only have `agent-header` test for rest and not realtime, you can check `http.test.js` file

owenpearson:
@sacOO7 I appreciate the input but I'm not gonna do any of those things because the purpose of this PR is to fix a bug, not to improve how we test the library. We could spend forever writing tests to check that creating multiple clients doesn't have any side-effects, but I don't think that would add enough value to justify the effort. As for testing that the agent is sent with websocket connections, sure, that would be nice, but it's not within the scope of this PR. Can you just review the bugfix here please?

sacOO7:
Actually it's difficult to guess if the bug is fixed, without a test available for the scenario to which the fix was applied. 
Actually one of the reasons the current bug exists is that we didn't have enough tests for `agents` being set properly.
Although adding real-time tests is not part of the bug fix, it can affect future bugs. 

I can at least expect one simple test as a part of the fix => 
**multiple rest instances with rest under http.test.js**

My sense of adding other tests is purely coming from a perspective where there are no bugs related to `agents` in the future. For me, I had to go through code for hours and check where things were going wrong. This is just to avoid future problems related to agents. Maybe you can create an issue to test` realtime` params (which might involve mocking the connection ) for the remaining 2 cases
1. single real-time instance with agent param being sent via websocket.
2. multiple real-time instances with proper checks for individual agent param being set (assertions would be same as 1st )

sacOO7:
> @sacOO7 I appreciate the input but I'm not gonna do any of those things because the purpose of this PR is to fix a bug, not to improve how we test the library. We could spend forever writing tests to check that creating multiple clients doesn't have any side-effects, but I don't think that would add enough value to justify the effort. As for testing that the agent is sent with websocket connections, sure, that would be nice, but it's not within the scope of this PR. Can you just review the bugfix here please?

Actually it's difficult to guess if the bug is fixed, without a test available for the scenario to which the fix was applied. 
Actually one of the reasons the current bug exists is that we didn't have enough tests for `agents` being set properly.
Although adding real-time tests is not part of the bug fix, it can affect future bugs. 

I can at least expect one simple test as a part of the fix => 
**multiple rest instances with rest under http.test.js**

My sense of adding other tests is purely coming from a perspective where there are no bugs related to `agents` in the future. For me, I had to go through code for hours and check where things were going wrong. This is just to avoid future problems related to agents. Maybe you can create an issue to test` realtime` params (which might involve mocking the connection ) for the remaining 2 cases
1. single real-time instance with agent param being sent via websocket.
2. multiple real-time instances with proper checks for individual agent param being set (assertions would be same as 1st )

sacOO7 (approved):

Peter-Maguire (approved):

PR #1094, opened 2022-12-13
fix: use correct sync event signature in base Transport class

### LINKED ISSUE #1093: React Native crash because connectionSerial type is object instead of number
It appears that [this commit](https://github.com/ably/ably-js/commit/2d3e0ae6c58cee90451e3e213afb21f520952d81#diff-9b7befa5663f39e9deaf9c8500109e233f464e51b87242ec7c74c56ec5820485R1137) introduced a regression because in `transport.ts`, the sync event is emitted like this:
```typescript
this.emit('sync', message.connectionId, message);
```
and in `connectionmanager.ts`, the callback signature is now different:

```typescript
transport.once('sync', function (connectionId: string, syncSerial: number) {
      clearTimeout(timeout);
      callback(null, connectionId, syncSerial);
    });
```

This is actually causing a crash with React Native on Android because the `connectionSerial` parameter is serialized as `&connectionSerial=[object Object]` when transformed into an URL. As a result, `WebSocketModule.java` in React Native is throwing a `java.lang.IllegalArgumentException`.

I believe the fix is to update the emitting code like this:
```typescript
this.emit('sync', message.connectionId, message.connectionSerial);
```

owenpearson:
Resolves #1093

### DIFF
--- a/src/common/lib/transport/transport.ts
+++ b/src/common/lib/transport/transport.ts
@@ -156,7 +156,7 @@ abstract class Transport extends EventEmitter {
       case actions.SYNC:
         if (message.connectionId !== undefined) {
           /* a transport SYNC */
-          this.emit('sync', message.connectionId, message);
+          this.emit('sync', message.connectionId, message.connectionId);
           break;
         }
         /* otherwise it's a channel SYNC, so handle it in the channel */


Peter-Maguire (approved):

src/common/lib/transport/transport.ts
```diff
@@ -156,7 +156,7 @@ abstract class Transport extends EventEmitter {
       case actions.SYNC:
         if (message.connectionId !== undefined) {
           /* a transport SYNC */
-          this.emit('sync', message.connectionId, message);
+          this.emit('sync', message.connectionId, message.connectionId);
```
sparga, line 159: Thanks for the fix! Is it expected to have `connectionId` twice?

PR #1110, opened 2023-01-04
[CAP-49] no connection serials

andydunstall:
Implements no connection serials as described by https://github.com/ably/specification/pull/88/files, and upgrade described by https://github.com/ably/realtime/issues/4216#issuecomment-1221407649.

### Protocol version
Updates the ably protocol version to `2.0`

**Spec**
* RSC7a: replaces version string `1.2` with `2.0`

**Testing**
This is already tested by `rest/http.test.js` and `realtime/init.test.js` which check the query string and headers are correct.

### Resume with `ATTACH`
Each channel tracks the latest `channelSerial` its received. Unlike with connection serials we no longer check for duplicates by comparing serials. This was only needed for the old upgrade scheme which could potentially have duplicates. We wanted to keep the `channelSerial` format opaque rather than having to parse it to use for comparison (such as is could be a legacy serial or a time serial).

When resuming, instead of including a `connectionSerial` in the query (which leaves the realtime backend to look up the channel serial for each channel and resume), the client just re-attaches the channels including the `channelSerial` in the `ATTACH` message.

This also removes the `checkAttachedMsgIndicators` check, which will re-attach if the channel has not received an `ATTACHED` after reconnecting, which is now redundant given we always re-attach.

**Spec**
* RTN15c6/RTN15c7: The client already sends an `ATTACH` on the new connection for channels in the `ATTACHING` and `SUSPENDED` states - so this just moves channels in `ATTACHED` states back to `ATTACHING` when we get a new connection so they re-attach
* RTL4c1: Includes the tracked `channelSerial` in the `ATTACH` message
* RTL15b: Updates `RealtimeChannel.properties.channelSerial` when receive `MESSAGE`/`PRESENCE`/`ATTACHED` actions and `channelSerial` is defined
* RTP5a1: Clears `RealtimeChannel.properties.channelSerial` when the channel enters `DETACHED`/`SUSPENDED`/`FAILED` states
* RTN15b2: Removes the `connectionSerial` query param on resume

**Testing**
`realtime/resume.test.js` already tests resume works under a number of scenarios

### Presence Re-Enter

**Spec**
* RTP17h: Stores own members by `clientId` only (rather than member key)
* RTP17f: When the channel becomes attached `RealtimePresence.onAttached` is called (though not if already attached as its `RealtimeChannel` already checks `state === this.state`). When attached `_ensureMyMembersPresent` runs to perform re-entry
* RTP17g:
  * Adds the `member` `id` attribute to the re-entry
  * No longer removes the member from the internal presence map
  * No longer checks that member key is not already a member of the main presence map

**Testing**
* With the new presence model `presence_auto_reenter` now re-enters both members (even the member that was already in the presence set) and `channel.presence.get` returns the same member (`one`) twice with different connection IDs
  * @SimonWoolf I think when we talked about it before this is expected? (I guess before `one` wouldn't have been re-entered as it was in the main presence map?)

### Recovery
The recover now needs to store the `channelSerials` for each attached channel (since this state is no longer kept by the realtime backend), which is encoded as JSON for simplicity rather than some custom format. This adds a new method `createRecoveryKey` to create the key rather than recalculating each time (which could get expensive). To keep compatibility also adds a `recoveryKey` getter that returns `createRecoveryKey`, which required updating `tsconfig` to target `es5` (agreed in https://github.com/ably/specification/pull/88/files).

**Spec**
* RTN16f: Already does this so is updated to decode the new recovery key format
* RTN16j: `recoverChannels` sets the `channelSerial` for all channels in the recovery key before connecting
* RTN16k: Already does this so is updated to decode the new recovery key format
* RTN16g: Adds a `createRecoveryKey` which returns the `msgSerial`, `connectionKey` and `name` to `channelSerial` pairs for each channel, encoded as JSON
  * RTN16g2: Checks if it has a connection key, if not returns `null` (if `CLOSED`, `CLOSING`, `FAILED`, or `SUSPENDED`, `connectionKey` is cleared
* RTN16d: Adds `recover multiple channels` which tests the conn id stays the same but the conn key changes after recover

**Testing**
* Adds a `recover multiple channels`, which is similar to `resume_active` except uses recover rather than resume. This also tests the connection id stays the same but the connection key changes as described by RTN16d.

### Upgrade
Implements the new upgrade scheme as described in option 1 of https://github.com/ably/realtime/issues/4216#issuecomment-1221407649 (not in the spec as ably-js specific). Also rather than keeping the `SYNC` action, which is also used by presence which causes some confusion, upgrade instead now uses the `ACTIVATE` action to activate the new transport.

1. Before activating the new transport after an upgrade we send an empty `ACTIVATE` message to the server and assumes the transport is active immediately after sending (since we'll reattach so doesn't matter if we drop messages),
2. When receiving the `ACTIVATE` the server sets the active transport and suspends the connections channels (the same as if it got a resume request),
3. Since the SDK has activated a new transport it re-attaches all channels as it would on a normal resume

The SDK no longer waits for `channels.onceNopending` since we re-attach once the new connection is activated.

**Testing**
* Already has lots of upgrade testing
* Removes the `unresponsive_upgrade_sync` test as this no longer makes sense (since we don't wait for `SYNC`)

## Spec
- [x] G4
- [x] RSC7a
- [x] RTN8c
- [x] RTN9c
- [x] RTN10
  - [x] RTN10a
  - [x] RTN10b
- [x] RTN15g3
- [x] RTN15b
  - [x] RTN15b2
- [x] RTN15c
  - [x] RTN15c1
  - [x] RTN15c2
  - [x] RTN15c3
  - [x] RTN15c6
  - [x] RTN15c7
- [x] RTN15f
- [x] RTN16
  - [x] RTN16a
  - [x] RTN16b
  - [x] RTN16c
  - [x] RTN16d
  - [x] RTN16e
  - [x] RTN16f
  - [x] RTN16i
  - [x] RTN16j
  - [x] RTN16k
  - [x] RTN16g
    - [x] RTN16g1
    - [x] RTN16g2
  - [x] RTN16d
  - [x] RTN16l
- [x] RTN19
  - [x] RTN19a1
  - [x] RTN19a2
- [x] RTN25
- [x] RTL4c1
- [x] RTL15b
- [x] RTP17
  - [x] RTP17h
  - [x] RTP17f
  - [x] RTP17g
  - [x] RTP17c
  - [x] RTP17d
- [x] RTP5
  - [x] RTP5a1
- [x] TR4f
- [x] CP2a
- [x] CP2b

### DIFF
--- a/src/common/lib/client/realtime.ts
+++ b/src/common/lib/client/realtime.ts
@@ -100,6 +100,10 @@ class Channels extends EventEmitter {
         channel.checkPendingState();
       } else if (channel.state === 'suspended') {
         channel._attach(false, null);
+      } else if (channel.state === 'attached') {
+        // Note explicity request the state, channel.attach() would do nothing
+        // as its already attached.
+        channel.requestState('attaching');
       }
     }
   }
--- a/src/common/lib/client/realtime.ts
+++ b/src/common/lib/client/realtime.ts
@@ -108,17 +108,6 @@ class Channels extends EventEmitter {
     }
   }
 
-  reattach(reason: ErrorInfo) {
-    for (const channelId in this.all) {
-      const channel = this.all[channelId];
-      /* NB this should not trigger for merely attaching channels, as they will
-       * be reattached anyway through the onTransportActive checkPendingState */
-      if (channel.state === 'attached') {
-        channel.requestState('attaching', reason);
-      }
-    }
-  }
-
   /* Connection interruptions (ie when the connection will no longer queue
    * events) imply connection state changes for any channel which is either
    * attached, pending, or will attempt to become attached in the future */
--- a/src/common/lib/client/realtime.ts
+++ b/src/common/lib/client/realtime.ts
@@ -119,30 +119,6 @@ class Channels extends EventEmitter {
     }
   }
 
-  resetAttachedMsgIndicators() {
-    for (const channelId in this.all) {
-      const channel = this.all[channelId];
-      if (channel.state === 'attached') {
-        channel._attachedMsgIndicator = false;
-      }
-    }
-  }
-
-  checkAttachedMsgIndicators(connectionId: string) {
-    for (const channelId in this.all) {
-      const channel = this.all[channelId];
-      if (channel.state === 'attached' && channel._attachedMsgIndicator === false) {
-        const msg =
-          '30s after a resume, found channel which has not received an attached; channelId = ' +
-          channelId +
-          '; connectionId = ' +
-          connectionId;
-        Logger.logAction(Logger.LOG_ERROR, 'Channels.checkAttachedMsgIndicators()', msg);
-        channel.requestState('attaching');
-      }
-    }
-  }
-
   /* Connection interruptions (ie when the connection will no longer queue
    * events) imply connection state changes for any channel which is either
    * attached, pending, or will attempt to become attached in the future */
--- a/src/common/lib/client/realtime.ts
+++ b/src/common/lib/client/realtime.ts
@@ -140,7 +140,7 @@ class Channels extends EventEmitter {
     }
   }
 
-  get(name: string, channelOptions: ChannelOptions) {
+  get(name: string, channelOptions?: ChannelOptions) {
     name = String(name);
     let channel = this.all[name];
     if (!channel) {
--- a/src/common/lib/client/realtime.ts
+++ b/src/common/lib/client/realtime.ts
@@ -8,7 +8,6 @@ import Defaults from '../util/defaults';
 import ErrorInfo from '../types/errorinfo';
 import ProtocolMessage from '../types/protocolmessage';
 import { ChannelOptions } from '../../types/channel';
-import { ErrCallback } from '../../types/utils';
 import ClientOptions, { DeprecatedClientOptions } from '../../types/ClientOptions';
 import * as API from '../../../../ably';
 import ConnectionManager from '../transport/connectionmanager';
@@ -55,13 +54,11 @@ class Realtime extends Rest {
 class Channels extends EventEmitter {
   realtime: Realtime;
   all: Record<string, RealtimeChannel>;
-  inProgress: Record<string, RealtimeChannel>;
 
   constructor(realtime: Realtime) {
     super();
     this.realtime = realtime;
     this.all = Object.create(null);
-    this.inProgress = Object.create(null);
     realtime.connection.connectionManager.on('transport.active', () => {
       this.onTransportActive();
     });
@@ -109,9 +106,7 @@ class Channels extends EventEmitter {
   }
 
   /* called when a transport becomes connected; reattempt attach/detach
-   * for channels that are attaching or detaching.
-   * Note that this does not use inProgress as inProgress is only channels which have already made
-   * at least one attempt to attach/detach */
+   * for channels that are attaching or detaching. */
   onTransportActive() {
     for (const channelName in this.all) {
       const channel = this.all[channelName];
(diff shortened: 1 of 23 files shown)


SimonWoolf (approved):

SimonWoolf:
Looked into failing ci, found a flaky test `setOptionsCallbackBehaviour` caused by the slight change in behaviour of channels going into the ATTACHING state after the transport upgrade until reattach, fix in https://github.com/ably/ably-js/pull/1110/commits/b1e0b4889bd8db5a3e6cfdf0aa367e484326f274

src/common/lib/client/realtime.ts
```diff
@@ -55,18 +54,34 @@ class Realtime extends Rest {
...
     this.all = Object.create(null);
-    this.inProgress = Object.create(null);
     realtime.connection.connectionManager.on('transport.active', () => {
       this.onTransportActive();
     });
   }
 
+  channelSerials(): { [name: string]: string } {
+    let serials: { [name: string]: string } = {};
+    for (const [name, channel] of Object.entries(this.all)) {
```
owenpearson, line 69: `Object.entries` would be a breaking change for IE11 and below. I'd suggest using:
```ts
for (const name of Utils.keysArray(this.all, true)) {
    const channel = this.all[name];
    // ...
}
```

src/common/lib/client/realtimechannel.ts
```diff
@@ -795,6 +798,11 @@ class RealtimeChannel extends Channel {
     );
     this.clearStateTimer();
 
+    // RTP5a1
+    if (['detached', 'suspended', 'failed'].includes(state)) {
```
owenpearson, line 802: Same goes for `Array.includes` wrt IE. We have a `Utils.arrIn` helper for this.

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es3",
+    "target": "es5",
```
owenpearson, line 3: What was the reason for this change? I think I set it to es3 because for some reason that resulted in a slightly smaller output size

test/realtime/resume.test.js
```diff
@@ -629,5 +624,87 @@ define(['shared_helper', 'async', 'chai'], function (helper, async, chai) {
         closeAndFinish(done, [sender_realtime, receiver_realtime, resumed_receiver_realtime], err);
       }
     });
+
+    // Tests recovering multiple channels only receives the expected messages.
+    it('recover multiple channels', async function () {
```
owenpearson, line 629: We don't use async/await in tests and if we need to use the Promise API we wrap it in an `if (typeof Promise !== 'undefined') just so that the tests run in older browsers. I know it's a pain writing tests using the `async` library so i'd be fine with just using `Promise.then` but as it stands i think this will break all of the tests on IE because of unsupported syntax

tsconfig.json
SimonWoolf: the `recoveryKey` getter, to allow it to generate `Object.defineProperty` with an accessor descriptor

src/common/lib/client/realtime.ts
SimonWoolf: :+1: (or `name in this.all` should be fine given it's a null-prototype ob)

tsconfig.json
andydunstall: (we discussed it in https://ably-real-time.slack.com/archives/C02P7F81DPS/p1669125337243929 - you said this branch still had a smaller output size than main)

src/common/lib/util/defaults.ts
```diff
@@ -76,7 +76,7 @@ const Defaults = {
   maxMessageSize: 65536,
 
   version,
-  apiVersion: '1.2',
+  apiVersion: '2.0',
```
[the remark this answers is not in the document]
andydunstall: thanks updated

src/common/lib/client/realtimechannel.ts
andydunstall: thanks updated

src/common/lib/client/realtime.ts
andydunstall: thanks updated

test/realtime/resume.test.js
andydunstall: Just to be clarify do you mean wrap in `if (typeof Promise !== 'undefined')` and replace with `Promise.then` - or remove promises entirely and use `async` library?

tsconfig.json
owenpearson: ah i remember now, thank you 👍

test/realtime/resume.test.js
owenpearson: either would be fine, but the former would probably be easier

andydunstall: ok sure replaced with `async` library (`.then` was being a pain)

owenpearson (changes requested):
Looks like [init_usetokenauth_defaulttokenparams_wildcard])https://github.com/ably/ably-js/blob/5bb2b3206b7549d31edbf09a1d77eb9c9c9fc976/test/realtime/init.test.js#L124-L151) is failing pretty consistently in the browser tests 😕

andydunstall:
> Looks like [init_usetokenauth_defaulttokenparams_wildcard](https://github.com/ably/ably-js/blob/5bb2b3206b7549d31edbf09a1d77eb9c9c9fc976/test/realtime/init.test.js#L124-L151) is failing pretty consistently in the browser tests 😕

Just ran again - seemed ok this time

That test seems to all be auth which I don't see how these changes could have broken (or only affected browsers)

SimonWoolf:
Re init_usetokenauth_defaulttokenparams_wildcard, I can repro, fixed by https://github.com/ably/ably-js/pull/1110/commits/85d143222bc9c4bc8f0b8787abf817ca6de8a65b

It was a race -- with a 12s ttl, realtime was asking the lib to start a reauth as soon as it connected, and with unlucky timings, the original token details could be cleared just before the test code looked at it. Not caused by this pr, but seems likely that small timings changes in this pr could have caused the race to show up more often

SimonWoolf:
history_until_attach flake fixed serverside by Andy (unrelated to this pr, was from the channeltimeserial change on sandbox)

latest run had two failures, both rest/push/Publish, which Owen says is a known flake

owenpearson (approved):
LGTM, thanks!

PR #1113, opened 2023-01-12
fix: use singleton agent for default restAgentOptions

owenpearson:


src/platform/nodejs/lib/util/http.ts
```diff
@@ -192,14 +194,26 @@ const Http: typeof IHttp = class {
...
+    const userAgentOptions = rest && rest.options.restAgentOptions;
+    const agentOptions = userAgentOptions || Defaults.restAgentOptions;
     // const doOptions: RequestOptions = {uri, headers: headers ?? undefined, encoding: null, agentOptions: agentOptions};
     const doOptions: Options = { headers: headers || undefined, responseType: 'buffer' };
     if (!this.agent) {
-      this.agent = {
-        http: new http.Agent(agentOptions),
-        https: new https.Agent(agentOptions),
-      };
+      if (!userAgentOptions) {
```
SimonWoolf, line 202: It's not great that just passing in that you want an eg max-20 socket pool instead of max-40, which seems innocuous, could trigger the behaviour to change from global to per-instance agents.

But I appreciate full per-pool-options-keyed meta-pool would be overkill. As a compromise, WDYT about adding a `Utils.deepEqual`, storing the options the global-agent was constructed with with the global-agent, and only making a per-lib instance if it differs? (That'd still be not great, eg if someone makes a first rest instance with one set of agentOtions and then hundreds more with a different set they'd get per-instance agents... but that'd be a sufficiently weird/rare thing to do that I think we could be ok with that?)

(maybe also an `agentOptions.global: boolean` that could be set to `false` to ignore the global agent, so someone could easily force per-instance agents if they want to?)

owenpearson: I actually went ahead with options keyed pool in [272e5f4](https://github.com/ably/ably-js/pull/1113/commits/272e5f45c42e35b40b21066d879a357de8be4ef9) and it ended up being fairly simple so I'm currently thinking we should just go ahead with that, wdyt?

SimonWoolf: Sure :+1:

SimonWoolf (approved):
Thanks!

Peter-Maguire (approved):

PR #1122, opened 2023-02-02
Node/http: stop got from doing its own retries

SimonWoolf:
Got has an actually quite nice and configurable [retry API](https://github.com/sindresorhus/got/blob/main/documentation/7-retry.md).

But so do we, and having both stacking is definitely undesirable. Got's defaults aren't great (retrying 429s?); that's fixable, but we in any case prefer ours because we want to have retries go to fallback endpoints / other datacenters, which got can't do. So this disables got retries.

(Might be worth scanning through got docs to see if it's doing anything else overly-helpfully which we want to disable in favour of our own existing mechanisms)


owenpearson (approved):
LGTM, thanks

PR #1124, opened 2023-02-09
Fix message interactions signature

cruickshankpg:
Message reactions can only be published with the publish(Message) or publish(Message[]) signatures, not with the publish(name, data) signature.


SimonWoolf (approved):

cruickshankpg:
@SimonWoolf I don't have merge permissions on sdks, would you mind doing the honours?

PR #1128, opened 2023-02-16
fix: don't wrap client provided reauth if failing due to clientId mismatch

owenpearson:


src/common/lib/transport/connectionmanager.ts
```diff
@@ -2086,6 +2086,10 @@ class ConnectionManager extends EventEmitter {
       Logger.logAction(Logger.LOG_ERROR, 'ConnectionManager.actOnErrorFromAuthorize()', msg);
       this.notifyState({ state: 'failed', error: new ErrorInfo(msg, 80019, 403, err) });
     } else {
+      if (err.code === 40102) {
+        this.notifyState({ state: this.state.failState as string, error: err });
+        return;
+      }
```
SimonWoolf, line 2092: This clause will never fire here, (mismatch, 40102, 403) will already go down the path of the previous clause -- needs to be above that

(also unnecessarily double-nested if when it could be part of the flat if/elseif/elseif/else)

owenpearson: ah good point. very unintuitive that we trigger behaviour by sending different status codes with the same error code.

SimonWoolf: shrug, yeah. it's cause we wanted to give the customer an easy way of refusing a token request from someone in a way that will stop that person from retrying, and having their auth server just return a particular http status code (403) is convenient for that -- we don't want to rely on the customer's server giving us a well-formed ErrorInfof

Peter-Maguire (approved):

PR #1142, opened 2023-03-02
fix: align exposed type of ErrorInfo with internal type

JCB-K:
The type of the ErrorInfo class was incorrect when using the library, due to the `ably.d.ts` not being in sync with 

Not sure if I did this right -  is it okay to edit ably.d.ts manually?


owenpearson (approved):
Ah good spot and thanks for the PR 🙂 and yes, `ably.d.ts` isn't generated or anything so manually editing it is fine 👍

PR #1169, opened 2023-03-30
RealtimeChannel: populate invalid state error cause with the current errorReason, if set

SimonWoolf:
per customer feedback that the `cause` not being populated in this error is confusing/unhelpful


owenpearson (approved):
LGTM

PR #1171, opened 2023-03-31
RTN20c: Retry connection attempt if 'online' while `CONNECTING`

owenpearson:
See ably/specification#136

Upon an 'online' event, if the connectionmanager state is `CONNECTING`, this will silently dispose of all pending transports and call `startConnect` again. All listeners are detached from the transports before disposal in order to prevent connection state side effects (otherwise the state will temporarily become `DISCONNECTED`).

### DIFF
--- a/src/common/lib/transport/connectionmanager.ts
+++ b/src/common/lib/transport/connectionmanager.ts
@@ -364,8 +364,18 @@ class ConnectionManager extends EventEmitter {
             'reattempting connection'
           );
           this.requestState({ state: 'connecting' });
+        } else if (this.state == this.states.connecting) {
+          // RTN20c: if 'online' event recieved while CONNECTING, abandon connection attempt and retry
+          this.pendingTransports.forEach(function (transport) {
+            // Detach transport listeners to avoid connection state side effects from calling dispose
+            transport.off();
+          });
+          this.disconnectAllTransports();
+
+          this.startConnect();
         }
       });
+
       addEventListener('offline', () => {
         if (this.state == this.states.connected) {
           Logger.logAction(
--- a/test/browser/connection.test.js
+++ b/test/browser/connection.test.js
@@ -182,6 +182,35 @@ define(['shared_helper', 'chai'], function (helper, chai) {
         });
       });
 
+      it('device_going_online_causes_connecting_connection_to_retry_attempt', function (done) {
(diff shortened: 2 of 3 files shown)


src/common/lib/transport/connectionmanager.ts
```diff
@@ -364,8 +364,20 @@ class ConnectionManager extends EventEmitter {
             'reattempting connection'
           );
           this.requestState({ state: 'connecting' });
+        } else if (this.state == this.states.connecting) {
+          // RTN20c: if 'online' event recieved while CONNECTING, abandon connection attempt and retry
+          this.pendingTransports.forEach(function (transport) {
```
SimonWoolf, line 369: can just call `this.disconnectAllTransports()`, no need to reimplement

(in particular -- there's a subtle bug here cause forEach isn't safe when the callback mutates the array being iterated over, which `disconnectAllTransports()` avoids by using `Utils.safeArrForEach`)

src/common/lib/transport/connectionmanager.ts
```diff
@@ -364,8 +364,20 @@ class ConnectionManager extends EventEmitter {
...
+            // Detach transport listeners to avoid connection state side effects from calling dispose
+            transport.off();
+
+            transport.dispose();
+          });
+          this.pendingTransports = [];
+
+          this.startConnect();
```
SimonWoolf, line 377: (was about to complain that if this happens near the end of the connecting sequence it might not have enough time to actually connect, but I've just noticed that startConnect() does reset the transition timer so that's fine)

SimonWoolf (approved):
:+1:  approved with one change request

src/common/lib/transport/connectionmanager.ts
owenpearson: i did try `disconnectAllTransports` in my initial attempt at this but that caused connection state side effects, i will check again if there's any easy way around that but otherwise will just move to safeArrForEach

owenpearson: OK I've changed[f22e9af](https://github.com/ably/ably-js/pull/1171/commits/f22e9af9b82f21b5c05982b867c8d4eb02353f24) to still use `forEach` to call `.off` on all pending transports but now it uses `disconnectAllTransports` to clean up cm state

PR #1174, opened 2023-04-04
refactor: fix ClientOptions.agents typing

owenpearson:
This was always parsed as a `Record<string, string>` but for some reason was typed as a `string[]`. I guess it was like that from the original implementation and typescript didn't complain since you can do a `for in` loop on an array.

### DIFF
--- a/src/common/types/ClientOptions.ts
+++ b/src/common/types/ClientOptions.ts
@@ -9,7 +9,7 @@ export type RestAgentOptions = {
 export default interface ClientOptions extends API.Types.ClientOptions {
   restAgentOptions?: RestAgentOptions;
   pushFullWait?: boolean;
-  agents?: string[];
+  agents?: Record<string, string | undefined>;
 }
 
 export type DeprecatedClientOptions = Modify<


lawrence-forooghian:
Is there a reason we decided not to include this property in `ably.d.ts`? As I understand it, it's meant to be part of the public API of the library, [just documented to say it's only for use by other Ably-authored SDKs](https://sdk.ably.com/builds/ably/specification/main/features/#RSC7d6b). It would be good for our SDKs to be able to take advantage of type checking when populating this property.

lawrence-forooghian:
([Example of this property in public API in ably-cocoa](https://github.com/ably/ably-cocoa/blob/e1ff8ead277ddf99e33bced2cca1499afb7c0001/Source/include/Ably/ARTClientOptions.h#L195-L198))

lawrence-forooghian (approved):
> I guess it's just a trade off between whether to keep `ClientOptions` concise or to allow other Ably SDKs to get a typed `ClientOptions.agents`.

Personally I'm in favour of the latter. But the change in this PR is a good one regardless 👍

Peter-Maguire (approved):

PR #1185, opened 2023-04-12
Allow channel get with filter option

mohyour:


src/common/lib/client/realtime.ts
```diff
@@ -161,6 +161,19 @@ class Channels extends EventEmitter {
...
+    name = String(name);
+    if (deriveOptions.filter) {
+      const filter = encodeURIComponent(deriveOptions.filter)
+      const match = Utils.matchDerivedChannel(name)
+      name = `[filter=${filter}${match}`
+    }
+    if (channelOptions) {
+      return this.get(name, channelOptions)
+    }
+    return this.get(name)
```
owenpearson, line 174: You can simplify this a bit - if `channelOptions` is undefined then `get()` will just ignore it either way
```suggestion
    return this.get(name, channelOptions)
```

src/common/lib/client/realtime.ts
```diff
@@ -161,6 +161,19 @@ class Channels extends EventEmitter {
     return channel;
   }
 
+  getDerived(name: string, deriveOptions: API.Types.DeriveOptions, channelOptions?: ChannelOptions) {
+    name = String(name);
```
owenpearson, line 165: I don't think this is necessary. If anything it would be better to throw an error if the value passed isn't already a string

src/common/lib/util/utils.ts
```diff
@@ -547,3 +547,20 @@ export function shallowEquals(source: Record<string, unknown>, target: Record<st
...
+export function matchDerivedChannel(name: string) {
+  /**
+   * This warning is triggered because the RegExp uses nested quantifiers,
+   * but it does not create any situation where the regex engine has to
+   * explore a large number of possible matches so it’s safe to ignore
+   */
+  const regex = /^(\[([^?]*)(?:(.*))\])?(.+)$/; // eslint-disable-line
+  const match = name.match(regex)
+  if (!match || !match.length || match.length < 5) {
+    throw new Error(`regex match failed`);
```
owenpearson, line 560: These should be `ErrorInfo`s

src/common/lib/util/utils.ts
```diff
@@ -547,3 +547,20 @@ export function shallowEquals(source: Record<string, unknown>, target: Record<st
...
+  }
+  if (match[2]) {
+    throw new Error(`cannot use a derived option with a ${match[2]} channel`);
```
owenpearson, line 563: Might be worth adding some comments around the logic in this function - it's quite hard to read the regex and work out exactly what causes this error message

ably.d.ts
```diff
@@ -986,6 +986,17 @@ declare namespace Types {
     modes?: ChannelModes;
   }
 
+
+  /**
+   * Passes an attachment to a {@link ChannelBase} name to give a new derived channel
+   */
+  interface DeriveOptions {
+    /**
+     * The filter string to be used to derive new channel.
```
owenpearson, line 995: This is public facing documentation so it would probably help to mention here that the filter string should be a JMESPath query.

Also, might be worth adding the new docstrings to https://github.com/ably/sdk-api-reference/blob/main/descriptions.md

owenpearson (changes requested):
Looks good so far, a few suggestions:

Peter-Maguire (approved):
It's worth mentioning that some work I did adding local filtered subscriptions (https://github.com/ably/ably-js/pull/1003) should probably be replaced with this system at some point

src/common/lib/util/utils.ts
```diff
@@ -547,3 +547,26 @@ export function shallowEquals(source: Record<string, unknown>, target: Record<st
...
+  const match = name.match(regex);
+  if (!match || !match.length || match.length < 5) {
+    throw new ErrorInfo('regex match failed', 400, 40010);
+  }
+  // Fail if there is already a channel qualifier, eg [meta]foo should fail instead of just overriding with [filter=xyz]foo
+  if (match![2]) {
+    throw new ErrorInfo(`cannot use a derived option with a ${match![2]} channel`, 400, 40010);
+  }
+  // Return match values to be added to derive channel quantifier.
+  return `${match![3] || ''}]${match![4]}`;
```
mschristensen, line 571: It seems a unexpected to return a string here with a closing bracket `]` in the middle of it. I would suggest we returned two values `match![3] || ''` and `match![4]` in an object with reasonable names as keys (e.g. `qualifierParam` and `channelName`), and then use those from the caller and make the caller responsible for closing their parentheses.

test/common/ably-common
mschristensen, line 1: should this submodule be updated in a separate PR?

ably.d.ts
```diff
@@ -986,6 +986,16 @@ declare namespace Types {
     modes?: ChannelModes;
   }
 
+  /**
+   * Passes an attachment to a {@link ChannelBase} name to give a new derived channel
```
mschristensen, line 990: I think this is always RealtimeChannelBase in this case?

What is meant by "passes an attachment"? I might be missing some context, but to me this should be something like "passes additional properties to a RealtimeChannelBase to produce a new derived channel"

mschristensen (changes requested):
Looking pretty good to me, nice one Moyo!

Added some small comments.

test/common/ably-common
mohyour: Yes, the submodule contain changes to the test app setup to update api key with capabilities for filtered subscription. Not updating it will mean the test for `getDerived` will fail with access denied based on capability. We can always have a PR to update the submodule first and have that merged before this but I thought since the change in submodule relates to the new method, we can have it in one PR.

mschristensen: Ah got it, thanks Moyo!

Linking ably-common PR for reference: https://github.com/ably/ably-common/pull/196

mschristensen:
Linking ably-common PR for reference: https://github.com/ably/ably-common/pull/196

andydunstall (approved):

test/realtime/message.test.js
```diff
@@ -1183,5 +1183,47 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
...
+              try {
+                expect(msg.data).to.equal(testData.data, 'Unexpected msg text received');
+              } catch (err) {
+                closeAndFinish(done, realtime, err);
+                return;
+              }
+              closeAndFinish(done, realtime);
+            });
+
+            rtChannel.publish("filtered", "no filter on message");
```
mattheworiordan: Given you are publishing the non-filtered message first, `closeAndFinish(done, realtime)` i.e. the successful path, will be reached first, and thus the test will pass before it's had a chance to receive the filtered message. I strongly recommend you take a more robust approach where perhaps you publish 10 messages, 5 of which are filtered, then when a non-filtered channel receives 10, you check that the filtered channel has received 5 and continue.

mschristensen: I think the message data here is misleading, the first message is actually the one that will be filtered, as it does not include `headers.number: 26095`. Therefore, if the filter was not working, the subscription should receive this message first and fail.

So, for clarity, I'd suggest we change "no filter on message" to e.g. "no header on message".

mattheworiordan: Ah, yes, you are right... good spot. Thanks for clearing up.

I do think it would be good regardless to have a filtered & non-filtered channel open at the same time to ensure this is possible and we're not somehow mistakenly applying filters to that channel as opposed to that instance of the channel.

mschristensen (changes requested):
@mattheworiordan made a good point that we should add a test to ensure that a filtered channel does not return the same results as an unfiltered channel within the SDK itself (i.e. to ensure we’re not mistakenly sharing a Channel instance for example). cc @mohyour 

Additionally, should this PR also have a spec change @owenpearson ? cf. https://ably.atlassian.net/wiki/spaces/product/pages/2560098305/DR2%3A+Filtered+Subscription+SDK+Implementation+Decision?focusedCommentId=2562785361

owenpearson:
@mschristensen yep, we should definitely have a spec change landed before we merge these changes 👍

test/realtime/message.test.js
```diff
@@ -1183,5 +1183,119 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
...
+      };
+
+      try {
+        /* set up realtime */
+        var realtime = helper.AblyRealtime({ key: helper.getTestApp().keys[5].keyStr });
+        var rest = helper.AblyRest();
+
+        realtime.connection.on('connected', function () {
+          var rtFilteredChannel = realtime.channels.getDerived('chan', filterOption);
+          var rtUnFilteredChannel = realtime.channels.get('chan');
```
owenpearson, line 1236:
```suggestion
          var rtUnfilteredChannel = realtime.channels.get('chan');
```

test/realtime/message.test.js
```diff
@@ -1183,5 +1183,119 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
           });
       });
     }
+
+    it('subscribes to filtered channel', function (done) {
+      var testData = [
+        {
+          name: 'filtered',
+          data: 'This message can be filtered',
```
owenpearson, line 1191: For clarity I think this would be better
```suggestion
          data: 'These headers match the JMESPath expression so this message should not be filtered',
```

test/realtime/message.test.js
```diff
@@ -1183,5 +1183,119 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
...
+            },
+          },
+        },
+        {
+          name: 'filtered',
+          data: 'Random data with no extras for filter',
+        },
+        {
+          name: 'filtered',
+          data: 'Another filtered message',
```
owenpearson, line 1206: I think the use of the word 'filtered' here is kind of misleading
```suggestion
          data: 'This message should also not be filtered',
```

test/realtime/message.test.js
```diff
@@ -1183,5 +1183,119 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
...
+          extras: {
+            headers: {
+              name: 'John Bull',
+              number: 26095,
+              bool: false,
+            },
+          },
+        },
+        {
+          name: 'filtered',
```
owenpearson, line 1216: It would be good to have a message with headers but with `number !== 26095` just to make sure that the JMESPath engine is being used correctly

owenpearson (changes requested):

test/realtime/message.test.js
```diff
@@ -1183,5 +1183,119 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
...
+        var rest = helper.AblyRest();
+
+        realtime.connection.on('connected', function () {
+          var rtFilteredChannel = realtime.channels.getDerived('chan', filterOption);
+          var rtUnFilteredChannel = realtime.channels.get('chan');
+
+          var filteredMessages = [];
+          var unFilteredMessages = [];
+          /* subscribe to event */
+          rtFilteredChannel.subscribe('filtered', function (msg) {
```
mschristensen, line 1241: We don't need the client-side name event name filters, let's just look at everything that comes back on the channel, since

```
name == `"filtered"`
```

in the expression will do this server-side.

test/realtime/message.test.js
```diff
@@ -1183,5 +1183,119 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
...
+              // Push received unfiltered messages into an array
+              unFilteredMessages.push(msg);
+            } catch (err) {
+              closeAndFinish(done, realtime, err);
+              return;
+            }
+          });
+
+          // Subscription to check all messages were received as expected
+          rtUnFilteredChannel.subscribe('end', function (msg) {
```
mschristensen, line 1262: When we remove the client-side filter on the "filtered" event name, we can split this logic into two (for filtered vs unfioltered assertions) and apply this logic in the subscription callbacks when the message name is "end".

mschristensen (changes requested):

src/common/lib/client/realtime.ts
```diff
@@ -161,6 +161,15 @@ class Channels extends EventEmitter {
     return channel;
   }
 
+  getDerived(name: string, deriveOptions: API.Types.DeriveOptions, channelOptions?: ChannelOptions) {
+    if (deriveOptions.filter) {
+      const filter = encodeURIComponent(deriveOptions.filter);
```
mschristensen, line 166: We are making a change so that we now use a base64 encoding instead of URL encoding. See this PR: https://github.com/ably/realtime/pull/4835/files

Can we please update this PR to reflect these changes before we merge?

mschristensen (changes requested):

test/realtime/message.test.js
```diff
@@ -1183,5 +1183,136 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
...
+            } catch (err) {
+              closeAndFinish(done, realtime, err);
+              return;
+            }
+          });
+
+          // Subscription to check all messages were received as expected
+          rtUnfilteredChannel.subscribe('end', function (msg) {
+            try {
+              expect(msg.data).to.equal(testData[5].data, 'Unexpected msg data received');
```
mschristensen, line 1275: minor, but probably more robust to do `testData[testData.length-1]` here in case we add new messages

test/realtime/message.test.js
```diff
@@ -1183,5 +1183,136 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
...
+              expect(filteredMessages[1].data).to.equal(testData[2].data, 'Unexpected data received');
+              expect(filteredMessages[0].extras.headers.name).to.equal(
+                testData[0].extras.headers.name,
+                'Unexpected header value received'
+              );
+              expect(filteredMessages[1].extras.headers.name).to.equal(
+                testData[2].extras.headers.name,
+                'Unexpected header value received'
+              );
+              var count = 0;
```
mschristensen, line 1289: we should prefer `let`

mschristensen (changes requested):
Looking good, very minor comments, but once we have the base64 change let's wait for tests to pass against sandbox and then we can merge :)

test/realtime/message.test.js
owenpearson: currently the ably-js tests are designed to run on platforms without ES6 support so we should keep this as `var`

mohyour: I was meant to actually take this out 😄

mschristensen:
Quick note that when base64 encoding the way you do that in the browser vs node may differ - something to watch out for :)

test/realtime/message.test.js
```diff
@@ -1183,5 +1183,135 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
...
+                testData[2].extras.headers.name,
+                'Unexpected header value received'
+              );
+              // Check that message with header that doesn't meet filtering condition is not received.
+              for (msg of filteredMessages) {
+                expect(msg.extras.headers.number).to.equal(26095, 'Unexpected header filtering value received');
+              }
+
+              // Check that we receive expected messages on unfiltered channel
+              expect(unFilteredMessages.length).to.equal(6, 'Expect only 6 unfiltered message to be received');
```
mschristensen, line 1295: possibly worth adding a comment to say that the extra message here is the `end` message

mschristensen (approved):

andydunstall (approved):

mschristensen (approved):

Issue #1184, opened 2023-04-12
Migrate to webpack v5


owenpearson:
See the official [migration guide](https://webpack.js.org/migrate/5/) and [previous attempt](https://github.com/ably/ably-js/pull/839).

This work should land in a new integration branch called `integration/v2`

As part of this work we will need to manually test that the lib still works in react-native and web workers, since these aren't covered by automated tests.

lawrence-forooghian:
Closed in #1201.

PR #1204, opened 2023-04-24
Fix mis-accessing of `util.inspect` method

lawrence-forooghian:
This was pointed out by webpack 5 whilst doing the upgrade for #1184.


owenpearson (approved):
LGTM, thanks

PR #1201, opened 2023-04-20
Upgrade webpack to version 5

lawrence-forooghian:
Resolves #1184. See commit messages for more details.

As described by #1184, this targets the newly-created `integration/v2` branch.

PR #1205, opened 2023-04-24
Update outdated references to ES3

lawrence-forooghian:
Whilst working on the webpack 5 upgrade, it wasn’t clear to me which version of ECMAScript we’re actually targeting. But f2014f7 ("Update to ES5") and #839 ("Now we no longer advertise support for IE8 (or ES3)") suggested that we’ve dropped ES3 support. Owen subsequently [confirmed this](https://github.com/ably/ably-js/pull/1201#issuecomment-1520018754). So let’s update the outdated references to ES3.


owenpearson (approved):
LGTM, thanks

PR #1211, opened 2023-04-25
Fix the flag for running a single test file

lawrence-forooghian:
`--test` clashes with the [flag of the same name](https://nodejs.org/docs/v16.17.0/api/cli.html#--test) introduced in Node 16.17.0, so rename it to `--file`.

I’ve also:

- removed the mention of running multiple test files, because there’s no indication of how you’re actually meant to do this, and from a brief glance at the code it’s not clear whether you can;
- fixed the invocation in the task description (missing necessary `=`)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,9 +40,9 @@ Run the Mocha test suite
 
     npm run test:node
 
-Or run just one or more test files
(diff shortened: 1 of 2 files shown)


owenpearson (approved):
LGTM, thanks!

PR #1215, opened 2023-04-26
Remove support for JSONP

### LINKED ISSUE #1198: Remove support for JSONP in 2.0 branch
Internal queries show that no one is using JSONP and due to this, we have decided to remove it from the library in the next major version update.

lawrence-forooghian:
Resolves #1198.

Note that in both the `web` and `nodejs` platforms, `Defaults.baseTransportOrder` and `Defaults.transportPreferenceOrder` now have identical contents. I considered combining them into a single property, but I couldn’t think of a way of expressing the two separate concepts that I think these properties are trying to express — that is, "transport most likely to be supported" and "transport we’d most like to use".  So I’ve kept them separate.


owenpearson (approved):
Looks good! Honestly, I would just supersede `baseTransportOrder` and `transportPreferenceOrder` with a single `transportOrder` but it's not significant enough to block this PR anyway

Peter-Maguire (approved):

Issue #1197, opened 2023-04-20
Remove soft-deprecated APIs in 2.0 branch


mikelee638:
Remove currently soft-deprecated features from the library. Examples include but are not necessarily limited to:

- #365
- Soft-deprecated option(s) in `authOptions`
- Soft-deprecated `ClientOptions` from `common/types/ClientOptions.ts`

lawrence-forooghian:
Closed in #1221, #1222, #1227.

PR #1221, opened 2023-04-27
Remove the `DeprecatedClientOptions` type

lawrence-forooghian:
This is part of #1197 (removing deprecated APIs for v2). See commit messages for more details. Would be good to hear opinions on the approach of the final commit and whether it's safe.

### DIFF
--- a/promises.js
+++ b/promises.js
@@ -3,7 +3,11 @@ function promisifyOptions(options) {
   if (typeof options == 'string') {
     options = options.indexOf(':') == -1 ? { token: options } : { key: options };
   }
-  options.promises = true;
+  if (options.internal === undefined) {
+    options.internal = { promises: true }
+  } else {
+    options.internal.promises = true
+  }
   return options;
 }
 
--- a/src/common/lib/client/channel.ts
+++ b/src/common/lib/client/channel.ts
@@ -94,7 +94,7 @@ class Channel extends EventEmitter {
       envelope = this.rest.http.supportsLinkHeaders ? undefined : format,
       headers = Utils.defaultGetHeaders(rest.options, format);
 
-    if (rest.options.headers) Utils.mixin(headers, rest.options.headers);
+    Utils.mixin(headers, rest.options.headers);
 
     const options = this.channelOptions;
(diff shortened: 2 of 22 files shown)


owenpearson (approved):
Changes here look good to me 👍

It should be possible to remove the two `InternalClientOptions` altogether:

For promises you would need to create a new test constructors (`helper.AblyRestPromise` and `helper.AblyRealtimePromise`) although there's not much point since I'm removing the callback API at the moment anyway.

For maxMessageSize you can add a spy to `transport.onProtocolMessage` which checks if the `ProtocolMessage.action == 4` (connected) and then mutates the `connectionDetails.maxMessageSize`. To get a reference to the transport during the test you can listen for the `client.connection.connectionManager.on('transport.pending')` event. That being said, I'm happy to leave this for another PR or even to just leave it entirely.

lawrence-forooghian:
> For promises you would need to create a new test constructors (`helper.AblyRestPromise` and `helper.AblyRealtimePromise`) although there's not much point since I'm removing the callback API at the moment anyway.

I'm not sure what you mean by this — the reason I've had to keep the `internal.promises` option is that the [Rest.Promise](https://github.com/ably/ably-js/blob/6778ff0227c4b7759b13c2218cfa62420f02ec04/src/common/lib/client/rest.ts#L231-L235) and [Realtime.Promise](https://github.com/ably/ably-js/blob/6778ff0227c4b7759b13c2218cfa62420f02ec04/src/common/lib/client/realtime.ts#L39-L43) functions need to pass it to the constructors of `Rest` and `Realtime` respectively. Do you think there's a different way of handling this? (I ask more out of curiosity than necessity — I think it's best we keep the `promises` property for now and you remove it as part of removing callbacks.)

> For maxMessageSize you can add a spy to `transport.onProtocolMessage` (…)

I had to keep it there for a [`Rest` test](https://github.com/ably/ably-js/blob/6778ff0227c4b7759b13c2218cfa62420f02ec04/test/rest/message.test.js#L161-L166), where that option doesn't exist. The alternative would be to change the 64-character string used in that test to one of length 65536 (`Defaults.maxMessageSize`). Do you have a preference?

PR #1228, opened 2023-04-27
Remove mentions of `browser/static` directory

lawrence-forooghian:
I think we stopped generating this directory as part of PR #982.


owenpearson (approved):
Yep you're right, `browser/static` was removed a while ago 👍

PR #1237, opened 2023-05-03
Rename `IPlatform` to `IPlatformConfig`

lawrence-forooghian:
Whilst trying to understand how the platform configuration mechanism works I noticed that we have multiple things called `Platform` and found it rather confusing.

The new naming matches how these objects are being used in the platforms’ `index.ts` files — that is, imported as `Config` and used as the value of `Platform.Config`.


owenpearson (approved):
LGTM 👍 the naming weirdness is just an artefact of when we used to use webpack aliases for most of the platform fields which was changed in https://github.com/ably/ably-js/pull/982

PR #1238, opened 2023-05-04
Remove mention of `Platform` from `README`

lawrence-forooghian:
It’s not part of the public API of the SDK (e.g. doesn’t appear in `ably.d.ts`). That does now lead to a question — now that it’s been accidentally documented, do we need to remember to maintain the `Ably.Realtime.Platform.Crypto` property or are we at liberty to rename / change as we see fit without considering it a breaking change?

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,7 +290,7 @@ When a 128 bit or 256 bit key is provided to the library, the `data` attributes 
 ```javascript
 // Generate a random 256-bit key for demonstration purposes (in
 // practice you need to create one and distribute it to clients yourselves)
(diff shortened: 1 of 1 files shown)


owenpearson (approved):
> do we need to remember to maintain the `Ably.Realtime.Platform.Crypto` property or are we at liberty to rename / change as we see fit without considering it a breaking change?

Good question. I'd say we can still change it - anyone who came across this snippet would likely have noticed that something's awry since `Realtime.Platform` isn't present in the type declarations.

PR #1242, opened 2023-05-04
Rename `IBufferUtils.toBuffer` to `toUint8Array`

lawrence-forooghian:
And change its signature to just return a `Uint8Array` ([`Buffer` inherits from `Uint8Array` since Node 3.0.0]( https://nodejs.org/api/buffer.html#buffers-and-typedarrays)).


lawrence-forooghian:
Putting this one back to draft.

PR #1243, opened 2023-05-04
Remove unnecessary property setting in platforms’ `index.ts` files

lawrence-forooghian:
This removes the setting of various properties on `Rest` and `Realtime` in the platforms’ `index.ts` files. See commit messages for more details.


owenpearson (approved):
LGTM, thanks

PR #1245, opened 2023-05-04
Fix mis-setting of `{Rest, Realtime}.Crypto` in `web-noencryption`

lawrence-forooghian:
This is unintentionally referring to the [browser’s Crypto interface](https://w3c.github.io/webcrypto/#crypto-interface). Looks like a copy and paste error from the other platform files, in which there is a local import named `Crypto`.

### DIFF
--- a/src/platform/web-noencryption/index.ts
+++ b/src/platform/web-noencryption/index.ts
@@ -26,8 +26,8 @@ Platform.Config = Config;
 Platform.Transports = Transports;
 Platform.WebStorage = WebStorage;
 
-Rest.Crypto = Crypto;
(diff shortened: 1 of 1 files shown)


owenpearson (approved):
Interesting 😄  I think maybe it would be more appropriate to remove these two lines but same difference ig.

lawrence-forooghian:
> I think maybe it would be more appropriate to remove these two lines but same difference ig.

Fair point — I did `null` just for consistency with the `Platform.Crypto = null` line above. I guess maybe it's good to be explicit 🤷‍♂️

PR #1246, opened 2023-05-04
Convert `crypto.js` files to TypeScript

lawrence-forooghian:


src/platform/nodejs/lib/util/crypto.ts
```diff
@@ -0,0 +1,279 @@
...
+    }
+  }
+
+  function isInstCipherParams(
+    params: API.Types.CipherParams | API.Types.CipherParamOptions
+  ): params is API.Types.CipherParams {
+    /* In node, can't use instanceof CipherParams due to the vm context problem (see
+     * https://github.com/nwjs/nw.js/wiki/Differences-of-JavaScript-contexts).
+     * So just test for presence of all necessary attributes */
+    return params.algorithm && params.key && params.keyLength && params.mode ? true : false;
```
owenpearson, line 122: very much a personal preference but i prefer:
```suggestion
    return !!(params.algorithm && params.key && params.keyLength && params.mode);
```

lawrence-forooghian: Have changed it 👍

PR #1251, opened 2023-05-10
Tell Prettier to ignore `ably-common` submodule

lawrence-forooghian:
This submodule contains some JSON files that don’t conform to Prettier style. We want `format:check` to not fail because of them, and `format` to not modify them. (I guess that the `format:check` invocation in the `check` GitHub workflow has been succeeding because that workflow isn’t configured to check out submodules.)

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+test/common/ably-common/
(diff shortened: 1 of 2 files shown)


owenpearson (approved):
I didn't realise you could have more than one `--ignore-path`, nice 🙂

PR #1258, opened 2023-05-11
Upgrade TypeScript to 4.9.5

lawrence-forooghian:
Ran `npm install typescript@4.9.5 typedoc@latest`. Want to use [the `satsifies` operator introduced in 4.9](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator).

Introduced one compilation error, which I’ve fixed here.

I’m targeting the v2 integration branch (which uses the latest webpack) since trying to upgrade TypeScript on `main` is giving some webpack errors that don’t seem worth trying to understand.


owenpearson (approved):
LGTM 👍

PR #1261, opened 2023-05-12
Feature/subscription filters

mschristensen:
We are creating a new method called getDerived so we can create derived/synthetic channels. Presently, we only have filter as an option. The filter option lets us implement filtered subscription on a channel.
DR can be found [here](https://ably.atlassian.net/wiki/spaces/product/pages/2560098305/DR2+Filtered+Subscription+SDK+Implementation+Decision)
Ticket: [DSC-54](https://ably.atlassian.net/browse/DSC-54)

Recreating this PR originally merged in https://github.com/ably/ably-js/pull/1185 and reverted in https://github.com/ably/ably-js/pull/1259 to apply these changes on an integration branch.

[DSC-54]: https://ably.atlassian.net/browse/DSC-54?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/realtime/message.test.js
+++ b/test/realtime/message.test.js
@@ -1183,5 +1183,145 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
           });
       });
     }
+
+    it('subscribes to filtered channel', function (done) {
+      var testData = [
+        {
+          name: 'filtered',
+          data: 'These headers match the JMESPath expression so this message should not be filtered',
+          extras: {
+            headers: {
+              name: 'Lorem Choo',
+              number: 26095,
+              bool: true,
+            },
+          },
+        },
+        {
+          name: 'filtered',
+          data: 'Random data with no extras for filter',
+        },
+        {
+          name: 'filtered',
+          data: 'This message should also not be filtered',
+          extras: {
+            headers: {
+              name: 'John Bull',
+              number: 26095,
+              bool: false,
+            },
+          },
+        },
+        {
+          name: 'filtered',
+          data: 'No header on message',
+        },
+        {
+          name: 'filtered',
+          data: 'Can be filtered because it does not meet filter condition on headers.number',
+          extras: {
+            headers: {
+              name: 'John Bull',
+              number: 12345,
+              bool: false,
+            },
+          },
+        },
+        {
+          name: 'end',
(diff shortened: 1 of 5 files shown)


test/realtime/message.test.js
```diff
@@ -1183,5 +1183,145 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
...
+                }
+              } catch (err) {
+                closeAndFinish(done, realtime, err);
+                return;
+              }
+              closeAndFinish(done, realtime);
+            });
+          });
+          var restChannel = rest.channels.get('chan');
+          restChannel.publish(testData);
```
mschristensen, line 1319: @mohyour it looks like there still might be a race here, we are publishing when we are connected but we are not guaranteed to publish after we have attached to the filtered & unfiltered channel. If messages are published before this attachment succeeds, we will lose messages. Instead, we should wait for the attachments before publishing.

test/realtime/message.test.js
```diff
@@ -1183,5 +1183,145 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
...
+                  expect(unFilteredMessages[i].data).to.equal(testData[i].data, 'Unexpected data received');
+                }
+              } catch (err) {
+                closeAndFinish(done, realtime, err);
+                return;
+              }
+              closeAndFinish(done, realtime);
+            });
+            var restChannel = rest.channels.get('chan');
+            restChannel.publish(testData);
```
mschristensen, line 1318: Hmm, almost there. While this fixes the issue for the unfiltered channel attachment, we still have a race condition where the test might fail if the filtered channel has not yet attached before we start publishing messages. We need to ensure both attachments have succeeded before we publish :)

owenpearson (commented):
LGTM, thanks!

owenpearson (approved):
LGTM, thanks!

owenpearson (approved):
LGTM, thanks!

owenpearson (approved):
LGTM, thanks!

owenpearson (approved):
LGTM, thanks!

PR #1263, opened 2023-05-15
fix(WebSocketTransport): use encodeURI for react-native

### LINKED ISSUE #1262: Unable to set as default origin header
Hi!

I'm trying to connect ably to react native and encountered the problem that after a successful connection I get an error - Unable to set as default origin header. At the same time, all requests are successful, connection status changes to connected.

Can you please tell me how to fix this?

```
const client = new Realtime({
          clientId: name,
          log: { level: 2 },
          useTokenAuth: true,
          token: ablyToken.data,
          authCallback: async (_params, callback) => {
           ....
          },
        });
```

Screenshot_1684136306

owenpearson:
Resolves #1262

Based on https://stackoverflow.com/questions/47486413/passing-parameter-in-websocket-uri-on-react-native-is-not-working

### DIFF
--- a/src/common/lib/transport/websockettransport.ts
+++ b/src/common/lib/transport/websockettransport.ts
@@ -37,7 +37,7 @@ class WebSocketTransport extends Transport {
     if (connectParams) {
       for (const key in connectParams) uri += (paramCount++ ? '&' : '?') + key + '=' + connectParams[key];
     }
-    this.uri = uri;
+    this.uri = Platform.Config.encodeWsUri ? Platform.Config.encodeWsUri(uri) : uri;
     return new Platform.Config.WebSocket(uri);
   }
 
--- a/src/common/types/IPlatformConfig.d.ts
+++ b/src/common/types/IPlatformConfig.d.ts
@@ -50,4 +50,5 @@ export interface IPlatformConfig {
     callback: (err: Error, result: boolean | CryptoJS.lib.WordArray) => void
   ) => void;
   isWebworker?: boolean;
+  encodeWsUri?: (uri: string) => string;
 }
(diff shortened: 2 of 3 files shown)


lawrence-forooghian (commented):
I'm not convinced by this one — the Stack Overflow post you've referenced doesn't explain why this would fix the issue. It would be good to understand whether the issue is that we are indeed trying to instantiate a `WebSocket` with a URI that is actually invalid (in which case we should be fixing this for all platforms) or if we're working around some sort of bug in React Native. I do think we probably need to go back to the user to get more information to help us reproduce.

sacOO7:
> I'm not convinced by this one — the Stack Overflow post you've referenced doesn't explain why this would fix the issue. It would be good to understand whether the issue is that we are indeed trying to instantiate a `WebSocket` with a URI that is actually invalid (in which case we should be fixing this for all platforms) or if we're working around some sort of bug in React Native. I do think we probably need to go back to the user to get more information to help us reproduce.

Adding to this, if we have a failing test without `encodeURI`, that should prove the solution.

owenpearson:
@lawrence-forooghian the `WebSocket` constructor should accept any valid URI without it needing to be url encoded, the issue described in that stackoverflow thread is that, on some version of react native websockets, the constructor isn't doing the url encoding in java. There isn't anything we need to fix inside the library because there's nothing wrong with passing an un-encoded url into the WebSocket constructor usually - this is just a patch to make ably-js compatible with whatever versions of react-native have this issue.

@sacOO7 that's true but we don't have automated tests for react-native

owenpearson:
It was actually raised on the react-native repo [here](https://github.com/facebook/react-native/issues/16041) but it was asked like a question instead of a bug report so they just closed the issue and directed the reporter to stack overflow 🙃

PR #1273, opened 2023-05-17
Fix incorrect accessing of web’s `getRandomWordArray` and `getRandomValues`

### LINKED ISSUE #1272: Web crypto code is always using non-cryptographically secure random data generator
The [code that picks the random data generator in web’s `crypto.js`](https://github.com/ably/ably-js/blob/d9573c2e5423f0ae15bd8d20a60d91dfe57f27a4/src/platform/web/lib/util/crypto.js#L21-L54) is trying to access properties on `Platform` that don't exist — they exist only on `Platform.Config`. This means that it always ends up falling back to using the non-cryptographically secure `Math.random`.

lawrence-forooghian:
We were trying to access properties on `Platform` that actually exist on `Platform.Config`. This means that we always ended up falling back to using the non-cryptographically secure `Math.random`.

Resolves #1272.

### DIFF
--- a/src/platform/nativescript/index.ts
+++ b/src/platform/nativescript/index.ts
@@ -6,7 +6,7 @@ import Platform from '../../common/platform';
 // Platform Specific
 import BufferUtils from '../web/lib/util/bufferutils';
 // @ts-ignore
-import Crypto from '../web/lib/util/crypto';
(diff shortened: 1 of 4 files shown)


owenpearson (approved):
good spot!

PR #1279, opened 2023-05-18
Fix Node’s `Crypto.getDefaultParams` with `ArrayBuffer` key

### LINKED ISSUE #1278: Node’s `Crypto.getDefaultParams` does not accept an `ArrayBuffer` key
`getDefaultParams` is meant to take an argument of type `ably.d.ts`’s `Types.CipherParamOptions`, whose `key` property is of type `ArrayBuffer | Uint8Array | string`.

However, the following test case, which attempts to call `getDefaultParams` with a key of type `ArrayBuffer`, is failing:

```js
it('getDefaultParams_ArrayBuffer_key', function (done) {
  Crypto.generateRandomKey(function (err, key) {
    if (err) {
      done(err);
    }
    var arrayBufferKey = Ably.Realtime.Platform.BufferUtils.toArrayBuffer(key);
    var params = Crypto.getDefaultParams({ key: arrayBufferKey });
    try {
      expect(BufferUtils.bufferCompare(params.key, key)).to.equal(0);
      done();
    } catch (err) {
      done(err);
    }
  });
});
```

It fails because the key length check in `validateCipherParams` is failing. But I believe that even beyond that, attempting to use the resulting `CipherParams` to encrypt or decrypt a message would fail on Node < 15, since support for passing Node’s `crypto.createCipheriv` a key of type `ArrayBuffer` was only added in Node 15.0.0.

lawrence-forooghian:
`getDefaultParams` is meant to take an argument of type `ably.d.ts`’s `Types.CipherParamOptions`, whose `key` property is of type `ArrayBuffer | Uint8Array | string`.

However, the test case that I’ve added here, which attempts to call `getDefaultParams` with a key of type `ArrayBuffer`, was failing due to the key length check in `validateCipherParams` failing.

But I believe then even beyond that, attempting to use the resulting `CipherParams` to encrypt or decrypt a message would fail on Node < 15, since support for passing Node’s `crypto.createCipheriv` a key of type `ArrayBuffer` was only added in Node 15.0.0.

Resolves #1278.

### DIFF
--- a/src/common/types/IBufferUtils.ts
+++ b/src/common/types/IBufferUtils.ts
@@ -4,6 +4,7 @@ export default interface IBufferUtils<Bufferlike, Output, ToBufferOutput, Compar
   base64CharSet: string;
   hexCharSet: string;
   isBuffer: (buffer: unknown) => buffer is Bufferlike;
+  isArrayBuffer: (buffer: unknown) => buffer is ArrayBuffer;
   // On browser this returns a Uint8Array, on node a Buffer
   toBuffer: (buffer: Bufferlike) => ToBufferOutput;
   toArrayBuffer: (buffer: Bufferlike) => ArrayBuffer;
--- a/src/platform/nodejs/lib/util/bufferutils.ts
+++ b/src/platform/nodejs/lib/util/bufferutils.ts
@@ -36,7 +36,7 @@ class BufferUtils implements IBufferUtils<Bufferlike, Output, ToBufferOutput, Co
     return this.toBuffer(buffer).toString('hex');
   }
 
-  isArrayBuffer(ob: unknown) {
+  isArrayBuffer(ob: unknown): ob is ArrayBuffer {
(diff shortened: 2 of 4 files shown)


owenpearson (commented):
Seems like a possibly-breaking change here - would it make sense to target the integration branch instead?

lawrence-forooghian:
Hey @owenpearson, would you mind explaining what you see as a breaking change here? The intention is just to work correctly when passed a certain type of key. The only public-facing change, I think, is that `Crypto.getDefaultParams` might return a copy of the passed key instead of maintaining a reference to it — do you consider that a breaking change?

owenpearson (approved):
Looks fine - I'll leave it to you to make the call on the above conversation

lawrence-forooghian:
Fair point — people aren't meant to be relying on the type of `CipherParams.key`, but to be on the safe side I've changed it to only perform a conversion in the case where the passed key is an `ArrayBuffer`.

PR #1280, opened 2023-05-18
Fix Node encryption of `ArrayBuffer` plaintext

### LINKED ISSUE #1281: Node implementation of `Crypto` is unable to encrypt plaintext of type `ArrayBuffer`
Trying to encrypt a message whose `data` is of type `ArrayBuffer` gives `TypeError: Cannot read properties of undefined (reading 'length')`.

As I understand it, the library is meant to be able to handle plaintext of type `NodeBufferlike = Buffer | ArrayBuffer | TypedArray`.

lawrence-forooghian:
This fixes an error that is thrown in Node when attempting to encrypt a message whose `data` property is of type `ArrayBuffer`. See commit messages for more details.

Resolves #1281.


owenpearson (approved):
LGTM

PR #1284, opened 2023-05-22
Added new action to push JS SDK into new CDN bucket

Peter-Kubik:
We are moving CDN S3 bucket into a new account, and with a new name (cdn.ably.com). To enable the migration, we need this GHA to temporarily push builds into both the old and new buckets. I have set up a new AWS Role in the SDK account that should be used to publish builds to the new CDN bucket


owenpearson (approved):
Looks good. Have you already copied the contents of cdn.ably.io to the new bucket?

Peter-Kubik:
@owenpearson it looks like I am not authorized to merge this PR. Could you please press the button for me, when you have time?

Issue #1295, opened 2023-05-24
Stop using CryptoJS for signing token requests


lawrence-forooghian:
Split from #1239.

Replace the web token signing code’s [use of `crypto-js`](https://github.com/ably/ably-js/blob/bea306561b7e9681f2445fcfa68462f33be8f2aa/src/common/lib/client/auth.ts#L59) with the [Web Crypto API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API) (or something else that doesn't increase bundle size much, if we decide we want to access HMAC in non-secure contexts).

lawrence-forooghian:
Closed by #1325.

Issue #1239, opened 2023-05-04
Remove "`WordArray' stuff" in next major release


lawrence-forooghian:
As suggestion by @SimonWoolf [here](https://github.com/ably/ably-js/issues/1210#issuecomment-1522395398) as a way of reducing bundle size. I am not yet sufficiently familiar with the library to know exactly what this work entails, but will look into it.

lawrence-forooghian:
Putting this on hold whilst I work on #1252, which I've split from this task.

lawrence-forooghian:
Closed by #1333.

PR #1299, opened 2023-05-24
Use Web Crypto API for encrypting and decrypting

### LINKED ISSUE #1292: Use Web Crypto API for encrypting and decrypting
Split from #1239.

Replace the web crypto code’s current use of `crypto-js` with the [Web Crypto API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API). Depends on #1293.

### LINKED ISSUE #1296: Choose and update TS definitions for the type of key returned by web’s `Crypto.generateRandomKey`
Split from #1292.

On web, this method currently returns a `WordArray` — a type that will no longer exist after #1292. We need to change it to something else. 

Given my understanding of the intended use case of this method —  to allow a user to generate a key which they can then share with other parties through some side channel in order to establish encrypted communication with those parties — it is important that whatever type we return here be something that a user of the library will know how work with, and especially that it be something that are able to easily serialise for transmission to other parties. With this in mind, I think that the obvious candidate would be an `ArrayBuffer` (and to continue returning `Buffer` for Node, as is idiomatic for that platform).

I think that this would also be a good moment to revisit the return type described in the TypeScript declarations file, too — it currently returns a [`CipherKey`](https://github.com/ably/ably-js/blob/bea306561b7e9681f2445fcfa68462f33be8f2aa/ably.d.ts#L2928-L2929), which is an alias to `unknown` because "user should not be interacting with it - output of getDefaultParams should be used opaquely". Whilst it is true that the output of `getDefaultParams` should be used opaquely, I do not think that this is true of the output of `generateRandomKey` — as described above, the user needs to know how to interact with this value so that they can serialise it.

What we'd ideally like to communicate in the type definitions is that it returns `ArrayBuffer` on web, and `Buffer` on Node, without requiring the user to perform any type assertions when trying to use the key. I do not know if this is possible — @owenpearson, might you have any thoughts on what we could do here?

### LINKED ISSUE #1300: Change web `BufferUtils` to not make use of `crypto-js` encoding / decoding
Split from #1292. <del>This also includes removing any functionality that isn't needed by the library (or moving it to the test suite if just used by tests).</del> (I've split this latter part to #1314)

lawrence-forooghian:
**Note: This is based on top of #1311; please review that one first.**

This PR changes the web crypto code to use the [Web Crypto API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API) instead of the CryptoJS library. See commit messages for more details.

Resolves #1292, resolves #1296, resolves #1300.

### DIFF
--- a/src/common/types/IBufferUtils.ts
+++ b/src/common/types/IBufferUtils.ts
@@ -4,7 +4,6 @@ export default interface IBufferUtils<Bufferlike, Output, ToBufferOutput, WordAr
   base64CharSet: string;
   hexCharSet: string;
   isBuffer: (buffer: unknown) => buffer is Bufferlike;
-  isArrayBuffer: (buffer: unknown) => buffer is ArrayBuffer;
   isWordArray: (val: unknown) => val is WordArrayLike;
   // On browser this returns a Uint8Array, on node a Buffer
   toBuffer: (buffer: Bufferlike) => ToBufferOutput;
--- a/src/common/types/IBufferUtils.ts
+++ b/src/common/types/IBufferUtils.ts
@@ -5,7 +5,7 @@ export default interface IBufferUtils<Bufferlike, Output, ToBufferOutput, WordAr
   isWordArray: (val: unknown) => val is WordArrayLike;
   // On browser this returns a Uint8Array, on node a Buffer
   toBuffer: (buffer: Bufferlike) => ToBufferOutput;
-  toArrayBuffer: (buffer: Bufferlike) => ArrayBuffer;
+  toArrayBuffer: (buffer: Bufferlike | WordArrayLike) => ArrayBuffer;
   base64Encode: (buffer: Bufferlike) => string;
   base64Decode: (string: string) => Output;
   hexEncode: (buffer: Bufferlike) => string;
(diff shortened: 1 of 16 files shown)


src/common/types/IBufferUtils.ts
```diff
@@ -12,7 +12,7 @@ export default interface IBufferUtils<Bufferlike, Output, ToBufferOutput, WordAr
   hexDecode: (string: string) => Output;
   utf8Encode: (string: string) => Output;
   utf8Decode: (buffer: Bufferlike) => string;
-  bufferCompare: (buffer1: Bufferlike, buffer2: Bufferlike) => number;
+  areBuffersEqual: (buffer1: Bufferlike, buffer2: Bufferlike) => boolean;
```
owenpearson, line 15: don't really mind this change but worth mentioning - compare functions are a thing in javascript because they're used as [a parameter for Array.sort](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#parameters)

owenpearson (changes requested):
the crypto op implementations look good to me 👍 

at this point is it worth getting rid of any references to the crypto-js api eg `WordArrayLike`?

src/common/types/IBufferUtils.ts
lawrence-forooghian: Ah, I see — I had suspected it might be something like that, but given we're not using the compare functionality anywhere I reckon the change is OK?

lawrence-forooghian:
> at this point is it worth getting rid of any references to the crypto-js api eg WordArrayLike?

We've still got some CryptoJS usage, that will be removed in #1295 and #1239

src/common/types/IBufferUtils.ts
owenpearson: yep, that's fine by me 👍

owenpearson (approved):

PR #1306, opened 2023-05-26
Integration/subscription filters

mschristensen:
Adds a new **experimental** `getDerived` method for obtaining an instance of a "derived
channel" which applies a filter to messages in the upstream channel.

The channel filter is specified via a base64 encoded JMESPath expression
which is given as the value of a new channel qualifier called "filter".


owenpearson (approved):
LGTM, thanks

PR #1311, opened 2023-05-30
Make `ICipher.decrypt` async

### LINKED ISSUE #1293: Make `ICipher.decrypt` asynchronous
The [`SubtleCrypto.decrypt`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/decrypt) method that we wish to start using in #1292 is an asynchronous-only API — it returns a `Promise`.

Our current internal cipher API (as described by the `ICipher` interface added in #1246) exposes a synchronous `decrypt` method. In light of the above, we will need to change this to an asynchronous method.

This will necessitate internal changes to the library but will also affect its public API, since the `{Message, PresenceMessage}.{fromEncoded, fromEncodedArray}` static methods, which are currently synchronous, may need to decrypt data and hence in light of the above may now need to perform an asynchronous operation.

We need to decide exactly how we'd like to change the public API — we will definitely need to have an asynchronous version of the aforementioned methods, but @owenpearson suggested that we might wish to also maintain synchronous variants (which, presumably, would fail in the case that the method needs to perform a decryption).

I've created #1302, in order to investigate the effort involved in the internal changes this change would require, so that we can then put an estimate on this issue.

lawrence-forooghian:
**Note: This is based on top of #1246; please review that one first.**

This makes `ICipher.decrypt` async, in preparation for using the (async-only) `SubtleCrypto.decrypt` method in #1292.

The changes of note are:

- making `RealtimeChannel#processMessage` async — involves adding a queue to synchronise access to this method
- making the public `{Message, PresenceMessage}#{fromEncoded, fromEncodedArray}` methods async

Resolves #1293.

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -2841,14 +2841,14 @@ declare namespace Types {
      * @param JsonObject - The deserialized `PresenceMessage`-like object to decode and decrypt.
      * @param channelOptions - A {@link ChannelOptions} object containing the cipher.
      */
-    static fromEncoded: (JsonObject: any, channelOptions?: ChannelOptions) => PresenceMessage;
+    static fromEncoded: (JsonObject: any, channelOptions?: ChannelOptions) => Promise<PresenceMessage>;
     /**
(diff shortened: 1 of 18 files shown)


owenpearson (approved):
LGTM, nicely done :)

PR #1316, opened 2023-06-02
INT-2325 upload to correct cdn bucket

Peter-Kubik:
In previous PR, I have accidentally used incorrect bucket name. This PR is fixing that. I have verified that the prod-cdn.ably.com S3 bucket does exist in the correct account.

### DIFF
--- a/.github/workflows/publish-cdn.yml
+++ b/.github/workflows/publish-cdn.yml
@@ -48,4 +48,4 @@ jobs:
         with:
           node-version: 14.x
       - run: npm ci
-      - run: node scripts/cdn_deploy.js --bucket cdn.ably.com --skipCheckout --tag=${{ github.event.inputs.version }}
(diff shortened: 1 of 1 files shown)


owenpearson (approved):
LGTM, thanks. I've also opened https://github.com/ably/ably-js/issues/1319 the gh workflow didn't fail even though the s3 upload failed.

Peter-Kubik:
Thanks for the review @owenpearson . Would you mind merging it for me? Also, could you then please run the cdn action when you have time, to see if it gets uploaded successfully? Thanks

owenpearson:
@Peter-Kubik I'm not entirely comfortable with running the action again on a tag we've already published to the CDN from, would likely be fine but it's still slightly risky. If you want me to run the CDN action against an existing release tag I'd rather we split this into two workflow files so I can upload to the new bucket without touching the existing one (we would also need to update the release process to make it clear that both workflows should be run on new release tags)

owenpearson:
as discussed with @Peter-Kubik there's no rush on this so we'll just wait for the next release

owenpearson (approved):

PR #1322, opened 2023-06-06
Make `ErrorInfo.fromValues`’s signature correspond to its expectations

lawrence-forooghian:
It expects the received object to have certain keys of certain types, and throws an error if not. So, get the compiler’s help in enforcing this.

### DIFF
--- a/src/platform/nodejs/lib/util/crypto.ts
+++ b/src/platform/nodejs/lib/util/crypto.ts
@@ -198,7 +198,10 @@ var CryptoFactory = function (bufferUtils: typeof BufferUtils) {
 
       generateRandom((keyLength || DEFAULT_KEYLENGTH) / 8, function (err, buf) {
         if (callback !== undefined) {
-          callback(err ? ErrorInfo.fromValues(err) : null, buf);
+          const errorInfo = err
+            ? new ErrorInfo('Failed to generate random key: ' + err.message, 500, 50000, err)
+            : null;
+          callback(errorInfo, buf);
         }
       });
     }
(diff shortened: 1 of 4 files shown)


src/platform/nodejs/lib/util/crypto.ts
```diff
@@ -198,7 +198,10 @@ var CryptoFactory = function (bufferUtils: typeof BufferUtils) {
 
       generateRandom((keyLength || DEFAULT_KEYLENGTH) / 8, function (err, buf) {
         if (callback !== undefined) {
-          callback(err ? ErrorInfo.fromValues(err) : null, buf);
+          const errorInfo = err
+            ? new ErrorInfo('Failed to generate random key: ' + err.message, 500, 50000, err)
```
lawrence-forooghian, line 202: I had absolutely no idea what would be an appropriate code and status code to use here — any thoughts @owenpearson?

sacOO7: You can give status code used for client error

sacOO7: HttpStatusCode.BadRequest

lawrence-forooghian: Hey @sacOO7, thanks for that, I've changed the `code` to 50000 to match what you linked to. I'm a bit confused about why "bad request" is the right thing to use, though, but if that's what we're doing then I guess I won't question it  🤷‍♂️

owenpearson (approved):

PR #1325, opened 2023-06-07
Stop using CryptoJS for signing token requests

### LINKED ISSUE #1295: Stop using CryptoJS for signing token requests

lawrence-forooghian:
**Note: This is based on top of #1320; please review that one first.**

This introduces a lightweight implementation of HMAC-SHA-256 and uses that for signing token requests instead of CryptoJS. See commit messages for more details.

Resolves #1295.

### DIFF
--- a/src/common/lib/client/auth.ts
+++ b/src/common/lib/client/auth.ts
@@ -130,11 +130,6 @@ class Auth {
 
     if (useTokenAuth(options)) {
       /* Token auth */
-      if (options.key && !hmac) {
-        const msg = 'client-side token request signing not supported';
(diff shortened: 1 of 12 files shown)


owenpearson (approved):
Looks good!

PR #1333, opened 2023-06-12
Remove CryptoJS library

### LINKED ISSUE #1239: Remove "`WordArray' stuff" in next major release

lawrence-forooghian:
We’re no longer making use of this library (in order to reduce bundle size, given that modern browsers provide all of the necessary crypto functionality).

Resolves #1239.


owenpearson (approved):
LGTM, nice

PR #1344, opened 2023-06-21
Refactor incremental backoff

### LINKED ISSUE #1365: Refactor incremental backoff tests
- [Old implementation for incremental backoff](https://github.com/ably/ably-js/pull/1008) doesn't have proper test assertions in place
- Aim is to add meaningful test assertions and improve code readability.

sacOO7:
- Need to be refactored based on https://github.com/ably/ably-dotnet/pull/1176
- Fixed #1365


.nvmrc
```diff
@@ -0,0 +1 @@
+v16.20.1
```
owenpearson, line 1: I would rather not add this tbh, you can see the versions of node we support building the library on in the gh workflows and adding config to enforce a specific patch version would just be annoying

CONTRIBUTING.md
```diff
@@ -43,6 +43,10 @@ Or run just one test file
 
     npm run test:node -- --file=test/realtime/auth.test.js
 
+Or run just one test
```
owenpearson, line 46: I'm happy to add this but it's nothing to do with incremental backoff so should be in a separate PR

CONTRIBUTING.md
```diff
@@ -57,6 +61,10 @@ The included vscode launch config allows you to launch and attach the debugger i
 file you want to run and start debugging. Note that breakpoint setting for realtime code will be within the
 browser/static directory, not the raw source files, and breakpoints in files under test should work directly.
 
+VSCode allows you to directly launch `package.json` scripts in debug mode.
```
owenpearson, line 64: This is generic NPM project advice (and specific to one text editor) so I don't think it belongs in the ably-js CONTRIBUTING.md

test/realtime/utils.test.js
```diff
@@ -0,0 +1,43 @@
...
+      }
+
+      function calculateBounds(retryAttempt, initialTimeout) {
+        var upperBound = Math.min((retryAttempt + 2) / 3, 2) * initialTimeout;
+        var lowerBound = 0.8 * upperBound;
+        return { lowerBound, upperBound };
+      }
+
+      for (var i = 0; i < retryTimeouts.length; i++) {
+        var { lowerBound, upperBound } = calculateBounds(retryAttempts[i], initialTimeout);
```
owenpearson, line 38: this would cause a syntax error in web browsers which don't support object destructuring

owenpearson (changes requested):

CONTRIBUTING.md
sacOO7: Actually, there is already `VSCode` specific advice right above this. If I had to guess, `vscode` is the main editor used for `js` based development.

test/realtime/utils.test.js
sacOO7: Yeah, I felt it might not work in older versions, but browsers in our current test suite will support it right? 
Since we are not using this code as a part of source code, we should be good.

.nvmrc
sacOO7: Yeah, I felt the same. I will remove this. I think we should add recommended `js` versions in `CONTRIBUTING.md`

CONTRIBUTING.md
sacOO7: Okay, I might create a new PR for this : )

test/realtime/utils.test.js
owenpearson: yeah i know we don't test on older browsers in CI but we still support them so we need to be able to run the test suite on them

CONTRIBUTING.md
owenpearson: i don't disagree with anything you said. but still, we don't need to add documentation for how to run npm scripts with vscode to this repo because it's nothing to do with ably-js.

sacOO7: Okay, got it. I will remove this. I think it should be pretty common for vscode devs to configure editor properly.

test/realtime/utils.test.js
sacOO7: done

CONTRIBUTING.md
sacOO7: moved to #1364

CONTRIBUTING.md
sacOO7: done

test/realtime/failure.test.js
```diff
@@ -185,6 +185,23 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
...
 
+    function checkIfRetryTimeoutsAreCorrect(retryTimeouts, calculationDelay = 0) {
+      function checkIsBetween(value, min, max) {
+        expect(value).to.be.above(min);
+        expect(value).to.be.below(max);
+      }
+
+      // Upper bound = min((retryAttempt + 2) / 3, 2) * initialTimeout
+      // Lower bound = 0.8 * Upper bound
+      checkIsBetween(retryTimeouts[0], 120, 150 + calculationDelay);
```
owenpearson, line 196: I thought the whole point of this change was to make the tests work correctly even if we changed `disconnectedRetryTimeout` but you're still inlining timeout ranges here and elsewhere

test/realtime/failure.test.js
```diff
@@ -185,6 +185,23 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
       }
     });
 
+    function checkIfRetryTimeoutsAreCorrect(retryTimeouts, calculationDelay = 0) {
```
owenpearson, line 188: This wouldn't work on older browsers
```suggestion
    function checkIfRetryTimeoutsAreCorrect(retryTimeouts, calculationDelay) {
      calculationDelay = calculationDelay || 0;
```

owenpearson (changes requested):

test/realtime/failure.test.js
sacOO7: Okay, got it

test/realtime/failure.test.js
sacOO7: Actually, this makes tests easier to read. 
Also, it doesn't break the rule of value assertions. 
If we add formulas in the assertions, it adds the additional complexity of testing those formulas first. 
As far as I know, keeping formulas as assertions in unit tests are avoided.

sacOO7: I have already added a comment denoting the same.

owenpearson: Sorry but I honestly think this is less readable than before. The reason we wanted to change this test at all was because it included a magic number derived form the timeout configuration but this PR just adds even more of the same problem. Also, I don't think it changes the readability much, but extracting the assertions into functions just makes it harder to read the code from top to bottom so I preferred it before. IMO all we needed to do in the first place was just replace `50` with `disconnectedRetryTimeout / 3`

sacOO7: I am not sure, what changes do you want me to make here?

owenpearson: You should just make it so that we don't have inline numbers which make the test depend on a specific disconnectedRetryTimeout. I don't mind whether you refactor your existing changes or revert them and just change `50` to `disconnectedRetryTimeout / 3`

owenpearson (approved):
LGTM 👍

test/realtime/failure.test.js
```diff
@@ -185,9 +185,14 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
...
 
+    function checkIsBetween(value, min, max) {
+      expect(value).to.be.above(min);
+      expect(value).to.be.below(max);
+    }
+
     utils.arrForEach(availableTransports, function (transport) {
       it('disconnected_backoff_' + transport, function (done) {
-        var disconnectedRetryTimeout = 150;
+        var disconnectedRetryTimeout = ;
```
owenpearson, line 195: syntax error 😄

owenpearson (changes requested):

test/realtime/failure.test.js
sacOO7: How could this happen

owenpearson (approved):

PR #1347, opened 2023-06-22
feat: add `ChannelStateChange.hasBacklog` and return state change to attach promise/callback

### LINKED ISSUE #1328: add `ChannelStateChange.hasBacklog`
This flag is already send as part of the attached protocol message, but should be exposed on `ChannelStateChange` objects so that users can tell whether to expect rewind messages.

owenpearson:
Resolves #1328 

See [corresponding spec PR](https://github.com/ably/specification/pull/152)

These features in combination will allow users to check whether to expect a backlog of messages once attached to a realtime channel. In particular, when using `rewind=1`, this will allow users to create a promise which resolves once a channel is attached and, if applicable, the rewind message has been received.

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -2556,9 +2556,9 @@ declare namespace Types {
     /**
      * Attach to this channel ensuring the channel is created in the Ably system and all messages published on the channel are received by any channel listeners registered using {@link RealtimeChannelCallbacks.subscribe | `subscribe()`}. Any resulting channel state change will be emitted to any listeners registered using the {@link EventEmitter.on | `on()`} or {@link EventEmitter.once | `once()`} methods. As a convenience, `attach()` is called implicitly if {@link RealtimeChannelCallbacks.subscribe | `subscribe()`} for the channel is called, or {@link RealtimePresenceCallbacks.enter | `enter()`} or {@link RealtimePresenceCallbacks.subscribe | `subscribe()`} are called on the {@link RealtimePresenceCallbacks} object for this channel.
      *
-     * @param callback - A function which will be called upon completion of the operation. If the operation succeeded, then the function will be called with `null`. If it failed, the function will be called with information about the error.
+     * @param callback - A function which will be called upon completion of the operation. If the operation succeeded and the channel became attached, then the function will be called with a {@link ChannelStateChange} object. If the channel was already attached the function will be called with `null`. If it failed, the function will be called with information about the error.
      */
-    attach(callback?: errorCallback): void;
+    attach(callback?: StandardCallback<ChannelStateChange | null>): void;
     /**
      * Detach from this channel. Any resulting channel state change is emitted to any listeners registered using the {@link EventEmitter.on | `on()`} or {@link EventEmitter.once | `once()`} methods. Once all clients globally have detached from the channel, the channel will be released in the Ably service within two minutes.
      *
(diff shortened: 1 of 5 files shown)


SimonWoolf (commented):
No objection to the idea, seems like a reasonable feature.

There's nothing ably-js specific about this; if we think it's worth doing we should have a spec PR for this so it can eventually be done in all sdks (and added to api docs). 

> which resolves once a channel is attached and, if applicable, the rewind message has been received.

not sure I understand what the last part of that sentence means -- it resolves once the channel is attached full stop, no?

SimonWoolf:
>  my thinking was that this is unlikely to be needed in other SDKs so adding it to the spec would likely be futile Libraries may offer a ClientOptions#agents property, for use only by other Ably-authored SDKs, on a need-to-have basis.

SimonWoolf:
:+1: 

> my thinking was that this is unlikely to be needed in other SDKs so adding it to the spec would likely be futile

Shrug, I think there's still some value in formalising the API in the spec even if we're not imminently intending to implement in other sdks. there's other bits of the spec that are not-mandatory ("Libraries may offer a ClientOptions#agents property...", "Client libraries can optionally save a round-trip request to the Ably service for expired tokens by detecting when a token has expired...", etc)

>  it wouldn't really make sense to add this promise to the library since it only works for rewind=1, any other rewind value has the possibility of a partial backlog

yeah, really for that feature you'd need realtime to tell you when the backlog is finished, and tbh I'm not sure that that'd be a good idea. (also most sensible ways of returning another promise would involve a breaking api change, and would make it a very prominent part of the api, and istm it's not worth either of those)

lawrence-forooghian:
I wasn't aware of the rewind feature, popping a link to the [documentation](https://ably.com/docs/realtime/channels/channel-parameters/rewind#rewind-example-ably) here for future reference.

I think that exposing the `hasBacklog` property (which I can't find any documentation of, e.g. in the [protocol spec](https://sdk.ably.com/builds/ably/specification/main/protocol/) — how did you find out about it?) makes sense, and if I've understood the use case correctly then I think it's a broadly useful thing, and hence worth including in the spec.

As for the change to the signature of `attach`, I wonder whether that needs to come alongside this change. For example, a user calling `attach` has, until now, had no direct way to access the `ChannelStateChange.resumed` property, which you can imagine a user might want to use to know whether they need to fetch channel history. I _guess_ (please correct me if wrong) that our advice to users has been that to find this information out they need to listen for a channel state change using `on`, so would the same advice not apply here too? That said, if we _are_ going to change the signature of `attach` then I think it should be a spec change, yeah.

ably.d.ts
```diff
@@ -2590,32 +2590,44 @@ declare namespace Types {
      *
      * @param event - The event name.
      * @param listener - An event listener function.
-     * @param callbackWhenAttached - A function which will be called upon completion of the channel {@link RealtimeChannelCallbacks.attach | `attach()`} operation. If the operation succeeded, then the function will be called with `null`. If it failed, the function will be called with information about the error.
+     * @param callbackWhenAttached - A function which will be called upon completion of the channel {@link RealtimeChannelCallbacks.attach | `attach()`} operation. If the operation succeeded, then the function will be called with a {@link ChannelStateChange} object. If it failed, the function will be called with information about the error.
```
lawrence-forooghian, line 2597: This needs to explain `null`, too.

ably.d.ts
```diff
@@ -2694,32 +2706,32 @@ declare namespace Types {
      *
      * @param event - The event name.
      * @param listener - An event listener function.
-     * @returns A promise which resolves upon success of the channel {@link RealtimeChannelPromise.attach | `attach()`} operation and rejects with an {@link ErrorInfo} object upon its failure.
+     * @returns A promise which, upon successful attachment to the channel, will be fulfilled with a {@link ChannelStateChange} object. Upon failure, the promise will be rejected with an {@link ErrorInfo} object.
      */
-    subscribe(event: string, listener?: messageCallback<Message>): Promise<void>;
+    subscribe(event: string, listener?: messageCallback<Message>): Promise<ChannelStateChange>;
```
lawrence-forooghian, line 2715: Don't all these promise `subscribe` calls need to have `null` in their return type too?

src/common/lib/client/realtimechannel.ts
```diff
@@ -296,14 +299,18 @@ class RealtimeChannel extends Channel {
        * current mode differs from requested mode */
       this._requestedFlags = _flags as API.Types.ChannelMode[];
     } else if (this.state === 'attached') {
-      callback();
+      callback(null, new ChannelStateChange('attached', 'attached'));
```
lawrence-forooghian, line 302: Shouldn't this be `null`?

owenpearson: oops, fixed

ably.d.ts
```diff
@@ -2694,32 +2710,32 @@ declare namespace Types {
      *
      * @param event - The event name.
      * @param listener - An event listener function.
-     * @returns A promise which resolves upon success of the channel {@link RealtimeChannelPromise.attach | `attach()`} operation and rejects with an {@link ErrorInfo} object upon its failure.
+     * @returns A promise which, upon successful attachment to the channel, will be fulfilled with a {@link ChannelStateChange} object. Upon failure, the promise will be rejected with an {@link ErrorInfo} object.
```
lawrence-forooghian, line 2713: This one is still missing `null` explanation?

owenpearson: fixed

lawrence-forooghian (approved):

PR #1352, opened 2023-06-22
Fix the type definition of `Crypto.getDefaultParams`

### LINKED ISSUE #1350: TypeScript definition of `Crypto.getDefaultParams` is asynchronous, but should be synchronous

lawrence-forooghian:
The implementations of this method return their result synchronously, as does the feature spec IDL.

I’m not sure why the compiler didn’t catch this.

Resolves #1350.

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -2970,9 +2970,9 @@ declare namespace Types {
      * Returns a {@link CipherParams} object, using the default values for any fields not supplied by the {@link CipherParamOptions} object.
      *
      * @param params - A {@link CipherParamOptions} object.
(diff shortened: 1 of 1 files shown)


owenpearson (approved):
LGTM

> I’m not sure why the compiler didn’t catch this.

Most of ably.d.ts isn't referenced by the source code at all so we have to manually keep it in sync with the actual lib API

lawrence-forooghian:
> Most of ably.d.ts isn't referenced by the source code at all so we have to manually keep it in sync with the actual lib API

In this case it is, though — [our internal `ICryptoStatic` extends the public `Crypto` interface](https://github.com/ably/ably-js/blob/4931665c67fd45a42d95f4c69d5f0f13e04bcbd3/src/common/types/ICryptoStatic.ts#L11).

PR #1357, opened 2023-06-27
fix(ConnectionManager): pass connectCount through connectImpl

owenpearson:
Fixes an issue where host-fallback logic is initially skipped when falling back to base transport


SimonWoolf (approved):
lgtm!

PR #1368, opened 2023-06-30
docs: fix description of AuthOptions.token

owenpearson:
This incorrectly claimed that it was possible to pass a `TokenRequest` object, which is not supported.

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -634,7 +634,7 @@ declare namespace Types {
     queryTime?: boolean;
 
     /**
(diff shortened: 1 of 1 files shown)


lawrence-forooghian (commented):
This docstring comes from https://github.com/ably/sdk-api-reference/blob/main/descriptions.md, which in turn is based on [`TO3j2`](https://sdk.ably.com/builds/ably/specification/main/features/#TO3j2), which states that it can be a `TokenRequest`. Do you think that this is a mistake in the spec?

owenpearson:
@lawrence-forooghian nope, it's not a mistake in the spec - it's just that ably-js doesn't fully implement `TO3j2`

lawrence-forooghian (approved):

PR #1369, opened 2023-06-30
build: use esbuild for web bundles

### LINKED ISSUE #1367: Use esbuild for web bundles
to be done in the integration/v2 branch

owenpearson:
Resolves #1367


src/platform/web/lib/util/webstorage.ts
```diff
@@ -14,16 +16,16 @@ class Webstorage implements IWebStorage {
      * somewhat roundabout way. (If unsupported or no global object,
      * will throw on accessing a property of undefined) */
     try {
-      global.sessionStorage.setItem(test, test);
```
lawrence-forooghian, line 17: How was this even working, given that `global` doesn't exist on the web?

src/common/lib/transport/connectionmanager.ts
```diff
@@ -21,6 +21,8 @@ import HttpStatusCodes from 'common/constants/HttpStatusCodes';
 type Realtime = any;
 type ClientOptions = any;
 
+let globalObject = typeof global !== 'undefined' ? global : typeof window !== 'undefined' ? window : self;
```
lawrence-forooghian, line 24: Is there a reason we couldn't switch all of the use of `global` to [`globalThis`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) instead?

lawrence-forooghian (approved):
I'm not knowledgeable about esbuild, but given that the tests are passing I'm assuming it's all good 👍

src/platform/web/lib/util/webstorage.ts
owenpearson: Because webpack scans for `global` access and replaces it depending on the platform it compiles for

src/common/lib/transport/connectionmanager.ts
owenpearson: This is a good idea, it's a bit of effort to check how this affects platforms support. I don't know, for example, if `globalThis` is widely supported by webworkers since in webworkers it's more common to use `self`. I've opened #1401 to do this separately.

PR #1385, opened 2023-07-11
Auth: adhere to RSA4c3

SimonWoolf:
which requires us to remain connected if we do an authorize which fails for some reason other than an explicit 403 being returned.

([slack discussion](https://ably-real-time.slack.com/archives/C8SPU4589/p1686825790902119))


owenpearson (approved):
LGTM (although slightly disappointing that this doesn't require a test change)

PR #1391, opened 2023-07-13
Make `Utils#inspectError` use `toString` for `Error`-like values

### LINKED ISSUE #1390: `Utils#inspectError` doesn’t include error’s `message` on web
## Description

On the web, `Utils#inspectError` uses `JSON.stringify` to serialize the error. However, `JSON.stringify` will not include the error’s `message` property, as described and explained [here](https://stackoverflow.com/questions/18391212/is-it-not-possible-to-stringify-an-error-using-json-stringify).

We should change the serialization such that the result includes the `message` property (there are some suggestions on the above-linked page).

lawrence-forooghian:
This used to be the logic but was (presumably unintentionally) swapped in 1430d68.

Resolves #1390.

### DIFF
--- a/src/common/lib/util/utils.ts
+++ b/src/common/lib/util/utils.ts
@@ -413,8 +413,8 @@ export function inspectError(err: unknown): string {
     (err as ErrorInfo)?.constructor?.name === 'ErrorInfo' ||
     (err as PartialErrorInfo)?.constructor?.name === 'PartialErrorInfo'
   )
-    return Platform.Config.inspect(err);
-  return (err as Error).toString();
+    return (err as Error).toString();
+  return Platform.Config.inspect(err);
 }
 
 export function inspectBody(body: unknown): string {


sacOO7 (changes requested):
Small request, can we add a simple test for this in `utils.test.js`, I created that test file in the last PR. Just to make sure this doesn't happen again : )

SimonWoolf (approved):
Nice, thanks

owenpearson (approved):
LGTM, thanks

lawrence-forooghian:
@sacOO7 Thank you for the suggestion! I don't really want to spend time on writing a test for this — the codebase does not in general have unit tests for utility things, and I don't think that the consequences of a regression are particularly severe nor is another regression very likely.

sacOO7:
> @sacOO7 Thank you for the suggestion! I don't really want to spend time on writing a test for this — the codebase does not in general have unit tests for utility things, and I don't think that the consequences of a regression are particularly severe nor is another regression very likely.

Actually, we started adding tests recently -> https://github.com/ably/ably-js/blob/main/test/realtime/utils.test.js.
I think utils are also somewhat derived from the spec, so there's nothing wrong with writing tests for them, it makes sure future changes are cross-checked for expected behavior. 

In the current case, we cannot deny the chance of regression happening, since it has happened once. Because we don't know why it happened, there is still a chance it might happen again.

Also, there's not much effort in checking if the `toString` method is called.

sacOO7 (approved):

lawrence-forooghian:
> Actually, we started adding tests recently -> https://github.com/ably/ably-js/blob/main/test/realtime/utils.test.js.

I was speaking to Owen and he said that this is not something we usually do.

> I think utils are also somewhat derived from the spec

I don't think that this piece of logging logic is.

> Also, there's not much effort in checking if the `toString` method is called.

It's a non-trivial amount of effort, which is the main reason I’m resistant to do it — I'd have to find a way of exposing the `ErrorInfo` class to the tests so that I can construct one. Maybe it's easy, maybe it's not, but I don't really want to spend the time thinking about it.

lawrence-forooghian:
> Hey @lawrence-forooghian
> 
> can we remember to link PRs to the jira issue they are fixing please. Happy to have a discussion about why this is required if needed.
> 
> It can be done by tagging the issue key in the title. see [ably/ably-go#596](https://github.com/ably/ably-go/pull/596) Or by adding it to the PR description. see [ably/ably-java#966](https://github.com/ably/ably-java/pull/966)
> 
> This ask is now reflected in: https://github.com/ably/engineering/blob/main/best-practices/pull-requests.md

Will do 👍

PR #1392, opened 2023-07-18
Rename some misnamed parameters

lawrence-forooghian:
This renames a couple of parameters that seem to be misnamed and whose names confused me whilst I was trying to figure out how to make a REST request from inside the codebase. See commit messages for more details.


owenpearson (approved):
Nice one, yeah I've noticed some of these are wrong before, thanks for fixing 👍

PR #1400, opened 2023-07-19
[SDK-3717] Add tree-shakable BaseClient and Rest exports

### LINKED ISSUE #1374: Create tree-shakeable `BaseClient` and `Rest` modules
- Create two classes, `BaseClient` and `Rest` which are available from a separate import path from the main ably-js module
- `BaseClient` optionally takes `Rest` as a second constructor argument, and if so, allows the REST API to be accessed
- Should include a typescript interface for the new import path

owenpearson:
Resolves #1374 

Refactors the existing `Rest` class into `BaseClient` and an encapsulated `Rest` class which can be passed in to the `BaseClient` constructor to allow access to the rest interface.

The tree-shakable modules are currently available as an esm module from 'ably/modules', example usage:
```ts
import { BaseClient, Rest } from 'ably/modules';

const client = new BaseClient(options, { Rest });
```

A new npm script (`npm run modulereport`) is added which displays information about the size of each module in the bundle-report workflow. Currently ([76d2ab0](https://github.com/ably/ably-js/pull/1400/commits/76d2ab0eb9237cc0f0f863a5b0a500b9a5ad50b7)) the output shows:

```
BaseClient: 61.63kb
BaseClient + Rest: 78.56kb
```

These numbers will be lower after we encapsulate further functionality into tree-shakable modules.

### DIFF
--- a/scripts/moduleReport.js
+++ b/scripts/moduleReport.js
@@ -1,0 +1,37 @@
+const esbuild = require('esbuild');
+
+// List of all modules accepted in ModulesMap
+const moduleNames = ['Rest'];
+
+function formatBytes(bytes) {
+  const kb = bytes / 1024;
+  const formatted = kb.toFixed(2);
+  return `${formatted}kb`;
+}
+
+// Gets the bundled size of an array of named exports from 'ably/modules' formatted as a string
+function getImportSize(modules) {
+  const outfile = modules.join('');
+  const result = esbuild.buildSync({
+    stdin: {
+      contents: `export { ${modules.join(', ')} } from './build/modules'`,
+      resolveDir: '.',
+    },
+    metafile: true,
+    minify: true,
+    bundle: true,
+    outfile,
+    write: false,
+  });
+
+  return formatBytes(result.metafile.outputs[outfile].bytes);
+}
+
+// First display the size of the BaseClient
+console.log(`BaseClient: ${getImportSize(['BaseClient'])}`);
+
+// Then display the size of each module together with the BaseClient
+moduleNames.forEach((moduleName) => {
+  const sizeInBytes = getImportSize(['BaseClient', moduleName]);
+  console.log(`BaseClient + ${moduleName}: ${sizeInBytes}`);
+});
--- a/src/common/lib/client/baseclient.ts
+++ b/src/common/lib/client/baseclient.ts
@@ -75,7 +75,7 @@ class BaseClient {
     this.http = new Platform.Http(normalOptions);
     this.auth = new Auth(this, normalOptions);
 
-    if (modules?.Rest) {
+    if (modules.Rest) {
       this._rest = new modules.Rest(this);
     }
   }
--- a/src/common/lib/client/baseclient.ts
+++ b/src/common/lib/client/baseclient.ts
@@ -1,0 +1,128 @@
+import Logger, { LoggerOptions } from '../util/logger';
+import Defaults from '../util/defaults';
+import Auth from './auth';
+import ErrorInfo from '../types/errorinfo';
+import Stats from '../types/stats';
+import { StandardCallback } from '../../types/utils';
+import { IHttp, RequestParams } from '../../types/http';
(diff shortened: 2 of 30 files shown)


src/common/lib/client/realtime.ts
```diff
@@ -14,17 +14,21 @@ import Platform from 'common/platform';
 import Message from '../types/message';
 
 class Realtime extends Rest {
-  channels: any;
```
lawrence-forooghian, line 17: I'm not sure I understand why this change is needed here?

lawrence-forooghian (approved):
I'm approving (with one comment), but as discussed offline, I'm still not sure whether it makes sense to combine the realtime and REST functionality into a single `BaseClient` class. My preference would be to resolve this matter before merging this PR, but if you'd prefer to defer the decision until a bit later then that's cool too 👍

jamienewcomb:
@owenpearson can we remember to link PRs to the jira issue they are fixing please. Happy to have a discussion about why this is required need be.

It can be done by tagging the issue key in the title. see https://github.com/ably/ably-go/pull/596
Or by adding it to the PR description. see https://github.com/ably/ably-java/pull/966

This ask is now reflected in: https://github.com/ably/engineering/blob/main/best-practices/pull-requests.md

src/common/lib/client/baseclient.ts
```diff
@@ -0,0 +1,128 @@
...
+  _currentFallback: null | {
+    host: string;
+    validUntil: number;
+  };
+  serverTimeOffset: number | null;
+  http: IHttp;
+  auth: Auth;
+  _rest?: Rest;
+
+  constructor(options: ClientOptions | string, modules: ModulesMap) {
```
lawrence-forooghian, line 31:
```suggestion
  constructor(options: ClientOptions | string, modules?: ModulesMap) {
```

(given the usage without passing modules in `modules.test.js` and the `if (modules?.Rest)` below)

test/browser/modules.test.js
```diff
@@ -0,0 +1,32 @@
+import { BaseClient, Rest } from '../../build/modules/index.js';
```
lawrence-forooghian, line 1: Does this need to be added to `test/support/browser_file_list.js`?

test/browser/modules.test.js
lawrence-forooghian, line 1: Does this need to be added to `test/support/browser_file_list.js`?

src/common/lib/client/defaultrest.ts
```diff
@@ -0,0 +1,24 @@
...
+ * Preloaded BaseClient with all REST features, used as the default non-treeshakeable Rest export
+ */
+class DefaultRest extends BaseClient {
+  constructor(options: string | ClientOptions) {
+    super(options, {
+      Rest,
+    });
+  }
+
+  static Platform = Platform;
```
lawrence-forooghian, line 18: Shouldn't these static properties all go on `BaseClient`? They're part of the API that it's meant to satisfy.

lawrence-forooghian: (and `BaseClient` already has a static `Platform` property)

test/browser/modules.test.js
lawrence-forooghian: My mistake — fixed as part of our call

src/common/lib/client/realtime.ts
owenpearson: tsc errors if you extend a class and override a dynamic accessor to be a static field, even if the return type of the accessor is the same as the static field

src/common/lib/client/defaultrest.ts
owenpearson: i've removed `Platform` here though, nice spot 👍

scripts/moduleReport.js
lawrence-forooghian, line 1: Can we update the `format` task to include this file?

lawrence-forooghian:
Replaced by #1417.

PR #1406, opened 2023-07-24
fix: ensure ErrorInfo.message is included in JSON output

owenpearson:
fixes an issue where, in web browsers, `Platform.config.inspect(error)` (which is just JSON.stringify)  would not include the ErrorInfo.message since ErrorInfo extends Error for which message is not an enumerable property.


lawrence-forooghian:
Was this not already resolved by https://github.com/ably/ably-js/pull/1391?

owenpearson:
@lawrence-forooghian ah good point, i forgot we did that recently 👍

Issue #1405, opened 2023-07-20
Many callbacks are potentially invoked with an error that doesn't conform to public API


lawrence-forooghian:
There are a couple of issues here:

- Many of our public methods which are meant to emit an `ErrorInfo` (as [defined by our type definitions](https://github.com/ably/ably-js/blob/92003d88d89f5d165748396944cc51bc5d00863f/ably.d.ts#L733-L746)) do not do so (at least, according to static type information — whether this actually is occurring at runtime I don't know) — they only emit an [`IPartialErrorInfo`](https://github.com/ably/ably-js/blob/92003d88d89f5d165748396944cc51bc5d00863f/src/common/lib/types/errorinfo.ts#L5-L10) (and before this type was introduced in 2395b88a1e9dc16965a5f9753429a889f72a7da3 they were instead emitting an `ErrorInfo` class, however that class did not actually conform to the public `ErrorInfo` interface).
- This issue is not being caught by the compiler, because the implementations of these methods are referring to [these internally-defined](https://github.com/ably/ably-js/blob/92003d88d89f5d165748396944cc51bc5d00863f/src/common/types/utils.d.ts#L1-L3) callback types, which refer to an "`ErrorInfo`" type that they do not in fact import. The result of this is that (for reasons I don't understand), instead of causing a compilation failure as one might expect, references to this `ErrorInfo` type just seem to be ignored by the compiler. If you add `import ErrorInfo from '../lib/types/errorinfo';` to the top of `src/common/types/utils.d.ts` you'll see that a whole bunch of compilation errors related to the previous bullet point suddenly appear.

So, we should probably:

1. Fix the fact that this issue is not causing a compilation error. My suggestion would be that we remove the callback types defined in `src/common/types/utils.d.ts` and instead directly refer to the callback types defined in the public `ably.d.ts` interface. (In fact, I’d go further and say that we should also be making sure that all of our types which are intended to implement an interface defined in `ably.d.ts` should explicitly declare that they implement this interface. That is, for example, the [`Auth` class](https://github.com/ably/ably-js/blob/main/src/common/lib/client/auth.ts) should declare `implements API.Types.Auth` etc.)

   And then, once the compilation errors appear, temporarily suppress them with type assertions until we do the next step.
2. Make sure that all of the public methods that are meant to emit an `ErrorInfo` do in fact so do. This is the tricker step and requires us to figure out how to work around the fact, as described in 2395b88a1e9dc16965a5f9753429a889f72a7da3, that:

[...]

PR #1410, opened 2023-07-25
[SDK-1904] Implement batch operations

### LINKED ISSUE #989: Implement batch operations feature
Implement batch operations feature as described in the specification:
https://github.com/ably/docs/blob/main/content/client-lib-development-guide/features.textile#batch-operations 

For reference please see: https://ably.com/docs/rest/batch 

The original PR for spec creation: https://github.com/ably/docs/pull/1421 

Acceptance criteria: feature has been implemented and unit test has been written.

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-1904) by [Unito](https://www.unito.io)

lawrence-forooghian:
This implements the following operations that were added in https://github.com/ably/specification/pull/139.

- batch publish
- batch presence
- token revocation

Resolves #989.

### DIFF
--- a/src/common/lib/client/rest.ts
+++ b/src/common/lib/client/rest.ts
@@ -24,6 +24,9 @@ type BatchPublishSpec = API.Types.BatchPublishSpec;
 type BatchPublishSuccessResult = API.Types.BatchPublishSuccessResult;
 type BatchPublishFailureResult = API.Types.BatchPublishFailureResult;
 type BatchPublishResult = BatchResult<BatchPublishSuccessResult | BatchPublishFailureResult>;
+type BatchPresenceSuccessResult = API.Types.BatchPresenceSuccessResult;
+type BatchPresenceFailureResult = API.Types.BatchPresenceFailureResult;
+type BatchPresenceResult = BatchResult<BatchPresenceSuccessResult | BatchPresenceFailureResult>;
 
 const noop = function () {};
 class Rest {
@@ -294,6 +297,48 @@ class Rest {
         } else {
           (callback as API.Types.StandardCallback<BatchPublishResult[]>)(null, batchResults);
         }
+      }
+    );
+  }
+
+  batchPresence(channels: string[], callback: API.Types.StandardCallback<BatchPresenceResult>): void;
+  batchPresence(channels: string[]): Promise<BatchPresenceResult>;
+  batchPresence(
+    channels: string[],
+    callbackArg?: API.Types.StandardCallback<BatchPresenceResult>
+  ): void | Promise<BatchPresenceResult> {
+    if (callbackArg === undefined) {
+      if (this.options.promises) {
+        return Utils.promisify(this, 'batchPresence', [channels]);
+      }
+      callbackArg = noop;
+    }
+
+    const callback = callbackArg;
+
+    const format = this.options.useBinaryProtocol ? Utils.Format.msgpack : Utils.Format.json,
+      headers = Utils.defaultPostHeaders(this.options, format);
+
+    if (this.options.headers) Utils.mixin(headers, this.options.headers);
+
+    const channelsParam = channels.join(',');
+
+    Resource.get(
+      this,
+      '/presence',
+      headers,
+      { newBatchResponse: 'true', channels: channelsParam },
+      null,
+      (err, body, headers, unpacked) => {
+        if (err) {
+          // TODO remove this type assertion after fixing https://github.com/ably/ably-js/issues/1405
+          callback(err as API.Types.ErrorInfo);
+          return;
+        }
+
+        const batchResult = (unpacked ? body : Utils.decodeBody(body, format)) as BatchPresenceResult;
+
+        callback(null, batchResult);
       }
     );
   }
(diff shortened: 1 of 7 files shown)


src/common/lib/client/rest.ts
```diff
@@ -228,6 +230,155 @@ class Rest {
...
+      API.Types.BatchResult<API.Types.BatchPublishSuccessResult | API.Types.BatchPublishFailureResult>[]
+    >
+  ): void;
+  batchPublish(
+    specs: API.Types.BatchPublishSpec
+  ): Promise<API.Types.BatchResult<API.Types.BatchPublishSuccessResult | API.Types.BatchPublishFailureResult>>;
+  batchPublish(
+    specs: API.Types.BatchPublishSpec[]
+  ): Promise<API.Types.BatchResult<API.Types.BatchPublishSuccessResult | API.Types.BatchPublishFailureResult>[]>;
+  batchPublish(
```
owenpearson, line 251: could simplify this signature a tiny bit using generics
```typescript
  batchPublish<T extends BatchPublishSpec | BatchPublishSpec[]>(specs: T): void 
  | Promise<T extends BatchPublishSpec ? BatchResult<BatchPublishSuccess | BatchPublishFailure> : BatchResult<BatchPublishSuccess | BatchPublishFailure>[]>
```

src/common/lib/client/rest.ts
```diff
@@ -228,6 +230,155 @@ class Rest {
     }
   }
 
+  batchPublish(
+    specs: API.Types.BatchPublishSpec,
```
owenpearson, line 234: Is it worth destructuring these types for the sake of brevity, ie:
```ts
type BatchPublishSpec = API.Types.BatchPublishSpec;
```

src/common/lib/client/rest.ts
```diff
@@ -228,6 +230,155 @@ class Rest {
...
+          | Promise<API.Types.BatchResult<API.Types.BatchPublishSuccessResult | API.Types.BatchPublishFailureResult>[]>;
+      }
+      callbackArg = noop;
+    }
+
+    const callback = callbackArg;
+
+    let requestBodyDTO: API.Types.BatchPublishSpec[];
+    let singleSpecMode: boolean;
+    if (Array.isArray(specOrSpecs)) {
```
owenpearson, line 277: we use a more portable version of this for compatibility
```suggestion
    if (Utils.isArray(specOrSpecs)) {
```

src/common/lib/client/rest.ts
```diff
@@ -228,6 +230,155 @@ class Rest {
...
+          // TODO remove this type assertion after fixing https://github.com/ably/ably-js/issues/1405
+          callback(err as API.Types.ErrorInfo);
+          return;
+        }
+
+        const batchResults = (unpacked ? body : Utils.decodeBody(body, format)) as API.Types.BatchResult<
+          API.Types.BatchPublishSuccessResult | API.Types.BatchPublishFailureResult
+        >[];
+
+        // I don't love the below type assertions for `callback` but not sure how to avoid them
```
owenpearson, line 309: when using the above generic suggestion:
```typescript
const callback = callbackArg as StandardCallback<T extends BatchPublishSpec ? BatchResult<BatchPublishSuccess | BatchPublishFailure> : BatchResult<BatchPublishSuccess | BatchPublishFailure>[]>;
```

owenpearson (changes requested):
Implementation and tests both LGTM, a few comments mainly just about typescript stuff

src/common/lib/client/rest.ts
lawrence-forooghian: Nice, thanks! Have updated.

src/common/lib/client/rest.ts
lawrence-forooghian: Good idea, have updated.

src/common/lib/client/rest.ts
lawrence-forooghian: Cool, have updated.

src/common/lib/client/rest.ts
owenpearson: ah yeah you're right, that wouldn't help much. fwiw i think it's possible to fix this using a typeguard but in a way that's no better than a type assertion anyway

src/common/lib/client/rest.ts
lawrence-forooghian: > declare an intermediate `BatchPublishResult` type

Cool, have done for presence and revocation too.

> use the `StandardCallback` type already available in this module (rather than `API.Types.StandardCallback`

The problem is that these two types are not equivalent. If you update the definition of `StandardCallback` in `src/common/types/utils.d.ts` to be `API.Types.StandardCallback` you'll see that it introduces a bunch of compilation errors. We have issue https://github.com/ably/ably-js/issues/1405 for fixing this. That is, we should be using `API.Types.StandardCallback` throughout the codebase to ensure we're conforming to the public API, and I think it hence makes sense to use it for any new code.

owenpearson: ah ok that makes sense, thanks for explaining!

owenpearson (approved):
LGTM, nice one 👍

Issue #1394, opened 2023-07-18
Create tree-shakable transport modules


owenpearson:
- depends on #1374 and #1393 
- create new  `WebSocketTransport`, `XHRPolling` and `XHRStreaming` classes
- These transports may be used by a `BaseClient` if and only if they are provided as constructor arguments

lawrence-forooghian:
Closed in #1432.

PR #1417, opened 2023-08-02
[SDK-3768] Add tree-shakable `Base*` and `Rest` exports 

### LINKED ISSUE #1415: Update class hierarchy on `tree-shaking` branch
Introduce classes `BaseRest`, `BaseRealtime`, `DefaultRealtime` into #1400, such that:

- `DefaultRest` extends `BaseRest`
- `DefaultRealtime` extends `BaseRealtime` extends `BaseRest`

The aim of this (vs the single `BaseClient` class introduced in #1400) is to give us a place to put the functionality that’s common to `BaseClient` and `DefaultRest`.

lawrence-forooghian:
This PR replaces #1400. It copies a lot from that PR but uses the class hierarchy described by #1415.

The output of `npm run modulereport` is as follows:

```
BaseRest: 74.28 KiB
BaseRest + Rest: 91.14 KiB
BaseRealtime: 148.03 KiB
BaseRealtime + Rest: 156.68 KiB
```

(Note that `BaseRest` is currently somewhat larger than `BaseClient` in #1415. This is because it includes the transports and `EventEmitter`, which are included due to [this line](https://github.com/ably/ably-js/blob/77e48be4dce3c501359d703137065dbd14435753/src/platform/web/modules.ts#L25). This line is absent in #1415 because that PR takes the approach of stripping everything out to add it back in again later, whereas I've taken the approach of keeping everything as it is until we explicitly decide to split it out. We'll be splitting out the transports in https://github.com/ably/ably-js/issues/1394.)

Resolves #1415.

### DIFF
--- a/.github/workflows/bundle-report.yml
+++ b/.github/workflows/bundle-report.yml
@@ -37,4 +37,5 @@ jobs:
           sourcePath: bundle-reports
           githubToken: ${{ secrets.GITHUB_TOKEN }}
           artifactName: bundle-report
+      # This step performs some validation and may fail, so it should come after the steps that we want to always execute.
       - run: npm run modulereport
--- a/.github/workflows/bundle-report.yml
+++ b/.github/workflows/bundle-report.yml
@@ -37,3 +37,4 @@ jobs:
           sourcePath: bundle-reports
           githubToken: ${{ secrets.GITHUB_TOKEN }}
           artifactName: bundle-report
+      - run: npm run modulereport
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -111,6 +111,17 @@ module.exports = function (grunt) {
       target: 'es6',
     };
 
+    var modulesConfig = {
+      ...baseConfig,
+      entryPoints: ['src/platform/web/modules.ts'],
+      outfile: 'build/modules/index.js',
+      format: 'esm',
+      plugins: [],
+    };
+
+    // For reasons I don't understand this build fails when run asynchronously
+    esbuild.buildSync(modulesConfig);
+
     Promise.all([
       esbuild.build(baseConfig),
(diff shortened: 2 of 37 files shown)


owenpearson (changes requested):
Code changes look fine, I thought we opted not to export a rest client without rest apis though - is there any reason someone would want to use a rest client but not call any rest apis?

lawrence-forooghian:
> Code changes look fine, I thought we opted not to export a rest client without rest apis though - is there any reason someone would want to use a rest client but not call any rest apis?

Yep, you're right. I'll look into what the best way to handle this is — we want a way for `BaseRest` to include `Rest` without it being included in its subclass `BaseRealtime`.

lawrence-forooghian:
I've split the `BaseRest` class into `BaseClient` and `BaseRest` classes, the latter of which always includes the `Rest` module. See updated class structure in cd5706290269ebafc95544df9b9d6c7692d39c3f.

I also added new commit f87184e02d4221f5e9a127a23491a7171495cf4e which removes some redundant type annotations.

lawrence-forooghian:
> i'm wondering if this could be solved by (…)

@owenpearson What's the "this" that you're referring to?

owenpearson (approved):
Looks good!

And re your last comment - I can't remember what I was referring to but it sounds like i had missed your previous comments and thought that that `BaseRest` still didn't include the rest module by default?

lawrence-forooghian:
> And re your last comment - I can't remember what I was referring to but it sounds like i had missed your previous comments and thought that that `BaseRest` still didn't include the rest module by default?

That's right — we [discussed it on Slack](https://ably-real-time.slack.com/archives/D01CGFCMG68/p1693902818761259).

PR #1418, opened 2023-08-03
Support presence message extras

lmars:
As per [RTP8a](https://sdk.ably.com/builds/ably/specification/main/features/#RTP8a) and [TP3i](https://sdk.ably.com/builds/ably/specification/main/features/#TP3i), presence messages support the extras field, but there is currently no way to set presence extras in the API (even though the IDL indicates the `enter` function signature as `enter(Data?, extras?: JsonObject) => io`).

Rather than making a breaking change to the `enter` function, this PR introduces an `enterMessage` function which supports passing a `PresenceMessage`, similar to how `publish` supports passing a `Message`, for example:

```javascript
channel.presence.enterMessage(PresenceMessage.fromValues({
  extras: { headers: { key: 'value' } },
}));
```

This functionality is required for an upcoming release of the [Spaces SDK](https://github.com/ably-labs/spaces).

[MMB-212]

[MMB-212]: https://ably.atlassian.net/browse/MMB-212?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -2454,7 +2454,7 @@ declare namespace Types {
     /**
      * Enters the presence set for the channel, passing a `data` payload. A `clientId` is required to be present on a channel.
      *
-     * @param data - The payload associated with the presence member.
+     * @param data - The data payload or {@link PresenceMessage} associated with the presence member.
      * @param callback - A function which will be called upon completion of the operation. If the operation succeeded, then the function will be called with `null`. If it failed, the function will be called with information about the error.
      */
     enter(data?: any, callback?: errorCallback): void;
@@ -2467,14 +2467,14 @@ declare namespace Types {
     /**
      * Updates the `data` payload for a presence member. If called before entering the presence set, this is treated as an {@link PresenceAction.ENTER} event.
      *
-     * @param data - The payload to update for the presence member.
+     * @param data - The data payload or {@link PresenceMessage} object to update for the presence member.
      * @param callback - A function which will be called upon completion of the operation. If the operation succeeded, then the function will be called with `null`. If it failed, the function will be called with information about the error.
      */
     update(data?: any, callback?: errorCallback): void;
     /**
      * Leaves the presence set for the channel. A client must have previously entered the presence set before they can leave it.
      *
-     * @param data - The payload associated with the presence member.
+     * @param data - The data payload or {@link PresenceMessage} associated with the presence member.
      * @param callback - A function which will be called upon completion of the operation. If the operation succeeded, then the function will be called with `null`. If it failed, the function will be called with information about the error.
      */
     leave(data?: any, callback?: errorCallback): void;
@@ -2488,7 +2488,7 @@ declare namespace Types {
      * Enters the presence set of the channel for a given `clientId`. Enables a single client to update presence on behalf of any number of clients using a single connection. The library must have been instantiated with an API key or a token bound to a wildcard `clientId`.
      *
      * @param clientId - The ID of the client to enter into the presence set.
-     * @param data - The payload associated with the presence member.
+     * @param data - The data payload or {@link PresenceMessage} associated with the presence member.
      * @param callback - A function which will be called upon completion of the operation. If the operation succeeded, then the function will be called with `null`. If it failed, the function will be called with information about the error.
      */
     enterClient(clientId: string, data?: any, callback?: errorCallback): void;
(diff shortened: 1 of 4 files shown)


paddybyers:
It's unfortunate to have the asymmetry of having `publish(Message)` but no `enter(PresenceMessage)`, but I don't see an easy way around that whilst remaining compatible.

owenpearson (commented):
According to the features spec, extras should be the second argument to `enter`, `leave`, `update`, similarly it's supposed to be the final argument to `enterClient` et al, is there a reason for breaking from the spec here?

lmars:
> extras should be the second argument to `enter`, `leave`, `update`, similarly it's supposed to be the final argument to `enterClient` et al, is there a reason for breaking from the spec here?

A few reasons:

1. I don't think the spec should be so prescriptive about function signatures, I think it's enough to say "there should be a function which supports entering a channel with data and/or extras" and leave it up to each SDK to implement that in whatever way seems reasonable (this was discussed internally in ADR57), so coming from working on ably-go and ably-rust, I personally ignore the IDL in the spec
2. I think the API should be closer to `publish` which either takes name+data or a `Message`, and I also don't like the function overloading pattern in general
3. I wanted to avoid a breaking change, unless you think this can be supported in a non-breaking way?

owenpearson:
@lmars fair enough, i don't disagree with (1) and (2). As for supporting this without a breaking change we could support both signatures for `enter` using `instanceof`?

```typescript
enter(data: any): Promise<void>;
enter(presenceMessage: PresenceMessage): Promise<void>;

enter(dataOrPresenceMessage: any): Promise<void> {
  if (dataOrPresenceMessage instanceof PresenceMessage) {
    // normalise arguments
  }
  // ...
}
```

owenpearson (changes requested):
also - any changes/additions to the public api need to be added to `ably.d.ts`, so with the changes in their current state we would need to add the `enterMessage` signature.

lmars:
@owenpearson ah thanks, I wasn't aware you could do `x instanceof y`, updated to use that approach in 76624d41, meaning this now supports:

```js
presence.enter(PresenceMessage.fromValues({
  data,
  extras: { headers: { key: 'value' } },
}));
```

Given that `data` has type `any` in ably.d.ts, I assume nothing needs to change there?

owenpearson:
> Given that data has type any in ably.d.ts, I assume nothing needs to change there?
 
Yeah I guess this means we can't add a new signature, would be good to update the docstrings to mention this though

src/common/lib/client/realtimepresence.ts
```diff
@@ -104,41 +104,56 @@ class RealtimePresence extends Presence {
     if (isAnonymousOrWildcard(this)) {
       throw new ErrorInfo('clientId must be specified to enter a presence channel', 40012, 400);
     }
-    return this._enterOrUpdateClient(undefined, undefined, data, 'enter', callback);
+    if (typeof data === 'function') {
+      callback = data as ErrCallback;
+      data = null;
+    }
+    const msg = PresenceMessage.fromData(data);
+    msg.action = 'enter';
```
owenpearson, line 112: could we just move these 5 lines to `_enterOrUpdateClient` and take `msg.action` as an argument rather than repeating it 4 times?

owenpearson (changes requested):

src/common/lib/client/realtimepresence.ts
lmars: Ah yes, I've moved the `fromData` call into `_enterOrUpdateClient` which removes the duplication (and reduces the diff), see 5baf94d9

lmars:
@owenpearson 

> would be good to update the docstrings to mention this though

Done in 5baf94d9.

Comments addressed, PTAL.

ably.d.ts
```diff
@@ -3236,6 +3240,14 @@ declare namespace Types {
      * @param channelOptions - A {@link ChannelOptions} object containing the cipher.
      */
     fromEncodedArray: (JsonArray: any[], channelOptions?: ChannelOptions) => PresenceMessage[];
+
+    /**
+     * Initialises a `PresenceMessage` from a `PresenceMessage`-like object.
+     *
+     * @param values - The values to intialise the `PresenceMessage` from.
+     * @param stringifyAction - Whether to convert the `action` field from a number to a string.
```
SimonWoolf, line 3248: We should not be exposing things like a `stringifyAction` param in the public api. There is no reason a user of the lib should have to know or care that we use numerical encoding in our internal wire protocol, and this doc string is meaningless to such a user.

I'm not a fan of having the one method to do conceptually quite different actions controlled by that param anyway, so in https://github.com/ably/ably-js/pull/1923 I've split it into two different methods, `fromValues(values: Properties<PresenceMessage>): Message`  suitable for public use, and `fromWireProtocol(values: WireProtocolPresenceMessage): PresenceMessage` as an internal method.

PR #1428, opened 2023-08-23
[SDK-3732] Make realtime presence functionality tree-shakable

### LINKED ISSUE #1375: Create tree-shakeable `Msgpack` module
- depends on #1374 
- create a class called `Msgpack` which can be passed to the `BaseClient` constructor
- if and only if the `Msgpack` module is not supplied, `BaseClient` will use json encoding for all requests and the `msgpack` dependency will not be imported

lawrence-forooghian:
**Note: This PR is based on top of #1425; please review that one first.**

We move the realtime presence functionality into a tree-shakable `RealtimePresence` module.

Resolves #1375.

### DIFF
--- a/src/common/lib/client/baserealtime.ts
+++ b/src/common/lib/client/baserealtime.ts
@@ -10,17 +10,20 @@ import { ChannelOptions } from '../../types/channel';
 import ClientOptions from '../../types/ClientOptions';
 import * as API from '../../../../ably';
 import { ModulesMap } from './modulesmap';
+import RealtimePresence from './realtimepresence';
 
 /**
  `BaseRealtime` is an export of the tree-shakable version of the SDK, and acts as the base class for the `DefaultRealtime` class exported by the non tree-shakable version.
  */
 class BaseRealtime extends BaseClient {
+  readonly _RealtimePresence: typeof RealtimePresence | null;
   _channels: any;
   connection: Connection;
 
   constructor(options: ClientOptions, modules: ModulesMap) {
     super(options, modules);
     Logger.logAction(Logger.LOG_MINOR, 'Realtime()', '');
+    this._RealtimePresence = modules.RealtimePresence ?? null;
     this.connection = new Connection(this, this.options);
     this._channels = new Channels(this);
     if (options.autoConnect !== false) this.connect();
--- a/scripts/moduleReport.js
+++ b/scripts/moduleReport.js
@@ -1,7 +1,7 @@
 const esbuild = require('esbuild');
 
 // List of all modules accepted in ModulesMap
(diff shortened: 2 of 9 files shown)


src/common/lib/client/baserealtime.ts
```diff
@@ -10,17 +10,20 @@ import { ChannelOptions } from '../../types/channel';
...
  */
 class BaseRealtime extends BaseClient {
+  readonly _RealtimePresence: typeof RealtimePresence | null;
   _channels: any;
   connection: Connection;
 
   constructor(options: ClientOptions, modules: ModulesMap) {
     super(options, modules);
     Logger.logAction(Logger.LOG_MINOR, 'Realtime()', '');
+    this._RealtimePresence = modules.RealtimePresence ?? null;
```
subashcs, line 26: How is this line different from the following?
``` this._RealtimePresence = modules.RealtimePresence;```

Comment (lawrence-forooghian) [reply src/common/lib/client/baserealtime.ts:26]: It would lead to `_RealtimePresence` being set to `undefined` instead of `null`. As written, `_RealtimePresence` doesn’t accept `undefined` as a value. If your question is "why not make `_RealtimePresence` of type `typeof RealtimePresence | undefined`, then it's just a matter of personal preference; I prefer to use `null` to indicate the absence of a value.

File: src/platform/web/modules/presencemessage.ts
Comment (owenpearson) [inline src/platform/web/modules/presencemessage.ts:1]: empty module?

File: test/common/modules/shared_helper.js [hunk 2]
```diff
@@ -249,6 +249,10 @@ define([
     expect(json1 === json2, 'JSON data contents mismatch.').to.be.ok;
   }
 
+  function randomString() {
```
owenpearson, line 252: Not sure how simple it would be to access from the modules test suite but `Utils.cheapRandStr` is essentially the exact same function

owenpearson (changes requested):

src/platform/web/modules/presencemessage.ts
lawrence-forooghian: Oops, will remove.

test/common/modules/shared_helper.js
lawrence-forooghian: This is a merge gone wrong, the helper function already exists — will remove this copy.

src/platform/web/modules/presencemessage.ts
lawrence-forooghian: Have removed.

test/common/modules/shared_helper.js
lawrence-forooghian: Removed.

owenpearson (approved):
Looks good!

PR #1430, opened 2023-08-23
Export the ErrorInfo class

lmars:
So that wrapper SDKs like Spaces can use the ErrorInfo class, for example:

```typescript
import * as Ably from 'ably';

throw new Ably.ErrorInfo('some error', 12345, 400);
```


owenpearson (approved):
LGTM 👍

PR #1438, opened 2023-09-14
[SDK-3734] Make HTTP request implementations tree-shakable

### LINKED ISSUE #1395: Create tree-shakable HTTP client modules
- depends on #1374 
- create new `FetchRequest` and `XHRRequest` classes
- these clients are used by the client if and only if they are provided to the `BaseClient` constructor.
- if both clients are provided XHR is preferred

lawrence-forooghian:
**Note: This PR is based on top of https://github.com/ably/ably-js/pull/1432; please review that one first.**

This exposes tree-shakable `XHRRequest` and `FetchRequest` modules.

The user is required to provide an HTTP module, even for Realtime, since it’s used for the internet connectivity check and for making a token request to the `authUrl`.

Resolves #1395.


owenpearson (approved):
Looks good! I think we maybe ought to consider making fetch the default in v2 but that's a separate conversation anyway

PR #1453, opened 2023-09-28
[SDK-3883] Remove ability to pass array of event names to `EventEmitter.prototype.once`

### LINKED ISSUE #1452: `EventEmitter.prototype.once(['event1', 'event2'], listener)` discards `listener`’s identity
The case where `once` is passed an array of event names is implemented by wrapping the listener argument in another listener:

https://github.com/ably/ably-js/blob/76cc74fcfa66beff40f2a367c8b9b69b8aabd60e/src/common/lib/util/eventemitter.ts#L260-L273

This means that the event emitter does not hold a reference to the listener argument (other than that held indirectly through the wrapper) and so it is not possible to remove this listener using `off`.

The client library specification does not specify a version of `once` which accepts an array of event names, and we do not advertise it as part of the public API. So, I think the simplest thing would be to remove this functionality.

lawrence-forooghian:
This functionality is implemented by wrapping the listener argument in another listener.  This means that the event emitter does not hold a reference to the listener argument (other than that held indirectly through the wrapper) and so it is not possible to remove this listener using `off(..., listener)`.

The client library specification does not specify a version of `once` which accepts an array of event names, and we do not advertise it as part of the public API. So, I think the simplest thing is to remove this functionality.

Resolves #1452.

### DIFF
--- a/src/common/lib/client/realtimechannel.ts
+++ b/src/common/lib/client/realtimechannel.ts
@@ -165,9 +165,11 @@ class RealtimeChannel extends Channel {
       // Ignore 'attaching' -- could be just due to to a resume & reattach, should not
       // call back setOptions until we're definitely attached with the new options (or
       // else in a terminal state)
-      this._allChannelChanges.once(
+      const self = this;
+      this._allChannelChanges.on(
         ['attached', 'update', 'detached', 'failed'],
-        function (this: { event: string }, stateChange: ConnectionStateChange) {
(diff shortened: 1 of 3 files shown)


lawrence-forooghian:
Thanks @owenpearson, I've created https://github.com/ably/ably-js/issues/1470 and will put it in the v2 epic.

lawrence-forooghian:
Closing this since we decided in #1470 that we'll fix and keep.

PR #1455, opened 2023-09-29
fix: add missing id param for `useStateErrors` call to `useChannelStateListener`

owenpearson:
Resolves a bug where using `useChannel` or `usePresence` wouldn't work with an id'd `AblyProvider`


ttypic (approved):
nice catch!

dpiatek (approved):

PR #1459, opened 2023-10-04
[SDK-3870] refactor: send react-hooks agent as channel param

owenpearson:
Sends the `react-hooks` agent as a channel param (workaround since typically the ably client used by react hooks already starts connecting before the hooks wrapper has an opportunity to add the agent as a transport param).

This branch also contains a change to make the client more selective about which options provided to `channels.get` will be interpreted as requiring reattachment. More precisely, the channel params and modes will now be shallow compared to existing params and modes, and only if they differ will the call the `chnanels.get` throw an error (provided of course that the channel is already attached or attaching).

### DIFF
--- a/src/platform/react-hooks/src/AblyReactHooks.ts
+++ b/src/platform/react-hooks/src/AblyReactHooks.ts
@@ -16,3 +16,15 @@ export type ChannelNameAndId = {
   id?: string;
 };
 export type ChannelParameters = string | ChannelNameAndOptions;
+
+export const version = '1.2.45';
+
+export function channelOptionsWithAgent(options?: Types.ChannelOptions) {
+  return {
+    ...options,
+    params: {
+      ...options?.params,
+      agent: `react-hooks/${version}`,
+    },
+  };
+}
--- a/src/platform/react-hooks/src/hooks/useChannel.ts
+++ b/src/platform/react-hooks/src/hooks/useChannel.ts
@@ -1,6 +1,6 @@
 import { Types } from '../../../../../ably.js';
 import { useEffect, useMemo, useRef } from 'react';
-import { ChannelParameters } from '../AblyReactHooks.js';
+import { channelOptionsWithAgent, ChannelParameters } from '../AblyReactHooks.js';
 import { useAbly } from './useAbly.js';
 import { useStateErrors } from './useStateErrors.js';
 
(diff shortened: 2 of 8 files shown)


src/platform/react-hooks/src/hooks/useChannel.ts
```diff
@@ -45,13 +45,16 @@ export function useChannel(
   const channelOptionsRef = useRef(channelOptions);
   const ablyMessageCallbackRef = useRef(ablyMessageCallback);
 
-  const channel = useMemo(() => ably.channels.get(channelName, channelOptionsRef.current), [ably, channelName]);
+  const channel = useMemo(
+    () => ably.channels.get(channelName, channelOptionsWithAgent(channelOptionsRef.current)),
```
ttypic, line 49: I am a little bit concerned that if we always provide options parameter and you somehow use same channel twice it will throw an exception. And it's not react way, because it prevents hooks composition (with composition it is easy to render the same hook twice, without even realizing it). Can we do smth like this? 

```js
if (ably.channel.exists(channelName) {
  const channel = ably.channel.get(channelName);
  channel.setOptions(channelOptionsWithAgent(channelOptionsRef.current));
  return channel;
} else {
  return ably.channels.get(channelName, channelOptionsWithAgent(channelOptionsRef.current)),
}
```

or maybe we can introduce additional option, similar for [react-use-websocket's share](https://www.npmjs.com/package/react-use-websocket#share-boolean)?

owenpearson: yeah i was thinking about this and really it seems like a limitation of ably-js, the code snippet you shared would stop an error but it would still interfere with other components using the same channel so doesn't seem ideal :/

As for the `share` option, this is kinda what we already do - if you call `channels.get` with a channel that's already being used then it will use that channel instead of creating a new instance.

ttypic: >As for the share option, this is kinda what we already do - if you call channels.get with a channel that's already being used then it will use that channel instead of creating a new instance.

yeah, but with current changes, `options` can't be empty it will always contain `agent` param, and will throw an exception, won't it?

owenpearson: it will only throw an exception if you call `useChannel` with different options in two or more different components, which was always how it behaved - i don't think this changes the error cases

src/platform/react-hooks/src/AblyReactHooks.ts
```diff
@@ -16,3 +16,15 @@ export type ChannelNameAndId = {
   id?: string;
 };
 export type ChannelParameters = string | ChannelNameAndOptions;
+
+export const version = '1.2.45';
+
+export function channelOptionsWithAgent(options?: Types.ChannelOptions) {
+  return {
+    ...options,
+    params: {
```
ttypic, line 25: I think shallowEquals that we are using in `_shouldReattachToSetOptions` won't work this way :( because params will be new for each `useChannel` hook :( Could you add test that:

```
useChannel('my-channel');
useChannel('my-channel');
```
will be rendered without exception?

ttypic (changes requested):

src/platform/react-hooks/src/AblyReactHooks.ts
owenpearson: I'm pretty sure it will work? It's a shallow equal comparison so even if the object reference is the same it will still return true if they have the same properties. You can test it manually with:
```js
console.log(
  Utils.shallowEquals(
    {
      agent: 'ably-js/1.2.3',
    },
    {
      agent: 'ably-js/1.2.3',
    }
  )
);
```

Adding tests for this won't really help with our current testing setup since the react-hooks test use a mock version of the ably-js SDK 😕

ttypic: oh, sorry, you are right! for some reason I thought we compare whole options, not params

ttypic (approved):

PR #1460, opened 2023-10-05
Publish to new CDN bucket only

surminus:
We've finally migrated to the new bucket, so we can stop publishing to the old bucket, which I intend to delete.


owenpearson (changes requested):
Looks good, but we should also update the bucket name in `scripts/README.md`

surminus:
Thanks @owenpearson, updated

owenpearson (approved):
LGTM, thanks

surminus:
@owenpearson, can you merge? I am not authorized to do so :cry:

PR #1462, opened 2023-10-12
fix: don't double url encode agent transport param

owenpearson:
This was never needed - encoding the connect params is the responsibility of each individual transport and the transports already do this.


ttypic (approved):
Nice catch!

PR #1466, opened 2023-10-20
[SDK-3910] Fix RealtimePresence#leave handling of PresenceMessage argument

### LINKED ISSUE #1465: `RealtimePresence#leave` is broken when passing a `PresenceMessage` instance
`RealtimePresence`’s `enter*`, `update*` and `leave*` methods all advertise themselves as accepting a `PresenceMessage` instance instead of just a data object. This is the only way that ably-js supports passing message extras to these methods.

However, `leave` does not handle a `PresenceMessage` argument properly. Update the `presenceMessageExtras` test case in `test/realtime/presence.test.js` and you'll see it fail:

<details>
<summary>
Test case
</summary>

```javascript
/*
 * Attach to channel, enter presence channel with extras and check received
 * PresenceMessage has extras. Then do the same for leaving presence.
 */
it('presenceMessageExtras', function (done) {
  var clientRealtime = helper.AblyRealtime({ clientId: testClientId, tokenDetails: authToken });
  var channelName = 'presenceEnterWithExtras';
  var clientChannel = clientRealtime.channels.get(channelName);
  var presence = clientChannel.presence;

  async.series(
    [
      function (cb) {
        clientChannel.attach(cb);
      },
      // Test entering with extras
      function (cb) {
        presence.subscribe(function presenceListener(presenceMessage) {
          presence.unsubscribe(presenceListener);
          try {
            expect(presenceMessage.extras).to.deep.equal(
              { headers: { key: 'value' } },
              'extras should have headers "key=value"'
            );
          } catch (err) {
            cb(err);
            return;
          }
          cb();
        });
        presence.enter(
          PresenceMessage.fromValues({
            extras: { headers: { key: 'value' } },
          })
        );
      },
      // Test leaving with extras
      function (cb) {
        presence.subscribe(function presenceListener(presenceMessage) {
          presence.unsubscribe(presenceListener);
          try {
            expect(presenceMessage.extras).to.deep.equal(
              { headers: { otherKey: 'otherValue' } },
              'extras should have headers "otherKey=otherValue"'
            );
          } catch (err) {
            cb(err);
            return;
          }
          cb();
        });
        presence.leave(
          PresenceMessage.fromValues({
            extras: { headers: { otherKey: 'otherValue' } },
          })
        );
      },
    ],
    function (err) {
      if (err) {
        closeAndFinish(done, clientRealtime, err);
        return;
      }
      closeAndFinish(done, clientRealtime);
    }
  );

  monitorConnection(done, clientRealtime);
});
```
</details>

It fails because the received presence message looks like this:

```
PresenceMessage {
  id: 'K92Dg-VUnC:1:0',
  clientId: 'testclient',
  connectionId: 'K92Dg-VUnC',
  timestamp: 1697827112736,
  encoding: null,
  data: { action: -1, extras: { headers: [Object] } },
  action: 'leave'
}
```

that is, the extras end up incorrectly being nested inside the message’s data.

lawrence-forooghian:
`RealtimePresence`’s `enter*`, `update*` and `leave*` methods all advertise themselves as accepting a `PresenceMessage` instance instead of just a data object. This is the only way that ably-js supports passing message extras to these methods.

However, `leave` does not handle a `PresenceMessage` argument properly; it ends up nesting the `PresenceMessage`’s top-level fields inside the payload’s `data` property.

To fix this, we change the argument handling to match that used by `enter`.

Resolves #1465.

### DIFF
--- a/src/common/lib/client/realtimepresence.ts
+++ b/src/common/lib/client/realtimepresence.ts
@@ -223,10 +223,8 @@ class RealtimePresence extends Presence {
       'RealtimePresence.leaveClient()',
       'leaving; channel = ' + this.channel.name + ', client = ' + clientId
     );
-    const presence = PresenceMessage.fromValues({
-      action: 'leave',
-      data: data,
-    });
+    const presence = PresenceMessage.fromData(data);
+    presence.action = 'leave';
     if (clientId) {
       presence.clientId = clientId;
     }
--- a/test/realtime/presence.test.js
+++ b/test/realtime/presence.test.js
@@ -373,38 +373,69 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
 
     /*
      * Attach to channel, enter presence channel with extras and check received
-     * PresenceMessage has extras.
+     * PresenceMessage has extras. Then do the same for leaving presence.
      */
     it('presenceMessageExtras', function (done) {
       var clientRealtime = helper.AblyRealtime({ clientId: testClientId, tokenDetails: authToken });
-      var channelName = 'presenceMessageExtras';
+      var channelName = 'presenceEnterWithExtras';
       var clientChannel = clientRealtime.channels.get(channelName);
       var presence = clientChannel.presence;
-      presence.subscribe(
-        function (presenceMessage) {
-          try {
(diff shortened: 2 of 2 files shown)


owenpearson (approved):
Looks good, thanks!

PR #1478, opened 2023-11-01
fix: don't compare `agent` channel param when determining whether channel options require reattach

### LINKED ISSUE #1477: 1.2.46 breaks useChannel hook
[1.2.46](https://github.com/ably/ably-js/releases/tag/1.2.46) is now breaking my React setup.

```javascript
const { channel } = useChannel(
        `private:users.${auth.user.id}`,
        'App\\Events\\MyEvent',
        (
            message: AblyEventMessage<{
                user: User
            }>,
        ) => {
            get(route('myroute'), {
                preserveState: true,
            })
        },
)
```

```
helpers.ts:120 Uncaught Error: Channels.get() cannot be used to set channel options that would cause the channel to reattach. Please, use RealtimeChannel.setOptions() instead.
```

owenpearson:
Resolves #1477 

Excludes the agent param is excluded when checking whether channel options provided to `Channels.get` will require re-attachment

### DIFF
--- a/src/common/lib/client/realtimechannel.ts
+++ b/src/common/lib/client/realtimechannel.ts
@@ -189,7 +189,14 @@ class RealtimeChannel extends Channel {
       return false;
     }
     if (options?.params) {
-      if (!this.params || !Utils.shallowEquals(this.params, options.params)) {
+      // Don't check against the `agent` param - it isn't returned in the ATTACHED message
+      const requestedParams = Object.assign({}, options.params);
+      delete requestedParams.agent;
+      if (Object.keys(requestedParams).length !== 0 && !this.params) {
+        return true;
+      }
+
+      if (this.params && !Utils.shallowEquals(this.params, requestedParams)) {
         return true;
       }
     }
--- a/src/platform/react-hooks/src/hooks/useChannelStateListener.ts
+++ b/src/platform/react-hooks/src/hooks/useChannelStateListener.ts
@@ -1,5 +1,6 @@
+import { useEffect, useRef } from 'react';
 import { Types } from '../../../../../ably.js';
-import { ChannelNameAndId } from '../AblyReactHooks.js';
+import { ChannelNameAndId, ChannelNameAndOptions, channelOptionsWithAgent } from '../AblyReactHooks.js';
 import { useAbly } from './useAbly.js';
 import { useEventListener } from './useEventListener.js';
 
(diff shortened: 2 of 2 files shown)


src/common/lib/client/realtimechannel.ts
```diff
@@ -630,6 +637,7 @@ class RealtimeChannel extends Channel {
       case actions.ATTACHED: {
         this.properties.attachSerial = message.channelSerial;
         this._mode = message.getMode();
+        console.log('Channel.onMessage ATTACHED, setting params to:', (message as any).params);
```
lawrence-forooghian, line 640: Is this meant to be here? Should it be using the internal logging API if so?

src/common/lib/client/realtimechannel.ts
```diff
@@ -189,7 +189,14 @@ class RealtimeChannel extends Channel {
       return false;
     }
     if (options?.params) {
-      if (!this.params || !Utils.shallowEquals(this.params, options.params)) {
+      // Don't check against the `agent` param - it isn't returned in the ATTACHED message
```
lawrence-forooghian, line 192: I'm not sure I understand what's going on here, could you explain please? The relevant spec point is [RTS3c1](https://sdk.ably.com/builds/ably/specification/main/features/#RTS3c1), which doesn't mention any special case for the `agent` param. And I don't think I understand why it's relevant that `agent` "it isn't returned in the ATTACHED message".

src/platform/react-hooks/src/hooks/useChannelStateListener.ts
```diff
@@ -8,22 +8,24 @@ type ChannelStateListener = (stateChange: Types.ChannelStateChange) => any;
...
-    typeof channelNameOrNameAndId === 'string' ? channelNameOrNameAndId : channelNameOrNameAndId.channelName;
+  const channelHookOptions =
+    typeof channelNameOrNameAndId === 'object' ? channelNameOrNameAndId : { channelName: channelNameOrNameAndId };
   const id = (channelNameOrNameAndId as ChannelNameAndId)?.id;
 
+  const { channelName, options: channelOptions } = channelHookOptions;
+
   const ably = useAbly(id);
-  const channel = ably.channels.get(channelName);
+  const channel = ably.channels.get(channelName, channelOptionsWithAgent(channelOptions));
```
lawrence-forooghian, line 28: Is the user allowed to pass options to `useChannelStateListener` that would trigger a reattach? If so, then don't we need to use `setOptions` here instead of `get` with options?

src/common/lib/client/realtimechannel.ts
owenpearson: whoops, my bad, thanks

src/common/lib/client/realtimechannel.ts
owenpearson: When you call `Channels.get` to get an already-attached channel, the client will compare the requested params with the params it received in the `ATTACHED` message to determine whether it would need to re-attach. Since realtime currently doesn't send the `agent` param in the `ATTACHED` message, the client will always believe that it needs to re-attach if the `agent` param is passed in to `Channels.get` but in reality the client will never actually need to re-attach because of the `agent` param, so this change just makes the client ignore this param from the check.

src/platform/react-hooks/src/hooks/useChannelStateListener.ts
owenpearson: Good point - I've updated this commit to use a ref for the channel options and use `setOptions` if they change

ttypic (approved):
LGTM

lawrence-forooghian (approved):

PR #1479, opened 2023-11-01
Turn off `continue-on-error` for browser jobs

lawrence-forooghian:
I _think_ this is preventing us from receiving email notifications when one of the browser jobs fails.

The motivation for the current configuration is not clear to me, but I’m guessing the idea was that the failure of a single browser test run shouldn’t cancel the others. Turning off `fail-fast` should do the same thing.


owenpearson (approved):
LGTM

PR #1480, opened 2023-11-01

lawrence-forooghian:
Other than fixing merge conflicts, the notable changes which bring the recently-added stuff in `main` in line with the changes from v2 are:

- Removing mentions of the `ably/promises` and `Realtime.Promise` API from the React hooks code and documentation

- Updating presence message extras test to use the promise-based API

- Fixing a newly-introduced compilation error calling `wsConnection.send` in `WebSocketTransport` (not sure exactly why, but I’m guessing something to do with the upgrade to the `ws` library in `main` and perhaps some changes to TypeScript version in v2) — the fix, which uses a type assertion to always pretend we’re always in the Node case, is a bit dodgy but that’s because the signature of `ProtocolMessage.serialize` is also a bit dodgy (it refers to `Buffer` even though the code that uses it is also called on web).

The `package-lock.json` merge conflicts were handled by fixing the `package.json` merge conflicts then running `npm install --lockfile-version 2` using Node 16.20.0. Using Node 18.18.2 seemed to lead to a lockfile that caused an error `@esbuild/android-arm not accessible from esbuild` when running `npm ci` in the Node 12 and 14 CI jobs.


owenpearson (approved):
Looks fine! Can we make an issue for `ProtocolMessage.serialize` having an incorrect typing?

lawrence-forooghian:
> Looks fine! Can we make an issue for `ProtocolMessage.serialize` having an incorrect typing?

Sure — created https://github.com/ably/ably-js/issues/1486.

PR #1485, opened 2023-11-03
Upgrade Playwright to 1.39.0

lawrence-forooghian:
This requires us to now explicitly install the browsers (see [this section of Playwright release notes](https://playwright.dev/docs/release-notes#breaking-changes-playwright-no-longer-downloads-browsers-automatically)).

I’m doing this upgrade not because it’s necessary but because NPM tried to make it for me when resolving a `package-lock.json` merge conflict in an upcoming commit. Given that the upgrade involves this CI change, thought it best to do it as a separate commit so it doesn’t get lost inside a merge commit.

### DIFF
--- a/.github/workflows/test-browser.yml
+++ b/.github/workflows/test-browser.yml
@@ -25,8 +25,8 @@ jobs:
         with:
           node-version: 16.x
       - run: npm ci
-      - name: Install playwright dependencies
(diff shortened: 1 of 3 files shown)


owenpearson (approved):
LGTM

ttypic (approved):

PR #1476, opened 2023-10-30
[SDK-3920] Add a basic test of NPM package

lawrence-forooghian:
**Note: This PR is based on top of #1473; please review that one first.**

This adds a package script `test:package`, which performs some basic testing to confirm that a browser-based TypeScript app is able to import the NPM package and use Ably functionality.

This hopefully gives us further confidence that we’ve correctly configured the package’s exports and typings. Next, we’ll build on top of this to add similar testing for the tree-shakable version of the library, once we’ve added typings for it in #1442.

Some of the approach here is copied from that used for [testing the CDN bundle in the Spaces SDK](https://github.com/ably/spaces/commit/fa95f9f).

Resolves #1474.

PR #1496, opened 2023-11-10
[SDK-3936] Remove remaining REST code from `BaseRealtime`

### LINKED ISSUE #1489: Tree-shakable library without `Rest` module shouldn't include `Resource` or `PaginatedResource`
These classes are only used for providing REST functionality, so if the library without the `Rest` module includes them then it means that there are some probably some further APIs that we need to make unavailable in the `Rest`-less client library.

One culprit is the token revocation functionality added in https://github.com/ably/ably-js/pull/1410, but I believe there are others, since this issue also exists in 7e21e56, i.e. before the merge of `main` which introduced the token revocation functionality.

lawrence-forooghian:
**Note: This PR is based on top of #1476; please review that one first.**

This adds further testing around how `BaseRealtime` behaves when you do and don't provide the `Rest` module, and then addresses the REST-related code that `BaseRealtime` was pulling in even when the `Rest` module was not provided. See commit messages for more details.

Resolves #1489.

### DIFF
--- a/src/common/lib/client/auth.ts
+++ b/src/common/lib/client/auth.ts
@@ -11,7 +11,6 @@ import ClientOptions from '../../types/ClientOptions';
 import HttpMethods from '../../constants/HttpMethods';
 import HttpStatusCodes from 'common/constants/HttpStatusCodes';
 import Platform from '../../platform';
-import Resource from './resource';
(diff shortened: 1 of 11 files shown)


owenpearson (approved):
LGTM, thanks

PR #1498, opened 2023-11-10
[SDK-3857] Add typings for the tree-shakable version of the library

### LINKED ISSUE #1442: Add TypeScript typings for tree-shakable version of the library
Expose typings for the tree-shakable version of the library added in https://ably.atlassian.net/browse/SDK-3719 — that is, the `BaseRest` and `BaseRealtime` classes, as well as the tree-shakable modules such as `Rest` etc. And expand the testing added in #1474 to cover the tree-shakable version of the library.

lawrence-forooghian:
**Note: This PR is based on top of #1496; please review that one first.**

This changes the way that the package does its exports, so that the user can import `ably/modules`, and then adds typings for the library that you get when you do this import. See commit messages for more details.

Resolves #1442.


owenpearson (approved):
Looks great!

PR #1502, opened 2023-11-14
[SDK-3858] Add documentation for modular variant of library

### LINKED ISSUE #1443: Add documentation for tree-shakable version of the library
Add documentation for the tree-shakable version of the library added in https://ably.atlassian.net/browse/SDK-3719. This includes:

- README
- generated TypeDoc documentation

lawrence-forooghian:
Adds documentation (TypeDoc and in `README`) for the modular variant of the library. See commit messages for more details.

Resolves #1443.


modules.d.ts
```diff
@@ -10,35 +10,247 @@ export declare const decodePresenceMessage: Types.PresenceMessageStatic['fromEnc
...
+ * ```javascript
+ * import { BaseRealtime, WebSocketTransport, FetchRequest } from 'ably/modules';
+ * const realtime = new BaseRealtime(options, { XHRPolling, FetchRequest });
+ * ```
+ *
+ * Provide this module if you wish the client to have an alternative mechanism for connecting to Ably if it’s unable to establish a WebSocket connection.
+ */
 export declare const XHRPolling: unknown;
+
+// TODO how is it different to XHRPolling? How should the user choose between the two?
```
lawrence-forooghian, line 111: Do you have any knowledge / thoughts that would help here, @owenpearson?

modules.d.ts
```diff
@@ -10,35 +10,247 @@ export declare const decodePresenceMessage: Types.PresenceMessageStatic['fromEnc
...
+ * ```javascript
+ * import { BaseRealtime, WebSocketTransport, FetchRequest } from 'ably/modules';
+ * const realtime = new BaseRealtime(options, { XHRPolling, FetchRequest });
+ * ```
+ *
+ * Provide this module if you wish the client to have an alternative mechanism for connecting to Ably if it’s unable to establish a WebSocket connection.
+ */
 export declare const XHRStreaming: unknown;
+
+// TODO what are the conditions in which somebody would do this? Is Fetch not always available in the browsers we’re targeting?
```
lawrence-forooghian, line 124: Do you have any knowledge / thoughts that would help here, @owenpearson?

modules.d.ts
owenpearson: `xhr_streaming` uses http streaming, ie the request is held open for a long time (60s per request i believe) and realtime messages are sent as chunks, whereas `xhr_polling` is just long polling, ie the request is held open until the server responds and then a new request is opened immediately after.

HTTP streaming is sometimes blocked by corporate firewalls so `xhr_polling` is the more likely transport to succeed but is less preferable because it more requests.

modules.d.ts
owenpearson: Yeah `fetch` is supported by the evergreen browsers, but there are still all sorts of weird web-like js platforms out there. Let's not talk about any specific scenario where someone wants to avoid `fetch` but just list it as an alternative. Also, `XHRRequest` will of course still be needed to use the web comet transports.

README.md
```diff
@@ -81,6 +81,25 @@ WebPack will search your `node_modules` folder by default, so if you include `ab
 
 If that doesn't work for some reason (e.g. you are using a custom webpack target), you can reference the `ably.js` static file directly: `require('ably/build/ably.js');` (or `import * as Ably from 'ably/build/ably.js'` for typescript / ES6 modules).
 
+#### Modular (tree-shakable) variant
+
+Aimed at those who are concerned about their app’s bundle size, the modular variant of the library allows you to create a client which has only the functionality that you choose. Unused functionality can then be tree-shaken by your module bundler.
+
+To use this variant of the library, import `ably/modules`. It exposes a `BaseRealtime` export, as well as various modules that provide functionality, such as `Rest`, `RealtimePresence`, etc. To create a client, choose which modules you wish to use and pass them to the constructor:
```
owenpearson, line 88: I don't like "import `ably/modules`" as a clause, it feels a bit imprecise to me? I tried to rewrite the paragraph to avoid it, but I still don't love it.

```suggestion
To use this variant of the library, import the `BaseRealtime` class from `ably/modules`, along with the additional modules you wish to use. Various modules such as `Rest`, `RealtimePresence`, etc. are available. To create a client, simply pass the additional modules to the `BaseRealtime` constructor as shown in the example below:
```

README.md
```diff
@@ -81,6 +81,25 @@ WebPack will search your `node_modules` folder by default, so if you include `ab
 
 If that doesn't work for some reason (e.g. you are using a custom webpack target), you can reference the `ably.js` static file directly: `require('ably/build/ably.js');` (or `import * as Ably from 'ably/build/ably.js'` for typescript / ES6 modules).
 
+#### Modular (tree-shakable) variant
+
+Aimed at those who are concerned about their app’s bundle size, the modular variant of the library allows you to create a client which has only the functionality that you choose. Unused functionality can then be tree-shaken by your module bundler.
```
owenpearson, line 86: I do think we should emphasise here that the API is the same as for the default clients, and be explicit that you can develop an application with a default client and then switch over to the modular client only when you want to start using it in production.

owenpearson: I also think it's worth being clear here exactly what's required for a working realtime client, ie at least one transport and at least one http client

owenpearson (changes requested):

README.md
```diff
@@ -81,6 +81,25 @@ WebPack will search your `node_modules` folder by default, so if you include `ab
...
+
+To use this variant of the library, import `ably/modules`. It exposes a `BaseRealtime` export, as well as various modules that provide functionality, such as `Rest`, `RealtimePresence`, etc. To create a client, choose which modules you wish to use and pass them to the constructor:
+
+```javascript
+import { BaseRealtime, WebSocketTransport, FetchRequest, RealtimePresence } from 'ably/modules';
+
+const options = { key: 'foo' /* Replace with a real key from the Ably dashboard */ };
```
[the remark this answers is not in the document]
lawrence-forooghian: Thanks, have updated it.

README.md
lawrence-forooghian: I've tweaked the wording based on your suggestion; see what you think.

README.md
lawrence-forooghian: > I do think we should emphasise here that the API is the same as for the default clients, and be explicit that you can develop an application with a default client and then switch over to the modular client only when you want to start using it in production.

Done.

> I also think it's worth being clear here exactly what's required for a working realtime client, ie at least one transport and at least one http client

I already mentioned this in the typings (and hence the TypeDoc docs), but have added it to the README too.

modules.d.ts
lawrence-forooghian: OK, I've left the wording as it is.

modules.d.ts
lawrence-forooghian: Thanks for the information. I've added some wording; see if you think it's enough?

README.md
```diff
@@ -81,6 +81,37 @@ WebPack will search your `node_modules` folder by default, so if you include `ab
...
+  WebSocketTransport,
+  FetchRequest,
+  RealtimePresence
+});
+```
+
+You must at least provide:
+
+- one HTTP request implementation; that is, one of `FetchRequest` or `XHRRequest`;
+- one realtime transport implementation; that is, one of `WebSocketTransport`, `XHRStreaming`, or `XHRPolling`.
```
owenpearson, line 109: hard to explain why i think this is better, but:
```suggestion
You must provide:

- at least one HTTP request implementation; that is, one of `FetchRequest` or `XHRRequest`;
- at least one realtime transport implementation; that is, one of `WebSocketTransport`, `XHRStreaming`, or `XHRPolling`.
```

modules.d.ts
owenpearson: That looks great, thanks 👍

owenpearson (approved):
Looks excellent, nice one 👍 I've got one additional suggested tweak but feel free to ignore if you disagree. I'll approve now so you can merge when you've resolved the eslint ignore comment

README.md
lawrence-forooghian: Yeah, I agree it reads better — will change.

PR #1503, opened 2023-11-16
Remove `static` stuff from `*Message` classes

lawrence-forooghian:
**Note: This PR is based on top of #1443; please review that one first.**

Replace them by standalone exports, which can be tree-shaken.

(My eye was in particular caught by the `Message.encrypt` function, which I was thinking it would be good to — at some point — bundle only if the `Crypto` module is used. This change isn’t made here though.)

### DIFF
--- a/src/common/lib/client/realtimechannel.ts
+++ b/src/common/lib/client/realtimechannel.ts
@@ -1,12 +1,27 @@
-import ProtocolMessage from '../types/protocolmessage';
+import ProtocolMessage, {
+  actions,
+  channelModes,
+  fromValues as protocolMessageFromValues,
+} from '../types/protocolmessage';
 import EventEmitter from '../util/eventemitter';
 import * as Utils from '../util/utils';
 import Logger from '../util/logger';
 import RealtimePresence from './realtimepresence';
-import Message, { CipherOptions } from '../types/message';
+import Message, {
+  fromValues as messageFromValues,
+  fromValuesArray as messagesFromValuesArray,
+  encodeArray as encodeMessagesArray,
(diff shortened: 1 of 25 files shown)


src/common/lib/types/defaultprotocolmessage.ts
```diff
@@ -0,0 +1,10 @@
+import ProtocolMessage, { fromDeserialized } from './protocolmessage';
+
+/**
+ `DefaultPresenceMessage` is the class returned by `DefaultRealtime`’s `PresenceMessage` static property. It exposes the static methods that are used by the tests.
```
owenpearson, line 4: Is it not possible to just use the `fromDeserialized` function in the tests?

lawrence-forooghian:
@owenpearson I've just pushed a change to fix a typo in a comment.

src/common/lib/types/defaultprotocolmessage.ts
lawrence-forooghian: How would you imagine the tests getting hold of that function? As things stand, the `fromDeserialized` function is only exported by the modular version of the SDK. Ditto for the `// Used by tests` stuff in `DefaultMessage`.

owenpearson: Is it possible to just expose the `fromDeserialized` standalone function in the non-modular version too? that would be my preference

lawrence-forooghian: Sure thing, have exported it as `protocolMessageFromDeserialized`.

owenpearson (approved):
Looks good 👍

PR #1504, opened 2023-11-16
[SDK-3937] Create a tree-shakable module for realtime publishing

### LINKED ISSUE #1491: Investigate creating tree-shakable module for message publishing functionality
@owenpearson mentioned that we have many browser use cases which only require subscriptions, and no publishing. He suggested that we create a separate tree-shakable module for this functionality.

This task exists to look into how much effort this would be, since the publishing functionality is quite entangled. I guess that, even if we were not able to fully split it into a separate module now, it would be good to have the module there in the API and we can work on reducing the size of the library without the module later.

lawrence-forooghian:
@owenpearson mentioned that we have many browser use cases which only require subscriptions, and no publishing. He suggested that we create a separate tree-shakable module for this functionality.

This commit introduces the API, but the bundle size savings are minimal since it only pulls out the very low-hanging fruit. I think that we could return to this at some point to see what further size savings we could achieve, but I didn’t want to spend too much time on this now.

Resolves #1491.

### DIFF
--- a/modules.d.ts
+++ b/modules.d.ts
@@ -163,6 +163,20 @@ export declare const FetchRequest: unknown;
 export declare const MessageInteractions: unknown;
 
 /**
+ * Provides a {@link BaseRealtime} instance with the ability to publish messages.
+ *
+ * To create a client that includes this module, include it in the `ModulesMap` that you pass to the {@link BaseRealtime.constructor}:
+ *
+ * ```javascript
+ * import { BaseRealtime, WebSocketTransport, FetchRequest, RealtimePublishing } from 'ably/modules';
+ * const realtime = new BaseRealtime(options, { WebSocketTransport, FetchRequest, RealtimePublishing });
+ * ```
+ *
+ * If this module is not provided, then calling {@link Types.RealtimeChannel.publish} on a channel will cause a runtime error.
+ */
+export declare const RealtimePublishing: unknown;
+
+/**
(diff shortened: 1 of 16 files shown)


owenpearson (commented):
Code changes look fine but I'm not really sure if this is worth it if the savings are only ~2kb 😕

Did you look into whether there's more stuff we could extract from `protocol.ts` or `connectionmanager.ts` for a client which doesn't need to do realtime publishing? Stuff like `msgSerial` and ack/nack handling are specific to publishing so I would imagine there's more we can do here but I also realise most of this stuff is quite baked in so it might be difficult to untangle

lawrence-forooghian:
Yeah, I think you're right that there's more that can be done. I think the mistake that I made is that I looked at the code and thought "oh, well, even if you're not publishing, you still have to send messages like `ATTACH`, so most of this stuff to do with sending messages can't be removed", but you're right, `ACK` and all that stuff only applies to publishing and presence, so I'll take a fresh look.

lawrence-forooghian:
@owenpearson I've pushed another commit which moves the `ACK` handling into a separate module, which is used by the `RealtimePublishing` and `RealtimePresence` modules. The size savings are still tiny, though; it takes `BaseRealtime` from 102.88KiB on the integration branch down to 100.59KiB.

There is still a bit of `msgSerial`-related code hanging around in `ConnectionManager` and `MessageQueue`, but I think the savings that we'd get from removing it would be even more marginal.

What do you think?

lawrence-forooghian:
@owenpearson should I go ahead and close the issue then?

owenpearson:
@lawrence-forooghian yes please 👍

PR #1505, opened 2023-11-16
Exclude `PresenceMessage` from `BaseRealtime` bundle

lawrence-forooghian:
**Note: This PR is based on top of #1503; please review that one first.**

Now `PresenceMessage` only gets included if you provide the `RealtimePresence` module.

PR #1507, opened 2023-11-17
[SDK-3943] Add some regression testing for bundle size and included files

### LINKED ISSUE #1497: Add some regression testing for bundle size and included files
e.g. make CI check that:

- the bundle size of just `BaseRealtime` is at most $x$ KiB
- a fixed list of files is included in the `BaseRealtime` bundle (to make sure e.g. `PaginatedResource` doesn't sneak in)

lawrence-forooghian:
**Note: This PR is based on top of #1505; please review that one first.**

This enhances the `modulereport` script, which gets run in CI, to do the following:

- check which files make a non-trivial contribution to `BaseRealtime` bundle size (against a hardcoded allow list)
- check that the bundle size of a minimal useful Realtime client (i.e. one that can subscribe) does not exceed a hardcoded threshold

The idea being to have some safeguards against accidental regressions in bundle size.

Resolves #1497.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "cors": "^2.8.5",
     "esbuild": "^0.18.10",
     "esbuild-plugin-umd-wrapper": "^1.0.7",
+    "esbuild-runner": "^2.2.2",
     "eslint": "^7.13.0",
     "eslint-plugin-import": "^2.28.0",
     "eslint-plugin-jsdoc": "^40.0.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -87,6 +87,7 @@
     "shelljs": "~0.8",
     "source-map-explorer": "^2.5.2",
     "ts-loader": "^9.4.2",
+    "ts-node": "^10.9.1",
```
owenpearson, line 90: very minor but since we already depend on esbuild i would prefer `esbuild-runner` over `ts-node`, it's waaay faster since it doesn't compile, just strips types

owenpearson (approved):
Looks great! I'm a big fan of the test to ensure we don't accidentally re-include modules in the minimal client. The latter test I think having a static threshold is a bit arbitrary and maybe we could consider something like snapshot testing in future but it's better than nothing so happy to approve anyway

package.json
lawrence-forooghian: Sure, have changed it. Still want to type-check the file though so have added a `tsc --noEmit` first.

owenpearson (approved):

PR #1509, opened 2023-11-20
fix: use 'ably' as import path from react-hooks

owenpearson:
The import paths here only worked before by accident - tsc would compile then to `from '../../../ably'` which in most cases would resolve to `node_modules/ably` however in certain environments (pnpm) this would not be the case and therefore the types would not be imported correctly. I've updated the tsconfig compilerOptions.paths field so that these are resolved correctly for development, and when consumed as a library these will just import directly from the 'ably' module.


ttypic (approved):
LGTM!

PR #1512, opened 2023-11-21
[SDK-3938] Remove the existing plugins mechanism

### LINKED ISSUE #1492: Decide what to do about existing plugins mechanism in light of v2 modularisation
See [this question](https://github.com/ably/ably-js/issues/1210#issuecomment-1523764005) from a while ago, and @mattheworiordan’s more recent question [here](https://ably.atlassian.net/wiki/spaces/SDK/pages/2773221397/Ably+JS+v2?focusedCommentId=2773319736).

One thing that I’m not sure about — as things stand, the v2 modular version of the library is only for web. If we were to replace the `plugins` mechanism with modules, what would that mean for non-web platforms? Would they always come with delta encoding functionality? Or do we need to extend the modules mechanism to other platforms?

lawrence-forooghian:
This removes the `ClientOptions.plugins` mechanism and moves the Vcdiff functionality to a new tree-shakable `Vcdiff` module. See commit messages for more details.

Resolves #1492.


owenpearson (approved):
LGTM, nice 👍

PR #1515, opened 2023-11-23
[SDK-3916] Update publishing methods to accept a `Message`-shaped object

### LINKED ISSUE #1472: Update `publish` methods to accept a `Message`-like type (without specification changes)
We decided today that we would like to implement the change described in #1398 for ably-js v2.

However, there doesn't seem to have been much appetite so far for making a spec change (https://github.com/ably/specification/pull/156 hasn't gained any traction). So, as @SimonWoolf suggested in that PR, let's tweak the typings just in ably-js, perhaps inspired by the approach of that PR.

I do not yet know exactly what form the final result will take.

lawrence-forooghian:
This fixes the optionality of some properties of the `Message` class and then updates the publishing methods to accept a `Message`-shaped object. See commit messages for more details.

Resolves #1472.

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -2055,12 +2055,12 @@ declare namespace Types {
      */
     presence: Presence;
     /**
-     * Retrieves a {@link Types.PaginatedResult} object, containing an array of historical {@link Message} objects for the channel. If the channel is configured to persist messages, then messages can be retrieved from history for up to 72 hours in the past. If not, messages can only be retrieved from history for up to two minutes in the past.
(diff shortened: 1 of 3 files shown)


owenpearson (changes requested):
Looks cool but I'm a bit unsure why we would even keep `Types.Message` here when it's now only referenced in a couple of places and is essentially just an alias to `MessageLike` (I realise that one's a class and one's an interface but I can't think of any situation where it would be useful to have both)? I think it might actually be better developer experience to drop`class Message` and then rename `MessageLike` to `Message` - mainly on the basis that `MessageLike` might be a slightly non-obvious name for some, wdyt?

.

lawrence-forooghian:
@owenpearson makes sense; I've incorporated your suggestions.

PR #1518, opened 2023-11-30
[SDK-1640] Remove the `Types` namespace

### LINKED ISSUE #909: Export library types outside the `Types` namespace
* Using the `Types` namespace is idiosyncratic and generally not what other libraries do.
* Ideally in v2 of the library we would move the types over completely, but for now we should export types from outside that namespace while keeping the namespace there for compat.
* See [internal slack thread](https://ably-real-time.slack.com/archives/C017C5EPYG0/p1637318700006800) for some discussion.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/SDK-1640) by [Unito](https://www.unito.io)

lawrence-forooghian:
We’re removing it because we believe that it’s more consistent with other JS libraries for users to write `import * as Ably from 'ably'` and then access types on the `Ably` namespace.

See commit messages for more details. In particular, I'd welcome feedback on the rename of the tree-shakable `RealtimePresence` module to `RealtimePresenceModule` to avoid a clash, which I’m not very happy about.

Resolves #909.


owenpearson (changes requested):
Looks good on the whole, I do agree though that the naming should be consistent so `RealtimePresenceModule` isn't great. tbh I can't think of a better way to name them to avoid a conflict but I do wonder - do we really need to export all the types from both 'ably/modules' and 'ably' separately? If someone only imports types from 'ably' then they will all get stripped by the compiler anyway so it's not like we need a separate tree-shakable way to access the types?

owenpearson:
> I guess I thought that there were two separate variants of the library, and you used one or the other. You mean you think it'd be better for the user to just import the modules-specific stuff from ably/modules and get the types from ably?

Yeah I mean it's the same library at the end of the day, just different entry points for importing code. I think it's fine to import types from `'ably'` and import the client from `'ably/modules'`

> I'm a bit confused by the word "but" though — is there a connection between your question and the naming of RealtimePresenceModule or have I misunderstood?

That I meant was that if the `RealtimePresence` type was in `'ably'` and the `RealtimePresence` module was in `'ably/modules'` then there wouldn't be a naming conflict, and if someone needed to import both in the same module they can choose how to resolve it themselves by using `as RealtimePresenceModule` or whatever

> Thinking about it more, the types for the modular version are closely based on the way we did them for the previous Promise-based variant of the library. Correspondingly, we also have two separate versions of the TypeDoc documentation, as if they were two separate libraries. If we were to only export the types from the default variant of the library, I'm not sure what that would mean for the TypeDoc documentation for the modular variant of the library — like, where would the links to the referenced types point to if these types are not exported by the library and hence not contained in the documentation?

OK I don't know the workings of our typedoc setup but I'm not a fan of having this one thing named differently from everything else so if we can find a solution to this I would prefer that, do you think it's def not possible?

lawrence-forooghian:
Tests are all failing, I believe it's sandbox related. Will re-run once told it's back up.

owenpearson (approved):
LGTM, thanks

PR #1522, opened 2023-12-01
[SDK-3590] Implement untyped stats API

### LINKED ISSUE #1269: Implement untyped stats API
see: https://github.com/ably/specification/issues/148

lawrence-forooghian:
This implements the API changes of spec commit `a731d12`, and the protocol version bump from spec commit `dfe9476`. Documentation taken from sdk-api-reference commit `570728b` (see https://github.com/ably/sdk-api-reference/pull/37/).

Also removes the `newBatchResponse` from the batch requests, per spec commit `c48464a` (it's no longer needed in protocol v3).

Resolves #1269.


types.d.ts
```diff
@@ -2532,42 +2416,26 @@ export declare class Connection extends EventEmitter<connectionEventCallback, Co
...
   /**
-   * A {@link StatsMessageTraffic} object containing the aggregate count of outbound message stats.
+   * For entries that are still in progress, such as the current month: the last sub-interval included in this entry (in format yyyy-mm-dd:hh:mm:ss), else undefined.
+   */
+  inProgress?: string;
+  /**
+   * The statistics for this time interval and time period. See the JSON schema which the {@link Stats.schema | `schema`} property points to for more information.
    */
-  outbound: StatsMessageTraffic;
+  entries: Record<string, number>;
```
owenpearson, line 2430: did you consider using a `Partial<Record<string, number>>` here? not necessarily saying we should but it's probably more accurate so would be good to make sure we get this right since it would be a breaking change to switch later

owenpearson (commented):
Implementation LGTM, one comment on typings:

lawrence-forooghian:
@owenpearson you mentioned that you'd seen some failing browser tests on this PR that worried you. Do you remember any in particular? After you mentioned it, I fixed some flakiness with the new stats test that this PR adds. Other than that, [here](https://test-observability.herokuapp.com/repos/ably/ably-js/uploads/compare?base-branches%5B%5D=1549-run-tests-in-loop&base-branches%5B%5D=1549-run-tests-in-loop-v2&alternative-branches%5B%5D=untyped-stats-upload-to-observability&alternative-createdBefore=&alternative-createdAfter=&alternative-failureMessage=) is a comparison of the failures on this branch compared to `main` and v2 integration. See "Failures introduced in alternative uploads", after running tests on all three branches many times. They are a bunch of presence-related things that failed a couple of times and I can't really see how they'd be related to the changes in the PR; I'd suspect it's some general flakiness.

lawrence-forooghian:
I've also rebased this PR onto `integration/v2` instead of #1518, because I'm not sure when I'll get round to addressing your feedback in that PR. So this PR is good to go if you're happy with it.

src/common/lib/client/rest.ts
```diff
@@ -210,31 +210,23 @@ export class Rest {
...
+        unpacked ? body : Utils.decodeBody(body, this.client._MsgPack, format)
+      ) as BatchPublishResult[];
 
-        // I don't love the below type assertions for `callback` but not sure how to avoid them
-        if (singleSpecMode) {
-          (callback as StandardCallback<BatchPublishResult>)(null, batchResults[0]);
-        } else {
-          (callback as StandardCallback<BatchPublishResult[]>)(null, batchResults);
-        }
+      // I don't love the below type assertions for `callback` but not sure how to avoid them
```
owenpearson, line 223: For internal typings I don't think we would really lose any strictness by just doing:
```ts
  batchPublish<T extends BatchPublishSpec | BatchPublishSpec[]>(
    specOrSpecs: T,
    callback?: StandardCallback<BatchPublishResult | BatchPublishResult[]>
  ): void | Promise<BatchPublishResult | BatchPublishResult[]> {
```

If we did that, there would be no need to use type assertions here

owenpearson (approved):
LGTM, thanks

src/common/lib/client/rest.ts
lawrence-forooghian: Fair point; on the other hand I guess it's nice to be able to have an overload that matches the type declared in `ably.d.ts`.

PR #1523, opened 2023-12-04
docs(react): use channel option for rewind

owenpearson:
The query string syntax no longer works since we are passing channel options to each channel in order to send the agent param

### DIFF
--- a/docs/react.md
+++ b/docs/react.md
@@ -1,6 +1,6 @@
 # ably-js React Hooks
 
-> [!IMPORTANT]  
+> [!IMPORTANT]
 > The ably-js React Hooks are currently in the release candidate phase, and there may be breaking changes in a future non-major release.
 
 Use Ably in your React application using idiomatic, easy to use, React Hooks!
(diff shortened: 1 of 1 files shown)


docs/react.md
```diff
@@ -120,21 +120,12 @@ const history = channel.history((err, result) => {
...
-    // This call will rewind 100 messages
-    console.log(message);
-});
-```
-
-We support providing [ChannelOptions](https://ably.com/docs/api/realtime-sdk/types#channel-options) to the `useChannel` hook:
-
-```javascript
-const { channel } = useChannel({ channelName: "your-channel-name", options: { ... } }, (message) => {
+const { channel } = useChannel({ channelName: "your-channel-name", options: {params: {rewind: '1'}}}, (message) => {
```
ttypic, line 128: Minor thing but I thing it will be consistent to code style:
```suggestion
const { channel } = useChannel({ channelName: "your-channel-name", options: { params: { rewind: '1' } } }, (message) => {
```

ttypic (approved):

PR #1524, opened 2023-12-04
[SDK-3967, SDK-3968] Remove false class exports in type declarations

### LINKED ISSUE #1519: Sort out the false class exports in our type declarations
(This issue is written with respect to the state of the type declarations after the changes made in #1518, but it applied equally before those changes, too.)

The type declarations claim that the SDK exports all manner of classes that it does not in fact export, such as `Auth`, `Presence`, `PushAdmin`, `AbstractRest`, `AbstractRealtime` etc. Trying to import these classes will result in a runtime (or bundler) error.

We should fix this. 

My suggestion would be to change these classes to interfaces. I think that only place where this will present a bit of a problem is with `AbstractRest` and `AbstractRealtime`; if we change those to interfaces then we'll have to copy all of their method declarations to the `Rest` and `Realtime` class declarations in `ably.d.ts`, and the `BaseRest` and `BaseRealtime` class declarations in `modules.d.ts`. That in itself is not much of an issue, but I’m not sure whether we'd also have to copy all of the documentation comments as well, or whether they'd somehow be inherited (would need to check the behaviour in TypeDoc and also in IntelliSense e.g. in VS Code). We'd also need to consider what to name them; the `Abstract` prefix probably wouldn't make sense any more, but we can't just call them `Rest` and `Realtime` because that clashes with the exports of `ably.d.ts`. Would we need to stick the word `I` or `Interface` in there somewhere?

### LINKED ISSUE #1520: `ClientOptions` documentation links to a nonexistent constructor for the `Abstract*` classes
(This issue is written with respect to the state of the type declarations after the changes made in https://github.com/ably/ably-js/pull/1518, but it applied equally before those changes, too.)

The documentation comment for `ClientOptions` says

```
/**
 * Passes additional client-specific properties to the REST {@link AbstractRest.constructor | `constructor()`} or the Realtime {@link AbstractRealtime.constructor | `constructor()`}.
 */
```

These links are nonsensical; these classes do not have a constructor that takes client options (I mean, they don't even actually exist; see #1519). In the generated TypeDoc HTML documentation, they link to an empty entry for the `new()` constructor for these classes. Ideally they should be linking to the constructor of the actual Rest or Realtime class exported by the SDK, but given that this is different for `ably.d.ts` and `modules.d.ts` I don't think there's an easy way of doing this. So we probably need to reword the comment in a way that removes these constructor links.

lawrence-forooghian:
**Note: This PR is based on top of #1518, please review that one first.**

Resolves #1519, resolves #1520. See commit messages for more details.

### DIFF
--- a/src/platform/react-hooks/src/hooks/useAbly.ts
+++ b/src/platform/react-hooks/src/hooks/useAbly.ts
@@ -2,8 +2,8 @@ import React from 'react';
 import { getContext } from '../AblyProvider.js';
 import * as API from 'ably';
 
-export function useAbly(id = 'default'): API.AbstractRealtime {
-  const client = React.useContext(getContext(id)) as API.AbstractRealtime;
+export function useAbly(id = 'default'): API.RealtimeClient {
+  const client = React.useContext(getContext(id)) as API.RealtimeClient;
 
   if (!client) {
     throw new Error(
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -375,7 +375,7 @@ export interface ChannelMetrics {
 }
 
 /**
- * Passes additional client-specific properties to the REST {@link AbstractRest.constructor | `constructor()`} or the Realtime {@link AbstractRealtime.constructor | `constructor()`}.
+ * Passes additional client-specific properties to the REST constructor or the Realtime constructor.
  */
 export interface ClientOptions extends AuthOptions {
   /**
@@ -1472,7 +1472,7 @@ export type recoverConnectionCallback = (
   callback: recoverConnectionCompletionCallback
 ) => void;
 
-// Internal Classes
+// Internal Interfaces
 
 // To allow a uniform (callback) interface between on and once even in the
 // promisified version of the lib, but still allow once to be used in a way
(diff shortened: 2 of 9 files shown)


src/platform/react-hooks/src/hooks/useAbly.ts
```diff
@@ -2,8 +2,8 @@ import React from 'react';
 import { getContext } from '../AblyProvider.js';
 import * as API from '../../../../../ably.js';
 
-export function useAbly(id = 'default'): API.AbstractRealtime {
-  const client = React.useContext(getContext(id)) as API.AbstractRealtime;
+export function useAbly(id = 'default'): API.RealtimeInterface {
```
owenpearson, line 5: Is there a reason to use `RealtimeInterface` over just using the `Realtime` class in the typing? I think having the word 'interface' in the name of an interface is often considered a code smell so it would be good to avoid having this in an important class in our typings, could we get away with only exposing the realtime class itself and letting people use that as the typing for any realtime client?

owenpearson (changes requested):

src/platform/react-hooks/src/hooks/useAbly.ts
lawrence-forooghian: The `Realtime` class is only exported by the default variant of the library. Presumably we want people to also be able to use React hooks with the modular variant of the library, which exports the `BaseRealtime` class instead. `RealtimeInterface` is the common interface that both of these classes conform to.

owenpearson: ok yeah that makes sense, so they're basically the same type but not compatible because of the constructor typing mismatch? In that case it's fine to keep this but i would prefer `RealtimeClient` over `RealtimeInterface` and to avoid using 'interface' as a suffix for interface names in the public API if possible

lawrence-forooghian: Sure, have changed to `*Client`.

owenpearson (approved):

PR #1541, opened 2023-12-11
Remove callbacks part 1

### LINKED ISSUE #1527: Convert auth code to use Promises internally
`src/common/lib/client/auth.ts` — quite a lot

### LINKED ISSUE #1528: Convert REST code to use Promises internally
src/common/lib/client/baseclient.ts — just calls through to rest on stats, time, request
src/common/lib/client/paginatedresource.ts — quite a lot
src/common/lib/client/resource.ts — quite a lot
src/common/lib/client/push.ts — quite a lot
src/common/lib/client/rest.ts — quite a lot
src/common/lib/client/restchannel.ts — not a lot
src/common/lib/client/restchannelmixin.ts — not a lot
src/common/lib/client/restpresence.ts — not a lot
src/common/lib/client/restpresencemixin.ts — not a lot

### LINKED ISSUE #1530: Convert `getRandomValues` and `getRandomArrayBuffer` to use Promises
src/platform/nativescript/config.js
src/platform/nodejs/config.ts
src/platform/web/config.ts

### LINKED ISSUE #1531: Convert `Message` encoding to use Promises internally
src/common/lib/types/message.ts — not a lot, for the encoding/decoding functions

### LINKED ISSUE #1532: Convert crypto code to use Promises internally
src/platform/nodejs/lib/util/crypto.ts — generateRandom, encrypt
src/platform/web/lib/util/crypto.ts

lawrence-forooghian:
This removes the internal use of callbacks in the codebase, except for the HTTP clients and the realtime functionality. I'll do these in subsequent PRs. See commit messages for more details.

Resolves #1527, resolves #1531, resolves #1532, resolves #1530, resolves #1528.

Jira links:

- https://ably.atlassian.net/browse/SDK-3976
- https://ably.atlassian.net/browse/SDK-3980
- https://ably.atlassian.net/browse/SDK-3981
- https://ably.atlassian.net/browse/SDK-3979
- https://ably.atlassian.net/browse/SDK-3977

### DIFF
--- a/src/common/lib/client/auth.ts
+++ b/src/common/lib/client/auth.ts
@@ -21,7 +21,6 @@ type TokenRevocationFailureResult = API.TokenRevocationFailureResult;
 type TokenRevocationResult = BatchResult<TokenRevocationSuccessResult | TokenRevocationFailureResult>;
 
 const MAX_TOKEN_LENGTH = Math.pow(2, 17);
-function noop() {}
 function random() {
   return ('000000' + Math.floor(Math.random() * 1e16)).slice(-16);
 }
@@ -297,14 +296,13 @@ class Auth {
 
     logAndValidateTokenAuthMethod(this.authOptions);
 
-    return new Promise((resolve, reject) => {
-      this._ensureValidAuthCredentials(true, (err: ErrorInfo | null, tokenDetails?: API.TokenDetails) => {
-        /* RSA10g */
-        delete this.tokenParams.timestamp;
-        delete this.authOptions.queryTime;
-        err ? reject(err) : resolve(tokenDetails!);
-      });
-    });
+    try {
+      return this._ensureValidAuthCredentials(true);
+    } finally {
+      /* RSA10g */
+      delete this.tokenParams.timestamp;
+      delete this.authOptions.queryTime;
+    }
   }
 
   /**
@@ -755,19 +753,13 @@ class Auth {
    */
   async getAuthParams(): Promise<Record<string, string>> {
     if (this.method == 'basic') return { key: this.key! };
-    else
-      return new Promise((resolve, reject) => {
-        this._ensureValidAuthCredentials(false, function (err: ErrorInfo | null, tokenDetails?: API.TokenDetails) {
-          if (err) {
-            reject(err);
-            return;
-          }
-          if (!tokenDetails) {
-            throw new Error('Auth.getAuthParams(): _ensureValidAuthCredentials returned no error or tokenDetails');
-          }
-          resolve({ access_token: tokenDetails.token });
-        });
-      });
+    else {
+      let tokenDetails = await this._ensureValidAuthCredentials(false);
+      if (!tokenDetails) {
+        throw new Error('Auth.getAuthParams(): _ensureValidAuthCredentials returned no error or tokenDetails');
+      }
+      return { access_token: tokenDetails.token };
+    }
   }
 
   /**
@@ -778,18 +770,11 @@ class Auth {
     if (this.method == 'basic') {
       return { authorization: 'Basic ' + this.basicKey };
     } else {
-      return new Promise((resolve, reject) => {
-        this._ensureValidAuthCredentials(false, function (err: ErrorInfo | null, tokenDetails?: API.TokenDetails) {
-          if (err) {
-            reject(err);
-            return;
-          }
-          if (!tokenDetails) {
-            throw new Error('Auth.getAuthParams(): _ensureValidAuthCredentials returned no error or tokenDetails');
-          }
-          resolve({ authorization: 'Bearer ' + Utils.toBase64(tokenDetails.token) });
-        });
-      });
+      const tokenDetails = await this._ensureValidAuthCredentials(false);
+      if (!tokenDetails) {
+        throw new Error('Auth.getAuthParams(): _ensureValidAuthCredentials returned no error or tokenDetails');
+      }
+      return { authorization: 'Bearer ' + Utils.toBase64(tokenDetails.token) };
     }
   }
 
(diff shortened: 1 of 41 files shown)


src/common/lib/client/auth.ts
```diff
@@ -116,7 +115,7 @@ class Auth {
   client: BaseClient;
   tokenParams: API.TokenParams;
   currentTokenRequestId: number | null;
-  waitingForTokenRequest: ReturnType<typeof Multicaster.create> | null;
+  waitingForTokenRequest: ReturnType<typeof Multicaster.create<API.TokenDetails>> | null;
```
owenpearson, line 118: any reason not to do this?
```suggestion
  waitingForTokenRequest: MulticasterInterface<API.TokenDetails> | null;
```

src/common/lib/client/resource.ts
```diff
@@ -94,126 +85,240 @@ function urlFromPathAndParams(path: string, params: Record<string, any>) {
...
+}
+
 class Resource {
-  static get<T = unknown>(
+  /**
+   * @param throwError Whether to throw any error returned by the underlying HTTP client.
+   *
+   * If you specify `true`, then this method will return a `ResourceResponse<T>`, and if the underlying HTTP client returns an error, this method call will throw that error. If you specify `false`, then it will return a `ResourceResult<T>`, whose `err` property contains any error that was returned by the underlying HTTP client.
+   */
+  static async get<T = unknown>(
```
owenpearson, line 139: unrelated to these changes but: i wonder if it's even worth keeping these helper methods. for Http, Resource, and PaginatedResource it would probably save a bit of bloat if we just used `do(method, ...args)` and got rid of these

src/common/lib/util/multicaster.ts
```diff
@@ -1,25 +1,37 @@
...
 
   // Private constructor; use static Multicaster.create instead
-  private constructor(members?: Array<AnyFunction | undefined>) {
-    this.members = (members as Array<AnyFunction>) || [];
+  private constructor(members?: Array<StandardCallback<T> | undefined>) {
+    this.members = (members as Array<StandardCallback<T>>) || [];
   }
 
-  call(...args: unknown[]): void {
+  call(err?: ErrorInfo | null, result?: T): void {
```
owenpearson, line 30: Is this still being used anywhere? If not I'd prefer marking it as `private` just to make it clear how it's intended to be used from now on

owenpearson (approved):
Looks good, a few minor comments:

src/common/lib/client/resource.ts
lawrence-forooghian: Sure, added a58e93bc1539dfe808ef0f6a8cea32c98f047f99.

src/common/lib/util/multicaster.ts
lawrence-forooghian: Sure, added a58e93bc1539dfe808ef0f6a8cea32c98f047f99.

src/common/lib/client/auth.ts
lawrence-forooghian: Nope, guess changing it just didn't seem in scope of the PR, but have added 9eeb7a2e1f32e2e186b317ab5d709dc6cf5666f7.

src/platform/nodejs/lib/util/crypto.ts
```diff
@@ -223,51 +214,58 @@ var createCryptoClass = function (bufferUtils: typeof BufferUtils) {
...
-      this.blockLength = this.iv.length;
     }
 
-    encrypt(plaintext: InputPlaintext, callback: (error: Error | null, data: OutputCiphertext | null) => void) {
+    async encrypt(plaintext: InputPlaintext): Promise<OutputCiphertext> {
       Logger.logAction(Logger.LOG_MICRO, 'CBCCipher.encrypt()', '');
+
+      const iv = await this.getIv();
+      if (!this.encryptCipher) {
+        this.encryptCipher = crypto.createCipheriv(this.algorithm, this.key, iv);
```
VeskeR, line 230: If I understand it correctly, `this.encryptCipher` is checked and `crypto.createCipheriv` is called here again (in additiona to `getIv`) because we may have provided iv value for CBCCipher constructor, so the first call to `getIv` will enter the first if and won't create encryptCipher.
I suggest moving this also to `getIv` method, inside `if (this.iv) {` statement to cover this specific case.
For future-proofing.

src/common/lib/client/auth.ts
```diff
@@ -887,7 +810,7 @@ class Auth {
     }
   }
 
-  _saveTokenOptions(tokenParams: API.TokenParams | null, authOptions: API.AuthOptions | null) {
+  _saveTokenOptions(tokenParams: API.TokenParams | null, authOptions: API.AuthOptions | undefined | null) {
```
VeskeR, line 813: can we mark `authOptions` param as optional here and in `_forceNewToken` instead of having to do `| undefined`?
the same would go for `tokenParams`.

src/common/lib/client/realtimepresence.ts
```diff
@@ -224,92 +201,72 @@ class RealtimePresence extends EventEmitter {
...
+          this.pendingPresence.push({
+            presence: presence,
+            callback: (err) => (err ? reject(err) : resolve()),
+          });
+          break;
+        case 'initialized':
+        case 'failed': {
+          /* we're not attached; therefore we let any entered status
+           * timeout by itself instead of attaching just in order to leave */
+          throw new PartialErrorInfo('Unable to leave presence channel (incompatible state)', 90001);
```
VeskeR, line 219: should call reject with this error

src/common/lib/client/realtimepresence.ts
```diff
@@ -224,92 +201,72 @@ class RealtimePresence extends EventEmitter {
...
+        }
+        default:
+          throw channel.invalidStateError();
```
VeskeR, line 222: should call reject with this error

src/common/lib/types/defaultmessage.ts
```diff
@@ -29,8 +29,8 @@ export class DefaultMessage extends Message {
   }
 
   // Used by tests
-  static encode(msg: Message | PresenceMessage, options: CipherOptions, callback: Function): void {
-    encode(msg, options, callback);
+  static encode<T extends Message | PresenceMessage>(msg: T, options: CipherOptions): Promise<T> {
```
VeskeR, line 32: add async keyword for method

VeskeR (changes requested):

src/common/lib/client/auth.ts
lawrence-forooghian: Have changed the call to `_forceNewToken` inside `authorize` so that it no longer requires a change of signature for `_saveTokenOptions` or `_forceNewToken`.

src/common/lib/types/defaultmessage.ts
lawrence-forooghian: Thanks, will do.

src/common/lib/client/realtimepresence.ts
lawrence-forooghian: Do you have an example of how this is inconsistent with the rest of the codebase? I can't, off the top of my head, think of other places where we wish to reject a promise directly inside the `Promise` constructor’s executor function (i.e. not as the result of an asynchronous operation). `throw`-ing an error seems like the most natural way to handle it.

src/common/lib/client/realtimepresence.ts
lawrence-forooghian: https://github.com/ably/ably-js/pull/1541#discussion_r1467782138

src/platform/nodejs/lib/util/crypto.ts
lawrence-forooghian: Yep, your understanding of the motivation is correct. You're right, it doesn't really make sense to spread the creation of `encryptCipher` across two separate places. But I think that inside `encrypt()` is the more appropriate place for it (so that method can be sure that `encryptCipher` is always non-null when it comes to use it, without having to understand the side effects of `getIv()`), so I've moved it there.

src/common/lib/client/auth.ts
VeskeR: cool!

src/platform/nodejs/lib/util/crypto.ts
VeskeR: Agree! Thank you

src/common/lib/client/realtimepresence.ts
VeskeR: > Do you have an example of how this is inconsistent with the rest of the codebase?

You're right, there are not so many of them. Actually, there is only [one place](https://github.com/ably/ably-js/pull/1541/files#diff-89e4e931a4a9fa77b03349795008725834fd3898ee8fdd31cc83509265929e64R238) that does it, and it's a part of this PR too.

When reviewing the PR the first time I thought these places in `realtimechannel` do that too: [1](https://github.com/ably/ably-js/pull/1541/files#diff-94e592083882d1fcef5156d5f972441c3932403b3f1fce9588802dac216e9142R386), [2](https://github.com/ably/ably-js/pull/1541/files#diff-94e592083882d1fcef5156d5f972441c3932403b3f1fce9588802dac216e9142R393) (since the code around there looks so similar to the one in question in `realtimepresence`), but no, they are rejecting inside the `.once` callback.

However, as a matter of fact, I also couldn't find any other place that is `throw`-ing an error directly inside the Promise constructor. Understandable, since Promises weren't used that much before.

So I guess that'a first and we can agree on a preferred style here going forward.

> `throw-ing` an error ... requires fewer changes upon later removing the use of the Promise constructor

True, but if we were to change the code inside the Promise in the opposite direction: add callback/asynchronous operation (for example wrap this code also in `.once()`, just like in `realtimechannel`), and we forgot to change `throw` statement to `reject` call, we would get uncaught promise errors now - which is much worse than needing to change a couple of lines when removing Promise constructor and changing `reject` to `throw`.

What do you think?

lawrence-forooghian: Thanks for pointing out that other place; I've changed it to also `throw`, too.

I see your point, but I think that we can live with that risk. I don't have strong opinions, though, happy to change them both to use `reject` if you prefer; let me know.

VeskeR: I would prefer we always use `reject` inside Promise constructors.

lawrence-forooghian: Sure; have changed all three to use `reject`.

src/common/lib/client/realtimepresence.ts
lawrence-forooghian: https://github.com/ably/ably-js/pull/1541#discussion_r1471715976

VeskeR (approved):

PR #1558, opened 2023-12-19
build: use `Ably` as UMD lib name

owenpearson:
Fixes a regression in which, when using UMD without commonjs/amd, the esbuild UMD wrapper would add each named export of ably to the global namespace rather than adding it in an object called `Ably` (as it did in v1).

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -119,7 +119,7 @@ module.exports = function (grunt) {
         sourcemap: true,
         format: 'umd',
         banner: { js: '/*' + banner + '*/' },
-        plugins: [umdWrapper.default()],
+        plugins: [umdWrapper.default({ libraryName: 'Ably' })],
         target: 'es6',
       };
     }
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -119,7 +119,7 @@ module.exports = function (grunt) {
         sourcemap: true,
         format: 'umd',
         banner: { js: '/*' + banner + '*/' },
-        plugins: [umdWrapper.default({ libraryName: 'Ably' })],
+        plugins: [umdWrapper.default({ libraryName: 'Ably', amdNamedModule: false })],
         target: 'es6',
       };
     }
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,7 +36,7 @@
         "cli-table": "^0.3.11",
         "cors": "^2.8.5",
         "esbuild": "^0.18.10",
-        "esbuild-plugin-umd-wrapper": "^1.0.7",
+        "esbuild-plugin-umd-wrapper": "ably-forks/esbuild-plugin-umd-wrapper#1.0.7-optional-amd-named-module",
         "esbuild-runner": "^2.2.2",
         "eslint": "^7.13.0",
         "eslint-plugin-import": "^2.28.0",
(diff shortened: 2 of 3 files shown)


lawrence-forooghian (approved):

VeskeR (approved):

VeskeR (changes requested):
`esbuild-plugin-umd-wrapper` outputs different UMD wrapper code for AMD resolution when `libraryName` option is provided compared to what webpack does on our v1 branch. That causes all our browser tests to fail with errors like: "TypeError: Cannot read properties of undefined (reading 'Ably')"

webpack from v1 branch (build/ably.js):
```javascript
	if(typeof exports === 'object' && typeof module === 'object')
		module.exports = factory();
	else if(typeof define === 'function' && define.amd)
		define([], factory);
	else if(typeof exports === 'object')
		exports["Ably"] = factory();
	else
		root["Ably"] = factory();
```

`esbuild-plugin-umd-wrapper` with `libraryName: 'Ably'` from this PR (build/ably.js):
```javascript
    if ("object" == typeof exports && "object" == typeof module) {
      module.exports = f();
    } else if ("function" == typeof define && define.amd) {
      define("Ably", [], f);
    } else if ("object" == typeof exports) {
      exports["Ably"] = f();
    } else {
      g["Ably"] = f();
    }
```

Note the difference for AMD resolution (`define` function call):
Webpack defines "anonymous" module, just `define([], factory);`
`esbuild-plugin-umd-wrapper` plugin defines a [named module](https://requirejs.org/docs/whyamd.html#namedmodules) `define("Ably", [], f);`

This causes all browser tests to fail.
I'm not sure how exactly named modules in AMD behave, and I couldn't find any meaningful explanation of this feature. Seems like it's [not recommended](https://requirejs.org/docs/whyamd.html#namedmodules:~:text=You%20should%20avoid%20naming%20modules%20yourself%2C%20and%20only%20place%20one%20module%20in%20a%20file%20while%20developing.) to use named modules anyway, and we should stick to how it was done in v1 to avoid problems.

There doesn't seem to be an option to change this behavior in `esbuild-plugin-umd-wrapper`.

@owenpearson WDYT?

VeskeR:
I've created [a fork](https://github.com/ably-forks/esbuild-plugin-umd-wrapper/tree/1.0.7-optional-amd-named-module) of an `esbuild-plugin-umd-wrapper` plugin that would allow us to not name our AMD module, thus preserving the behavior we had on v1 branch using webpack.

To use it, we need to change dependency line in package.json to:
```
"esbuild-plugin-umd-wrapper": "ably-forks/esbuild-plugin-umd-wrapper#1.0.7-optional-amd-named-module",
```
and [this line](https://github.com/ably/ably-js/pull/1558/files#diff-8e7e5f941c67b0cf481a9d650295119d4922b96117e7adc069ba30a81a78c65fR120) to:
```javascript
plugins: [umdWrapper.default({ libraryName: 'Ably', amdNamedModule: false })],
```

@owenpearson let me know if you want me to push those changes to this PR (and also resolve conflicts with [integration/v2](https://github.com/ably/ably-js/tree/integration/v2) branch, since it got updated a few times since this PR was opened)

owenpearson:
@VeskeR, nice one, thanks for not only identifying the issue but also creating the fork to fix it! If you're happy to update the PR and resolve conflicts too that would be amazing 🙂

VeskeR (approved):

PR #1567, opened 2024-01-09
Use correct request headers in `Resource` after re-auth

### LINKED ISSUE #1566: `Resource` reauth uses wrong request headers
Note that it's using the headers from the response that triggered the re-auth, instead of the request headers provided by the user:

https://github.com/ably/ably-js/blob/fdd5a566eafac0a7fae9592f890361a608794487/src/common/lib/client/resource.ts#L262

This regression was introduced in 601ebfe7d2808dfcb3bb2fb59d9a09c984eb3d47.

lawrence-forooghian:
It was incorrectly using the headers from the response that triggered the re-auth, instead of the request headers provided by the user. This regression was introduced in 601ebfe.

Resolves #1566.

### DIFF
--- a/src/common/lib/client/resource.ts
+++ b/src/common/lib/client/resource.ts
@@ -247,7 +247,7 @@ class Resource {
         function (
           err: ErrorInfo | ErrnoException | null | undefined,
           res: any,
-          headers: Record<string, string>,
+          resHeaders: Record<string, string>,
           unpacked?: boolean,
           statusCode?: number
         ) {
@@ -263,7 +263,7 @@ class Resource {
             });
             return;
           }
-          callback(err as ErrorInfo, res, headers, unpacked, statusCode);
+          callback(err as ErrorInfo, res, resHeaders, unpacked, statusCode);
         }
       );
     }


VeskeR (approved):
LGTM!

Good catch. Now I wonder if there might be other similar mistakes due to shadowed variables.
@lawrence-forooghian  How did you find this one? Was it causing an error somewhere, or you just noticed it in the code/logs?

I suggest we add [`no-shadow` typescript-eslint rule](https://typescript-eslint.io/rules/no-shadow/) with "error" level and change all such places. Quick check shows there are 45 places in the codebase where a variable is shadowing another one.
Going forward it will help us to name things in more explicit way, just like yours "resHeaders", so it's harder to make a mistake.

lawrence-forooghian:
> How did you find this one?

I just happened to notice it whilst I was converting that code to use promises as part of another task.

> I suggest we add [no-shadow typescript-eslint rule](https://typescript-eslint.io/rules/no-shadow/)

Definitely worth looking into (without looking at the existing violations I'm not sure whether it might get in our way or not, but I think it'll be fine). Created https://github.com/ably/ably-js/issues/1569 for it.

PR #1576, opened 2024-01-16
1529 convert realtime to use promises

lawrence-forooghian:


lawrence-forooghian:
Closing for now; we can pick this up again in the future if we want.

PR #1589, opened 2024-01-22
[SDK-3816] Document platform-specific nature of `useBinaryProtocol` default

### LINKED ISSUE #1424: Web build does not use MessagePack by default
The [currently-implemented behaviour](https://github.com/ably/ably-js/blob/92003d88d89f5d165748396944cc51bc5d00863f/src/common/lib/util/defaults.ts#L278-L282) is that if your options don’t contain the `useBinaryProtocol` key then it will set `useBinaryProtocol` using the `Platform.Config.preferBinary` config, which [is false on web](https://github.com/ably/ably-js/blob/92003d88d89f5d165748396944cc51bc5d00863f/src/platform/web/config.ts#L44). This means that by default web is not using MessagePack, it appears. I have no clue what the intention of this `preferBinary` is (versus `supportsBinary`) nor why it’s false for web.

I think that, unless the user explicitly tells us otherwise (or, post-#1375, if the `Msgpack` module is not supplied), then web should be using MessagePack.

lawrence-forooghian:
I had initially thought it a bug that web doesn’t use MessagePack by default, but [Paddy said](https://ably-real-time.slack.com/archives/CURL4U2FP/p1705674537763479):

> Yes it's intentional, in that browsers' support for binary types in general was historically poor, and JSON transport performance is significantly better in a browser than msgpack. In modern browsers then binary is supported consistently, but I'd still expect that JSON encode/decode performance is dramatically better than msgpack in a browser.

Resolves #1424.


lawrence-forooghian:
> I suggest we also add Paddy's explanation

Sure, done.

VeskeR (approved):
LGTM

PR #1593, opened 2024-01-24
[SDK-4048] Document that `queryTime` auth option requires `Rest` module

### LINKED ISSUE #1592: Test and document effect on token request of not providing `Rest` module
It will fail if the `queryTime` option is set since it needs to make a `time()` call.

lawrence-forooghian:
Because this option causes `time()` to be called internally.

Resolves #1592.


VeskeR (approved):
LGTM

PR #1601, opened 2024-01-29
[SDK-4011] Fix broken link text in generated documentation

### LINKED ISSUE #1560: TypeDoc is outputting incorrect text for links of the form `{@link ChannelState:type}`
We have a few `@link` tags that use a [`:type` meaning](https://typedoc.org/guides/declaration-references/#meaning) for disambiguation. For a tag of the form `{@link ChannelState:type}`, TypeDoc was previously generating HTML link text of "ChannelState", but as of commit 5f69795 (which updated TypeDoc) it now generates text of ":type" (the link's `href` is still correct, though).

Seems like a regression in TypeDoc. Either report it or work around it (I noticed that adding spaces and being explicit about the link text seems to generate the correct text, e.g. `{ @link ChannelState:type | ChannelState }`), or both.

lawrence-forooghian:
Resolves #1560, in which upgrading TypeDoc caused some link text in the generated HTML documentation to become incorrect. See commit messages for more details.

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -833,11 +833,6 @@ export type ChannelMode =
   | ChannelMode.PRESENCE
   | ChannelMode.PRESENCE_SUBSCRIBE
   | ChannelMode.ATTACH_RESUME;
-/**
- * Describes the possible flags used to configure client capabilities, using {@link ChannelOptions}.
- */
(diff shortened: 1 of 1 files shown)


VeskeR (approved):
LGTM

PR #1608, opened 2024-02-02
fix: should reattach channel logic

ttypic:
Resolves https://github.com/ably/ably-js/issues/1609
Jira links [SDK-4075](https://ably.atlassian.net/browse/SDK-4075)

`_shouldReattachToSetOptions` used `this.params` and `this.modes` when checking channel options, but these fields are only set after the channel is attached. This created a bug when `RealtimeChannels.get()` is invoked during `attaching` state.

[SDK-4075]: https://ably.atlassian.net/browse/SDK-4075?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/common/lib/client/realtimechannel.ts
+++ b/src/common/lib/client/realtimechannel.ts
@@ -135,6 +135,7 @@ class RealtimeChannel extends Channel {
   }
 
   setOptions(options?: API.Types.ChannelOptions, callback?: ErrCallback): void | Promise<void> {
+    const previousChannelOptions = this.channelOptions;
     if (!callback) {
       if (this.rest.options.promises) {
         return Utils.promisify(this, 'setOptions', arguments);
@@ -154,7 +155,7 @@ class RealtimeChannel extends Channel {
     }
     Channel.prototype.setOptions.call(this, options);
     if (this._decodingContext) this._decodingContext.channelOptions = this.channelOptions;
-    if (this._shouldReattachToSetOptions(options)) {
+    if (this._shouldReattachToSetOptions(options, previousChannelOptions)) {
       /* This does not just do _attach(true, null, callback) because that would put us
        * into the 'attaching' state until we receive the new attached, which is
        * conceptually incorrect: we are still attached, we just have a pending request to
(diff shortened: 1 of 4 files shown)


src/common/lib/client/realtimechannel.ts
```diff
@@ -189,19 +189,20 @@ class RealtimeChannel extends Channel {
       return false;
     }
     if (options?.params) {
-      // Don't check against the `agent` param - it isn't returned in the ATTACHED message
-      const requestedParams = Object.assign({}, options.params);
-      delete requestedParams.agent;
-      if (Object.keys(requestedParams).length !== 0 && !this.params) {
+      // Don't check against the `agent` param
+      const requestedParams = omitAgent(options.params);
+      const existingParams = omitAgent(this.channelOptions.params);
```
owenpearson, line 194: The problem here which is making tests fail is that in `RealtimeChannel.setOptions` we call the parent `Channel.setOptions` (which updates the value of `this.channelOptions`) before we call `_shouldReattachToSetOptions` so when updating options via `setOptions`, `existingParams` will always be equal to `requestedParams`.

owenpearson (changes requested):

src/common/lib/client/realtimechannel.ts
owenpearson: The simplest fix I can think of is to update `_shouldReattachToSetOptions` to take the existing options as an argument. That way you could copy the value of `this.ChannelOptions` before calling `Channel.prototype.setOptions` and pass them into `_shouldReattach...` explicitly

src/common/lib/client/realtimechannel.ts
```diff
@@ -189,19 +189,20 @@ class RealtimeChannel extends Channel {
       return false;
     }
     if (options?.params) {
-      // Don't check against the `agent` param - it isn't returned in the ATTACHED message
-      const requestedParams = Object.assign({}, options.params);
-      delete requestedParams.agent;
-      if (Object.keys(requestedParams).length !== 0 && !this.params) {
+      // Don't check against the `agent` param
```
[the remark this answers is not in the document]
ttypic: 👍 put it back

src/common/lib/client/realtimechannel.ts
ttypic: thanks, fixed this

owenpearson (approved):
LGTM, thanks!

PR #1613, opened 2024-02-07
Typings: update v2 typings to use createRecoveryKey() instead of recoveryKey

SimonWoolf:
(leave the recoveryKey getter there just so as not to unnecessarily break the code of people who don't use typescript, doesn't cost us much)

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -2386,17 +2386,13 @@ export declare interface Connection
    */
   id?: string;
   /**
-   * A unique private connection key used to recover or resume a connection, assigned by Ably. When recovering a connection explicitly, the `recoveryKey` is used in the recover client options as it contains both the key and the last message serial. This private connection key can also be used by other REST clients to publish on behalf of this client. See the [publishing over REST on behalf of a realtime client docs](https://ably.com/docs/rest/channels#publish-on-behalf) for more info.
+   * A unique private connection key used to recover or resume a connection, assigned by Ably. This private connection key can also be used by other REST clients to publish on behalf of this client. See the [publishing over REST on behalf of a realtime client docs](https://ably.com/docs/rest/channels#publish-on-behalf) for more info. (If you want to explicitly recover a connection in a different SDK instance, see createRecoveryKey() instead)
    */
   key?: string;
   /**
-   * The recovery key string can be used by another client to recover this connection's state in the recover client options property. See [connection state recover options](https://ably.com/docs/realtime/connection#connection-state-recover-options) for more information.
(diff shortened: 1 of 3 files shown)


ably.d.ts
```diff
@@ -2386,17 +2386,13 @@ export declare interface Connection
...
+   * A unique private connection key used to recover or resume a connection, assigned by Ably. This private connection key can also be used by other REST clients to publish on behalf of this client. See the [publishing over REST on behalf of a realtime client docs](https://ably.com/docs/rest/channels#publish-on-behalf) for more info. (If you want to explicitly recover a connection in a different SDK instance, see createRecoveryKey() instead)
    */
   key?: string;
   /**
-   * The recovery key string can be used by another client to recover this connection's state in the recover client options property. See [connection state recover options](https://ably.com/docs/realtime/connection#connection-state-recover-options) for more information.
+   * The recovery key string can be used by another client to recover this connection's state in the recover client options property. See [connection state recover options](https://ably.com/docs/connect/states?lang=javascript#connection-state-recovery) for more information.
    */
-  recoveryKey: string | null;
```
owenpearson, line 2395: the PR description says leave the recoveryKey getter there but this commit removes it (or at least removes it from the typings)

ably.d.ts
```diff
@@ -2386,17 +2386,13 @@ export declare interface Connection
...
+   * A unique private connection key used to recover or resume a connection, assigned by Ably. This private connection key can also be used by other REST clients to publish on behalf of this client. See the [publishing over REST on behalf of a realtime client docs](https://ably.com/docs/rest/channels#publish-on-behalf) for more info. (If you want to explicitly recover a connection in a different SDK instance, see createRecoveryKey() instead)
    */
   key?: string;
   /**
-   * The recovery key string can be used by another client to recover this connection's state in the recover client options property. See [connection state recover options](https://ably.com/docs/realtime/connection#connection-state-recover-options) for more information.
+   * The recovery key string can be used by another client to recover this connection's state in the recover client options property. See [connection state recover options](https://ably.com/docs/connect/states?lang=javascript#connection-state-recovery) for more information.
```
sacOO7, line 2393:
```suggestion
   * createRecoveryKey method returns string that can be used by another client to recover this connection's state in the recover client options property. See [connection state recover options](https://ably.com/docs/connect/states?lang=javascript#connection-state-recovery) for more information.
```

sacOO7 (changes requested):

ably.d.ts
SimonWoolf: yes. per pr description: "for people who don't use typescript"

the point is, if someone is using typescript, they will be alerted to the breaking change when their app fails to compile, so they can switch to the new method. the static typings let us safely be more aggressive with deprecations. but if someone is just using javascript, they get no such warning, the first time they notice the breaking change would be when their app code throws breaks in production. so was gonna keep the old getter (with a deprecated log warning) just for them

PR #1615, opened 2024-02-07
ci: fix pkg version obtained for attw check

owenpearson:
`npm show` just shows the version from NPM rather than the local package, and `npm pkg get version` returns a string with double quotes, so this seems to be the best way to just get the local package version


ttypic (approved):
LGTM!

PR #1618, opened 2024-02-09

lawrence-forooghian:
We need the fix from https://github.com/ably/ably-js/pull/1615.

### DIFF
--- a/src/platform/react-hooks/src/hooks/usePresence.ts
+++ b/src/platform/react-hooks/src/hooks/usePresence.ts
@@ -1,4 +1,4 @@
-import * as Ably from '../../../../../ably.js';
+import type * as Ably from 'ably';
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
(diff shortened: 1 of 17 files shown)


src/platform/react-hooks/src/hooks/usePresence.ts
```diff
@@ -1,4 +1,4 @@
-import * as Ably from '../../../../../ably.js';
+import type * as Ably from 'ably';
```
VeskeR, line 1: Not related to this PR, but I've noticed that v2 branch is importing 'ably' incorrectly in react hooks.
See opened issue for this: https://github.com/ably/ably-js/issues/1619.

This PR correctly resolved merge conflicts for 'usePresence.ts' and `usePresence.test.tsx`, so it has `import type * as Ably from 'ably';`.
I believe all other react hook files also should import in the same way

VeskeR (approved):
LGTM

PR #1583, opened 2024-01-19
[SDK-4003] Fix `multiple_send_binary_2_200` is failing intermittently in browser on v2 branch

VeskeR:
Fixes race condition error caused by byte array being shared in memory between different operations in web crypto.
Adds new tests to cover the case like this.
See commits for more details.

Resolves #1557

PR #1629, opened 2024-02-21
[ECO-4204] Remove code that's supporting older platforms (in v2)

### LINKED ISSUE #1332: Remove fallbacks to `Math.random`
In the absence of a cryptographically secure random number generator, the library falls back to using `Math.random` in two places:

- https://github.com/ably/ably-js/blob/d67eb7cd117bbe1dc7f27b8f45df62e3b0d68c08/src/common/lib/util/utils.ts#L457-L464
- https://github.com/ably/ably-js/blob/d67eb7cd117bbe1dc7f27b8f45df62e3b0d68c08/src/platform/web/lib/util/crypto.js#L36-L53

Since in ably-js 2.0 we are always targeting browsers that will have the [`Crypto.getRandomValues`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues) method, we should be able to assume that all platforms have access to a secure random number generator and hence can remove this fallback code.

The first thing to do, I think, would be to unify the current `IConfig.getRandomValues` and `IConfig.getRandomArrayBuffer` methods into a single method. I started doing this whilst working on #1239 (branch [https://github.com/ably/ably-js/commits/sort-out-myriad-random-2](`sort-out-myriad-random-2`)) but stopped because it didn't seem directly necessary for that task.

### LINKED ISSUE #1538: Remove code that's supporting older platforms (in v2)
In particular, owenpearson said there is a bunch of stuff in `utils.ts` that can be replaced by built-in ES stuff. (Once we've done #1537 we'll know what version of ES we're targeting.)

There may also be other stuff that's supporting older browsers; again, waiting for outcome of #1537 to know what we can remove.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4204) by [Unito](https://www.unito.io)

VeskeR:
Resolves #1538 , resolves #1332  

### DIFF
--- a/src/common/lib/util/logger.ts
+++ b/src/common/lib/util/logger.ts
@@ -45,27 +45,19 @@ const getDefaultLoggers = (): [Function, Function] => {
   let consoleLogger;
   let errorLogger;
 
-  /* Can't just check for console && console.log; fails in IE <=9 */
-  if (
-    (typeof Window === 'undefined' && typeof WorkerGlobalScope === 'undefined') /* node */ ||
-    typeof globalObject?.console?.log?.apply === 'function' /* sensible browsers */
-  ) {
+  // we expect ably-js to be run in environments which have `console` object available with its `log` function
+  if (typeof globalObject?.console?.log === 'function') {
     consoleLogger = function (...args: unknown[]) {
       console.log.apply(console, args);
     };
+
     errorLogger = console.warn
       ? function (...args: unknown[]) {
           console.warn.apply(console, args);
         }
       : consoleLogger;
-  } else if (globalObject?.console.log as unknown) {
-    /* IE <= 9 with the console open -- console.log does not
-     * inherit from Function, so has no apply method */
-    consoleLogger = errorLogger = function () {
-      Function.prototype.apply.call(console.log, console, arguments);
-    };
   } else {
-    /* IE <= 9 when dev tools are closed - window.console not even defined */
+    // otherwise we should fallback to noop for log functions
     consoleLogger = errorLogger = function () {};
   }
 
--- a/src/common/lib/util/utils.ts
+++ b/src/common/lib/util/utils.ts
@@ -322,13 +322,9 @@ export function arrChooseN<T>(arr: Array<T>, n: number): Array<T> {
   return result;
 }
 
-export const trim = (String.prototype.trim as unknown)
-  ? function (str: string) {
-      return str.trim();
-    }
-  : function (str: string) {
-      return str.replace(/^[\s\uFEFF\xA0]+|[\s\uFEFF\xA0]+$/g, '');
-    };
+export const trim = function (str: string) {
+  return str.trim();
+};
 
 /**
  * Uses a callback to communicate the result of a `Promise`. The first argument passed to the callback will be either an error (when the promise is rejected) or `null` (when the promise is fulfilled). In the case where the promise is fulfilled, the resulting value will be passed to the callback as a second argument.
--- a/src/common/lib/util/utils.ts
+++ b/src/common/lib/util/utils.ts
@@ -44,12 +44,6 @@ export function copy<T = Record<string, unknown>>(src: T | Record<string, unknow
 }
 
 /*
- * Determine whether or not a given object is
- * an array.
- */
-export const isArray = Array.isArray;
-
-/*
  * Ensures that an Array object is always returned
  * returning the original Array of obj is an Array
  * else wrapping the obj in a single element Array
@@ -58,7 +52,7 @@ export function ensureArray(obj: Record<string, unknown>): unknown[] {
   if (isEmptyArg(obj)) {
     return [];
   }
-  if (isArray(obj)) {
+  if (Array.isArray(obj)) {
     return obj;
   }
   return [obj];
@@ -154,7 +148,7 @@ export function containsValue(ob: Record<string, unknown>, val: unknown): boolea
 }
 
 export function intersect<K extends string, T>(arr: Array<K>, ob: K[] | Partial<Record<K, T>>): K[] {
-  return isArray(ob) ? arrIntersect(arr, ob) : arrIntersectOb(arr, ob);
+  return Array.isArray(ob) ? arrIntersect(arr, ob) : arrIntersectOb(arr, ob);
 }
 
 export function arrIntersect<T>(arr1: Array<T>, arr2: Array<T>): Array<T> {
--- a/src/common/lib/util/utils.ts
+++ b/src/common/lib/util/utils.ts
@@ -49,7 +49,7 @@ export function copy<T = Record<string, unknown>>(src: T | Record<string, unknow
  * else wrapping the obj in a single element Array
  */
 export function ensureArray(obj: Record<string, unknown>): unknown[] {
-  if (isEmptyArg(obj)) {
+  if (isNil(obj)) {
     return [];
   }
   if (Array.isArray(obj)) {
@@ -72,16 +72,12 @@ export function isEmpty(ob: Record<string, unknown> | unknown[]): boolean {
   return true;
 }
 
-/*
- * Determine whether or not an argument to an overloaded function is
- * undefined (missing) or null.
- * This method is useful when constructing functions such as (WebIDL terminology):
- *   off([TreatUndefinedAs=Null] DOMString? event)
- * as you can then confirm the argument using:
- *   Utils.isEmptyArg(event)
+/**
+ * Checks if `value` is `null` or `undefined`.
+ *
+ * Source: https://github.com/lodash/lodash/blob/main/src/isNil.ts
  */
-
-export function isEmptyArg(arg: unknown): arg is null | undefined {
+export function isNil(arg: unknown): arg is null | undefined {
   return arg === null || arg === undefined;
 }
 
--- a/src/common/lib/util/utils.ts
+++ b/src/common/lib/util/utils.ts
@@ -227,20 +227,9 @@ export function forInOwnNonNullProperties(ob: Record<string, unknown>, fn: (prop
   }
 }
 
-export const arrFilter = (Array.prototype.filter as unknown)
-  ? function <T>(arr: Array<T>, fn: (value: T, index?: number, arr?: Array<T>) => boolean) {
-      return arr.filter(fn);
-    }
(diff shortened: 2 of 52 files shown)


src/common/lib/util/logger.ts
```diff
@@ -45,10 +45,9 @@ const getDefaultLoggers = (): [Function, Function] => {
   let consoleLogger;
   let errorLogger;
 
-  /* Can't just check for console && console.log; fails in IE <=9 */
   if (
     (typeof Window === 'undefined' && typeof WorkerGlobalScope === 'undefined') /* node */ ||
-    typeof globalObject?.console?.log?.apply === 'function' /* sensible browsers */
+    typeof globalObject?.console?.log === 'function' /* sensible browsers */
```
lawrence-forooghian, line 50: This code is a bit confusing after the deletion in b00fb7c82aa4facaea70ea413c08e937c844373a:

- what is a "sensible browser"?
- under what circumstances is the `else` triggered now?

src/platform/web/lib/util/webstorage.ts
```diff
@@ -1,4 +1,3 @@
-import * as Utils from 'common/lib/util/utils';
```
lawrence-forooghian, line 1: If the intention was that this PR be reviewed commit by commit (which, given the "See commit messages for more details" instruction it presumably was), then 84a99908ad1e3b38c5c0acec6be4adacf0a727a2 is just kinda noise for reviewers — shouldn’t this change just go into the commit that introduced the linter error? Ditto 3d41166d5a7d6a0cbf8b361e9571a9b2f62c5b03.

src/platform/web/config.ts
```diff
@@ -76,6 +77,18 @@ const Config: IPlatformConfig = {
       crypto.getRandomValues(arr);
     };
   })(globalObject.crypto),
+  getRandomArrayBuffer: (function (crypto) {
+    const DEFAULT_BLOCKLENGTH_WORDS = 4; // 32-bit words
```
lawrence-forooghian, line 73: I'm a bit confused by this implementation — what's the idea of this "32-bit word" thing, and why do we have to go via `BufferUtils`? I haven't looked into the exact API calls needed but I assumed that something along the lines of the following would suffice:

1. create a `Int8Array` of the desired length
1. pass that to `crypto.getRandomValues`
1. return the underlying `ArrayBuffer`

lawrence-forooghian: Oh, I see now that it's been copied and pasted from the web crypto code. If you want to keep the implementation as it is, then I’d suggest one of:

1. creating an issue to revisit this implementation and putting a comment about it here, because it’s going to confuse people in the future I think, or 
2. at least mentioning in the commit message that this is copied from existing code

src/common/lib/util/utils.ts
```diff
@@ -227,14 +227,6 @@ export function forInOwnNonNullProperties(ob: Record<string, unknown>, fn: (prop
   }
 }
 
-/* Useful when the function may mutate the array */
```
lawrence-forooghian, line 273: Is this removal of `safeArrForEach` related to removing code that's supporting older platforms? If not, then I wonder whether it'd be better to keep it; the removed comment makes the intention a bit clearer. Or alternatively this commit could put a comment in `connectionmanager.ts` explaining _why_ we're using `slice()`?

src/platform/web/lib/http/request/xhrrequest.ts
```diff
@@ -91,7 +74,7 @@ class XHRRequest extends EventEmitter implements IXHRRequest {
     super();
     params = params || {};
     params.rnd = Utils.cheapRandStr();
-    if (needJsonEnvelope() && !params.envelope) params.envelope = 'json';
+    if (!params.envelope) params.envelope = 'json';
```
lawrence-forooghian, line 77: Replace `needJsonEnvelope()` with `false` (which is `needJsonEnvelope()`’s return value in the non-IE10 case)  and you'll see that this is a behaviour change. I think we want to remove this line.

lawrence-forooghian (commented):
There still seem to be some mentions of IE:

- https://github.com/ably/ably-js/blob/3d41166d5a7d6a0cbf8b361e9571a9b2f62c5b03/test/browser/connection.test.js#L21
- https://github.com/ably/ably-js/blob/3d41166d5a7d6a0cbf8b361e9571a9b2f62c5b03/test/common/modules/testapp_manager.js#L37

and of browser compatibility accommodations:

- https://github.com/ably/ably-js/blob/3d41166d5a7d6a0cbf8b361e9571a9b2f62c5b03/.eslintrc.js#L22 (related to [this](https://caniuse.com/?search=trailing) I think)


Also, just a heads-up that writing "Resolves #1538 , #1332" in your PR description doesn't do what you think it does — you need to write "Resolves #1538, resolves #1332".

VeskeR:
> There still seem to be some mentions of IE:

Initially I wasn't sure if I wanted to remove IE code from tests that would lead to those tests not even being able to run on IE, but now looking at it again it really doesn't make sense to have this code present anymore. We're never going to test against IE and if someone else would try to run those tests - well it is going to break for a reason - we don't support IE anymore.
So will remove those mentions from test cases too.

> and of browser compatibility accommodations:

Will take a look.

> Also, just a heads-up that writing "Resolves https://github.com/ably/ably-js/issues/1538 , https://github.com/ably/ably-js/issues/1332" in your PR description doesn't do what you think it does — you need to write "Resolves https://github.com/ably/ably-js/issues/1538, resolves https://github.com/ably/ably-js/issues/1332".

Right, thanks!

src/common/lib/util/logger.ts
VeskeR: Agree, it looks more confusing now without special cases for IE.
I think whole `if` statement can be rewritten to just check for `typeof globalObject?.console?.log === 'function'` (without typeofs for Window and WorkerGlobalScope for node), and call it a `sensible environment` - meaning that any reasonable environment we expect ably-js to be running on should have a `console` object available.
WDYT?

> under what circumstances is the else triggered now?

Nothing comes to mind immediately, but it is a safety check that if an environment doesn't have `console` for some reason, we default to noop for log operations and don't fail whole application because of this.
Would like to remove it?

src/platform/web/lib/util/webstorage.ts
VeskeR: > If the intention was that this PR be reviewed commit by commit

I didn't intend it from the beginning, but given the type of changes done in this PR it totally makes sense to go commit by commit instead of looking at everything at once.

I agree it's a noise, will squash those.

src/platform/web/config.ts
VeskeR: > Oh, I see now that it's been copied and pasted from the web crypto code.

Yeah, this is just a copy paste.
I actually wanted to further change this code after unifying it all under `IPlatformConfig.getRandomArrayBuffer`, but I think I kinda lost it with all other changes.

If you look closely at those lines:

```javascript
const DEFAULT_BLOCKLENGTH_WORDS = 4; // 32-bit words

return async function (byteLength: number): Promise<ArrayBuffer> {
  const blockRandomArray = new Uint32Array(DEFAULT_BLOCKLENGTH_WORDS);
  const words = byteLength / 4;
  const nativeArray = words === DEFAULT_BLOCKLENGTH_WORDS ? blockRandomArray : new Uint32Array(words);

  ...
```

You will notice that `const nativeArray = words === DEFAULT_BLOCKLENGTH_WORDS ? blockRandomArray : new Uint32Array(words);` doesn't actually accomplish anything. It's the same if we were always creating Uint32Array with `words` length in the first place.

It also didn't do anything in web crypto code. Before https://github.com/ably/ably-js/pull/1583/commits/fccd4ac950858634ccb7e0120ad5520261be63b8 it would actually have different behavior, since we were reusing pre-existing array, but after that change there is no point in `words === DEFAULT_BLOCKLENGTH_WORDS` check anymore.

So it can reduced to:

```javascript
return async function (byteLength: number): Promise<ArrayBuffer> {
  const words = byteLength / 4;
  const nativeArray = new Uint32Array(words);

  crypto.getRandomValues(nativeArray);
  return bufferutils.toArrayBuffer(nativeArray);
};
```

And now there is really now point to use `words` and Uint32Array anymore, so per your first comment we can change it to:

```javascript
return async function (byteLength: number): Promise<ArrayBuffer> {
  const nativeArray = new Uint8Array(byteLength);
  crypto.getRandomValues(nativeArray);
  return nativeArray.buffer;
};
```

Will add those changes with explanation in commit message.

src/platform/web/lib/http/request/xhrrequest.ts
VeskeR: Good catch, thank you! Yes, removing the line is the right thing here.

src/common/lib/util/utils.ts
VeskeR: It just didn't feel like a good thing to have a separate Utils method just to call two native Array methods back to back without any additional logic.

I do agree though that it is now less clear why we need to use `.slice()` in `connectionmanager.ts`, and adding comments there would mean duplicating that comment, whereas before it was only in one place in Utils.safeArrForEach.

I don't have strong opinion on this one, can omit this commit if you'd prefer that, let me know

src/platform/web/lib/util/webstorage.ts
VeskeR: Squashed

src/platform/web/config.ts
VeskeR: Simplified this function and added explanation to commit message in https://github.com/ably/ably-js/pull/1629/commits/5106b006944d06e66adc477f66e158a4752d70f1.

src/common/lib/util/utils.ts
VeskeR: Decided to add comments to `connectionmanager.ts` in https://github.com/ably/ably-js/pull/1629/commits/8a8e93f3caf4416575d49327cbd549fdf5508e02

VeskeR:
Applied changes from conversations, removed IE mentions from tests and removed browser compatibility eslint rule

lawrence-forooghian (approved):

VeskeR:
Added https://github.com/ably/ably-js/pull/1629/commits/2fc88a33cd6ef5fbf8b5147460bf13fabbcc7e1c to remove node 0.8.x specific code, please review

lawrence-forooghian (approved):

PR #1631, opened 2024-02-26
[CHA-138] paginatedresource: allow for no content

AndyTWF:
Handles the situation whereby an endpoint returns 204 that caused paginatedresource to throw an error. This does not check for the body actually being empty, but at least handles the situation whereby the response is obviously trying not to return one.

This does not fully address #676 in that it does not check the body itself for being empty and is not robust in this sense. However it at least respects the 204 status code - which will unblock other teams using the SDK.

### DIFF
--- a/src/common/lib/client/resource.ts
+++ b/src/common/lib/client/resource.ts
@@ -36,7 +36,7 @@ function unenvelope<T>(callback: ResourceCallback<T>, format: Utils.Format | nul
     }
 
     if (outerStatusCode === httpStatusCodes.NoContent) {
-      callback(err, [] as any, [] as any, true, outerStatusCode);
+      callback(err, [] as any, outerHeaders, true, outerStatusCode);
       return;
     }
 
--- a/src/common/lib/client/resource.ts
+++ b/src/common/lib/client/resource.ts
@@ -6,6 +6,7 @@ import HttpMethods from '../../constants/HttpMethods';
 import ErrorInfo, { IPartialErrorInfo, PartialErrorInfo } from '../types/errorinfo';
 import Rest from './rest';
 import { ErrnoException } from '../../types/http';
+import httpStatusCodes from '../../constants/HttpStatusCodes';
 
 function withAuthDetails(
   rest: Rest,
@@ -31,6 +32,11 @@ function unenvelope<T>(callback: ResourceCallback<T>, format: Utils.Format | nul
   return (err, body, outerHeaders, unpacked, outerStatusCode) => {
     if (err && !body) {
       callback(err);
+      return;
+    }
+
+    if (outerStatusCode === httpStatusCodes.NoContent) {
+      callback(err, [] as any, [] as any, true, outerStatusCode);
       return;
     }
 
--- a/test/rest/http.test.js
+++ b/test/rest/http.test.js
@@ -67,7 +67,7 @@ define(['ably', 'shared_helper', 'chai'], function (Ably, helper, chai) {
       //Intercept Http.do with test
 
       function testRequestHandler(_, __, ___, ____, _____, ______, callback) {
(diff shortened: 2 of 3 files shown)


ably.d.ts
```diff
@@ -3081,7 +3081,6 @@ declare namespace Types {
      *
      * Creates a new {@link ChannelBase} or {@link RealtimeChannelBase} object, with the specified channel {@link DeriveOptions}
      * and {@link ChannelOptions}, or returns the existing channel object.
-     *
```
owenpearson, line 3084: Guessing this was staged by mistake? Having a line between description and params is consistent with the other jsdoc annotations

test/rest/http.test.js
```diff
@@ -62,5 +62,26 @@ define(['ably', 'shared_helper', 'chai'], function (Ably, helper, chai) {
...
+      }
+
+      rest.http.do = testRequestHandler;
+
+      let callbackCalled = false;
+      rest.request('GET', '/foo', {}, null, {}, (error, response) => {
+        expect(error).to.be.null;
+        expect(response.statusCode).to.equal(204);
+        expect(response.items).to.be.empty;
+        callbackCalled = true;
```
owenpearson, line 80: i think it's nicer to just call `done` inside the callback here than to mutate a variable and test against it

test/rest/http.test.js
```diff
@@ -62,5 +62,26 @@ define(['ably', 'shared_helper', 'chai'], function (Ably, helper, chai) {
...
+        callback(null, null, null, false, 204);
+      }
+
+      rest.http.do = testRequestHandler;
+
+      let callbackCalled = false;
+      rest.request('GET', '/foo', {}, null, {}, (error, response) => {
+        expect(error).to.be.null;
+        expect(response.statusCode).to.equal(204);
+        expect(response.items).to.be.empty;
```
owenpearson, line 79: can you `try/catch` these assertions and call `done(err)` in the catch block? that will make the test fail fast rather than going to a timeout

test/rest/http.test.js
```diff
@@ -62,5 +62,26 @@ define(['ably', 'shared_helper', 'chai'], function (Ably, helper, chai) {
 
       done();
     });
+
+    it('Should handle no content responses', function (done) {
+      //Intercept Http.do with test
+
+      function testRequestHandler(_, __, ___, ____, _____, ______, callback) {
```
owenpearson, line 69: not really important but might slightly simpler to test this by using `https://echo.ably.io/respondWith?status=204`

owenpearson (changes requested):
looks good! couple of suggestions though:

test/rest/http.test.js
AndyTWF: Done!

ably.d.ts
AndyTWF: Whoops, removed!

test/rest/http.test.js
AndyTWF: I did try this but sadly it didn't like it :(

owenpearson (approved):
LGTM 👍

src/common/lib/client/resource.ts
```diff
@@ -34,6 +35,11 @@ function unenvelope<T>(callback: ResourceCallback<T>, format: Utils.Format | nul
       return;
     }
 
+    if (outerStatusCode === httpStatusCodes.NoContent) {
+      callback(err, [] as any, [] as any, true, outerStatusCode);
```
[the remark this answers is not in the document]
AndyTWF: Good shout! Used the outer headers in https://github.com/ably/ably-js/pull/1631/commits/088d67f3b6987cd412445c73f57ff9aeba5a7199

VeskeR (approved):
LGTM!

PR #1633, opened 2024-02-26
[ECO-4121] Publicly document v2 platform compatibility and update CI accordingly

### LINKED ISSUE #1605: Publicly document v2 platform compatibility and update CI accordingly
Now that we've decided on our platform compatibility guarantees for v2 (in https://github.com/ably/ably-js/issues/1537), update:

* the claims we make about platform support in the README
* the `engines` field in `package.json`
* CI to run on the Node versions we claim to support
* the version of ES that we target when bundling / compiling TypeScript

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4121) by [Unito](https://www.unito.io)

VeskeR:
Resolves #1605 

Changes:

- Use Node.js 20 in github workflows
- Run CI tests for node on Node.js 16, 18, 20
- Update engines in package.json to node >=16
- Update platform compatibility statements in README and CONTRIBUTING
- Change build/bundle ES target to ES2017
- Convert NodeCometTransport in nodejs platform to ES6 classes (otherwise it fails in runtime after building for ES2017)

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,11 +14,19 @@ This library currently targets the [Ably client library features spec](https://w
 
 This SDK supports the following platforms:
 
-**Browsers:** All major desktop and mobile browsers, including (but not limited to) Chrome, Firefox, IE (only version 9 or newer), Safari on iOS and macOS, Opera, and Android browsers.
+**Browsers:** All major desktop and mobile browsers, including (but not limited to) Chrome, Firefox, Edge, Safari on iOS and macOS, Opera, and Android browsers. IE is not supported. See compatibility table below for more information on minimum supported versions for major browsers:
+
+| Browser       | Minimum supported version | Release date  |
+| ------------- |:-------------------------:| -------------:|
+| Chrome        | 58                        | Apr 19, 2017  |
+| Firefox       | 52                        | Mar 7, 2017   |
+| Edge          | 79                        | Dec 15, 2020  |
+| Safari        | 11                        | Sep 19, 2017  |
+| Opera         | 45                        | May 10, 2017  |
 
 **Webpack:** see [using Webpack in browsers](#using-webpack), or [our guide for serverside Webpack](#serverside-usage-with-webpack)
 
-**Node.js:** version 8.17 or newer. (1.1.x versions work on Node.js 4.5 or newer). We do not currently provide an ESM bundle, please [contact us](https://www.ably.com/contact) if you would would like to use ably-js in a NodeJS ESM project.
+**Node.js:** version 16.x or newer. (1.1.x versions work on Node.js 4.5 or newer, 1.2.x versions work on Node.js 8.17 or newer). We do not currently provide an ESM bundle, please [contact us](https://www.ably.com/contact) if you would would like to use ably-js in a NodeJS ESM project.
 
 **React (release candidate):** We offer a set of React Hooks which make it seamless to use ably-js in your React application. See the [React Hooks documentation](./docs/react.md) for more details.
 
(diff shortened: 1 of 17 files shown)


VeskeR:
I'm not sure about [this place](https://github.com/ably/ably-js/blob/d0e9fbe4797ebb8520286592ba7b5d82a139eb39/README.md?plain=1#L37) in README:

```
Ably-js has fallback mechanisms in order to be able to support older browsers; specifically it supports comet-based connections for browsers that do not support websockets. Each of these fallback transport mechanisms is supported and tested on all the browsers we test against, even when those browsers do not themselves require those fallbacks. These mean that the library should be compatible with nearly any browser on most platforms.
Known browser incompatibilities will be documented as an issue in this repository using the ["compatibility" label](https://github.com/ably/ably-js/issues?q=is%3Aissue+is%3Aopen+label%3A%22compatibility%22).
```

Should we remove it entirely or reword it somehow? Currently it does imply that we support much older versions than we listed above. At the same time we do still have comet transport in codebase in case websockets doesn't work, even though all of the browser versions we support have websockets API available.

lawrence-forooghian:
> Ably-js has fallback mechanisms in order to be able to support older browsers; specifically it supports comet-based connections for browsers that do not support websockets. Each of these fallback transport mechanisms is supported and tested on all the browsers we test against, even when those browsers do not themselves require those fallbacks. These mean that the library should be compatible with nearly any browser on most platforms.

I think that this statement is misleading. My understanding is that browser compatibility is only one of the reasons that we support alternative (i.e. non-WebSocket) transports. The other is that there are potential network conditions (firewalls or proxies are the ones I’ve seen talked about) that might block WebSocket connections (see [the documentation for `WebSocketTransport`](https://github.com/ably/ably-js/blob/ec95b59e8b50cb78cf783dfec960619b09c31203/modules.d.ts#L143)). So, I think we should remove this statement from the compatibility section of the `README`. Might be good to explain about alternative transports somewhere else, but I don't think it needs to be in this issue.

> Known browser incompatibilities will be documented as an issue in this repository using the ["compatibility" label](https://github.com/ably/ably-js/issues?q=is%3Aissue+is%3Aopen+label%3A%22compatibility%22).

This statement seems to no longer be accurate — there are no issues using that label and it doesn't appear in our [guidance about which labels to use](https://github.com/ably/engineering/blob/583425887d6a5e099edc8663cefbc64a02ef037e/best-practices/github.md?plain=1#L207-L234). So I think we can remove that part of the statement.

README.md
```diff
@@ -30,7 +38,9 @@ This SDK supports the following platforms:
...
 
-We regression-test the library against a selection of those (which will change over time, but usually consists of the versions that are supported upstream, plus old versions of IE).
+We test the library against a selection of browsers using their latest versions. Please refer to [the test-browser GitHub workflow](./.github/workflows/test-browser.yml) for the set of browsers that currently undergo CI testing.
+
+We regression-test the library against a selection of Node.js version, which will change over time. We will always support and test against current LTS Node.js versions, and optionally some older versions that are supported upstream. We reserve the right to drop support for non-LTS versions in a non-major release. We will update the `engines` field in [package.json](./package.json) whenever we change the Node.js versions supported by the project. Please refer to [the test-node GitHub workflow](./.github/workflows/test-node.yml) for the set of versions that currently undergo CI testing.
```
lawrence-forooghian, line 43: - "selection of Node.js <del>version</del> versions"
- "optionally some older versions that are _supported upstream_" — I know this language was already there, but I don't really know what it means; do you?

VeskeR:
> The other is that there are potential network conditions (firewalls or proxies are the ones I’ve seen talked about) that might block WebSocket connections (see [the documentation for WebSocketTransport](https://github.com/ably/ably-js/blob/ec95b59e8b50cb78cf783dfec960619b09c31203/modules.d.ts#L143)).

I didn't know about that, thank you! Now it makes sense why we would still need comet transport even after switching to newer platforms.

Yea, I will reword that paragraph and find a better place for it then. No reason for it to be in the supported platforms section.

> This statement seems to no longer be accurate — there are no issues using that label and it doesn't appear in our [guidance about which labels to use](https://github.com/ably/engineering/blob/583425887d6a5e099edc8663cefbc64a02ef037e/best-practices/github.md?plain=1#L207-L234). So I think we can remove that part of the statement.

Will remove.

README.md
VeskeR: > "optionally some older versions that are supported upstream" — I know this language was already there, but I don't really know what it means; do you?

Yes, "upstream" refers to our dependencies we have in ably-js, in this case the ones we use in node.js bundle (`@ably/msgpack-js`, `got` and `ws` atm). Then, the entire sentence means that if those libraries still support older versions of node.js, we may decide to support them too (hence "optionally"). Conversely, if all our dependencies require node.js 18+, there is no way for us to support an older node.js 16, even if we wanted to.

I agree it's a bit confusing the way it's written right now, I also had to think about what it meant for a while..
I think it's a bit more cleared this way:
```
and optionally some older versions that are still supported by upstream dependencies
```

Will change to that.

VeskeR:
Removed fallback transport paragraph from this PR.
Reworded it and added to another section in https://github.com/ably/ably-js/pull/1664.

Also applied other changes.

PR #1640, opened 2024-02-27
Bring `whenState` in line with spec

### LINKED ISSUE #1597: Understand if `whenState` can be removed from v2 API
It's not clear to me why these methods are part of the SDK’s public API.

The first one was added in 55b1be4fde0ccf15535a89c9da2a79cd430ce181 as an "undocumented feature" used by a test case. It was then added to the typings in https://github.com/ably/ably-js/pull/444, again not sure why.

It's not in the specification, and it's not clear how the state change that it claims to emit should be populated (I guess it should probably return the state change that led to the object being in that state, but that's not what it does). This has led to bugs like #1598 and #1599.

I suggest that we remove it from the public API (i.e. the typings) in v2. (And if for some reason we do have to keep it, I'd suggest that we add it to the feature spec, properly defining what value it should emit.)

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4126) by [Unito](https://www.unito.io)

lawrence-forooghian:
Updates the `whenState` methods to conform to https://github.com/ably/specification/pull/183. See commit messages for more details.

Resolves #1597.
https://ably.atlassian.net/browse/ECO-4662

### DIFF
--- a/src/common/lib/util/eventemitter.ts
+++ b/src/common/lib/util/eventemitter.ts
@@ -286,19 +286,16 @@ class EventEmitter {
   }
 
   /**
-   * Private API
-   *
    * Listen for a single occurrence of a state event and fire immediately if currentState matches targetState
    * @param targetState the name of the state event to listen to
    * @param currentState the name of the current state of this object
-   * @param returnValue
    */
-  async whenState(targetState: string, currentState: string, returnValue?: unknown) {
+  async whenState(targetState: string, currentState: string) {
     if (typeof targetState !== 'string' || typeof currentState !== 'string') {
       throw new Error('whenState requires a valid state String argument');
     }
     if (targetState === currentState) {
-      return returnValue;
+      return null;
     } else {
       return this.once(targetState);
     }
--- a/src/common/lib/util/eventemitter.ts
+++ b/src/common/lib/util/eventemitter.ts
@@ -291,27 +291,16 @@ class EventEmitter {
    * Listen for a single occurrence of a state event and fire immediately if currentState matches targetState
    * @param targetState the name of the state event to listen to
    * @param currentState the name of the current state of this object
-   * @param listener the listener to be called
-   * @param listenerArgs
+   * @param returnValue
    */
-  whenState(targetState: string, currentState: string, listener: Function, ...listenerArgs: unknown[]) {
-    const eventThis = { event: targetState };
-
+  async whenState(targetState: string, currentState: string, returnValue?: unknown) {
     if (typeof targetState !== 'string' || typeof currentState !== 'string') {
       throw new Error('whenState requires a valid state String argument');
     }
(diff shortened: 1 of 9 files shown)


src/common/lib/util/eventemitter.ts
```diff
@@ -286,32 +286,18 @@ class EventEmitter {
...
    * @param currentState the name of the current state of this object
-   * @param listener the listener to be called
-   * @param listenerArgs
    */
-  whenState(targetState: string, currentState: string, listener: Function, ...listenerArgs: unknown[]) {
-    const eventThis = { event: targetState };
-
+  async whenState(targetState: string, currentState: string) {
     if (typeof targetState !== 'string' || typeof currentState !== 'string') {
       throw 'whenState requires a valid event String argument';
```
VeskeR, line 295: should change to `throw new Error(...)` instead of throwing a string.

also message is a bit misleading using the term `event` here.
maybe like this is better:
`whenState requires a valid state String argument`

src/common/lib/util/eventemitter.ts
```diff
@@ -286,32 +286,18 @@ class EventEmitter {
...
    * Listen for a single occurrence of a state event and fire immediately if currentState matches targetState
    * @param targetState the name of the state event to listen to
    * @param currentState the name of the current state of this object
-   * @param listener the listener to be called
-   * @param listenerArgs
    */
-  whenState(targetState: string, currentState: string, listener: Function, ...listenerArgs: unknown[]) {
-    const eventThis = { event: targetState };
-
+  async whenState(targetState: string, currentState: string) {
```
VeskeR, line 293: After getting familiar with `whenState` API it doesn't feel right to me to have it as part of the EventEmitter class.
It just has too much business logic: why EventEmitter knows about the `state`; knows that state can change and knows what to do when target and current state are the same.

Not sure how we would change it though, only duplicating this code both in Connection and RealtimeChannel comes to mind and maybe at least adding the definition to the separate interface and making sure that Connection and RealtimeChannel implement that interface.

Probably can do it in a separate PR or just create a refactoring issue for now.

VeskeR (changes requested):

src/common/lib/util/eventemitter.ts
lawrence-forooghian: Good catch, was like that already but will change.

src/common/lib/util/eventemitter.ts
lawrence-forooghian: Yeah, agree that `EventEmitter` isn't really the right place for it. Given that it's only an implementation detail (i.e. isn't in the public `EventEmitter` API), I don't think that it's very urgent to sort out. Created https://github.com/ably/ably-js/issues/1669 for it.

VeskeR (approved):

PR #1641, opened 2024-02-27
Remove Node 12 and es5 code that was supporting older platforms in v2

VeskeR:
This PR is based on https://github.com/ably/ably-js/pull/1633 so review that one first.

Removes more code that's supporting older platforms. These commits couldn't be included in https://github.com/ably/ably-js/pull/1629, since it would break CI for them due to it still being built and tested on es5 and node 12.


webpack.config.js
```diff
@@ -124,27 +124,13 @@ const reactNativeConfig = {
  *
  *   1. This package is designed for Node only and hence requires polyfills of Node libraries (e.g. `stream`, `path`) — webpack takes care of this for us.
  *   2. The package is not compatible with RequireJS and hence we don’t have any easy way to directly load it in our tests — the webpack bundle exposes it as a global named MochaJUnitReporter.
- *
- * - Node: The library uses optional chaining syntax, which is not supported by Node 12.
```
lawrence-forooghian, line 128: Do we need a Node bundle at all any more, or can `test/support/mocha_reporter.js` now just directly `require('mocha_unit_reporter')`? In removing the Node webpack config I think we can also remove the `webpack-node-externals` dependency introduced in dd7c83ea910c580a4ee20f72cf19babd59351b76.

webpack.config.js
```diff
@@ -124,27 +124,13 @@ const reactNativeConfig = {
...
 
   const baseConfig = {
     mode: 'development',
     entry: path.join(dir, 'index.js'),
-    module: {
-      rules: [
-        {
-          // The optional chaining syntax used by mocha-junit-reporter is not supported by Node 12 (see above). Hence, we transpile using Babel.
-          test: /\.js$/,
-          loader: 'babel-loader',
```
lawrence-forooghian, line 141: I don't think we're using `babel-loader` or `@babel/preset-env` any more, can we remove these dependencies?

VeskeR: True, will remove those, thanks!

webpack.config.js
VeskeR: Will check

VeskeR:
Removed `babel-loader`, `@babel/preset-env`, `webpack-node-externals` and Node.js bundle for MochaJUnitReporter.

webpack.config.js
```diff
@@ -124,27 +123,13 @@ const reactNativeConfig = {
  *
  *   1. This package is designed for Node only and hence requires polyfills of Node libraries (e.g. `stream`, `path`) — webpack takes care of this for us.
  *   2. The package is not compatible with RequireJS and hence we don’t have any easy way to directly load it in our tests — the webpack bundle exposes it as a global named MochaJUnitReporter.
- *
- * - Node: The library uses optional chaining syntax, which is not supported by Node 12.
  */
 function createMochaJUnitReporterConfigs() {
```
lawrence-forooghian, line 127: This function is still written as if supporting multiple configs. I think that the documentation comment, function name, and return value can all be simplified now that this is only for browser.

VeskeR: True, simplified

lawrence-forooghian (approved):

PR #1644, opened 2024-02-27
consumergroup: use modulo-based hashing scheme

mschristensen:
The `hashring` package is node-only as it depends on the native `crypto` package. Replaced with a simple modulo hash scheme for now.

Fixes the case where the channel is already attached and the channel is obtained with new rewind options that require a re-attach.

Updates the consumer group partitioning test to more robustly assert that channels are partitioned across consumers.

### DIFF
--- a/src/common/lib/client/baserealtime.ts
+++ b/src/common/lib/client/baserealtime.ts
@@ -13,7 +13,7 @@ import { ModulesMap, RealtimePresenceModule } from './modulesmap';
 import { TransportNames } from 'common/constants/TransportName';
 import Platform, { TransportImplementations } from 'common/platform';
 import { VcdiffDecoder } from '../types/message';
-import HashRing from 'hashring';
+import HashRing from '../util/hashring';
 
 /**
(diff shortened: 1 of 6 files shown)


mschristensen:
There are still some races and other failures due to use of presence, will work on this in another PR.

src/common/lib/client/baserealtime.ts
```diff
@@ -231,8 +255,9 @@ class ChannelGroup {
     });
   }
 
-  subscribe(cb: (channel: string, msg: any) => void) {
-    this.subsciptions.on('message', cb);
+  async subscribe(cb: (channel: string, msg: any) => void): Promise<void> {
+    await this.join();
+    this.subscriptions.on('message', cb);
```
zknill, line 260: If you subscribe twice then you'll join twice, and we will have duplicate processing of the presence set?

mschristensen: Presence set is deduplicated, so I think this is ok

zknill (approved):

PR #1645, opened 2024-02-27
feat: no upgrade

owenpearson:
Jira links:
- https://ably.atlassian.net/browse/REA-1752

removes XHRStreaming transport, replaces upgrade mechanism with a websocket + base fallback transport mechanism whereby no two transports may be simultaneously connected.


src/common/lib/transport/connectionmanager.ts
```diff
@@ -1508,129 +1261,160 @@ class ConnectionManager extends EventEmitter {
...
+      }
+      if (this.realtime.http.checkConnectivity) {
+        Utils.whenPromiseSettles(this.realtime.http.checkConnectivity(), (err, connectivity) => {
+          if (err || !connectivity) {
+            Logger.logAction(
+              Logger.LOG_MAJOR,
+              'ConnectionManager WebSocket slow timer',
+              'http connectivity check failed'
+            );
+            this.wsCheckResult = false;
```
SimonWoolf, line 1322: Not sure I see why're we'd be setting wsCheckResult to false here?

src/common/lib/transport/connectionmanager.ts
```diff
@@ -1508,129 +1261,160 @@ class ConnectionManager extends EventEmitter {
...
-    } else if (this.transports.length > 1 && this.getTransportPreference()) {
-      this.connectPreference(transportParams, connectCount);
+      return;
+    }
+
+    const transportPreference = this.getTransportPreference();
+
+    if (
+      (transportPreference && transportPreference === this.baseTransport) ||
+      (this.baseTransport && !this.webSocketTransport)
```
SimonWoolf, line 1284: I don't think it's right that we're now just never unpersisting the transport preference. If you ever go to a corporate network that breaks websockets, and you end up with an xhr_polling persisted preference, that's it's no more websockets for you ever, in this current instance of the lib or any future ones that share its localstorage.

Not sure what the best solution is. maybe something like, in `connectImpl()`, if the transport preference is the base transport but the ws transport is theoretically available, kick off a immediate ws connectivity check in parallel with the base connection sequence, that removes the persisted transport preference (and unsets `wsCheckResult`) if it succeeds?

src/common/lib/transport/connectionmanager.ts
```diff
@@ -1508,129 +1261,160 @@ class ConnectionManager extends EventEmitter {
...
-         * onconnected listener and get rid of it */
-        transport.off();
-        transport.disconnect();
-        Utils.arrDeleteValue(this.pendingTransports, transport);
-      } else if (!transport && !fatal) {
-        /* Preference failed in a transport-specific way. Try more */
-        this.unpersistTransportPreference();
-        this.connectImpl(transportParams, connectCount);
+        if (this.baseTransport) {
+          this.connectBase(transportParams);
```
SimonWoolf, line 1358: What happens if the websocket succeeds after we kick off connectBase? I guess whatever one succeeds second will just activate over the top of the first one? that doesn't seem ideal.

Why not cancel it here so you avoid the chance of both transports succeeding? I guess the answer is that you've removed the proposedTrasnports mechanism so you don't actually have a way to cancel it, what's the reasoning behind doing that?

src/common/lib/util/defaults.ts
```diff
@@ -28,7 +28,6 @@ type CompleteDefaults = IDefaults & {
     realtimeRequestTimeout: number;
     recvTimeout: number;
     preferenceConnectTimeout: number;
```
SimonWoolf, line 30: Doesn't really make sense to still have a `preferenceConnectTimeout` any more -- can we remove that, and add the two new timeouts you've added (the 10s and 4s) to the (undocumented part of the) clientOptions?

src/common/lib/transport/connectionmanager.ts
```diff
@@ -1508,129 +1261,160 @@ class ConnectionManager extends EventEmitter {
...
-   * - base: we try to connect with the best transport that we think will
-   *   never fail for this platform. If it doesn't work, we try fallback hosts.
-   * - upgrade: given a connected transport, we see if there are any better
-   *   ones, and if so, try to upgrade to them.
-   *
-   * connectImpl works out what stage you're at (which is purely a function of
-   * the current connection state and whether there are any stored preferences),
-   * and dispatches accordingly. After a transport has been set pending,
-   * tryATransport calls connectImpl to see if there's another stage to be done.
+   * connectImpl works out which transport to initially attempt a connection with
```
SimonWoolf, line 1264: The previous version documented how the connection sequence worked here, can we not do the same? it's not necessarily obvious

SimonWoolf (changes requested):
Good start, some feedback

various bits of thely added state doesn't ever seem to be reset, eg
- this.abandonedWebSocket can be set to true by the wsGiveUpTimer; if that ever happens it stays true forever, which has the effect that the library will never again try a fallback host with the websocket transport
- this.wsCheckResult is evaluated by the websocketSlowTimer if it's null; once it's been set to a bool it's never reevaluated. (except it's set to false if the http connectivity check fails, but I don't understand why)

ably-js lib instances can be long-lived indeed, we should mae sure they don't get in permanent suboptimal states

src/common/lib/transport/connectionmanager.ts
owenpearson: Was thinking about this, if we perform a ws connectivity check in `connectImpl` whenever there's a base transport preference and the connectivity check succeeds before the base transport is connected, should we just abandon the base transport attempt and try websocket immediately, or just unpersist the transport preference for next time? I'm leaning towards the former

src/common/lib/transport/connectionmanager.ts
```diff
@@ -204,18 +215,17 @@ class ConnectionManager extends EventEmitter {
   connectionStateTtl: number;
   maxIdleInterval: number | null;
   transports: TransportName[];
-  baseTransport: TransportName;
-  upgradeTransports: TransportName[];
+  baseTransport?: TransportName;
+  webSocketTransport?: TransportCtor;
```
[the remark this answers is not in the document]
owenpearson: yeah fair point, i think this is just an artefact from previous cm logic which i hadn't cleaned up. I've changed the field to a bool `webSocketTransportAvailable`.

src/common/lib/transport/connectionmanager.ts
```diff
@@ -1456,14 +1223,6 @@ class ConnectionManager extends EventEmitter {
 
     const auth = this.realtime.auth;
 
-    /* The point of the connectCounter mechanism is to ensure that the
-     * connection procedure can be cancelled. We want disconnectAllTransports
-     * to be able to stop any in-progress connection, even before it gets to
-     * the stage of having a pending (or even a proposed) transport that it can
-     * dispose() of. So we check that it's still current after any async stage,
-     * up until the stage that is synchronous with instantiating a transport */
-    const connectCount = ++this.connectCounter;
```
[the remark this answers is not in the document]
owenpearson: I thought so, but yeah from addressing your other comments I can see transport attempts getting stuck in the fallback host loop without being able to cancel them so I've restored the connectCounter to deal with that

src/common/lib/transport/connectionmanager.ts
owenpearson: I've re-implemented `connectCounter` and added a `proposedTransport` (which, if set, is disposed when we fallback from ws to base) so this should be solid now

src/common/lib/transport/connectionmanager.ts
owenpearson: added some detailed docstrings to `connectImpl` and `connectWs`

src/common/lib/transport/websockettransport.ts
```diff
@@ -207,12 +207,26 @@ class WebSocketTransport extends Transport {
       });
     }
   }
-}
 
-function initialiseTransport(transportStorage: TransportStorage): typeof WebSocketTransport {
-  if (WebSocketTransport.isAvailable()) transportStorage.supportedTransports[shortName] = WebSocketTransport;
+  async checkConnectivity() {
+    const ws = this.createWebSocket('wss://ws-up.ably-realtime-nonprod.com/', {});
```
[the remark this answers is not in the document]
owenpearson: good spot! i was originally using this endpoint for testing but i've changed it to the prod endpoint now

src/common/lib/transport/connectionmanager.ts
```diff
@@ -1511,130 +1335,132 @@ class ConnectionManager extends EventEmitter {
...
+          this.unpersistTransportPreference();
+          if (this.state === this.states.connecting) {
+            Logger.logAction(
+              Logger.LOG_MINOR,
+              'ConnectionManager.connectImpl():',
+              'web socket connectivity available, cancelling connection attempt with ' + this.baseTransport
+            );
+            if (this.proposedTransport) {
+              this.proposedTransport.dispose();
+            }
```
SimonWoolf, line 1383: also need to disconnect any pending transport here

actually probably easiest to just call disconnectAllTransports() (and that method should dispose of the proposedTransport if there is one, you forgot to add that back in)

src/common/lib/transport/connectionmanager.ts
```diff
@@ -1511,130 +1335,132 @@ class ConnectionManager extends EventEmitter {
...
+            this.connectWs(transportParams, this.connectCounter++);
```
SimonWoolf, line 1384: > `this.connectWs(transportParams, this.connectCounter++);`

this looks dodgy. Remember that `x++` increments x and returns the _old_ value -- I haven't had to think about the connect counter in a while but that seems unlikely to be what you intended

src/common/lib/transport/connectionmanager.ts
```diff
@@ -1297,6 +1038,86 @@ class ConnectionManager extends EventEmitter {
...
+        Logger.logAction(
+          Logger.LOG_MINOR,
+          'ConnectionManager WebSocket give up timer',
+          'websocket connection took more than 10s; ' + (this.baseTransport ? 'trying base transport' : '')
+        );
+        // if we don't have a base transport to fallback to, just let the websocket connection attempt time out
+        if (this.baseTransport) {
+          this.proposedTransport?.dispose();
+          this.pendingTransport?.dispose();
+          this.connectBase(transportParams, this.connectCounter++);
```
SimonWoolf, line 1108: (similar comment to above)

SimonWoolf (commented):
@owenpearson couple new comments, and the feedback in https://github.com/ably/ably-js/pull/1645#pullrequestreview-1914330752 doesn't seem to have been addressed yet?

src/common/lib/transport/connectionmanager.ts
owenpearson: oh good point, i've changed both to pre-increment now

owenpearson:
@SimonWoolf I've addressed the feedback in https://github.com/ably/ably-js/pull/1645#pullrequestreview-1914330752 by setting the value of `abandonedWebSocket` and `webSocketCheckResult` upon successful websocket connectivity result in `connectImpl`

src/common/lib/transport/connectionmanager.ts
lawrence-forooghian: I haven't been involved in this PR at all so sorry for the drive-by comment, but I'd suggest that this kind of confusion about the subtle distinctions between these operators means that they're perhaps best avoided, and we could just have an extra line of code to do the increment?

lawrence-forooghian: (They [were removed in Swift](https://github.com/apple/swift-evolution/blob/main/proposals/0004-remove-pre-post-inc-decrement.md) for this reason for example)

SimonWoolf:
> I've addressed the feedback in https://github.com/ably/ably-js/pull/1645#pullrequestreview-1914330752 by setting the value of abandonedWebSocket and webSocketCheckResult upon successful websocket connectivity result in connectImpl

that's within a `transportPreference === this.baseTransport` check -- what if we don't have a baseTransport, eg someone's starting the lib with `transports: ['web_socket']`? then we still set abandonedWeboscket=true if the giveUpTimer expires, but nothing ever sets it to false again?

owenpearson:
@SimonWoolf Good point, I've made it so that `abandonedWebSocket` is only set to true if there's a base transport to fall back to.

lawrence-forooghian:
@owenpearson just a heads up that depending on which PR gets merged first you might need to also update the migration guide (see https://github.com/ably/ably-js/pull/1670)

SimonWoolf (approved):

PR #1651, opened 2024-02-28
Feature/rea 1808 unsubscribe from channel group

mschristensen:


src/common/lib/client/baserealtime.ts
```diff
@@ -142,6 +143,33 @@ class ConsumerGroup extends EventEmitter {
...
+      }
+      this.channel.presence.unsubscribe(this.computeMembershipListener);
+      await this.channel.presence.leave(null);
+      await this.channel.detach();
+      this.channel = undefined;
+    } catch (err) {
+      Logger.logAction(
+        Logger.LOG_ERROR,
+        'ConsumerGroup.leave()',
+        'failed to leave presence set on consumer group channel:' + err
```
VeskeR, line 168: I think you usually do `Utils.inspectError(err)` if you want to append err to log message.
Might want to change other places too in another PR (saw it in ChannelGroup.addSubscriptions too).

src/common/lib/client/baserealtime.ts
```diff
@@ -281,6 +323,13 @@ class ChannelGroup {
       }
 
       this.subscribedChannels[channel].unsubscribe();
+      this.subscribedChannels[channel].detach().catch((err) => {
+        Logger.logAction(
+          Logger.LOG_ERROR,
+          'ChannelGroups.removeSubscriptions()',
+          'failed to detach from channel ' + channel + ': ' + err
```
VeskeR, line 330: ditto

VeskeR (changes requested):
On the surface level LGTM.
Only a comment about error logging

VeskeR (approved):

PR #1652, opened 2024-02-29
test: test ux improvements

owenpearson:
Includes a couple of improvements for the tests in this project:
- by default, outputs debug logs for failing tests and no logs for passing tests, making the test results much easier to read and allowing us to see full client logs for test flakes in CI (especially helpful if we aren't able to reproduce the flake locally). you can still configure all of the tests to emit logs to stdout by explicitly setting a log level (using `ABLY_LOG_LEVEL` in nodejs or the `log_level` query param in browser).
- use source map for node tests (no more stack traces pointing to 'ably-node:8392')
- uses the mocha CLI to run node tests, avoiding the grunt-mocha task. this makes it easier to leverage mocha's CLI options and means that the output works with colorterm, see screenshots:

example test stdout with colorterm:
image
example test failure output:
image


test/tasks/grunt-mocha.js
```diff
@@ -1,70 +0,0 @@
...
-
-  grunt.registerTask('mocha:webserver', 'Run the Mocha web server', function () {
-    kexec('test/web_server');
-  });
-
-  grunt.registerTask(
-    'mocha',
-    'Run the Mocha test suite.\nOptions:\n  --file=<file> e.g. --file=test/rest/auth.test.js\n  --debug will debug using standard node debugger\n  --inspector will start with node inspector',
-    function () {
-      var runTests = getRelativePath(helpers).concat(['test/realtime/*.test.js', 'test/rest/*.test.js']).join(' ');
```
VeskeR, line 37: previously we were providing a list of test files explicitly here: `test/realtime/*.test.js`, `test/rest/*.test.js`.
now we don't and `mocha` loads `test/all.js` file which imports `rest` and `realtime` tests. I assume `all.js` wasn't used at all while we had `grunt mocha` task.

and then, it seems like we can remove `all.js` too. we already require `modules_helper` and `test_helper` modules via `.mocharc.js` config and we can add `spec` prop to that to specify `test/realtime/*.test.js` and `test/rest/*.test.js` test files.

@owenpearson WDYT?

VeskeR (changes requested):

VeskeR (commented):
Overall LGTM, and these are very welcome changes! Definitely will improve dev experience working with tests going forward.

test/tasks/grunt-mocha.js
VeskeR: and for browsers it seems like everything is loaded using `test/support/browser_setup.js` script with data from `test/support/browser_file_list.js`, so it also doesn't use `all.js`.

Gruntfile.js
```diff
@@ -152,8 +157,6 @@ module.exports = function (grunt) {
     });
   });
 
-  grunt.loadTasks('test/tasks');
-
   grunt.registerTask('test', ['test:node']);
```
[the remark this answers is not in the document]
owenpearson: yeah we weren't using the `test` task even before this PR, there's actually quite a few grunt tasks we aren't using so I'll add a commit to deal with those together

.mocharc.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  require: ['source-map-support/register', 'test/support/modules_helper.js', 'test/support/test_helper.js'],
```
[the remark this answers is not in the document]
owenpearson: good point! these hooks weren't being run in the browser at all so i made a commit to move them into shared_helper.js in another branch (https://github.com/ably/ably-js/pull/1645/commits/359c5143c169839594b22f68a015a227beb64184), i might just rebase this PR onto that one for the sake of simplicity rather than duplicating the commit

test/tasks/grunt-mocha.js
owenpearson: nice, yeah you were right that `all.js` wasn't being used anymore. and I've added a `spec` field to the mocha node config

.mocharc.js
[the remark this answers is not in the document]
owenpearson: cool, yeah i'm gonna wait for #1645 to land on the integration branch and rebase this PR (which will resolve this comment) so we can hold off merging until then

Gruntfile.js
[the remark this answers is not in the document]
owenpearson: fixed this and cleaned up some other unused grunt stuff in [a0013b1](https://github.com/ably/ably-js/pull/1652/commits/a0013b1541987d2a50f69454a375dc38dd2d1c37)

VeskeR:
@owenpearson one more thing, is there a reason why `test/web_server` file is not named `test/web_server.js` (with .js extension) ?
if we can rename it, I think this PR is just the place for it

owenpearson:
@VeskeR I guess it's because it's supposed to be used as an executable i guess (hence the hashbang and execute permissions). I don't really mind either way but I doubt anyone's actually using it like that nowadays though so happy to change if you'd prefer

owenpearson:
@VeskeR I've added [61e28de](https://github.com/ably/ably-js/pull/1652/commits/61e28de7e80ff981d0cce5d893c80b7ae981f626) to make the web_server executable a module

.mocharc.js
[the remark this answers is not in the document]
VeskeR: Oh, I see!
I've noticed those new `afterEach` hooks in shared_helper and was wondering if that was a replacement for root_hooks.js, and turns out it is.

Now, I believe root_hooks.js was never registered in browser tests, because even though it is listed in `browser_file_list.js`, in `browser_setup.js` there are TEST_REGEXP and TEAR_DOWN_REGEXP regexps that filter out anything that is not `*.test.js` or `*tear_down.js` -> `root_hooks.js` doesn't match. And there is no other place where `root_hooks.js` would be imported in browser setup, so I guess that was the culprit.

I feel like separate `root_hooks.js` file was a better place for hooks like these, they seem a bit out of place for `shared_helper.js`.

That being said, current browser test setup is a maze and I don't want to drag on this PR for much longer. I believe Unified Test Suite will be the opportunity for us to re-implement browser tests with more modern setup.
So we can leave this as it is currently in this PR

test/support/runPlaywrightTests.js
```diff
@@ -85,7 +85,7 @@ const runTests = async (browserType) => {
     caughtError = error;
   }
 
-  mochaWebServerProcess.kill();
+  mochaServer.kill();
```
VeskeR, line 88: I think it should be `mochaServer.close()`? To call this method https://github.com/ably/ably-js/pull/1652/files#diff-a020bd360991ca253146ff21e077ceeec55ff07e573e0a679b385c09b4a02f3fR58

Gruntfile.js
```diff
@@ -152,9 +133,6 @@ module.exports = function (grunt) {
     });
   });
 
-  grunt.loadTasks('test/tasks');
-
-  grunt.registerTask('test', ['test:node']);
   grunt.registerTask(
     'test:node',
```
VeskeR, line 137: this one is also replace by npm script `test:node`, which does the same.
plus, there is no more `grunt mocha` task (see `['build:node', 'mocha']` below), so this command will fail

Gruntfile.js
```diff
@@ -167,59 +145,6 @@ module.exports = function (grunt) {
...
-      grunt.task.run([
-        'bump-only:' + versionType,
-        'release:refresh-pkgVersion',
-        'all',
-        'release:git-add-generated',
-        'bump-commit',
-      ]);
-    },
-  );
-
```
VeskeR, line 222: so all `release` grunt tasks are not used anymore?
probably they are not, there is no mention of grunt in `Release Process` in `CONTRIBUTING.md`

VeskeR:
> @VeskeR I've added [61e28de](https://github.com/ably/ably-js/pull/1652/commits/61e28de7e80ff981d0cce5d893c80b7ae981f626) to make the web_server executable a module

Thank you!
And removal of `kexec` is a nice side-effect of this change, that package was not compatible with Windows at all.

VeskeR (changes requested):

Gruntfile.js
owenpearson: yeah these tasks were for an old release process where we added generated build files to the git index, the steps in CONTRIBUTING.md is what we use now

VeskeR: got it

test/web_server.js
```diff
@@ -0,0 +1,63 @@
...
+class MochaServer {
+  constructor(playwrightTest) {
+    this.playwrightTest = playwrightTest;
+  }
+
+  async listen() {
+    const server = express();
+    this.server = server;
+    server.use(function (req, res, next) {
+      if (!playwrightTest) console.log('%s %s %s', req.method, req.url, req.path);
```
VeskeR, line 23:
```suggestion
      if (!this.playwrightTest) console.log('%s %s %s', req.method, req.url, req.path);
```

test/web_server.js
```diff
@@ -0,0 +1,63 @@
...
+      next();
+    });
+
+    server.use(cors());
+
+    server.get('/', function (req, res) {
+      if (playwrightTest) {
```
VeskeR, line 30:
```suggestion
      if (this.playwrightTest) {
```

VeskeR (changes requested):

VeskeR (approved):

Issue #1630, opened 2024-02-23
Add documentation for `ChannelProvider`


ttypic:
We need to update readme to show how to use `ChannelProvider`

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4635) by [Unito](https://www.unito.io)

PR #1654, opened 2024-02-29
[ECO-4635] docs: add `ChannelProvider` documentation

ttypic:
Resolves https://github.com/ably/ably-js/issues/1630
Jira Issue [ECO-4635](https://ably.atlassian.net/browse/ECO-4635)

[ECO-4635]: https://ably.atlassian.net/browse/ECO-4635?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/react.md
+++ b/docs/react.md
@@ -55,7 +55,7 @@ root.render(
 
 ### Define Ably channels
 
-Once you've done this, define Ably channels you want to use by utilizing the `ChannelProvider` component:
+Once you've set up `AblyProvider`, define Ably channels you want to use by utilizing the `ChannelProvider` component:
 
 ```jsx
 <ChannelProvider channelName="your-channel-name">
@@ -101,8 +101,9 @@ return (
 ```
 
 > [!NOTE]
-> Please note that attempts to publish to a derived channel (the one created or retrieved with a filter expression) will fail. In order to send messages to the channel called _"your-derived-channel-name"_ from the example above, you will need to create another channel instance without a filter expression.
-
+> Please note that attempts to publish to a derived channel (the one created or retrieved with a filter expression)
+> using channel instance will fail, since derived channels support only `subscribe` capability.
+> Use `publish` function returned by `useChannel` hook instead.
 
 ---
 
--- a/docs/react.md
+++ b/docs/react.md
@@ -53,7 +53,17 @@ root.render(
 )
 ```
 
-Once you've done this, you can use the `hooks` in your code. The simplest example is as follows:
+### Define Ably channels
+
+Once you've done this, define Ably channels you want to use by utilizing the `ChannelProvider` component:
+
+```jsx
+<ChannelProvider channelName="your-channel-name">
+  <Component />
+</ChannelProvider>
+```
+
+After setting up `ChannelProvider`, you can employ the provided `hooks` in your code. Here's a basic example:
 
 ```javascript
 const { channel } = useChannel("your-channel-name", (message) => {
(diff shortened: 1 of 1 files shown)


owenpearson (approved):
Looks good!

docs/react.md
```diff
@@ -53,7 +53,17 @@ root.render(
 )
 ```
 
-Once you've done this, you can use the `hooks` in your code. The simplest example is as follows:
+### Define Ably channels
+
+Once you've done this, define Ably channels you want to use by utilizing the `ChannelProvider` component:
```
VeskeR, line 58:
```suggestion
Once you've set up `AblyProvider`, define Ably channels you want to use by utilizing the `ChannelProvider` component:
```

docs/react.md
```diff
@@ -63,7 +73,36 @@ const { channel } = useChannel("your-channel-name", (message) => {
...
+
+return (
+  <ChannelProvider channelName="your-derived-channel-name" options={{ ... }} deriveOptions={deriveOptions}>
+    <Component />
+  </ChannelProvider>
+)
+```
+
+> [!NOTE]
+> Please note that attempts to publish to a derived channel (the one created or retrieved with a filter expression) will fail. In order to send messages to the channel called _"your-derived-channel-name"_ from the example above, you will need to create another channel instance without a filter expression.
```
VeskeR, line 104: I think this step for a user would also mean using a separate `AblyProvider`, because derived channel and regular one will have the same name, so they can't be placed inside one `AblyProvider`.
Given that this is a bit advanced and non-obvious, maybe we should add an example for this too?

VeskeR (changes requested):

docs/react.md
owenpearson: Oh sorry I didn't pick up on this in my initial review. Out of interest, what's the reason for not being able to publish to a derived channel? That seems like quite a disappointing limitation to me 🙁

If there really is no way we can fix that then I think recommending rest publishing is probably better than recommending client side filtering since the whole reason people use filtered channels is so that they don't incur the cost of receiving every message on the channel.

ttypic: >Out of interest, what's the reason for not being able to publish to a derived channel?

@owenpearson I think that's a question for rt team :) derived channel has only subscribe capability right now

owenpearson: tbh this should work fine? haven't tested though
```ts
const { ably } = useChannel(...opts);
ably.channels.get(channelNameWithoutDeriveExpression).publish();
```

ttypic: cool I haven't known this, could you look at this PR https://github.com/ably/ably-js/pull/1658/files ? WDYT? I think it will be better if we handle this situation ourselves

owenpearson: that looks like an ideal solution :)

VeskeR: I assume we decided to go with transient realtime publishing instead of rest to keep what's user is exposed to to one API (Realtime in this case), so user doesn't have to figure out how Rest works, what's the difference and how to set it up in the application?

docs/react.md
```diff
@@ -63,7 +73,38 @@ const { channel } = useChannel("your-channel-name", (message) => {
...
+
+return (
+  <ChannelProvider channelName="your-derived-channel-name" options={{ ... }} deriveOptions={deriveOptions}>
+    <Component />
+  </ChannelProvider>
+)
+```
+
+> [!NOTE]
+> Please note that attempting to publish to a derived channel (one created or retrieved with a filter expression) will result in failure.
```
VeskeR, line 104: this paragraph now clashes with https://github.com/ably/ably-js/pull/1658.
I suggest we make this PR to be based on https://github.com/ably/ably-js/pull/1658 and figure out the correct wording for this part there.

VeskeR (changes requested):

VeskeR (approved):

PR #1655, opened 2024-02-29
Feature/rea 1808 use separate channel object

mschristensen:
Depends on https://github.com/ably/ably-js/pull/1651

### DIFF
--- a/test/realtime/channelgroup.test.js
+++ b/test/realtime/channelgroup.test.js
@@ -270,7 +270,7 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
     });
 
     // wait for n consumers to appear in the given channel's presence set
-    function waitForConsumers(channel, n) {
+    function waitForNConsumersInPresenceSet(channel, n) {
       return new Promise(async (resolve, reject) => {
         const interval = setInterval(async () => {
           try {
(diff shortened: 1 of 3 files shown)


test/realtime/channelgroup.test.js
```diff
@@ -531,5 +517,80 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
...
+      let realtime1 = helper.AblyRealtime({ clientId: 'testclient' });
+      let realtime2 = helper.AblyRealtime({ clientId: 'consumer' });
+
+      // connect
+      Promise.all([
+        new Promise((resolve) => realtime1.connection.on('connected', resolve)),
+        new Promise((resolve) => realtime2.connection.on('connected', resolve)),
+      ])
+        .then(async () => {
+          // create 2 consumers in one group
```
VeskeR, line 522: I think the comment is wrong here? there are no 2 consumers in this test

test/realtime/channelgroup.test.js
```diff
@@ -531,5 +517,80 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, helper, async
...
+          });
+          
+          // publish active channels
+          const channelName = `${prefix}:channel`;
+          let activeChannel = realtime1.channels.get(activeChannelName);
+          await activeChannel.attach();
+          activeChannel.publish('event0', { active: [channelName] });
+
+          // wait for the consumer to appear in the group
+          await waitForConsumers(realtime1.channels.get(consumerGroupName), 1);
```
VeskeR, line 546: out of interest, why do we expect a consumer to appear in **realtime1**, if we've created a consumer previously on **realtime2** `const consumer = realtime2.channelGroups.get`?

VeskeR: oh, I see, `waitForConsumers` actually checks for presence in channel, so it's cross-client. makes sense now.
maybe a slight name change to `waitForConsumers` would be good in a separate PR? to make it more clear for what it waits.

VeskeR (approved):
LGTM, some minor test comments, but they are not blockers really

PR #1658, opened 2024-03-01
[ECO-4666] feat: add `publish` to the `useChannel` hook

ttypic:
Using publish method you can send messages to the derived channels (channels with filter qualifier, without attaching to the channel or using other workarounds)

### DIFF
--- a/docs/react.md
+++ b/docs/react.md
@@ -8,7 +8,7 @@ Use Ably in your React application using idiomatic, easy to use, React Hooks!
 Using this module you can:
 
 - Interact with [Ably channels](https://ably.com/docs/channels) using a React Hook.
-- [Publish messages](https://ably.com/docs/channels#publish) via Ably using the channel instances the hooks provide
+- [Publish messages](https://ably.com/docs/channels#publish) via Ably using publish function the hooks provide
 - Get notifications of user [presence on channels](https://ably.com/docs/presence-occupancy/presence)
 - Send presence updates
 
@@ -94,20 +94,27 @@ const messagePreviews = messages.map((msg, index) => <li key={index}>{msg.data.s
 `useChannel` supports all of the parameter combinations of a regular call to `channel.subscribe`, so you can filter the messages you subscribe to by providing a `message type` to the `useChannel` function:
 
 ```javascript
-const { channel } = useChannel("your-channel-name", "test-message", (message) => {
+useChannel("your-channel-name", "test-message", (message) => {
     console.log(message); // Only logs messages sent using the `test-message` message type
 });
 ```
 
-The `channel` instance returned by `useChannel` can be used to send messages to the channel. It's just a regular Ably JavaScript SDK `channel` instance.
+#### useChannel `publish` function
+
+The `publish` function returned by `useChannel` can be used to send messages to the channel.
 
 ```javascript
-channel.publish("test-message", { text: "message text" });
+const { publish } = useChannel("your-channel-name")
+publish("test-message", { text: "message text" });
 ```
 
-Because we're returning the channel instance, and Ably SDK instance from our `useChannel` hook, you can subsequently use these to perform any operations you like on the channel.
+#### useChannel `channel` instance
 
-For example, you could retrieve history like this:
+The `useChannel` hook returns an instance of the channel, which is part of the Ably JavaScript SDK. This allows you to access the standard Ably JavaScript SDK functionalities associated with channels.
+
+By providing both the channel instance and the Ably SDK instance through our useChannel hook, you gain the flexibility to execute various operations on the channel.
+
+For instance, you can easily fetch the history of the channel using the following method:
 
 ```javascript
 const { channel } = useChannel("your-channel-name", (message) => {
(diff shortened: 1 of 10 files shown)


owenpearson (commented):
Implementation looks good! I'm guessing the test failures are just issues with the FakeAblySdk class rather than anything wrong with the lib itself

VeskeR:
I got confused a bit initially by the term `transient realtime publishing`, but it seems like it's just the same old `.publish` method called on a realtime channel, and it is considered `transient` if we never call attach/subscribe methods on a channel. Correct?

docs/react.md
```diff
@@ -94,15 +94,26 @@ const messagePreviews = messages.map((msg, index) => <li key={index}>{msg.data.s
...
+publish("test-message", { text: "message text" });
+```
+
+#### useChannel `channel instance`
+
 The `channel` instance returned by `useChannel` can be used to send messages to the channel. It's just a regular Ably JavaScript SDK `channel` instance.
 
 ```javascript
-channel.publish("test-message", { text: "message text" });
+const { channel } = useChannel("your-channel-name")
```
VeskeR, line 116: the actual publish call is now missing. basically, keep the old and the new line, like this:
```jsx
const { channel } = useChannel("your-channel-name")
channel.publish("test-message", { text: "message text" });
```

docs/react.md
```diff
@@ -94,15 +94,26 @@ const messagePreviews = messages.map((msg, index) => <li key={index}>{msg.data.s
...
+The `publish` function returned by `useChannel` can be used to send messages to the channel.
+
+```javascript
+const { publish } = useChannel("your-channel-name")
+publish("test-message", { text: "message text" });
+```
+
+#### useChannel `channel instance`
+
 The `channel` instance returned by `useChannel` can be used to send messages to the channel. It's just a regular Ably JavaScript SDK `channel` instance.
```
VeskeR, line 113: so we have two ways to publish a message now, correct?
via dedicated `publish` function and via `channel.publish`.

it might get a bit confusing for users to have two APIs that are so similar.

couple of ideas on how to improve it:
1. rename `publish` to `transientPublish` in `useChannel`. this is our intention here - to provide a user with a dedicated publish function that, as soon as you use **only** this function, will guarantee that you won't attach to a channel and won't incur any additional costs associated with receiving messages. plus a separate name might signal that this is not the same thing as calling `channel.publish`, so they are not interchangeable - proccing the user to read the docs more carefully and find the difference.
2. explain the difference between `publish` function and `channel.publish` right here, since this is the first time we see it. for example, on [line 104](https://github.com/ably/ably-js/pull/1658/files#diff-8bbfe4c553487b2bc9e53a0c20c04233b9ee8d4dfe72129e89731fb509f00999R104) above, add something like:

`publish` function is useful when you you're working with derived channels, which do not support calling `channel.publish`. More about [publishing to derived channels](make this a link to a new #### section for [line 153](https://github.com/ably/ably-js/pull/1658/files#diff-8bbfe4c553487b2bc9e53a0c20c04233b9ee8d4dfe72129e89731fb509f00999R153), call the section something like `Publishing to a derived channel`)

VeskeR, line 113:
```suggestion
The `channel` instance returned by `useChannel` can also be used to send messages to the channel. It's just a regular Ably JavaScript SDK `channel` instance.
```

docs/react.md
```diff
@@ -139,19 +150,7 @@ const { channel } = useChannel({ channelName: "your-derived-channel-name", optio
...
-const options = { ... };
-const deriveOptions = { filter: 'headers.email == `"rob.pike@domain.com"` || headers.company == `"domain"`' }
-const callback = (message) => { ... };
-
-const { channel: readOnlyChannelInstance } = useChannel({ channelName, options, deriveOptions }, callback);
-const { channel: readWriteChannelInstance } = useChannel({ channelName, options }, callback); // NB! No 'deriveOptions' passed here
-
-readWriteChannelInstance.publish("test-message", { text: "message text" });
-```
+Please note that attempts to publish to a derived channel (the one created or retrieved with a filter expression) using channel instance will fail. Use `publish` function from `useChannel` hook instead.
```
VeskeR, line 153:
```suggestion
Please note that attempts to publish to a derived channel (the one created or retrieved with a filter expression) using channel instance will fail, since derived channels support only `subscribe` capability. Use `publish` function returned by `useChannel` hook instead.
```

VeskeR: my intention to the change above is to explain **why** you can't publish to a derived channel - it simply does not support that.

also, is there a page somewhere on ably.com/docs that explains derived channels? I couldn't find one. if there is, it would be beneficial to put that link here, after the `since derived channels support only subscribe capability` message with text like "more about derived channels".

if there is no such page, then we can link at least Capabilities page: https://ably.com/docs/auth/capabilities, with the text "more about channel capabilities".

VeskeR (changes requested):

docs/react.md
VeskeR: on the second thought, `transientPublish` may also confuse people to thinking "wait, the 'regular' publish is not transient then? it subscribes to something?".
@owenpearson wdyt?

owenpearson: Yeah i think `transientPublish` will confuse users, especially since we don't have great documentation for it (try googling 'ably transient publishing' and you won't find much). Also bear in mind the vast majority of users won't be using derived channels so there will be no difference. I think we should keep the name `publish` and add something in the docstring + readme which explains that if you're using a derived channel the `publish` function will perform a transient publish.

ttypic: I agree with @owenpearson . @VeskeR,  `publish` = `channel.publish`, with single exception if channel is derived, then it uses transient publish (and I believe SDK user shouldn't even think about it).

ttypic: > so we have two ways to publish a message now, correct?
> via dedicated publish function and via channel.publish.


Yes and no in documentation should be only one way, using hook's `publish`. But we return channel instance, so if SDK user have custom functionality that depends on AblyRealtimeClient they can use it as is. But we generally don't recommend using `channel` instance.

ttypic: Updated documentation a bit, removed mentions of using `channel` instance to publish messages

docs/react.md
```diff
@@ -8,7 +8,7 @@ Use Ably in your React application using idiomatic, easy to use, React Hooks!
 Using this module you can:
 
 - Interact with [Ably channels](https://ably.com/docs/channels) using a React Hook.
-- [Publish messages](https://ably.com/docs/channels#publish) via Ably using the channel instances the hooks provide
+- [Publish messages](https://ably.com/docs/channels#publish) via Ably using publish function the hooks provide
```
owenpearson, line 11:
```suggestion
- [Publish messages](https://ably.com/docs/channels#publish) via Ably using the publish function the hooks provide
```

owenpearson (approved):

docs/react.md
VeskeR: > I think we should keep the name publish and add something in the docstring + readme which explains that if you're using a derived channel the publish function will perform a transient publish.

>  in documentation should be only one way, using hook's publish

sounds good to me, it solves the main issue of showing to users two separate APIs for publishing, which could've caused issues.

VeskeR (approved):

PR #1659, opened 2024-03-01
Feature/rea 1808 modules

mschristensen:


VeskeR:
I think this commit 51722812df901db252d351c1abe35feff2297016 ended up here by mistake.
It has the same message as https://github.com/ably/ably-js/pull/1655/commits/07c12a2c4ff7765fb9dd08c9d767e03c35f8a433, from the actual PR https://github.com/ably/ably-js/pull/1655 where we started using separate channel object.
Could you please squash to remove noise?

PR #1661, opened 2024-03-01
Experimental v2/channel and consumer group

mschristensen:


src/common/lib/util/hashring.ts
```diff
@@ -0,0 +1,44 @@
+// HashRing currently implements a modulo-based hash partitioning scheme.
+// It is used to distribute keys across a set of nodes.
+// TODO(mschristensen): use consistent hashing instead of modulo-based hashing
+// to avoid re-assigning all keys when nodes are added or removed.
+export default class HashRing {
```
ttypic, line 5: Small suggestion: maybe it's better to rename class? I believe name HashRing implies consistent hashing, but until you look at implementation or read TODO, you never know it's module-based hash

modules.d.ts
```diff
@@ -130,6 +131,25 @@ export declare const MsgPack: unknown;
...
+/**
+ * The module is experimental and the API is subject to change.
+ *
+ * Provides a {@link BaseRealtime} instance with the ability to interact with the experiemental channel groups feature.
+ *
+ * To create a client that includes this module, include it in the `ModulesMap` that you pass to the {@link BaseRealtime.constructor}:
+ *
+ * ```javascript
+ * import { BaseRealtime, WebSocketTransport, FetchRequest, RealtimePresence, ChannelGroups } from 'ably/modules';
+ * const realtime = new BaseRealtime(options, { WebSocketTransport, FetchRequest, RealtimePresence, ChannelGroups });
```
ttypic, line 143: I thought in this release we'll pack `ChannelGroups` in default web client, so SDK user won't have to declare WebSocketTransport, FetchRequest, RealtimePresence

ttypic (commented):
Implementation looks good, I am happy to release with experiment tag

modules.d.ts
VeskeR: it is part of DefaultRealtime (import from `ably`) - part of default web client bundle
https://github.com/ably/ably-js/pull/1661/files#diff-5dd5bf8949d722411724b640db79f4d05ef91185d5b5c6e26af4b642520d2839R41

this are just typings for the modular variant

VeskeR (changes requested):
LGTM overall, small comment about polyfill.

Test checks are looking weird though:
All integration tests are failing with an abnormal time (100m - 300m) - can't even load logs from them, there are just so many after running for so long.
Maybe there is a flaky/failed test that is just stuck in a loop or doesn't have a timeout set or something.
Should probably make sure tests are passing before releasing

src/common/lib/util/utils.ts
```diff
@@ -174,6 +174,48 @@ export function arrSubtract<T>(arr1: Array<T>, arr2: Array<T>): Array<T> {
   return result;
 }
 
+export const arrIndexOf = (Array.prototype.indexOf as unknown)
```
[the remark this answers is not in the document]
mschristensen: Woops, yes this came back in a bad merge. Thanks, fixed.

mschristensen:
@VeskeR I pushed up some changes to tests. They pass reliably for me locally so not sure why we're getting issues in CI. Will wait for the run to finish and inspect the logs

VeskeR:
Still failed after 100+ minutes unfortunately. Hope you will be able to load some logs, there are a lot.

Quickly looked at previous PRs:
seems like it started happening with this PR and onwards: https://github.com/ably/ably-js/pull/1655 (see its CI result https://github.com/ably/ably-js/actions/runs/8109535381/job/22164928744). Probably something was added to the tests that is causing an infinite loop of some sorts.

PRs before that had their tests passing in CI. For example, https://github.com/ably/ably-js/pull/1651 precedes the PR above and its tests either pass, or fail in a reasonable time (~10 minutes)

mschristensen:
From the logs, there's a lot of 
```
2024-03-05T13:47:13.6564779Z 13:47:13.656 Ably: Connection state: failed; reason: [ErrorInfo: account restricted (connection limit exceeded); statusCode=401; code=40111; see https://help.ably.io/error/40111 ]
```

So I have bumped the default connection limits for the test account from 100 to 200, and will see what effect that has.

mschristensen:
Actually, looks like there's a lot of

```
3-06T10:36:33.3326370Z 10:36:33.332 Ably: Transport.onProtocolMessage(): received on web_socket: [ProtocolMessage; action=DISCONNECTED; error=[ErrorInfo: Key/token status changed (expire); statusCode=401; code=40142; see https://help.ably.io/error/40142 ]]; connectionId = IK5jju0wpn
2024-03-06T10:36:33.3330234Z 10:36:33.332 Ably: Transport.onDisconnect(): err = [ErrorInfo: Key/token status changed (expire); statusCode=401; code=40142; see https://help.ably.io/error/40142 ]
```

Which is causing a large number of connections to be terminated and reopened. Looking into it

mschristensen:
@ttypic @VeskeR Tests are now passing. The problem was due to some shared auth state causing conflicts between simultaneous usage of the two `BaseRealtime` instances. This was causing a large amount of connection churn due to auth issues. Either way, having separate `ConnectionManager` instances meant double the volume of connections even when not using channel groups. So, I've made changes to use the same `Channels` object from the `ChannelGroup` as well as the base client and prevent conflicts using a dummy channel option so that the library treats channels created by the channel group independently.

test/common/modules/shared_helper.js
```diff
@@ -44,6 +44,20 @@ define([
 
   async function monitorConnectionAsync(action, realtime, states) {
     const monitoringResultPromise = new Promise((resolve, reject) => {
+      if (Object.prototype.toString.call(realtime) == '[object Array]') {
```
ttypic, line 47: May be `Array.isArray(realtime)` here ?

src/common/lib/util/utils.ts
```diff
@@ -174,6 +174,33 @@ export function arrSubtract<T>(arr1: Array<T>, arr2: Array<T>): Array<T> {
...
+
+export function arrUnique<T>(arr: Array<T>): Array<T> {
+  return arr.filter((value, index) => arr.indexOf(value) === index);
+}
+
+export function arrUniqueBy<T>(arr: Array<T>, iteratee: (value: T) => unknown): Array<T> {
+  return arr.filter((value, index) => arrIndexOfBy(arr, iteratee, iteratee(value)) === index);
+}
+
+export function arrIn(arr: Array<unknown>, val: unknown): boolean {
```
ttypic, line 200: I think you can safely use `includes` instead of `arrIn`

ttypic (commented):
Looks good, left two small comments

test/common/modules/shared_helper.js
mschristensen: Yes probably better, I was following the pattern used in `closeAndFinish`

modular.d.ts
```diff
@@ -131,6 +132,25 @@ export declare const MsgPack: unknown;
  */
 export declare const RealtimePresence: unknown;
 
+/**
+ * The module is experimental and the API is subject to change.
```
VeskeR, line 136: we reverted the "modules" mechanism for modular variant of the library to use "plugins" again as it was in v1.
see: https://github.com/ably/ably-js/pull/1663
mainly it means that instead of "module" we call them "plugin", and to use a plugin in your modular variant you do:
```typescript
const realtime = new BaseRealtime({ ...options, plugins: { ChannelGroups } });
```
instead of
```typescript
const realtime = new BaseRealtime(options, { ChannelGroups });
```

you can take a look at the linked PR and jsdocs for other plugins in this file to find the documentation differences.
Basically, we just need to change documentation here.

modular.d.ts
```diff
@@ -240,6 +260,11 @@ export interface ModularPlugins {
    */
   RealtimePresence?: typeof RealtimePresence;
 
+  /**
+   * See {@link ChannelGroups | documentation for the `ChannelGroups` module}.
```
VeskeR, line 264:
```suggestion
   * See {@link ChannelGroups | documentation for the `ChannelGroups` plugin}.
```

src/common/lib/util/utils.ts
VeskeR: yes, `arrIn` polyfill were removed in https://github.com/ably/ably-js/pull/1629/commits/e5961ead813a6435988679f22911ec232b41ef9e

test/realtime/channelgroup.test.js
```diff
@@ -0,0 +1,701 @@
+'use strict';
```
VeskeR, line 1: I realized that test files are not automatically picked up by the playwright browser tests as they are for node.
If you check, for example, [this CI run for webkit](https://github.com/ably/ably-js/actions/runs/8177130978/job/22358085711) in this PR you can see there are no `realtime/channelgroup` tests completed.

It seems that we need to add new test file names to `test/support/browser_file_list.js` for it to be included in browser tests. I hope it's the only place we need to change for that.

You can check locally if it works if you do `describe.only('realtime/channelgroup'` below, and run `PLAYWRIGHT_BROWSER=chromium npm run test:playwright` (might need to do `npx playwright install --with-deps` before that to install browsers)

VeskeR (changes requested):

modular.d.ts
mschristensen: Yeah I updated the implementation, just need to update these docs :)

Issue #1330, opened 2023-06-08
Brainstorming design for future of browser transport upgrade in ably-js


SimonWoolf:
An issue for discussing and refining what the transport sequence should be in the next major version of ably-js.

One option is just https://github.com/ably/ably-js/issues/1327 -- starting the lib with websocket as a preferredTransport, and going back to the existing comet with fallback hosts + websocket upgrade attempt once connected  strategy that we use now if that doesn't connect within some small timeout.

Another is getting rid of transport upgrade and moving to a pure fallback model, where you stick with whatever transport you first succeeded with. One advantage of this is simplicity, both of the sdk connectionmanager code and at the conceptual level (upgrade now generally works pretty solidly but it has undeniably been a past source of bugs).

As previously discussed, we think that for this to mesh with fallback hosts without adding too long a delay before connection in the event of either an issue with the closed datacenter or a (corporate MITM'd) network which is screwing with websockets, we should have a websocket version of the 'is the internet up' check, maybe using cloudflare workers or something, that we can run in parallel with the existing cdn connectivity check.

So a possible sequence (not set in stone, just a starting point for discussion) might be something like:

1. connection attempt to primary host over websocket
1. if not yet succeeded after 4s, start parallel XHR and websocket connectivity checks
1. if websocket succeeds, continue pure-websocket host-fallback sequence (if connection attempt to primary host hasn't succeeded after 10s, cancel and try the first fallback host)
1. if websocket connectivity check has not succeeded after 6s (so 10s after entire connect sequence started) but the xhr one has, cancel the ws attempt and start an xhr polling attempt to the primary host. If that fails, continue the host fallback sequence sticking with xhr polling. Persist xhr polling as the preferred transport for all future connect attempts -- but also occasionally retry the websocket connectivity check (eg if the network changes); if that ever succeeds clear the persisted xhr transport preference (but otherwise take no immediate action and continue with xhr).
1. if neither check succeeds, just go into the `disconnected` state and wait 15s to retry, as now.

One question is, what role does this leave for xhr_streaming transport? I suggest: none. Any MITM'd corporate network fucky enough to block / not understand websockets is also quite likely to block / not understand xhr streaming, as we have seen in the past.

cc owenpearson @lawrence-forooghian  paddybyers

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4161) by [Unito](https://www.unito.io)

paddybyers:
[...]

PR #1670, opened 2024-03-07
[ECO-1416] Add migration guide for v2

lawrence-forooghian:
WIP in that I’m sure there will be some editing needed, but it’s ready for some eyes on it. I’m also adding @m-hulbert to take a look, since this is an important piece of user-facing documentation.

There's a couple of uninteresting formatting commits, and then the migration guide itself is in the final commit; see that commit message for more details.

Resolves ECO-1416.


docs/migration-guides/v2/lib.md
```diff
@@ -0,0 +1,284 @@
...
+
+- If you’re implicitly choosing the callbacks variant of the library by writing `require('ably')`, then update your code to `require('ably/promises')`.
+- If you’re explicitly choosing the callbacks variant through the subpath of the imported module — that is, if you’re writing `require('ably/callbacks')` — then update your code to `require('ably/promises')`.
+- If you’re explicitly choosing the callbacks variant at the moment of instantiating the client — that is, if you’re writing `new Ably.Rest.Callbacks(…)` or `new Ably.Realtime.Callbacks(…)` — then update your code to use the `Promise` property instead, i.e. write `new Ably.Realtime.Promise(…)` or `new Ably.Rest.Promise(…)`.
+
+Now, update your code to use ably-js’s promise-based API instead of the callback-based API. **(TODO what would be good guidance on how to do this? Is it just "look at the v1 documentation for the promises variant and figure it out, here's a link", or an example of a single method call plus some hand-wavey instructions, or something else?)**
```
lawrence-forooghian, line 64: Thoughts?

docs/migration-guides/v2/lib.md
```diff
@@ -0,0 +1,284 @@
...
+#### Update your usage of `Rest.request()` to pass a `version` argument
+
+**Note:** This section is only relevant if you’re using the `Rest.request()` method.
+
+The signature of this method has been changed; it now requires that you pass a `version` argument to specify the major version of the REST API to use. For more information about REST API versioning, see [this section of the REST API reference](https://ably.com/docs/api/rest-api#versioning).
+
+(**TODO this wording "major version" comes from https://github.com/ably/sdk-api-reference/pull/32#pullrequestreview-1408872163. I guess that it means you should take the current sentence 'By default, all requests receive the latest version of the API, which is currently 1.1' from the linked versioning page and infer that the "major version" is 1, and so that's what I should pass to `Rest.request()`. But I thought that we've moved away from using a `major.minor` protocol version to just a single integer? See https://sdk.ably.com/builds/ably/specification/main/features/#CSV2a. Should we update those versioning docs to say that the latest version of the API is 3, and then drop the "major" in the documentation for `Rest.request()`?**)
```
lawrence-forooghian, line 113: Thoughts?

docs/migration-guides/v2/lib.md
```diff
@@ -0,0 +1,284 @@
+# Migration guide for ably-js v2
+
+Here’s how to migrate from ably-js v1 to v2:
+
+1. [Stop using functionality that v1 deprecated and which v2 removes](#stop-using-v1-deprecated).
+1. [Stop using other functionality that v2 removes](#stop-using-v2-removed).
+1. [Update to v2 and handle its breaking changes](#update-to-v2-and-handle-breaking-changes).
+1. (Optional) [Stop using functionality that v2 deprecates](#stop-using-v2-deprecated).
+1. (Optional) [Take advantage of new features that v2 introduces](#use-v2-new-features).
```
VeskeR, line 9: it's probably temporary while WIP, but commenting to not forget change numbers to 1, 2, 3 ...

lawrence-forooghian: Not sure if you know, but Markdown actually doesn't care about the numbers that you write in your source for an ordered list. This still gets rendered as 1, 2, 3, …, and you don't have to care about keeping the numbers up to date in your source as you insert / remove / reorder list items.

docs/migration-guides/v2/lib.md
```diff
@@ -0,0 +1,284 @@
...
+- The `host` client option has been renamed to `restHost`. Update your code to use `restHost`.
+- The `wsHost` client option has been renamed to `realtimeHost`. Update your code to use `realtimeHost`.
+- The `queueEvents` client option has been renamed to `queueMessages`. Update your code to use `queueMessages`.
+- `RealtimePresence`’s `on` method has been renamed to `subscribe`. Update your code to use `subscribe`.
+- `RealtimePresence`’s `off` method has been renamed to `unsubscribe`. Update your code to use `unsubscribe`.
+- `Auth`’s `authorise` method has been renamed to `authorize`. Update your code to use `authorize`.
+- The `"xhr"` transport has been renamed to `"xhr_streaming"`. Update your client options code to use `transports: ["xhr_streaming"]`.
```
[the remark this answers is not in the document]
lawrence-forooghian: Yep, depending on which PR gets first either @owenpearson will need to update the migration guide in his PR (see https://github.com/ably/ably-js/pull/1645#issuecomment-2004963270), or I'll update it in this PR.

docs/migration-guides/v2/lib.md
```diff
@@ -0,0 +1,284 @@
...
+
+### Only TypeScript users
+
+- In v2, the `stats()` method on `Rest` and `Realtime` no longer accepts an argument of type `any`. Make sure that any argument you pass to this method implements the `StatsParams` interface.
```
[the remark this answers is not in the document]
lawrence-forooghian: Yes, definitely, I’m not sure why I missed this; will check my notes.

docs/migration-guides/v2/lib.md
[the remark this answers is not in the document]
owenpearson: just a heads up - i'm gonna merge that PR as soon as CI is all passing on it. still happy to contribute any changelog stuff needed though whether that be just as a comment here or as a commit on this branch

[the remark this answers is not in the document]
lawrence-forooghian: I see it's been merged now. As far as I'm aware, the only user-facing change — at least that would need to be mentioned in a migration guide — is the removal of the `xhr_streaming` transport, is that right? So it's just modifying this line to Andrii’s suggested text above?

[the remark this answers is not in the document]
owenpearson: I think we should put something more descriptive about how the transport connection sequence has changed too, since there will be a regression for anyone using xhr_polling, in that the initial connection will be slower until transport preference has been cached in the browser, although we should also make it clear that it's an overall improvement for anyone who isn't using xhr_polling by focussing on the fact that we no longer initially connect with comet and then try to upgrade, instead we always try web_socket first if it's available.

docs/migration-guides/v2/lib.md
```diff
@@ -0,0 +1,284 @@
...
+
+The v1 type declarations contained various sets of three related types in order to reflect the different asynchronicity styles offered by the SDK. For example, it had a `RealtimeChannelBase` type which contained functionality common to the callbacks-based and promises-based API of a realtime channel, and then `RealtimeChannelCallbacks` and `RealtimeChannelPromises` types which contained the APIs unique to those asynchronicity styles.
+
+Now that [the callbacks-based variant of the library has been removed](#switch-from-callbacks-to-promises), we’ve simplified the types, removing the `*Callbacks` types and combining each pair of `*Base` and `*Promise` types into a single type. For example, where previously there existed three types that described a realtime channel, now there is just one, named `RealtimeChannel`. You should update your code to stop referring to the suffixed types. In general, this is a case of removing these suffixes from the type names.
+
+#### Be aware that some properties might not be populated on messages recieved from Ably
```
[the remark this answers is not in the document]
lawrence-forooghian: Thanks, will fix.

docs/migration-guides/v2/lib.md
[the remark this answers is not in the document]
lawrence-forooghian: I'm not familiar with what was implemented in #1645 so I'll put some guidance in that's a paraphrase of what you've written above, then you can tell me if it looks right :)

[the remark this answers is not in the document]
lawrence-forooghian: Also what I said in https://github.com/ably/ably-js/pull/1670#discussion_r1530755508 is wrong; this needs to be moved from the "Stop using functionality that v1 deprecated and which v2 removes" section to the "Stop using other functionality that v2 removes" section, will do.

docs/migration-guides/v2/lib.md
```diff
@@ -0,0 +1,284 @@
...
+#### Stop explicitly selecting the promise-based variant of the library
+
+As [mentioned above](#switch-from-callbacks-to-promises), v2 of the library no longer offers a choice between a callbacks-based API and a promises-based API. This means that v1’s mechanism for choosing which variant to use has been removed, so you should stop using this mechanism.
+
+- If you’re explicitly choosing the promises variant of the library through the subpath of the imported module — that is, if you’re writing `require('ably/promises')` — then update your code to `require('ably')`.
+- If you’re explicitly choosing the promises variant at the moment of instantiating the client — that is, if you’re writing `new Ably.Rest.Promise(…)` or `new Ably.Realtime.Promise(…)` — then update your code remove the use of the `Promise` property, i.e. write `new Ably.Realtime(…)` or `new Ably.Rest(…)`.
+
+#### Be aware of platforms that are no longer supported
+
+v2 of the library drops support for some platforms that v1 supported. Most notably, it no longer supports Internet Explorer or non-LTS versions of Node.js. For more information, see [this section of the Readme](../../../README.md#supported-platforms).
```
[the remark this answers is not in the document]
lawrence-forooghian: Ah good spot, will change, thanks

docs/migration-guides/v2/lib.md
lawrence-forooghian: So, what the [ably.com REST API versioning documentation](https://ably.com/docs/api/rest-api#versioning) refers to as the "API version" — that is, the value of the `X-Ably-Version` request header — is the same concept as what the SDK specification calls the "protocol version" (referred to by [CSV2](https://sdk.ably.com/builds/ably/specification/main/features/#CSV2)), as evidenced by the fact that [RSC7e](https://sdk.ably.com/builds/ably/specification/main/features/#RSC7e) says that this CSV2 version is what the library should use to populate the `X-Ably-Version` header of the requests that it sends.

The Ably service accepts two different formats of protocol version:

- As the ably.com REST API versioning documentation suggests with its examples of "1.1" and "1.2", it accepts a protocol version of the format `major.minor`. This is confirmed by [CSV2a](https://sdk.ably.com/builds/ably/specification/main/features/#CSV2a), which implies that this format is accepted if `major` < 2.
- As CSV2a implies, it also accepts a single integer which is greater than or equal to 2. The maximum value permitted here will be incremented as new breaking changes are introduced to the REST or Realtime API. The implication here is that from now on, there will no longer be a concept of a "major" protocol version; there is now just "protocol version".

Now, since the `Rest.request()` mandatory version parameter was added after the introduction of protocol version 2, it accepts a single integer version. Yet, the documentation for this parameter refers to a "major" version, even though that concept no longer exists for future protocol versions. It should just refer to "the protocol version".

The problem is that in order to provide users with more information about REST API versioning, the documentation for this new `Rest.request()` version parameter links them to the ably.com REST API versioning documentation, which has not been updated recently and has a couple of problems:

- it refers to protocol versions 1.1 and 1.2, meaning that a user coming to this page from a `Rest.request()` documentation comment that mentions "the protocol version" wouldn't know how to populate this value, which is a single integer;
- it says that the latest API version is 1.1, even though [CSV2b](https://sdk.ably.com/builds/ably/specification/main/features/#CSV2b) tells us that the latest version must be at least 3

So, the ideal thing to do would be to:

- update the `Rest.request()` version parameter documentation to just refer to "the protocol version"
- update the ably.com REST API versioning documentation to say that the latest API version is 3

But, just saying "the latest version is 3" isn't very useful — what I think it really intends to say is "you are currently viewing documentation for REST API version 3". It's not very useful to know what the latest API version is without knowing how to interact with it.

So, then we'd need to update the REST API documentation to reflect the changes to the REST API made between protocol version 1.1 and 3. Off the top of my head, these are changes to the batch APIs (token revocation, batch presence, batch publish) and to the stats API (flattened, "untyped stats"). These REST API changes aren't yet documented anywhere as far as I’m aware; rather you can infer them from the corresponding changes that have been made to the SDK feature spec.

_But_, then, we'd have to figure out how to handle the problem of ably.com REST API documentation versioning. The ably.com REST API documentation isn't currently versioned, so what would happen to users who are making requests to the REST API using an `X-Ably-Version` of "1.1"? Suddenly the documentation for the API they're using would disappear.

As you can see, this simple-seeming question of how to document this new `Rest.request()` version parameter ends up requiring quite a lot of changes to the REST API documentation, so that's why I wanted to get @m-hulbert’s opinion on how to proceed.

lawrence-forooghian: <del>requiring</del> potentially requiring — keen to hear other ideas

docs/migration-guides/v2/lib.md
```diff
@@ -0,0 +1,284 @@
...
+The changes below are split into those that affect all users, and those that only affect TypeScript users.
+
+### All users
+
+#### Stop explicitly selecting the promise-based variant of the library
+
+As [mentioned above](#switch-from-callbacks-to-promises), v2 of the library no longer offers a choice between a callbacks-based API and a promises-based API. This means that v1’s mechanism for choosing which variant to use has been removed, so you should stop using this mechanism.
+
+- If you’re explicitly choosing the promises variant of the library through the subpath of the imported module — that is, if you’re writing `require('ably/promises')` — then update your code to `require('ably')`.
+- If you’re explicitly choosing the promises variant at the moment of instantiating the client — that is, if you’re writing `new Ably.Rest.Promise(…)` or `new Ably.Realtime.Promise(…)` — then update your code remove the use of the `Promise` property, i.e. write `new Ably.Realtime(…)` or `new Ably.Rest(…)`.
```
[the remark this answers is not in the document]
lawrence-forooghian: Fixed.

docs/migration-guides/v2/lib.md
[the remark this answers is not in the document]
lawrence-forooghian: Fixed.

docs/migration-guides/v2/lib.md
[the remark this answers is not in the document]
lawrence-forooghian: Fixed.

docs/migration-guides/v2/lib.md
lawrence-forooghian: Thanks for the suggestion; I've pushed some stuff along the lines you suggested.

docs/migration-guides/v2/lib.md
```diff
@@ -0,0 +1,284 @@
...
+<h4 id="switch-to-promise-based-generateRandomKey">Switch to using the new promise-based API of <code>Crypto.generateRandomKey()</code></h4>
+
+**Note:** This section is only relevant if you’re using the `Crypto.generateRandomKey()` method.
+
+If you’re using the `Crypto.generateRandomKey()` method, you’ll need to change how you call this method. In v1, this method required that you pass a callback. In v2, it communicates its result by returning a promise.
+
+So you need to change code that looks like this:
+
+```javascript
+Ably.Realtime.Crypto.generateRandomKey((key) => { … })
```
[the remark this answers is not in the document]
lawrence-forooghian: Ah yes, thanks re the `err` param, have changed. As for the "how to use the promise" thing, I didn't really want to keep writing out examples of these two different patterns. So I’ve just changed it to say "into code that makes use of the returned promise, for example by using the `await` keyword on it", think that’s OK?

docs/migration-guides/v2/lib.md
```diff
@@ -0,0 +1,284 @@
...
+
+**Note:** This section is only relevant if you’re using the `RealtimeChannel.whenState()` or `Connection.whenState()` methods.
+
+The `RealtimeChannel.whenState()` and `Connection.whenState()` methods now return `null` when the connection is already in the given state, instead of attempting to synthesize an artificial state change.
+
+#### Be aware that the `fromEncoded()` and `fromEncodedArray()` methods are now async
+
+**Note:** This section is only relevant if you’re using `Message` or `PresenceMessage`’s `fromEncoded` or `fromEncodedArray` methods.
+
+`Message` and `PresenceMessage`’s `fromEncoded` and `fromEncodedArray` methods now operate asynchronously. That is, instead of returning the decoding result, they return a promise. Update your code to `await` the result of these methods.
```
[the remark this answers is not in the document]
lawrence-forooghian: Changed it to "Update your code to retrieve the result of these promises, for example by using the `await` keyword.", happy with that?

docs/migration-guides/v2/lib.md
lawrence-forooghian: > and I believe it would be confusing to users as well

Yeah, agreed

lawrence-forooghian:
> JSONP support was also removed in https://github.com/ably/ably-js/pull/1215. It affected the client options transports field. I think it should be mentioned in either Stop using other functionality that v2 removes or Update to v2 and handle its breaking changes section.

Yes, definitely, thanks! I missed that there was a breaking API change so I just considered it part of the browser compatibility changes. I've added a section for it now.

lawrence-forooghian:
> If we agree to use this folder structure, then we should probably move existing react-migration-guide.md to the docs/migration-guides/v1 folder to keep it consistent.

Good point! I've added 9284ce88ed3220fbe820790a6e1ec6d94803acff which does this.

docs/migration-guides/v2/lib.md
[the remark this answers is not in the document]
lawrence-forooghian: Yeah, not sure how I missed this, thanks for pointing it out. I’ve added a section for it but I don't think the guidance is great; will add a comment here asking for thoughts.

docs/migration-guides/v2/lib.md
```diff
@@ -0,0 +1,375 @@
...
+
+- `all`
+- `apiRequests`
+- `channels`
+- `connections`
+- `inbound`
+- `outbound`
+- `persisted`
+
+You should migrate your code to use the `entries` property instead of these properties. `entries` is an object all of whose properties have numeric values. Here is an example of how this might look:
```
lawrence-forooghian, line 192: @SimonWoolf I've written this guidance for how to migrate to using the flattened stats API but it feels like a bit of a cop-out; it basically just says "Here's an example of how to migrate one property; beyond that you're own your own. Go fetch some stats, look at the schema they point to, figure out how the changes you need to make".

Can we do any better than this? Are there some clear rules for how the v1 stats API maps to the flattened stats response? For example, it seems that the message-related properties (`all`, `inbound`, `outbound`, `persisted`) now require you to prefix them with `messages`.

docs/migration-guides/v2/lib.md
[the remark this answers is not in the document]
lawrence-forooghian: Actually @owenpearson would you mind writing something that we can add to describe these changes? I think it'd be a lot quicker than me trying to understand it. Feel free to push a commit.

docs/migration-guides/v2/lib.md
m-hulbert: Nice write up @lawrence-forooghian 👍 

The plan is to version the REST API in docs to make this easier to handle. As I understand it, the intention is to try and avoid making protocol version yet another version devs need to consider. Whilst this will work to some extent by having the docs for each language link out to the correct version of the REST API, there will still need to be some explainer text within the REST API itself as to what the version means, and by extension what this means for `rest.request()`.

lawrence-forooghian: Thanks for the reply @m-hulbert. Sounds like this is stuff planned for the future. Do you have thoughts on what we should do then in this immediate context of telling users migrating to ably-js v2 how to use the version of `Rest.request()` that requires a version?

lawrence-forooghian: Basically, for somebody who is now thinking "I have to specify a number here, it says; what number?!"

docs/migration-guides/v2/lib.md
SimonWoolf: Ideally this should be in the main docs rather than ably-js specific, but sure, the main changes are:

- the various stats that count messages in different ways (all, inbound, outbound, persisted, processed) are now under `messages` instead of all in the top level
- `connections` is no longer broken down into plain and tls
- instead of top-level `apiRequests` and `tokenRequests`, which doesn't make much sense because token requests are api requests, you now have a top level `apiRequests` that's broken down currently into `tokenRequests` and `other` with an `all` aggregate, with potential to split `other` out into more specific types later
- `push` is completely reorganised in a way that makes more sense

You could also link to [the source of the stats schema](https://github.com/ably/ably-common/blob/main/json-schemas/src/app-stats.json) which is an up to date & accurate description of the new stats struct

docs/migration-guides/v2/lib.md
m-hulbert: I wonder if in the short term we should just say something along the lines of "set the version to `3` to use with ably-js v2". Anyone reading this guide will already be migrating to v2, so they shouldn't need to worry about the historical context of setting a version as `major.minor` rather than a single int. 

I appreciate this will be slightly contradictory to the REST API reference docs in the short term, but I think it's the simplest answer for anyone migrating.

lawrence-forooghian: I think it would actually make more sense to say to use version 2 for now, since that'll provide the same behaviour as v1’s `request()`?

lawrence-forooghian: i.e. when you call `request()` in v1 it sends an `X-Ably-Version: 2` header

lawrence-forooghian: So, sure, that makes sense for users migrating, we say in this migration guide to specify a version of 2. As for the docstring for the method, do you want to leave it as it is (i.e. referring to "major" version and linking to the versioning docs?)

docs/migration-guides/v2/lib.md
owenpearson: I actually didn't know this, and find it quite strange 😅 i do think one of the key advantages of markdown is that the source is human readable so you don't need a renderer (and i rarely, if ever, use one locally), so i marginally prefer using an actual sequence here

docs/migration-guides/v2/lib.md
```diff
@@ -0,0 +1,375 @@
...
+
+<h2 id="stop-using-v1-deprecated">Stop using functionality that v1 deprecated and which v2 removes</h2>
+
+Begin by updating to ably-js 1.2.50 or later, to make sure you see the deprecation log messages [described below](#stop-using-v1-deprecated-all-users).
+
+Now, you need to stop using the functionality that is deprecated by v1 and which is removed in v2. Here we explain how.
+
+The changes below are split into those that affect all users, and those that only affect TypeScript users.
+
+<h3 id="stop-using-v1-deprecated-all-users">Changes that affect all users</h3>
```
owenpearson, line 19: not gonna lie, on my first read of this my heart skipped a beat when i saw "changes that affect all users" followed by a long wall of text 😬 not sure exactly what to suggest but i think we could do better to make it immediately clear that these are breaking changes to individual features which users may or not be using

docs/migration-guides/v2/lib.md
```diff
@@ -0,0 +1,375 @@
...
+
+- Changes to `Crypto.getDefaultParams()`:
+  - The ability to pass a callback to this method has been removed in v2. This method now directly returns its result, instead of returning it asynchronously. Update your code so that it uses the return value of this method instead of passing a callback.
+  - The ability to call this method without specifying an encryption key has been removed in v2. Update your code so that it instead passes an object whose `key` property contains an encryption key. That is, replace `Crypto.getDefaultParams()` with `Crypto.getDefaultParams({ key })`, where `key` is an encryption key that you have generated (for example from the `Crypto.generateRandomKey()` method).
+  - The ability to pass the encryption key as the first argument of this method has been removed in v2. Update your code so that it instead passes an object whose `key` property contains the key. That is, replace `Crypto.getDefaultParams(key)` with `Crypto.getDefaultParams({ key })`.
+- The `headers` client option has been removed in v2. Remove your use of this client option.
```
owenpearson, line 29: I think we could maybe order these bullets a bit better in order of how likely users are to be using them? for example, `log`/`logLevel` is quite a common and important option whereas I'd be quite surprised if anyone was using `headers`

docs/migration-guides/v2/lib.md
```diff
@@ -0,0 +1,375 @@
+# Migration guide for ably-js v2
+
+Here’s how to migrate from ably-js v1 to v2:
+
+1. [Stop using functionality that v1 deprecated and which v2 removes](#stop-using-v1-deprecated).
+1. [Stop using other functionality that v2 removes](#stop-using-v2-removed).
+1. [Update to v2 and handle its breaking changes](#update-to-v2-and-handle-breaking-changes).
+1. (Optional) [Stop using functionality that v2 deprecates](#stop-using-v2-deprecated).
+1. (Optional) [Take advantage of new features that v2 introduces](#use-v2-new-features).
+
```
owenpearson, line 10:
```suggestion
## NB: changes to realtime connection sequence and transport order

In previous versions of ably-js, realtime clients with multiple available transports would initially attempt to connect with the transport most likely to succeed (`xhr_polling` in web browsers) and only once the initially connection was made, would we then attempt to 'upgrade' to a preferable transport such as `web_socket`.

In version 2 of the library, realtime clients will instead connect with `web_socket` first and only failover to alternative transports if the initial WebSocket connection attempt fails. For the vast majority of users this will result in a smoother initial connection sequence, however in environment where WebSocket connections are unavailable and time out instead of failing immediately (such as when using a corporate proxy which strips WebSocket headers) this may result in a slower initial connection. Once a connection is first established, transport preference will be cached in web browser local storage so subsequent connections will use the best available transport. If you expect web socket attempts to always fail in your enviornment, you can skip the WebSocket connection step by explicitly providing a list of transports which omits the `web_socket` transport via client options.
```

owenpearson (changes requested):

docs/migration-guides/v2/lib.md
m-hulbert: Oh sure, if there's nothing in v3 required then I agree it makes more sense to stick with `2` as you suggest. 

For the docstring I think we can drop "major" and just refer to it as "version" since this will be the case going forwards. We can leave the link in for now too even if it'll be a little wonky until we release the versioned REST API.

docs/migration-guides/v2/lib.md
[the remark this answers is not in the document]
owenpearson: I've added it as a suggestion here https://github.com/ably/ably-js/pull/1670#discussion_r1532206072, i'm not super happy with the long paragraphs and placement within the document so feel free to edit as much as you like, hopefully that's at least a helpful starting point.

docs/migration-guides/v2/lib.md
lawrence-forooghian: Sure, have changed

docs/migration-guides/v2/lib.md
owenpearson: as mentioned on slack, we should also mention that when using `web_socket` as a transport, version 2 of the realtime client will make use of the `wss://ws-up.ably-realtime.com` endpoint, so if you're using a firewall you'll need to ensure that this endpoint is reachable. we can link out to the documentation at https://faqs.ably.com/if-i-need-to-whitelist-ablys-servers-from-a-firewall-which-ports-ips-and/or-domains-should-i-add.

docs/migration-guides/v2/lib.md
lawrence-forooghian: Yeah, that's a good point. I've:

- changed the intro to each section to say:
 
  > The changes below are split into:
  > 
  > - general changes
  > - changes that only affect TypeScript users
- changed "All users" headings to "General changes"
- added the sentence "If you’re not using any of the deprecated functionality described below, then this section does not affect you." to the deprecated v1 functionality intro
- added the sentence "Some of these changes are only relevant if you’re using specific features of the library. The guidance below makes it clear when this is the case." to the v2 breaking changes section.

docs/migration-guides/v2/lib.md
lawrence-forooghian: I agree that the logging one is likely to affect a lot of people, so have put that first. And since nobody internally even seemed to know about `headers`, I've put that last. Beyond that I have no idea what things are more likely than others so I’ve kept the order as it is :)

docs/migration-guides/v2/lib.md
lawrence-forooghian: Thanks! I’ve incorporated what you said above.

docs/migration-guides/v2/lib.md
lawrence-forooghian: Resolving since Owen is away

docs/migration-guides/v2/lib.md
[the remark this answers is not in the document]
lawrence-forooghian: OK, have improved that section now; see https://github.com/ably/ably-js/pull/1670#discussion_r1532030255

docs/migration-guides/v2/lib.md
lawrence-forooghian: Thanks, have very lightly edited the above and added it, and also added a section about whitelisting the endpoint.

lawrence-forooghian: Resolving since Owen’s away.

docs/migration-guides/v2/lib.md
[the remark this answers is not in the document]
lawrence-forooghian: Thanks, have:

- very lightly edited what you wrote in that comment there and added it
- added section re removal of `xhr` / `xhr_streaming`

docs/migration-guides/v2/lib.md
```diff
@@ -0,0 +1,412 @@
...
+This means that the library is no longer capable of interacting with CryptoJS’s `WordArray` type. Specifically:
+
+- `Crypto.generateRandomKey()` now returns an `ArrayBuffer` instead of a `WordArray`.
+- `Crypto.getDefaultParams({ key })` no longer accepts a `WordArray` key; pass an `ArrayBuffer` or `Uint8Array` instead.
+
+#### Stop requesting the `xhr_streaming` and `xhr` transports
+
+**Note:** This section is only relevant if you’re explicitly writing `"xhr_streaming"` or `"xhr"` as part of your `transports` client option.
+
+The `xhr_streaming` transport, also known as `xhr`, has been removed. If your `transports` client option contains `"xhr_streaming"` or `"xhr"`, remove this value. If you still wish to explicitly request a non-WebSocket transport, request `xhr_polling` instead.
```
lawrence-forooghian, line 276: @owenpearson (probably actually @SimonWoolf since Owen is away now) do you have any further information on why `xhr_streaming` was removed? I’m telling users to use `xhr_polling` instead, but that’s worse (in terms of responsiveness), right? I’m just a bit out of the loop on this change

SimonWoolf: We're now doing fallback instead of upgrade. Say you try to connect with websocket, it times out, you try the health check endpoints, which tells you that there's internet connectivity but websockets are broken for some reason. So that likely means you're being MITM'd by some shitty corporate proxy that doesn't deal with websockets properly, and the guarantees that we'd normally get with TLS are worthless. So now you have to ask yourself: what transport should you try next? Because we know from the whole e*****a saga that corporate proxy that break websockets can break xhr streaming as well! We need to try something that we know will work even with the most basic of shitty web proxies, which means long polling.

Given which, what role is left for xhr streaming?

Of course not using it in the default connection strategy doesn't necessarily mean removing it from the lib altogether. Not sure if there was an explicit discussion on that or if owen just figured that if we're not using it by default then we should just remove it. Shrug, if some customer comes to us and says "We're using a very particular corporate firewall that can't cope with websockets but actually can do xhr streaming fine" then we can always re-add it

more background: https://github.com/ably/ably-js/issues/1330

docs/migration-guides/v2/lib.md
VeskeR: Awesome changes, thank you!

docs/migration-guides/v2/lib.md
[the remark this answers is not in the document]
VeskeR: I believe this is now resolved, so @lawrence-forooghian feel free to close this conversation if you've added everything you wanted with Owen.

docs/migration-guides/v2/lib.md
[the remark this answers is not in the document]
VeskeR: Yea, good enough, we had much more verbose examples for callbacks conversion above.

docs/migration-guides/v2/lib.md
VeskeR: Same, didn't know about this feature, I'm too used to IDEs automatically populating those lists now, had no idea markdown does it in the back too.
Anyway, it's changed now, closing conversation.

docs/migration-guides/v2/lib.md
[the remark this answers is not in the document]
VeskeR: Great!

VeskeR (approved):
LGTM! I'm happy to merge once all remaining conversations are resolved and `[skip ci]` is removed.

ttypic (approved):
Like it!

docs/migration-guides/v2/lib.md
lawrence-forooghian: > shitty corporate proxies that break websockets probably also break break xhr streaming

Ah, I didn’t know this!

> Of course not using it in the default connection strategy doesn't necessarily mean removing it from the lib altogether. Not sure if there was an explicit discussion on that or if owen just figured that if we're not using it by default then we should just remove it.

Yeah, that's the part that wasn't clear to me. Well, it's done now, and Owen’s not here to explain, and there don't seem to be strong opinions about it, so I guess we'll be sticking with it.

> Shrug, if some customer comes to us and says "We're using a very particular corporate firewall that can't cope with websockets but actually can do xhr streaming fine" then we can always re-add it

OK, cool.

So sounds like there isn't really else that I can add to the migration guide (explaining why we removed it, that is); let's see if anyone complains 🤷

lawrence-forooghian: Actually, it sounds like, given that it didn't help massively with firewalls, the main reason that `xhr_streaming` existed was to provide a WebSocket-like experience in browsers that didn't support WebSocket. With that in mind, I've updated this guidance:

> v1 offered the `xhr_streaming` transport, also known as `xhr`, to provide a performant realtime transport for browsers which did not support WebSocket connections. Since this limitation does not apply to any of the [browsers supported by v2](#supported-platforms), the `xhr_streaming` transport has been removed. If your `transports` client option contains `"xhr_streaming"` or `"xhr"`, remove this value. If you still wish to explicitly request a non-WebSocket transport, request `xhr_polling` instead.

PR #1674, opened 2024-03-08
[ECO-4115] Split `usePresence` hook into different hooks for entering presence and subscribing to presence events

### LINKED ISSUE #1621: Split `usePresence` hook into different hooks for entering presence and subscribing to presence events
`usePresence` hook is quite opinionated right now - it enters presence and subscribes to presence events automatically on component mount, and there is no documentation on how to modify this behavior (via combination of `modes` and `subscribeOnly` options). Also docs (both [on github](https://github.com/ably/ably-js/blob/main/docs/react.md#usepresence), and [Ably Docs](https://ably.com/docs/getting-started/react#usePresence)) don't say a word about `usePresence` entering presence.
It seems reasonable to have two separate hooks for these two behaviors. See Slack thread for more discussion: https://ably-real-time.slack.com/archives/C030C5YLY/p1703088239046009

This split depends on `ChannelProvider` from https://github.com/ably/ably-js/pull/1620, to avoid confusion when using two different presence hooks with the same channel name. Channel name and options would be set in one place in `ChannelProvider`, and other parts of the application will be just using that channel with set options.

Using `ChannelProvider` will fix the issue of implicit usage of channel in `usePresence`, which means that channel options we pass to `usePresence` matter and may cause channel to reattach, which causes errors if options are not changes via `setOptions`.

As part of the split, we can document or entirely remove next `usePresence` options (they are undocumented and not obvious how to use):

1. `subscribeOnly` - no docs, prevents entering presence on component mount.
Reasoning for 'subscribeOnly': https://github.com/ably-labs/react-hooks/issues/40:
* "modes" - decide how would your channel connection behave, but not necessary what API you will call
* since `usePresence` is quite opinionated currently (it enters and subscribes automatically), this means that if you don't have required capability set, you will get authorization error when trying to do operation you're not supposed to (e.g. trying to enter a presence when only subscribe capability is allowed via `modes`).

Probably should also create separate `ChannelNameAndOptions` interfaces for `useChannel` and new `usePresence*` hooks (depends on the options we end up having). Currently, `subscribeOnly` is not used in `useChannel`, but is present in `ChannelNameAndOptions` which is used by both `useChannel` and `usePresence` hooks.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4115) by [Unito](https://www.unito.io)

VeskeR:
Resolves #1621 

See commit messages for more details

### DIFF
--- a/src/platform/react-hooks/src/hooks/usePresence.ts
+++ b/src/platform/react-hooks/src/hooks/usePresence.ts
@@ -1,5 +1,5 @@
 import type * as Ably from 'ably';
-import { useCallback, useEffect } from 'react';
+import { useCallback, useEffect, useRef } from 'react';
 import { ChannelParameters } from '../AblyReactHooks.js';
 import { useAbly } from './useAbly.js';
 import { useChannelInstance } from './useChannelInstance.js';
@@ -26,27 +26,37 @@ export function usePresence<T = any>(
   const ably = useAbly(params.ablyId);
   const { channel } = useChannelInstance(params.ablyId, params.channelName);
   const { connectionError, channelError } = useStateErrors(params);
+  // we can't simply add messageOrPresenceObject to dependency list in our useCallback/useEffect hooks,
+  // since it will most likely cause an infinite loop of updates in cases when user calls this hook
+  // with an object literal instead of a state or memoized object.
+  // to prevent this from happening we store messageOrPresenceObject in a ref, and use that instead.
+  // note that it still prevents us from automatically re-entering presence with new messageOrPresenceObject if it changes.
+  // one of the options to fix this, is to use deep equals to check if the object has actually changed. see https://github.com/ably/ably-js/issues/1688.
+  const messageOrPresenceObjectRef = useRef(messageOrPresenceObject);
 
-  const onMount = async () => {
-    await channel.presence.enter(messageOrPresenceObject);
-  };
+  useEffect(() => {
+    messageOrPresenceObjectRef.current = messageOrPresenceObject;
+  }, [messageOrPresenceObject]);
 
-  const onUnmount = () => {
+  const onMount = useCallback(async () => {
+    await channel.presence.enter(messageOrPresenceObjectRef.current);
+  }, [channel.presence]);
+
+  const onUnmount = useCallback(() => {
     // if connection is in one of inactive states, leave call will produce exception
     if (channel.state === 'attached' && !INACTIVE_CONNECTION_STATES.includes(ably.connection.state)) {
       channel.presence.leave();
     }
-  };
+  }, [channel, ably.connection.state]);
 
-  const useEffectHook = () => {
-    if (!skip) onMount();
+  useEffect(() => {
+    if (skip) return;
+
+    onMount();
     return () => {
       onUnmount();
     };
-  };
-
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  useEffect(useEffectHook, [skip]);
+  }, [skip, onMount, onUnmount]);
 
   const updateStatus = useCallback(
     (messageOrPresenceObject: T) => {
--- a/src/platform/react-hooks/src/hooks/usePresence.ts
+++ b/src/platform/react-hooks/src/hooks/usePresence.ts
@@ -5,6 +5,10 @@ import { useAbly } from './useAbly.js';
 import { useStateErrors } from './useStateErrors.js';
 import { useChannelInstance } from './useChannelInstance.js';
 
+interface PresenceMessage<T = any> extends Ably.PresenceMessage {
+  data: T;
+}
+
 export interface PresenceResult<T> {
   presenceData: PresenceMessage<T>[];
   updateStatus: (messageOrPresenceObject: T) => void;
@@ -88,14 +92,4 @@ export function usePresence<T = any>(
   );
 
   return { presenceData, updateStatus, connectionError, channelError };
-}
-
-interface PresenceMessage<T = any> {
-  action: Ably.PresenceAction;
-  clientId: string;
-  connectionId: string;
-  data: T;
-  encoding: string;
-  id: string;
-  timestamp: number;
 }
--- a/src/platform/react-hooks/src/hooks/usePresence.ts
+++ b/src/platform/react-hooks/src/hooks/usePresence.ts
@@ -46,9 +46,7 @@ export function usePresence<T = any>(
   };
 
   const onMount = async () => {
-    channel.presence.subscribe('enter', updatePresence);
-    channel.presence.subscribe('leave', updatePresence);
-    channel.presence.subscribe('update', updatePresence);
+    channel.presence.subscribe(['enter', 'leave', 'update'], updatePresence);
 
     if (!subscribeOnly) {
       await channel.presence.enter(messageOrPresenceObject);
@@ -65,9 +63,7 @@ export function usePresence<T = any>(
         channel.presence.leave();
       }
     }
-    channel.presence.unsubscribe('enter', updatePresence);
-    channel.presence.unsubscribe('leave', updatePresence);
-    channel.presence.unsubscribe('update', updatePresence);
+    channel.presence.unsubscribe(['enter', 'leave', 'update'], updatePresence);
   };
 
   const useEffectHook = () => {
--- a/src/platform/react-hooks/src/hooks/usePresence.ts
+++ b/src/platform/react-hooks/src/hooks/usePresence.ts
@@ -1,77 +1,45 @@
 import type * as Ably from 'ably';
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useEffect } from 'react';
 import { ChannelParameters } from '../AblyReactHooks.js';
 import { useAbly } from './useAbly.js';
-import { useStateErrors } from './useStateErrors.js';
 import { useChannelInstance } from './useChannelInstance.js';
+import { useStateErrors } from './useStateErrors.js';
 
-interface PresenceMessage<T = any> extends Ably.PresenceMessage {
-  data: T;
-}
-
-export interface PresenceResult<T> {
-  presenceData: PresenceMessage<T>[];
+export interface PresenceEnterResult<T> {
   updateStatus: (messageOrPresenceObject: T) => void;
   connectionError: Ably.ErrorInfo | null;
   channelError: Ably.ErrorInfo | null;
 }
-
-export type OnPresenceMessageReceived<T> = (presenceData: PresenceMessage<T>) => void;
-export type UseStatePresenceUpdate = (presenceData: Ably.PresenceMessage[]) => void;
 
 const INACTIVE_CONNECTION_STATES: Ably.ConnectionState[] = ['suspended', 'closing', 'closed', 'failed'];
 
 export function usePresence<T = any>(
   channelNameOrNameAndOptions: ChannelParameters,
   messageOrPresenceObject?: T,
-  onPresenceUpdated?: OnPresenceMessageReceived<T>,
-): PresenceResult<T> {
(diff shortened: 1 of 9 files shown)


src/platform/react-hooks/src/hooks/usePresenceEnter.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export interface PresenceEnterResult<T> {
+  updateStatus: (messageOrPresenceObject: T) => void;
+  connectionError: Ably.ErrorInfo | null;
+  channelError: Ably.ErrorInfo | null;
+}
+
+const INACTIVE_CONNECTION_STATES: Ably.ConnectionState[] = ['suspended', 'closing', 'closed', 'failed'];
+
+export function usePresenceEnter<T = any>(
```
VeskeR, line 16: not sure about naming for new hooks, open to suggestions:
`usePresenceEnter` - for `presence.enter()` logic
`usePresenceListener` - for `presence.subscribe(['enter', 'leave', 'update']` logic

src/platform/react-hooks/src/hooks/usePresenceListener.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export interface PresenceListenerResult<T> {
+  presenceData: PresenceMessage<T>[];
+  connectionError: Ably.ErrorInfo | null;
+  channelError: Ably.ErrorInfo | null;
+}
+
+export type OnPresenceMessageReceived<T> = (presenceData: PresenceMessage<T>) => void;
+export type UseStatePresenceUpdate = (presenceData: Ably.PresenceMessage[]) => void;
```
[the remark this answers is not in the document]
ttypic: 🤷 I think it's better to delete if it's not used

ttypic (changes requested):
Looks good to me, I think this will be more clear for users than `subscribeOnly` option in `usePresence`. Added some comments and we need to merge https://github.com/ably/ably-js/pull/1676 first

src/platform/react-hooks/src/hooks/usePresenceEnter.ts
```diff
@@ -0,0 +1,59 @@
...
+  const { connectionError, channelError } = useStateErrors(params);
+
+  const onMount = async () => {
+    await channel.presence.enter(messageOrPresenceObject);
+  };
+
+  const onUnmount = () => {
+    // if connection is in one of inactive states, leave call will produce exception
+    if (channel.state === 'attached' && !INACTIVE_CONNECTION_STATES.includes(ably.connection.state)) {
+      channel.presence.leave();
```
owenpearson, line 37: I know this was the behaviour in the previous `usePresence` hook, but i'm not sure if this behaviour is more helpful than not - for example if you have two components mounted with this hook and one of them is umounted then you'll leave presence even though you still have a component with `usePresenceEnter` mounted. I do think it would be easier to support users explicitly entering/leaving presence, but if we are gonna provide a hook it should have a more intelligent way of automatically leaving presence than just every component unmount. WDYT @VeskeR?

VeskeR: I thought a bit about this earlier today too.
We may need something like it's done in `useChannel` with its channel detachment logic:
https://github.com/ably/ably-js/blob/46dcda7f69350657ef5579344dfe95bd2ca0b967/src/platform/react-hooks/src/hooks/useChannel.ts#L92
There we check the count of existing listeners on the channel, so if there are multiple `useChannel` hooks and some of them unmount, we won't detach the channel, since there are still listeners active.

So one option would be to keep track of mounted `usePresenceEnter` hooks and check that we unmounted every single one of them before leaving the presence.

> I do think it would be easier to support users explicitly entering/leaving presence

This is something I didn't see discussed before. It would be kinda similar to how we added 'publish' function to `useChannel` hook in https://github.com/ably/ably-js/pull/1658, we would have something like:
```tsx
const { enter, leave } = usePresence('channel-name');

useEffect(() => {
  enter();

  return () => {
    leave();
  };
}, [enter, leave]);
```
so users can have more granular control at which point in their app they want to enter presence, and when exactly they would like to leave it.

ttypic: Just for the record, `useChannel` detach logic is also flawed, e.g. I believe `channel.listeners.length` will be 0 if you only subscribe to presence updates, but channel shouldn't be detached. And I think now we should move this logic to the `ChannelProvider`.

Explicitly entering or leaving presence may work; the only thing that concerns me is that this way, we have less control, and the user should take care of this themselves, which could potentially lead to a wrong implementation.

I started feeling maybe we need `PresenceProvider` as well :) The problem with current `usePresence` hooks that even if we calculate all refs and leave only when it's necessary, we still have no control when they are rendered and if they have different payload for `enter` will have different results.

ttypic: But on second thoughts I don't know how to design `PresenceProvider`, I don't want it to enter presence immediately, this way we have to implement `skip` parameter and it will look weird

ttypic: I think I'll vote for some ref counter for entering/leaving

src/platform/react-hooks/src/hooks/usePresenceListener.ts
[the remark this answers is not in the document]
VeskeR: Removed

src/platform/react-hooks/src/hooks/usePresenceListener.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  const useEffectHook = () => {
+    if (!skip) onMount();
+    return () => {
+      onUnmount();
+    };
+  };
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useEffect(useEffectHook, [skip]);
```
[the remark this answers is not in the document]
VeskeR: Removed `eslint-disable-next-line react-hooks/exhaustive-deps` and fixed hook not reacting to `onPresenceMessageReceived` changes in https://github.com/ably/ably-js/pull/1674/commits/4906bda6cc9b6bb41a86a6ade2c5a369c90e7f28.

Also added tests to check this behavior in https://github.com/ably/ably-js/pull/1674/commits/4ca5e8f0eacee04198b8e300d12db661c140dd10

src/platform/react-hooks/src/hooks/usePresenceEnter.ts
VeskeR: Decided to rename `usePresenceEnter` to `usePresence`. Clashing with the previous meaning of `usePresence` is justifiable since we're releasing those changes in a major version
`usePresenceListener` is fine.

src/platform/react-hooks/src/hooks/usePresenceListener.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  const { channel } = useChannelInstance(params.id, params.channelName);
+  const { connectionError, channelError } = useStateErrors(params);
+  const [presenceData, updatePresenceData] = useState<Array<PresenceMessage<T>>>([]);
+
+  const updatePresence = async (message?: Ably.PresenceMessage) => {
+    const snapshot = await channel.presence.get();
+    updatePresenceData(snapshot);
+
+    onPresenceMessageReceived?.call(this, message);
```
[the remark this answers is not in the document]
VeskeR: `.call()` removed in https://github.com/ably/ably-js/pull/1674/commits/8eae2d8ab5fce29b9e903e62f76cd6d978397cd5.
Fixed `onPresenceMessageReceived` not being in a dependency list in the same commit, based on the conversation above https://github.com/ably/ably-js/pull/1674#discussion_r1517552560.

src/platform/react-hooks/src/hooks/usePresenceListener.ts
```diff
@@ -0,0 +1,62 @@
...
+      ? channelNameOrNameAndOptions
+      : { channelName: channelNameOrNameAndOptions };
+  const skip = params.skip;
+
+  const { channel } = useChannelInstance(params.id, params.channelName);
+  const { connectionError, channelError } = useStateErrors(params);
+  const [presenceData, updatePresenceData] = useState<Array<PresenceMessage<T>>>([]);
+
+  const updatePresence = async (message?: Ably.PresenceMessage) => {
+    const snapshot = await channel.presence.get();
```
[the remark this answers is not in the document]
VeskeR: Created https://github.com/ably/ably-js/issues/1686 for this. Won't be part of this PR

src/platform/react-hooks/src/hooks/usePresenceEnter.ts
VeskeR: Most likely will implement ref counter to handle leaving presence intelligently. Created https://github.com/ably/ably-js/issues/1687 for this, won't be a part of this PR.

VeskeR:
@ttypic when removing `eslint-disable-next-line react-hooks/exhaustive-deps` hooks, I noticed that `usePresence` hook also doesn't declare 'messageOrPresenceObject' parameter (initial state when entering presence) as dependency in its useEffect/useCallback, meaning it won't re-enter when this parameter changes. It gets complicated by the fact that this parameter can be an object (passed by a ref) and users can call `usePresence` with an object literal `usePresence('channel', {foo:'bar'})` in their components, which would cause an infinite loop of rerenders if we add it as a dependency as is - on each render the new object literal is a different reference to an object.
I think we can solve it by using useRef and using deep equals to check if the value actually changed - if we even need to fix this issue.
This was the behavior before and I don't remember seeing issues raised by users regarding this problem, so we can solve it later.
Created for https://github.com/ably/ably-js/issues/1688 this.

src/platform/react-hooks/src/hooks/usePresence.ts
```diff
@@ -1,105 +1,62 @@
...
-
-  const updatePresence = async (message?: Ably.PresenceMessage) => {
-    const snapshot = await channel.presence.get();
-    updatePresenceData(snapshot);
+  const onMount = useCallback(async () => {
+    await channel.presence.enter(messageOrPresenceObject);
+    // don't add messageOrPresenceObject to dependency list, since it will most likely cause an infinite loop of updates,
+    // in cases when user calls this hook with an object literal instead of a state or memoized object.
+    // it does however prevent us from re-entering presence with new data automatically if it changes.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
ttypic, line 35: maybe it will be better to put `messageOrPresenceObject` in ref? I don't `// eslint-disable-next-line react-hooks/exhaustive-deps` :) can be dangerous when you change code

ttypic (approved):
Looks good, left small comment

src/platform/react-hooks/src/hooks/usePresence.ts
VeskeR: > // eslint-disable-next-line react-hooks/exhaustive-deps :) can be dangerous when you change code

Yeah, I thought about it too, don't remember why I decided to leave the warning suppression for now (maybe I didn't want to imlement the half-measure? not sure).

Using useRef won't solve the underlying issue - we still won't re-enter if `messageOrPresenceObject` changes.
However, if we introduce any other dependencies in later PRs, `eslint-disable-next-line react-hooks/exhaustive-deps` would suppress those helpful warning.

Will change to `useRef` and move comment over there to note that we don't actually react to `messageOrPresenceObject` changes.

VeskeR: Added `useRef` in https://github.com/ably/ably-js/pull/1674/commits/2454a287ceebcad3008b3024509b26c1c1804657

owenpearson (approved):
LGTM 👍

Issue #1677, opened 2024-03-08
Add `maxMessageSize` client option to typings in v1 and undo its removal in v2


lawrence-forooghian:
We removed it from v2 in https://github.com/ably/ably-js/commit/fe3c9ce49df62ffa357dec0843a3f3325ca7c65e because it was inside the `DeprecatedClientOptions` type. But there’s nothing deprecated about it; it's still in the spec (TO3l8) but for some reason was not added to the public typings in 228eabf25f8f913eb16ee3d08f36e572e7b16310. So stop calling it deprecated it in v1 and add it to the typings there, and undo its removal in v2.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4681) by [Unito](https://www.unito.io)

PR #1682, opened 2024-03-08

lawrence-forooghian:
Here’s how we handle the following commits that merge commit 1262a77 brings in from `main`:

- 36162c4: The content of this commit already exists on v2 (in commit eaee6f6), except for the deprecation warnings for the old API, which we remove in the current commit.

- 8f725ee: This commit un-deprecates the `maxMessageSize` client option, so we undo the part of fe3c9ce that relates to this option (as described in #1677).


VeskeR (approved):
LGTM

PR #1685, opened 2024-03-11
Improve deprecation log messages

lawrence-forooghian:
Add more detail that addresses what the user might have been trying to achieve with their use of the deprecated API, and more detail about how to change their code.

### DIFF
--- a/src/common/lib/util/defaults.ts
+++ b/src/common/lib/util/defaults.ts
@@ -196,7 +196,7 @@ export function normaliseOptions(options: DeprecatedClientOptions): NormalisedCl
     options.queueMessages = options.queueEvents;
   }
   if (options.headers) {
-    Logger.deprecatedWithMsg(
+    Logger.deprecated(
       'the `headers` client option',
       '' /* there is no replacement; see DeprecatedClientOptions.headers */
     );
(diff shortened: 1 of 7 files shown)


src/common/lib/util/defaults.ts
```diff
@@ -219,12 +219,15 @@ export function normaliseOptions(options: DeprecatedClientOptions): NormalisedCl
 
     /* emit an appropriate deprecation warning */
     if (options.environment) {
-      Logger.deprecatedWithMsg(
-        'fallbackHostsUseDefault',
-        'There is no longer a need to set this when the environment option is also set since the library will now generate the correct fallback hosts using the environment option.'
+      Logger.deprecated(
+        'The `fallbackHostsUseDefault` client option',
+        'Are you using this client option to force the library to make use of fallback hosts even though you’ve set the `environment` client option? If so, this is no longer necessary: remove your usage of the `fallbackHostsUseDefault` client option and the library will then automatically choose the correct fallback hosts to use for the specified environment.'
```
lmars, line 224: I'm not a fan of this being a question, how about:

```suggestion
        'If you are you using this client option to force the library to use default fallback hosts even though you’ve set the `environment` client option, then this is no longer necessary: remove your usage of the `fallbackHostsUseDefault` client option and the library will then automatically choose the correct fallback hosts to use for the specified environment.'
```

lawrence-forooghian: Thanks, have updated.

VeskeR (approved):
LGTM

src/common/lib/util/defaults.ts
lawrence-forooghian: Going to resolve this conversation.

PR #1690, opened 2024-03-13

lawrence-forooghian:
The conflicts here all resulted from the tweaks to deprecation messages (messages which v2 already removes) in `main`.

I’ve also changed the `Connection.recoveryKey` deprecation warning to be consistent with the warnings that are on `main`, and to account for 76200a6’s change to the output of the `Logger.deprecated` method.

### DIFF
--- a/src/common/lib/client/connection.ts
+++ b/src/common/lib/client/connection.ts
@@ -59,7 +59,9 @@ class Connection extends EventEmitter {
   }
 
   get recoveryKey(): string | null {
-    Logger.deprecated('Connection.recoveryKey attribute', 'Connection.createRecoveryKey() method');
+    Logger.deprecationWarning(
+      'The `Connection.recoveryKey` attribute has been replaced by the `Connection.createRecoveryKey()` method. Replace your usage of `recoveryKey` with the return value of `createRecoveryKey()`. `recoveryKey` will be removed in a future version.',
+    );
     return this.createRecoveryKey();
   }
 
--- a/src/common/lib/util/logger.ts
+++ b/src/common/lib/util/logger.ts
@@ -108,20 +108,27 @@ class Logger {
     }
   }
 
-  static deprecated = function (original: string, replacement: string) {
-    Logger.deprecatedWithMsg(original, "Please use '" + replacement + "' instead.");
+  static deprecated = (description: string, msg: string) => {
+    Logger.deprecationWarning(`${description} is deprecated and will be removed in a future version. ${msg}`);
   };
 
-  static deprecatedWithMsg = (funcName: string, msg: string) => {
+  static renamedClientOption(oldName: string, newName: string) {
+    Logger.deprecationWarning(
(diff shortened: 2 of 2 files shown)


src/common/lib/client/connection.ts
```diff
@@ -59,7 +59,9 @@ class Connection extends EventEmitter {
   }
 
   get recoveryKey(): string | null {
-    Logger.deprecated('Connection.recoveryKey attribute', 'Connection.createRecoveryKey() method');
+    Logger.deprecationWarning(
+      'The `Connection.recoveryKey` attribute has been replaced by the `Connection.createRecoveryKey()` method. Replace your usage of `recoveryKey` with the return value of `createRecoveryKey()`. `recoveryKey` will be removed in a future version.',
+    );
     return this.createRecoveryKey();
```
sacOO7, line 65: `createRecoveryKey` is public right? According to spec it should be public

lawrence-forooghian: Yes, it is — why do you ask?

src/common/lib/client/connection.ts
```diff
@@ -59,7 +59,9 @@ class Connection extends EventEmitter {
   }
 
   get recoveryKey(): string | null {
-    Logger.deprecated('Connection.recoveryKey attribute', 'Connection.createRecoveryKey() method');
+    Logger.deprecationWarning(
```
VeskeR, line 62: Isn't it better to use `Logger.deprecated` for deprecation messages? It adds `is deprecated and will be removed in a future version` automatically so there is no difference in wording between deprecation messages. Based on the recent commits it seemed to me like that was the intention.
So it would be
```javascript
Logger.deprecated(
  'The `Connection.recoveryKey` attribute',
  'It has been replaced by the `Connection.createRecoveryKey()` method. Replace your usage of `recoveryKey` with the return value of `createRecoveryKey()`.'',
)
```

VeskeR (commented):
LGTM, just a quick question about the intended Logger method to call for deprecation messages

src/common/lib/client/connection.ts
VeskeR: I see, so basically you choose .deprecated / .deprecationWarning / .renamedClientOption / .renamedMethod whichever sound best in the current context.

SGTM

src/common/lib/client/connection.ts
sacOO7: Actually, it wasn't before as per https://ably-real-time.slack.com/archives/C8SPU4589/p1707301881553349?thread_ts=1707230865.713579&cid=C8SPU4589

src/common/lib/client/connection.ts
lawrence-forooghian: Yeah, exactly! It's all pretty subjective, and depending on the time of day and phase of the moon I'd probably be easily convinced that what you suggested actually sounds better 🤷

PR #1691, opened 2024-03-13
build: use esbuild for nodejs bundle

owenpearson:


lawrence-forooghian:
Thanks @owenpearson, would you mind explaining the motivation for this PR? Is it something that needs to be merged before v2 release?

owenpearson:
@lawrence-forooghian The motivation is simply just faster build times for quick feedback when testing locally, I've been using these changes locally for a while now and figured it would be a nice improvement for any developers working in the project.

It's not a breaking change or anything but since it's a change to how we build for nodejs users I think it would make sense to land this before the v2 release as long as that's not gonna interrupt the release. I don't think it would be particularly risky to merge it later on either, so totally up to you both. I did previously speak to Andrii to check if there was any interest in having this in the v2 branch and he was in favour.

Gruntfile.js
```diff
@@ -102,12 +101,36 @@ module.exports = function (grunt) {
...
+
+    const nodeConfig = {
+      platform: 'node',
+      entryPoints: ['src/platform/nodejs/index.ts'],
+      outfile: 'build/ably-node.js',
+      bundle: true,
+      sourcemap: true,
+      format: 'umd',
+      banner: { js: '/*' + banner + '*/' },
+      footer: { js: 'module.exports = module.exports.default;' },
```
lawrence-forooghian, line 119: What does this line do?

Gruntfile.js
```diff
@@ -102,12 +101,36 @@ module.exports = function (grunt) {
...
-  grunt.registerTask('build:node', ['checkGitSubmodules', 'webpack:node']);
-
   grunt.registerTask('build:browser', ['checkGitSubmodules', 'webpack:browser']);
 
   grunt.registerTask('all', ['build', 'requirejs']);
 
+  grunt.registerTask('build:node', function () {
+    const done = this.async();
+
+    const nodeConfig = {
```
lawrence-forooghian, line 111: Is it possible to DRY up this and the stuff in `build:browser`, the same way as `webpack.config.js` has a `baseConfig` etc?

Gruntfile.js
owenpearson: So there's some boring issue with es module + commonjs interop where any default esm exports get exported on the `module.exports.default` field so that named exports can also be exported in commonjs. Since we're exporting everything necessary on the default export object and all our docs say to `const Ably = require('ably');` then this is fine and not a breaking change.

I agree that this is totally non-obvious though so I'll add a comment 👍

Gruntfile.js
owenpearson: good idea, will do!

lawrence-forooghian (approved):

Gruntfile.js
owenpearson: actually i thought about this a bit more and found a much more sensible solution :) [a676a7b](https://github.com/ably/ably-js/pull/1691/commits/a676a7b61ae4c61783995c409c15a32b72b25153)

lawrence-forooghian: Oh nice!

PR #1697, opened 2024-03-15
Change `PresenceEnterResult` to `PresenceResult` in react hooks

VeskeR:
This was wrongfully renamed in 730c621d6de3f76be727c18e528071937936a967. It is a remnant of a `usePresenceEnter` name for the new hook, which was temporary used during development in https://github.com/ably/ably-js/pull/1674, but we decided to keep the name `usePresence`.


ttypic (approved):
Nice spot!

PR #1701, opened 2024-03-19
Fix stray second arg to modular constructor

lawrence-forooghian:
Missed this in e8c3012.


owenpearson (approved):
LGTM 👍 

Also, since this test suite is an es module istm we should easily be able to enable type checking for development here. Might be different for other editors but in my neovim i can get a type warning for this issue by running `cp modular.d.ts build/modular/index.d.ts` and adding a `// @ts-check` comment to the top of `modular.test.js`.

lawrence-forooghian:
Thanks @owenpearson, created #1703 for that.

PR #1702, opened 2024-03-19
Improve errors for incorrect constructor args

### LINKED ISSUE #1700: Improve v2 experience when not passing client options
See [internal Slack message](https://ably-real-time.slack.com/archives/CURL4U2FP/p1710788497142189) from @owenpearson:

> minor regression in the js v2 branch - if you don't provide any arguments to the realtime constructor you get an error about "Cannot read properties of undefined (reading 'plugins')" so you'd never see [this error](https://github.com/ably/ably-js/blob/a3389a859ba71a894f18d7f25c4f7ad3296fd198/src/common/lib/client/baseclient.ts#L51-L55) which is more informative

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4702) by [Unito](https://www.unito.io)

lawrence-forooghian:
This fixes the regression that I introduced in e8c3012, which stopped the user from getting a useful error message when not passing any args to the constructor. I’ve also introduced a wider range of error messages to give the user useful guidance depending on which variant of the library they’re using.

Resolves #1700.

### DIFF
--- a/src/common/lib/client/defaultrealtime.ts
+++ b/src/common/lib/client/defaultrealtime.ts
@@ -22,14 +22,15 @@ import Defaults from '../util/defaults';
  `DefaultRealtime` is the class that the non tree-shakable version of the SDK exports as `Realtime`. It ensures that this version of the SDK includes all of the functionality which is optionally available in the tree-shakable version.
  */
 export class DefaultRealtime extends BaseRealtime {
-  constructor(options: ClientOptions | string) {
+  // The public typings declare that this requires an argument to be passed, but since we want to emit a good error message in the case where a non-TypeScript user does not pass an argument, tell the compiler that this is possible so that it forces us to handle it.
+  constructor(options?: ClientOptions | string) {
     const MsgPack = DefaultRealtime._MsgPack;
     if (!MsgPack) {
       throw new Error('Expected DefaultRealtime._MsgPack to have been set');
     }
 
     super(
-      Defaults.objectifyOptions(options, {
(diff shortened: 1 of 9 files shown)


src/common/lib/client/defaultrealtime.ts
```diff
@@ -22,14 +22,15 @@ import Defaults from '../util/defaults';
  `DefaultRealtime` is the class that the non tree-shakable version of the SDK exports as `Realtime`. It ensures that this version of the SDK includes all of the functionality which is optionally available in the tree-shakable version.
  */
 export class DefaultRealtime extends BaseRealtime {
-  constructor(options: ClientOptions | string) {
+  // The public typings declare that this requires an argument to be passed, but since we want to emit a good error message in the case where a non-TypeScript user does not pass an argument, tell the compiler that this is possible so that it forces us to handle it.
```
owenpearson, line 25: (this is super minor so feel free to ignore) we can probably dedupe comments in `DefaultRealtime` and `DefaultRest` by declaring an interface which they both implement

owenpearson (approved):
Looks good, thanks for fixing!

src/common/lib/client/defaultrealtime.ts
lawrence-forooghian: I started trying to implement this and it wasn't immediately obvious how to do it in a friction-free way so I gave up; I don't really want to spend more time on this so think we can live with the duplication.

PR #1716, opened 2024-03-27
build(node): mark ws as external module

owenpearson:
fixes an issue with realtime connections using the bun runtime

(also marks `got` as external just so we aren't pinned to a specific version)

### DIFF
--- a/grunt/esbuild/build.js
+++ b/grunt/esbuild/build.js
@@ -51,7 +51,7 @@ const nodeConfig = {
    * where the bun runtime has its own 'ws' module, and the npm 'ws' module
    * doesn't work at all.
    */
-  external: ['ws'],
+  external: ['ws', 'got'],
 };
 
 module.exports = {
--- a/grunt/esbuild/build.js
+++ b/grunt/esbuild/build.js
@@ -43,6 +43,15 @@ const nodeConfig = {
   platform: 'node',
   entryPoints: ['src/platform/nodejs/index.ts'],
   outfile: 'build/ably-node.js',
+  /**
+   * externals are not bundled into the esbuild output, instead they are left
+   * as commonjs imports. This means that the version installed via package.json
+   * dependencies is used rather than being pinned to whatever version we built
+   * the package release with. This is especially important in the case of 'ws'
+   * where the bun runtime has its own 'ws' module, and the npm 'ws' module
+   * doesn't work at all.
+   */
+  external: ['ws'],
 };
 
 module.exports = {


grunt/esbuild/build.js
```diff
@@ -43,6 +43,7 @@ const nodeConfig = {
   platform: 'node',
   entryPoints: ['src/platform/nodejs/index.ts'],
   outfile: 'build/ably-node.js',
+  external: ['ws', 'got'],
```
VeskeR, line 46: I think it would be beneficial to add a comment here explaining why we want to do this (based on that [slack convo](https://ably-real-time.slack.com/archives/C030C5YLY/p1711542833244969?thread_ts=1711541241.653079&cid=C030C5YLY)).
We already had 'ws' and 'got' marked as externals in v1: https://github.com/ably/ably-js/blob/1.2.50/webpack.config.js#L60-L61 when we were building node using webpack (I assume that's why v1 works in bun), but we lost this config when migrated to esbuild.
With a comment explaining what issues it fixes it would be less likely we remove it accidentally.

owenpearson: good idea, i've added a comment explaining this now 👍

grunt/esbuild/build.js
```diff
@@ -43,6 +43,15 @@ const nodeConfig = {
   platform: 'node',
   entryPoints: ['src/platform/nodejs/index.ts'],
   outfile: 'build/ably-node.js',
+  /**
+   * externals are not bundled into the esbuild output, instead they are left
+   * as commonjs imports. This means that the version installed via package.json
+   * dependencies is used rather than being pinned to whatever version we built
+   * the package release with. This is especially import in the case of 'ws' where
```
VeskeR, line 50:
```suggestion
   * the package release with. This is especially important in the case of 'ws' where
```

VeskeR: might need to fix line width after this too

VeskeR (changes requested):

grunt/esbuild/build.js
VeskeR: great!

grunt/esbuild/build.js
owenpearson: whooops, fixed it now

VeskeR (approved):
LGTM

VeskeR:
Merging this PR on Owen's behalf, as he is out of the office, and we want to release this fix as soon as possible.

PR #1722, opened 2024-04-05
[ECO-4113] Add missing `signal` option when using AbortController with fetch API

### LINKED ISSUE #1623: `fetchRequest` incorrectly uses `AbortController`
VeskeR pointed this out in https://github.com/ably/ably-js/pull/1603#discussion_r1489865984:

> I've noticed that we don't actually use AbortController correctly here.
> We're supposed [to pass AbortController.signal to fetch function](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) options (`fetch(url, { signal })`), but we don't do that, we only call controller.abort() when timeout occurs which does nothing for us.
> Seems like adding `signal` prop to `requestInit` object below would solve the issue:
> ```javascript
>   const requestInit: RequestInit = {
>     method: _method,
>     headers: fetchHeaders,
>     body: body as any,
>     signal: controller.signal,
>   };
> ```

┆Issue is synchronized with this [Jira Bug](https://ably.atlassian.net/browse/ECO-4113) by [Unito](https://www.unito.io)

VeskeR:
Resolves #1623

### DIFF
--- a/src/platform/web/lib/http/request/fetchrequest.ts
+++ b/src/platform/web/lib/http/request/fetchrequest.ts
@@ -44,6 +44,8 @@ export default async function fetchRequest(
     timeout = setTimeout(
       () => {
         controller.abort();
+        // When AbortController.abort() is called, the fetch() promise rejects with a DOMException named AbortError (source: https://developer.mozilla.org/en-US/docs/Web/API/AbortController)
+        // However, we beat it in the Promise.race() by resolving our custom 'Request timed out' error on the next line, thereby exposing users to the better-formatted error.
         resolve({ error: new PartialErrorInfo('Request timed out', null, 408) });
       },
       client ? client.options.timeouts.httpRequestTimeout : Defaults.TIMEOUTS.httpRequestTimeout,
@@ -54,6 +56,7 @@ export default async function fetchRequest(
     method: _method,
     headers: fetchHeaders,
     body: body as any,
+    signal: controller.signal,
   };
 
   if (!Platform.Config.isWebworker) {


src/platform/web/lib/http/request/fetchrequest.ts
```diff
@@ -54,6 +54,7 @@ export default async function fetchRequest(
     method: _method,
     headers: fetchHeaders,
     body: body as any,
+    signal: controller.signal,
```
lawrence-forooghian, line 57: What is the point of `controller`? We don't ever call `abort()` on it so I’m not sure what it's achieving.

Ways in which it might be sensible to use `abort()`:

- to cancel in-flight requests when, for example, the underlying `Realtime` instance is closed — I’m not sure if we do this for any of our other HTTP request implementations though
- to provide users with a way to explicitly cancel an in-flight request — we have no API for this

lawrence-forooghian: Oops, I’m talking nonsense — we do call `abort()` on it, sorry!

lawrence-forooghian: Another thing, from [here]():

> When `abort()` is called, the `fetch()` promise rejects with a `DOMException` named `AbortError`.

Just wanted to check that our `resolve({ error: new PartialErrorInfo('Request timed out', null, 408) });` beats the `AbortError` in becoming the error that is exposed to the user? Looks like it should, might be worth putting a comment alongside it though.

VeskeR: > Oops, I’m talking nonsense — we do call abort() on it, sorry!
Correct! That was the reason behind the issue, we did call abort() before, but never passed the controller signal, so it did nothing.

VeskeR: > Just wanted to check that our resolve({ error: new PartialErrorInfo('Request timed out', null, 408) }); beats the AbortError in becoming the error that is exposed to the user? Looks like it should, might be worth putting a comment alongside it though.

I tested it locally before, `resolve({ error: new PartialErrorInfo('Request timed out', null, 408) });` resolves first so we get our formatted error as expected. Will add a comment for that, thanks!

VeskeR: Added comment in https://github.com/ably/ably-js/compare/ca0b1db206eaa758fd1ab45a4aa0253a6f6583ec..d418952a56cfe168e515ab9455661991c6456939

Issue #1699, opened 2024-03-18
Compilation Failure with Ably on Next.js --turbo


Elmosh:
I've been integrating Ably with a Next.js application and encountered a lot of development warnings initially, which escalated to a compilation error when using Next.js turbo. This issue persists regardless of the package manager used; switching from pnpm to npm did not resolve the problem.

Given that Turbo is achieving a [99% pass rate](https://areweturboyet.com
) on next dev tests, it seems likely that Ably might soon face compatibility issues with a significant portion of Next.js projects if this issue remains unaddressed.

**Error Details:**
When attempting to compile, i get this error:

```
Failed to compile
./node_modules/keyv/src/index.js:22:14
Module not found
  20 | 	if (options.adapter || options.uri) {
  21 | 		const adapter = options.adapter || /^[^:+]*/.exec(options.uri)[0];
> 22 | 		return new (require(adapters[adapter]))(options);
     | 		            ^^^^^^^^^^^^^^^^^^^^^^^^^^
  23 | 	}
  24 |
  25 | 	return new Map();

```

**Attempted Resolutions:**

* Switching between pnpm and npm as the package manager.
* Ensuring that the latest versions of all related packages are in use.
* Ignoring or Overrides didn't fix the errors: 

```
config.ignoreWarnings = [{ module: /node_modules\/keyv\/src\/index\.js/ }];
```

```
"overrides": {
    "cacheable-request": {
      "keyv": "npm:@keyvhq/core@~1.6.6"
    }
  }
```

* Uninstalling Ably fixes the errors.

I am reaching out for guidance on resolving this issue or to report it in case it's not already on your radar. Any advice on potential workarounds or fixes would be greatly appreciated, as the inability to use --turbo significantly impacts our development process.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4701) by [Unito](https://www.unito.io)

VeskeR:
Hello @Elmosh , sorry for the late reply. Thank you for bringing our attention to this issue!

I was able to reproduce the issue locally when using turbopack with npm/pnpm, specifically when using next.js App Router (not Pages). I was able to fix the issue with adding `ably` to the list of next.js `serverComponentsExternalPackages` in `next.config.js` (read more [here](https://nextjs.org/docs/app/api-reference/next-config-js/serverComponentsExternalPackages)):
```javascript
const nextConfig = {
  // ...
  experimental: {
    serverComponentsExternalPackages: ['ably'],
[...]

PR #1723, opened 2024-04-05
Add README explaining possible issues when using `ably-js` with Next.js App Router and Turbopack

VeskeR:
This is based on the issue and solution found in https://github.com/ably/ably-js/issues/1699#issuecomment-2022702584


ttypic (approved):
Good research!

PR #1725, opened 2024-04-08
channelgroup: emit channel assigned/active events

mschristensen:
Makes the `ChannelGroup` object an event emitter that emits events when the set of active or assigned channels changes. This allows users to observe the set of changes to the active channels to drive UI updates.


VeskeR (approved):
LGTM

Issue #1732, opened 2024-04-15
Make ably-js throw a readable error when using Realtime client in the Vercel Edge runtime


VeskeR:
Currently, when trying to use the `Ably.Realtime` instance in the [Vercel Edge runtime](https://vercel.com/docs/functions/runtimes/edge-runtime) - whether within serverless functions or in SSR when rendering the frontend in Edge runtime - we get the error `Error: A Node.js API is used (setImmediate) which is not supported in the Edge Runtime.`.
This error either prevents the serverless function from running or results in a white screen of death when rendering the frontend page.

The issue comes from the assignment of the `setImmediate` function to the `IPlatformConfig.nextTick` method in the web platform bundle. We do this when `setImmediate` function is defined on the global object on the [next lines](https://github.com/ably/ably-js/blob/40d362198d12beba4f4db03ddecee4503aed1918/src/platform/web/config.ts#L53-L58). Consequently, `IPlatformConfig.nextTick` is called in various transport and websocket-related classes like `Connection` and `WebSocketTransport` during the regular `Ably.Realtime` workflow, leading to the error above.

Here's exactly what's happening:

1. When importing a package in the Vercel Edge runtime, Vercel selects a `browser` bundle from the available bundles provided by the package in node_modules, as explained by a Vercel staff member in this [GitHub issue](https://github.com/vercel/next.js/discussions/41235#discussioncomment-3825756):
> So you are correct in that we decided to pick the `browser` field for the edge runtime, as we felt it was closer to what the runtime was capable of than Node.js (since you cannot use Node.js APIs in Edge, and browsers are usually fully spec-compliant also, compared to polyfills/packages.)

For ably-js, this means that in Vercel Edge runtimes, our web platform bundle will be selected, which includes the optional usage of the `setImmediate` function.

2. The Vercel Edge runtime defines `setImmediate` function instead of leaving it undefined. It sets it to be the function that just throws `Error: A Node.js API is used (setImmediate) which is not supported in the Edge Runtime.` error immediately when called.

3. Since `setImmediate` function is actually defined, no polyfill will work out of the box, as polyfills usually rely on condition checks like `if (!global.setImmediate) { // define polyfill }`. For example, [this](https://github.com/YuzuJS/setImmediate/tree/master) popular `setImmediate` polyfill won't work.

4. Because `setImmediate` function is actually defined, our web bundle code sets `IPlatformConfig.nextTick` to the value of the `setImmediate` function, which eventually throws an error.

[...]

Issue #1731, opened 2024-04-15
Fix errors when receiving WebSocket data in Vercel Edge runtime and indefinitely creating new connections


VeskeR:
When using an `Ably.Realtime` instance in the [Vercel Edge runtime](https://vercel.com/docs/functions/runtimes/edge-runtime) - whether within serverless functions or in SSR when rendering the frontend in Edge runtime - the connection to Ably servers sets up correctly. However, we begin to get `TypeError: Cannot read properties of null (reading 'length')` errors continuously after the connection has been set up. Here's the full error log:

```
TypeError: Cannot read properties of null (reading 'length')
    at WebSocketTransport.onWsData (webpack-internal:///(middleware)/./node_modules/ably/build/ably.js:6373:131)
    at wsConnection.onmessage (webpack-internal:///(middleware)/./node_modules/ably/build/ably.js:6342:31)
    at [nodejs.internal.kHybridDispatch] (node:internal/event_target:826:20)
    at WebSocket.dispatchEvent (node:internal/event_target:761:26)
    at fireEvent (eval at requireWithFakeGlobalScope (C:\Users\Andrew\dev\ably\ably-js-next-example\node_modules\next\dist\compiled\edge-runtime\index.js:1:657096), <anonymous>:12687:14)
    at websocketMessageReceived (eval at requireWithFakeGlobalScope (C:\Users\Andrew\dev\ably\ably-js-next-example\node_modules\next\dist\compiled\edge-runtime\index.js:1:657096), <anonymous>:12709:7)
    at _ByteParser.run (eval at requireWithFakeGlobalScope (C:\Users\Andrew\dev\ably\ably-js-next-example\node_modules\next\dist\compiled\edge-runtime\index.js:1:657096), <anonymous>:13122:17)
    at _ByteParser._write (eval at requireWithFakeGlobalScope (C:\Users\Andrew\dev\ably\ably-js-next-example\node_modules\next\dist\compiled\edge-runtime\index.js:1:657096), <anonymous>:13001:14)
    at writeOrBuffer (node:internal/streams/writable:564:12)
    at _write (node:internal/streams/writable:493:10)
 ⨯ uncaughtException: TypeError: Cannot read properties of null (reading 'length')
    at WebSocketTransport.onWsData (webpack-internal:///(middleware)/./node_modules/ably/build/ably.js:6373:131)
    at wsConnection.onmessage (webpack-internal:///(middleware)/./node_modules/ably/build/ably.js:6342:31)
[...]

PR #1736, opened 2024-04-15
Throw error when instantiating Realtime instances in Vercel Edge runtime

VeskeR:
This is a temporary fix for https://github.com/ably/ably-js/issues/1732, until we fix fully https://github.com/ably/ably-js/issues/1731


ttypic (approved):
LGTM

Issue #1739, opened 2024-04-16
Update docs about using Ably in Chrome Extension 


ttypic:
We mentioned some unnecessary workarounds to use Ably in Chrome extensions. However, starting from Chrome 116, nothing needs to be done, and Ably works with Chrome extensions out-of-the-box.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4749) by [Unito](https://www.unito.io)

VeskeR:
More context in internal slack thread: https://ably-real-time.slack.com/archives/C030C5YLY/p1702996856568239

PR #1743, opened 2024-04-17
docs: update Ably JS in Chrome extension

ttypic:
Resolves https://github.com/ably/ably-js/issues/1739
Jira issue [ECO-4749](https://ably.atlassian.net/browse/ECO-4749)

[ECO-4749]: https://ably.atlassian.net/browse/ECO-4749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


README.md
```diff
@@ -511,9 +511,18 @@ This library currently does not support being the [target of a push notification
 
 - ["Unable to parse request body" error when publishing large messages from old versions of Internet Explorer](https://support.ably.com/solution/articles/3000062360-ably-js-unable-to-parse-request-body-error-when-publishing-large-messages-from-old-browsers).
 
-#### Manifest v3 Chrome Extensions
-Chrome extensions built with Manifest v3 require service workers instead of background pages.
-This is supported in Ably via the [Web Worker build](#supported-platforms), however [workarounds](docs/chrome-mv3.md) are required to ensure Chrome does not mark the service worker as inactive.
+#### Chrome Extensions
+Ably supports Chrome extensions out-of-the-box since v2. However, extension service workers get improved support for
```
ttypic, line 515: Probably we need to rephrase that it's only if you use Ably in extension's service worker (any suggestions how to rewrite will be very welcome)

owenpearson, line 515: It's ambiguous whether you're talking about ably-js v2 or manifest v2, moreover we should mention that it works because you can load ably-js in background scripts.
```suggestion
ably-js works out-of-the-box in background scripts for Chrome extensions using manifest v2. However, extension service workers get improved support for
```

Moreover,

README.md
```diff
@@ -511,9 +511,18 @@ This library currently does not support being the [target of a push notification
...
+WebSockets only starting with Chrome 116. Therefore, it is necessary to ensure that the extension runs only on Chrome
+versions supporting WebSockets in service workers by setting the minimum Chrome version to 116 in the manifest:
```
owenpearson, line 517: A few things I think we needed to change here:
- make it clear that the reason for this advice is that manifest v3 only supports service workers, not background scripts
- make it clear that this advice only affect realtime clients, not rest clients
- explain what the caveats are if you run a realtime client in a service worker in older versions of chrome
```suggestion
Ably supports Chrome extensions out-of-the-box since v2. However, since manifest v3 background pages are no longer supported so you will need to run ably-js inside a service worker. 
If you are using an ably-js realtime client in a service worker, note that in version of Chrome before 116 active WebSockets would not reset the 30s service worker idle timer which would result in the client being closed prematurely, however in versions 116 and above, service workers will stay active as long as a client is connected.
You can ensure that your extension only runs in versions 116 and above by adding the following to your `manifest.json`:
```

owenpearson (changes requested):

README.md
```diff
@@ -511,9 +511,18 @@ This library currently does not support being the [target of a push notification
 
 - ["Unable to parse request body" error when publishing large messages from old versions of Internet Explorer](https://support.ably.com/solution/articles/3000062360-ably-js-unable-to-parse-request-body-error-when-publishing-large-messages-from-old-browsers).
 
-#### Manifest v3 Chrome Extensions
-Chrome extensions built with Manifest v3 require service workers instead of background pages.
-This is supported in Ably via the [Web Worker build](#supported-platforms), however [workarounds](docs/chrome-mv3.md) are required to ensure Chrome does not mark the service worker as inactive.
+#### Chrome Extensions
+ably-js works out-of-the-box in background scripts for Chrome extensions using manifest v2. However, since manifest v3 background pages are no longer supported so you will need to run ably-js inside a service worker.
+If you are using an ably-js realtime client in a service worker, note that in version of Chrome before 116 active WebSockets would not reset the 30s service worker idle timer which would result in the client being closed prematurely, however in versions 116 and above, service workers will stay active as long as a client is connected.
```
VeskeR, line 516:
```suggestion
If you are using an ably-js realtime client in a service worker, note that in versions of Chrome before 116, active WebSockets would not reset the 30s service worker idle timer, resulting in the client being closed prematurely, however, in versions 116 and above, service workers will stay active as long as a client is connected.
```

VeskeR (approved):
LGTM, minor wording change suggestion

owenpearson (approved):

PR #1746, opened 2024-04-17
Test app manager: respect ablyGlobals settings re port/tls

SimonWoolf:
allows tests to run against a local realtime with `ABLY_ENV=local ABLY_PORT=8081 ABLY_USE_TLS=false`
(which used to work, I think it was broken by the change to get rid of the use of self-signed certs for local dev)


ttypic (approved):
LGTM

PR #1750, opened 2024-04-24
[ECO-4737, ECO-4756] Fix auth token encoding in React Native

### LINKED ISSUE #1730: Invalid accessToken in request using history in react native
Description:
When using Ably token authentication in ably-js v2 in react native and calling history the request fails with an error
`Invalid accessToken in request`. This is true whether an ably token or tokenRequest is returned to the client

Steps to reproduce:

1. Instantiate an ably client in react native (tested with "react-native": "0.73.6" and expo": "^50.0.15") using authUrl or authCallback client options which returns an Ably token to the client
2. Make a call to history on a channel

Expected behaviour:
No error and history response is returned to the client

Actual behaviour:
The bearer token in the authorization request header is partly corrupt and is refused by the ably REST API
image

Workaround: 

1. Downgrade to 1.2.50 where issue is not present
2. Use JWT authentication https://ably.com/docs/auth/token?lang=javascript#standard

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4737) by [Unito](https://www.unito.io)

### LINKED ISSUE #1749: History API error: Invalid accessToken in request: xyz.F5UkPAOLTbMV4emxqgxzCb5...
Since v2, I've been getting this invalid accessToken error whenever I call this `channel.history()` API.

Note that the same accessToken is used across all other APIs in my app and all work well, except the history API call, so yes, I think my API key is valid. I also tried with a newly created API key but to no avail.

```bash
01:04:49.540 Ably: PaginatedResource.handlePage(): Unexpected error getting resource: err = [ErrorInfo: Invalid accessToken in request: xyz.F5UkPAOLTbMV4emxqgxzCb5...; statusCode=400; code=40005; see https://help.ably.io/error/40005 ]
```

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4756) by [Unito](https://www.unito.io)

VeskeR:
Resolves #1730 , resolves #1749 

Fixes incorrect usage of the underlying ArrayBuffer buffer in web BufferUtils, where we didn't account for the `byteOffset` and `byteLength` of the original array buffer view. This resulted in us incorrectly encoding auth tokens in the React Native bundle, where we use a polyfill for the TextEncoder class, which returned Uint8Array as a subarray of the underlying ArrayBuffer (see [polyfill's code](https://github.com/anonyco/FastestSmallestTextEncoderDecoder/blob/f04c88cf11024672e4d3e3dd59b9530b50eefbed/EncoderDecoderTogether.src.js#L268)).

### DIFF
--- a/src/common/types/IBufferUtils.ts
+++ b/src/common/types/IBufferUtils.ts
@@ -2,7 +2,9 @@ export default interface IBufferUtils<Bufferlike, Output, ToBufferOutput> {
   base64CharSet: string;
   hexCharSet: string;
   isBuffer: (buffer: unknown) => buffer is Bufferlike;
-  // On browser this returns a Uint8Array, on node a Buffer
+  /**
+   * On browser this returns a Uint8Array, on node a Buffer
+   */
   toBuffer: (buffer: Bufferlike) => ToBufferOutput;
   toArrayBuffer: (buffer: Bufferlike) => ArrayBuffer;
   base64Encode: (buffer: Bufferlike) => string;
@@ -13,6 +15,9 @@ export default interface IBufferUtils<Bufferlike, Output, ToBufferOutput> {
   utf8Decode: (buffer: Bufferlike) => string;
   areBuffersEqual: (buffer1: Bufferlike, buffer2: Bufferlike) => boolean;
   byteLength: (buffer: Bufferlike) => number;
+  /**
+   * Returns ArrayBuffer on browser and Buffer on Node.js
+   */
   arrayBufferViewToBuffer: (arrayBufferView: ArrayBufferView) => Bufferlike;
   hmacSha256(message: Bufferlike, key: Bufferlike): Output;
 }
--- a/src/platform/nodejs/lib/util/bufferutils.ts
+++ b/src/platform/nodejs/lib/util/bufferutils.ts
@@ -56,7 +56,7 @@ class BufferUtils implements IBufferUtils<Bufferlike, Output, ToBufferOutput> {
   }
 
   arrayBufferViewToBuffer(arrayBufferView: ArrayBufferView): Buffer {
-    return this.toBuffer(arrayBufferView.buffer);
+    return this.toBuffer(arrayBufferView);
   }
 
   utf8Decode(buffer: Bufferlike): string {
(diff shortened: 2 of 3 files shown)


ttypic (approved):
LGTM

lawrence-forooghian:
Continuing our [conversation on Slack](https://ably-real-time.slack.com/archives/CURL4U2FP/p1714493728320409?thread_ts=1714483172.339229&cid=CURL4U2FP) where I asked why you'd switched web’s output type from `ArrayBuffer` to `Uint8Array`.

You said that going back and forth between `Uint8Array` and `ArrayBuffer` is error-prone because you have to make sure to remember to use `byteOffset` and `byteLength` when accessing a `Uint8Array`’s underlying storage. Fair enough, but I guess that could be solved by just having a method in `BufferUtils` for converting a `Uint8Array` to an `ArrayBuffer` and making sure that we always use it?

In terms of what's the more "correct" type to use, you said that using `Uint8Array` makes it more consistent with Node’s `Buffer`, which is a subclass of `Uint8Array`. Fair enough, although it does seem like that, out of `ArrayBuffer` and `Uint8Array`, the web platform usually uses `ArrayBuffer` as its binary data type — see, e.g. [`WebSocket`’s `binaryType` property](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/binaryType) or [`SubtleCrypto`’s `exportKey`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/exportKey). [MDN’s documentation for `ArrayBuffer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer) also says "The `ArrayBuffer` object is used to represent a generic raw binary data buffer", which sounds like what we want.

Just some thoughts, will leave it to you to decide what to do 🙂

VeskeR:
> I guess that could be solved by just having a method in BufferUtils for converting a Uint8Array to an ArrayBuffer and making sure that we always use it?

Agree, I thought about this yesterday after your comments and started leaning towards this approach as well.
It will achieve the same goal and additionally ensure that if the consumer of the `BufferUtils` gets an array, they won't be able to access its `.buffer` property incorrectly (without using `byteOffset` and `byteLength`), as there is simply no such property on `ArrayBuffer` (and this could happen if we return `Uint8Array` from `BufferUtils`).

I'll go ahead and change it so we always convert to `ArrayBuffer` using the dedicated method which knows about `byteOffset` and `byteLength`.

VeskeR:
@lawrence-forooghian Updated PR to always convert to ArrayBuffer using dedicated method in https://github.com/ably/ably-js/pull/1750/commits/a20f20b6ba40d5abffb3c982fdae1ebcd449117c

lawrence-forooghian (approved):

PR #1758, opened 2024-05-03
[WIP] Update `got` to v12

VeskeR:
`got@12` is the latest version of `got` we can update to without making very drastic changes to the codebase.
Installing `got@13` will cause our TS import statements to fail with:
```
Cannot find module 'got' or its corresponding type declarations.
  There are types at 'node_modules/got/dist/source/index.d.ts', but this result could not be resolved under your current 'moduleResolution' setting. Consider updating to 'node16', 'nodenext', or 'bundler'.
```
and tinkering with `module` and `moduleResolution` in `tsconfig.json` opens up another can of worms which I'm not sure justifies trying to upgrade to `got@13`

### DIFF
--- a/grunt/esbuild/build.js
+++ b/grunt/esbuild/build.js
@@ -52,6 +52,8 @@ const nodeConfig = {
    * doesn't work at all.
    */
   external: ['ws', 'got'],
+  format: 'cjs',
+  plugins: [],
 };
 
 module.exports = {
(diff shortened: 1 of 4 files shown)


grunt/esbuild/build.js
```diff
@@ -52,6 +52,8 @@ const nodeConfig = {
    * doesn't work at all.
    */
   external: ['ws', 'got'],
+  format: 'cjs',
```
VeskeR, line 55: If we bundle node.js as an UMD, we will end up with calling `require('got')` in UMD header, which inevitably fails with `[ERR_REQUIRE_ESM]` error since got@12 is pure ESM and can't be `require`d.
So one option is to stop using UMD for node.js bundle and just bundle it as CJS. Then we will only `await import('got')` whenever we actually need it

Issue #1759, opened 2024-05-04
`usePresence` errors on v2 after staying on the same page for a long time


Elmosh:
Hey @VeskeR ,  following up on this issue Error: Connection closed #1753.

I have been using ably for a while, but after upgrading to v2 there’s a lot of errors popped up. 

Along with the error in the mentioned issue that happens on page reload, there’s two more errors, they happens suddenly after staying on the same page for a long time (30+minutes) or to reproduce much faster; open on mobile then minimizing the browser for a couple of minutes and once it’s back opened the two errors appear:

```
Uncaught (in promise) Error: Unable to perform operation on channel: online-channel (not currently attached)
    at _ErrorInfo.fromValues (errorinfo.ts:54:34)
    at fromDeserialized (protocolmessage.ts:84:45)
    at deserialize (protocolmessage.ts:76:10)
    at WebSocketTransport.onWsData (websockettransport.ts:132:9)
    at wsConnection.onmessage (websockettransport.ts:81:18)
```

```
Presence Channel Error: Error: Channel detached
    at _RealtimeChannel.processMessage (realtimechannel.ts:524:13)
    at Channels2.processChannelMessage (baserealtime.ts:143:19)
    at _ConnectionManager.processChannelMessage (connectionmanager.ts:1809:34)
    at _ConnectionManager.processNextPendingChannelMessage (connectionmanager.ts:1793:12)
    at _ConnectionManager.onChannelMessage (connectionmanager.ts:1784:12)
    at WebSocketTransport.onProtocolMessage (transport.ts:198:32)
    at WebSocketTransport.onWsData (websockettransport.ts:131:12)
    at wsConnection.onmessage (websockettransport.ts:81:18)
```

Here’s the ably code (I’m using next with turbo):

provider:
```js
"use client";

import * as Ably from "ably";
import { AblyProvider, ChannelProvider } from "ably/react";
import { ReactNode, useEffect, useState } from "react";

interface Props {
  currentUser: string | null;
  children: ReactNode;
}
const AblyContext: React.FC<Props> = ({ currentUser, children }) => {
  const [client, setClient] = useState<InstanceType<
    typeof Ably.Realtime
  > | null>(null);
  
    useEffect(() => {
    const newClient = new Ably.Realtime({
      authUrl: "/api/ablyToken",
      authMethod: "POST",
      clientId: currentUser!,
    });

    setClient(newClient);

    return () => {
      if (newClient) {
              newClient.close();
      }
    };
[...]

Issue #1753, opened 2024-04-26
usePresence error - Uncaught (in promise) Error: Connection closed


bookercodes:
When using `usePresence`, if I reload the page, I see the following error (with preserve log on) 

```
ably.js:260 Uncaught (in promise) Error: Connection closed
    at _ErrorInfo.fromValues (webpack-internal:///(app-pages-browser)/./node_modules/ably/build/ably.js:260:34)
    at Object.closed (webpack-internal:///(app-pages-browser)/./node_modules/ably/build/ably.js:3697:27)
    at WebSocketTransport.close (webpack-internal:///(app-pages-browser)/./node_modules/ably/build/ably.js:3749:52)
    at _ConnectionManager.closeImpl (webpack-internal:///(app-pages-browser)/./node_modules/ably/build/ably.js:5113:42)
    at _ConnectionManager.requestState (webpack-internal:///(app-pages-browser)/./node_modules/ably/build/ably.js:4898:12)
    at eval (webpack-internal:///(app-pages-browser)/./node_modules/ably/build/ably.js:4197:16)
```

I wonder if maybe ably-js is trying to send an event like "leave" but the connection is already closed? Possibly related to [this](https://github.com/ably/ably-js/blob/main/src/platform/react-hooks/src/hooks/usePresence.ts#L45) LOC?

The code to reproduce is very simple: 

```js
import { usePresence } from "ably/react"

const WhosOnlineList = ({ channelName }) => {

  usePresence(channelName, { fullName: Math.random() })


  return null
}
export default WhosOnlineList
```

Why does this matter?

It might not. I don't think you can technically be "present" on a channel if the connection has closed, but I'd like to understand the error better in case I should be structuring my React component differently. 

Thank you!

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4780) by [Unito](https://www.unito.io)

VeskeR:
Hello @bookercodes , @nikitadubyk !

Thank you for reporting this and providing steps to reproduce. We're going to investigate this and get back with an update as soon as possible!

VeskeR:
Hey @bookercodes , @nikitadubyk !

I was able to reproduce the issue with the provided code sample. I'm currently looking into its cause.

Elmosh:
Same errors on page reload after updating to v2.

VeskeR:
Hey @bookercodes !
[...]

Issue #1565, opened 2024-01-08
"Error: Connection closed" even when connection is open


TranquilMarmot:
In our Next.js app, we're creating / closing an Ably realtime client inside of a `useEffect`.

```ts
const createAblyRealtimeClient = (authParams?: Record<string, string>) => {
  return new Realtime({
    authUrl: AUTH_URL,
    authMethod: "POST",
    authParams,
  });
};

export const useAblyRealtimeClient = (authParams?: Record<string, string>) => {
  const clientRef = useRef<Realtime | null>(null);

  useEffect(() => {
    // don't create the ably client on the server because it will cause errors
    if (typeof window !== "undefined") {
      clientRef.current = createAblyRealtimeClient(authParams);
    }

    return () => {
      // close the Ably connection when the component unmounts
      if (
        clientRef.current &&
        ["initialized", "connected", "connecting"].includes(
          clientRef.current?.connection.state,
        )
      ) {
        try {
          clientRef.current?.close();
        } catch (error) {
          console.error("Error closing Ably client", error);
        }
      }

      clientRef.current = null;
    };
  }, [authParams?.chatId]);

  return clientRef.current;
};
```

This client is then passed as a prop to an `AblyProvider`.

However, this still causes the Ably client to throw an error. Next.js will show us this:
![image](https://github.com/ably/ably-js/assets/504259/9e1e4921-f382-433e-a6d2-a72bc218064e)

More of a stack trace...
```
Uncaught (in promise) Error: Connection closed
    at new ErrorInfo (webpack-internal:///(app-pages-browser)/../../node_modules/.pnpm/ably@1.2.48_react-dom@18.2.0_react@18.2.0/node_modules/ably/build/ably-commonjs.js:1154:28)
    at ErrorInfo.fromValues (webpack-internal:///(app-pages-browser)/../../node_modules/.pnpm/ably@1.2.48_react-dom@18.2.0_react@18.2.0/node_modules/ably/build/ably-commonjs.js:1171:36)
    at Object.closed (webpack-internal:///(app-pages-browser)/../../node_modules/.pnpm/ably@1.2.48_react-dom@18.2.0_react@18.2.0/node_modules/ably/build/ably-commonjs.js:5172:36)
    at Transport.close (webpack-internal:///(app-pages-browser)/../../node_modules/.pnpm/ably@1.2.48_react-dom@18.2.0_react@18.2.0/node_modules/ably/build/ably-commonjs.js:5513:58)
[...]

PR #1760, opened 2024-05-14
Fix `ConnectionManager` reusing error object stored in `errorReason`

### LINKED ISSUE #1573: Re-used `errorReason`s have misleading stack traces
Currently we re-use certain state errors by storing them in an `errorReason` field on the relevant state machine class. In the case of `ConnectionManager` we actually re-use these errors (run `git grep getError` to see examples), but since javascript error stack traces point to where the `Error` was _constructed_ rather than where it was _thrown_, this means that whenever we throw the stored `errorReason` it has a misleading stack trace.

For example, if you close a client and then attempt to attach to a channel, the promise will reject with an exception pointing to where you closed the client (rather than where you tried to attach to the channel, as it should do).

┆Issue is synchronized with this [Jira Bug](https://ably.atlassian.net/browse/ECO-4130) by [Unito](https://www.unito.io)

VeskeR:
Resolves #1573

This helps with troubleshooting issues like https://github.com/ably/ably-js/issues/1753, https://github.com/ably/ably-js/issues/1565, https://github.com/ably/ably-js/issues/1759


src/common/lib/transport/connectionmanager.ts
```diff
@@ -916,7 +916,13 @@ class ConnectionManager extends EventEmitter {
    *********************/
 
   getError(): IPartialErrorInfo | string {
-    return this.errorReason || this.getStateError();
+    if (this.errorReason) {
+      // create new PartialErrorInfo so it has the correct stack trace
+      // which points to the place which caused us to return this error.
+      return PartialErrorInfo.fromValues(this.errorReason);
```
owenpearson, line 922: Would it be better to wrap the existing error as the `cause` of the returned error, so that you have both stack traces? (not saying it would be, just interested to know your thoughts)

VeskeR: It didn't feel like it would provide any useful additional information here.
To me, it seems that you would want to use the `cause` field in situations where you are `try-catch`ing some expected error and then wrapping it to add more contextual information about what happened. In this specific case, we just want to create a new instance of `PartialErrorInfo` to get the correct stack trace; we don't have any additional context to provide.

owenpearson: well i was imagining the additional context in this case being where the connection error originated. for example, if you're trying to attach to a channel when your connection is closed, the error would tell you where you tried to attach to the channel, and the error.cause would tell you where you closed the connection - both useful bits of information for debugging.

VeskeR: Well, I guess there is no harm in adding that info. Currently, though, only Firefox displays the `cause` in the console ([source](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause#browser_compatibility)), so it wouldn't be much help by default for everyone, as devs would need to manually try/catch the error and access the `cause` property in other browsers. But at this point, yeah, it's just another bit of information that might help, I guess?
Added `cause` [here](https://github.com/ably/ably-js/compare/eda9c4195e8522cae3ebb37e15d5a59176e05865..f15aca3cc2b9d2f1c253aa7ff3fc87154434721c)

owenpearson: we actually override `Error.cause` in `ErrorInfo` with our own `cause` field so i'm not sure if that will apply, but yeah just having the information accessible will be useful either way, if someone raises an issue we can tell them to check `Error.cause` rather than having to look through their code and find where the connection error originates from

VeskeR: We don't do anything special with the `cause` field in `ErrorInfo` classes, so in the end it's just a regular `Error.cause` property by js spec. So it's not displayed in chromimum browsers and safari, but on firefox it looks like this:
image

PR #1761, opened 2024-05-14
[ECO-4780] Fix `Connection closed` error when using `usePresence` hook

### LINKED ISSUE #1753: usePresence error - Uncaught (in promise) Error: Connection closed

VeskeR:
Resolves #1753
Also seems to help with #1759 , ideally would confirm with the user that this fix helps with the problem once released.

`ably.connection.state` dependency in `onUnmount` callback caused the next issue:

When `ably.connection.state` is added as a dependency to the `onUnmount` callback, changes to the connection state will update this callback, which in turn triggers our mounting `useEffect` hook. This became problematic on page reloads: when the connection state changes to `closed`, the `onUnmount` callback updates, and this triggers our mounting `useEffect` hook, which calls `onMount` and tries to enter presence. Since the connection is closed at that point, we get a "Connection closed" error. Although it's not critical and doesn't seem to cause any application behavior issues, polluting the log with "Connection closed" errors should be avoided.

### DIFF
--- a/src/platform/react-hooks/src/hooks/usePresence.ts
+++ b/src/platform/react-hooks/src/hooks/usePresence.ts
@@ -1,9 +1,10 @@
 import type * as Ably from 'ably';
-import { useCallback, useEffect, useRef } from 'react';
+import { useCallback, useEffect, useRef, useState } from 'react';
 import { ChannelParameters } from '../AblyReactHooks.js';
 import { useAbly } from './useAbly.js';
 import { useChannelInstance } from './useChannelInstance.js';
 import { useStateErrors } from './useStateErrors.js';
+import { useConnectionStateListener } from './useConnectionStateListener.js';
 
 export interface PresenceResult<T> {
   updateStatus: (messageOrPresenceObject: T) => void;
@@ -26,6 +27,7 @@ export function usePresence<T = any>(
   const ably = useAbly(params.ablyId);
   const { channel } = useChannelInstance(params.ablyId, params.channelName);
   const { connectionError, channelError } = useStateErrors(params);
+
   // we can't simply add messageOrPresenceObject to dependency list in our useCallback/useEffect hooks,
   // since it will most likely cause an infinite loop of updates in cases when user calls this hook
   // with an object literal instead of a state or memoized object.
@@ -33,30 +35,40 @@ export function usePresence<T = any>(
   // note that it still prevents us from automatically re-entering presence with new messageOrPresenceObject if it changes.
   // one of the options to fix this, is to use deep equals to check if the object has actually changed. see https://github.com/ably/ably-js/issues/1688.
   const messageOrPresenceObjectRef = useRef(messageOrPresenceObject);
-
   useEffect(() => {
     messageOrPresenceObjectRef.current = messageOrPresenceObject;
   }, [messageOrPresenceObject]);
 
-  const onMount = useCallback(async () => {
-    await channel.presence.enter(messageOrPresenceObjectRef.current);
-  }, [channel.presence]);
-
-  const onUnmount = useCallback(() => {
-    // if connection is in one of inactive states, leave call will produce exception
-    if (channel.state === 'attached' && !INACTIVE_CONNECTION_STATES.includes(ably.connection.state)) {
-      channel.presence.leave();
-    }
-  }, [channel, ably.connection.state]);
+  // we need to listen for the current connection state in order to react to it.
+  // for example, we should enter presence when first connected, re-enter when reconnected,
+  // and be able to prevent entering presence when the connection is in an inactive state.
+  // all of that can be achieved by using the useConnectionStateListener hook.
+  const [connectionState, setConnectionState] = useState(ably.connection.state);
+  useConnectionStateListener((stateChange) => {
+    setConnectionState(stateChange.current);
+  });
+  const shouldNotEnterPresence = INACTIVE_CONNECTION_STATES.includes(connectionState) || skip;
 
   useEffect(() => {
-    if (skip) return;
+    if (shouldNotEnterPresence) {
+      return;
+    }
 
-    onMount();
-    return () => {
-      onUnmount();
+    const onMount = async () => {
+      await channel.presence.enter(messageOrPresenceObjectRef.current);
     };
-  }, [skip, onMount, onUnmount]);
+    onMount();
+
+    return () => {
+      // here we use the ably.connection.state property, which upon this cleanup function call
+      // will have the current connection state for that connection, thanks to us accessing the Ably instance here by reference.
+      // if the connection is in one of the inactive states or the channel is not attached, a presence.leave call will produce an exception.
+      // so we only leave presence in other cases.
+      if (channel.state === 'attached' && !INACTIVE_CONNECTION_STATES.includes(ably.connection.state)) {
+        channel.presence.leave();
+      }
+    };
+  }, [shouldNotEnterPresence, channel, ably]);
 
   const updateStatus = useCallback(
     (messageOrPresenceObject: T) => {


ttypic (approved):
Nice catch!

src/platform/react-hooks/src/hooks/usePresence.ts
```diff
@@ -33,21 +33,33 @@ export function usePresence<T = any>(
...
   useEffect(() => {
     messageOrPresenceObjectRef.current = messageOrPresenceObject;
   }, [messageOrPresenceObject]);
 
+  useEffect(() => {
+    ablyConnectionStateRef.current = ably.connection.state;
+  }, [ably.connection.state]);
+
   const onMount = useCallback(async () => {
     await channel.presence.enter(messageOrPresenceObjectRef.current);
```
ttypic, line 54: Maybe we should add checks for entering as well?

VeskeR: That was my initial idea in order to fix this error, but I wasn't sure at a time that there is no situation where such a check might backfire. Like it won't try to enter a presence when connection is in a transitive state or something.
However, all of that comes down to [this check](https://github.com/ably/ably-js/blob/1753/fix-usePresence-connection-closed/src/common/lib/client/realtimepresence.ts#L133-L135) in `RealtimePresence._enterOrUpdateClient`. So basically, if we try to enter a presence and the connection is currently not active - we throw an error which would break everything in React anyway.
So I guess there is no downside to adding a check for connection state to that `onMount` method too.

VeskeR: One more thing that came to mind:
With fix from this PR, what would happen now if we manually close connection and then open it again? Or that matter, if we loose connection to Able server due to network issues and then reconnect. Will we automatically rejoin presence set on reconnection?

VeskeR: Since previously, the interaction between `ably.connection.state` in `useEffect` and our mounting hooks described in the [comment here](https://github.com/ably/ably-js/pull/1761/files#diff-63f026c2db9f54ace72a1df10dd2228a651f632b4d3f70594e6da465c5737786R36-R42) would actually benefit us.
On any connection state change that causes a re-render, we would leave and enter presence every time, thus guaranteeing that if the last connection state was active, we would be in the presence set.

I hope that makes sense 😅

VeskeR: Updated to use `useConnectionStateListener` to monitor connection state changes and added inactive connection check in `useEffect` hook in https://github.com/ably/ably-js/compare/e295b6673c227be4f28ec8b1c575f985d2a9d1ef..78633ff8ccdca6e6e85a6ced8c5133d24187cdf1

src/platform/react-hooks/src/hooks/usePresence.ts
```diff
@@ -44,19 +59,23 @@ export function usePresence<T = any>(
 
   const onUnmount = useCallback(() => {
     // if connection is in one of inactive states, leave call will produce exception
-    if (channel.state === 'attached' && !INACTIVE_CONNECTION_STATES.includes(ably.connection.state)) {
+    if (channel.state === 'attached' && !INACTIVE_CONNECTION_STATES.includes(connectionState)) {
       channel.presence.leave();
     }
-  }, [channel, ably.connection.state]);
+  }, [channel, connectionState]);
```
VeskeR, line 65: Not sure about this dependency on `connectionState`. Seems like there is no way to avoid it.
Also, since this is a cleanup funciton, I think it would "lag behind" on the actual connection state, e.g. when connection state changes, we call `onUnmount` as our cleanup function in `useEffect` below, which has previous connection state stored (has stored `connected` value but the actual connection state is `closing` or something)

src/platform/react-hooks/src/hooks/usePresence.ts
```diff
@@ -44,19 +59,23 @@ export function usePresence<T = any>(
...
 
   useEffect(() => {
+    if (INACTIVE_CONNECTION_STATES.includes(connectionState)) {
```
ttypic, line 68: I would change this into:

```jsx
const shouldNotEnterPresence = INACTIVE_CONNECTION_STATES.includes(connectionState) || skip;

useEffect(() => {
   if (shouldNotEnterPresence) return;

   await channel.presence.enter(messageOrPresenceObjectRef.current);
   
   return () => {
      if (channel.state === 'attached' && !INACTIVE_CONNECTION_STATES.includes(connectionState)) {
        channel.presence.leave();
      }
   };
}, [shouldNotEnterPresence]);

```

this way we invoke effect only then it's necessary

ttypic (changes requested):

src/platform/react-hooks/src/hooks/usePresence.ts
VeskeR: Agree, removed callbacks in https://github.com/ably/ably-js/compare/2743834c0521378015334b8d6af3de4b47058e03..2f9fadd5f8dc7fbc840e2e56f98b3340c721b808

src/platform/react-hooks/src/hooks/usePresence.ts
VeskeR: I liked the idea, thank you! Implemented it in https://github.com/ably/ably-js/compare/2743834c0521378015334b8d6af3de4b47058e03..2f9fadd5f8dc7fbc840e2e56f98b3340c721b808.

It works as expected, no "Connection closed" errors and now I finally can't see a problem with the solution. I hope that was the last change and we would be able to merge now.

src/platform/react-hooks/src/hooks/usePresence.ts
```diff
@@ -26,37 +27,48 @@ export function usePresence<T = any>(
...
+      // here we use the ably.connection.state property, which upon this cleanup function call
+      // will have the current connection state for that connection, thanks to us accessing the Ably instance here by reference.
+      // if the connection is in one of the inactive states or the channel is not attached, a presence.leave call will produce an exception.
+      // so we only leave presence in other cases.
+      if (channel.state === 'attached' && !INACTIVE_CONNECTION_STATES.includes(ably.connection.state)) {
+        channel.presence.leave();
+      }
     };
-  }, [skip, onMount, onUnmount]);
+  }, [shouldNotEnterPresence, channel, ably.connection.state]);
```
ttypic, line 71: let's put `ably` here instead of  `ably.connection.state` to avoid unnecessary effect remounts

ttypic (changes requested):

src/platform/react-hooks/src/hooks/usePresence.ts
VeskeR: Updated

ttypic (approved):
LGTM

PR #1765, opened 2024-05-17
types/capability: allow wildcard capability when presented in object form, add privileged-headers capability

AndyTWF:
At the moment, it's not possible (according to TypeScript) to specify a capability as `{"resource": ["*"]}` using the map-form - which is still valid. This change also allows the explicit specification of a wildcard.

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -683,7 +683,8 @@ export type capabilityOp =
   | 'stats'
   | 'channel-metadata'
   | 'push-subscribe'
-  | 'push-admin';
+  | 'push-admin'
+  | 'privileged-headers';
 
 /**
  * Capabilities which are available for use within {@link TokenParams}.
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -684,6 +684,7 @@ export type capabilityOp =
   | 'channel-metadata'
   | 'push-subscribe'
   | 'push-admin';
+
 /**
  * Capabilities which are available for use within {@link TokenParams}.
  */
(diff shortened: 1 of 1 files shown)


ably.d.ts
```diff
@@ -698,7 +699,7 @@ export interface TokenParams {
    *
    * @defaultValue `'{"*":["*"]}'`
    */
-  capability?: { [key: string]: capabilityOp[] } | string;
+  capability?: { [key: string]: capabilityOp[] | ['*'] } | string;
```
VeskeR, line 702: Shouldn't `*` value be added to `capabilityOp` type above?
Based on the new empty line added near that type in this PR, is there a chance you already tried it and something went wrong?

VeskeR: Also, OOI, is this the correct list of all capabilities available for token params: https://ably.com/docs/auth/capabilities? I couldn't find the list in the spec, only in the docs on a website.
If that list is correct, it appears that we're also missing `privileged-headers` capability type. If there's nothing special about it, maybe we should add it too and expand this PR a bit

VeskeR (changes requested):

ably.d.ts
AndyTWF: > Shouldn't * value be added to capabilityOp type above?

We could achieve the same result that way, yes! The reason I thought it would be worth making it separate is because if you specify `*`, you're essentially making any other specified capabilities redundant. So it doesn't really make sense to have `*` with some other `capabilityOp`. Happy to go with whatever you'd prefer, though!

> Also, OOI, is this the correct list of all capabilities available for token params

Yes that's the correct list, we could definitely add `privileged-headers` - which I'll do now

VeskeR: > We could achieve the same result that way, yes! The reason I thought it would be worth making it separate is because if you specify *, you're essentially making any other specified capabilities redundant. So it doesn't really make sense to have * with some other capabilityOp. Happy to go with whatever you'd prefer, though!

Oh, I see now what you did there. Cool stuff! I'm happy to leave it as `capabilityOp[] | ['*']` then.

VeskeR (approved):
LGTM

PR #1768, opened 2024-05-20
[ECO-4788] Make sure we tear down connections in `modular.test.js`

lawrence-forooghian:
(My mistake when I wrote these tests.)


ttypic (approved):
LGTM

Issue #1772, opened 2024-05-21
Consider running browser tests against `FetchRequest` in addition to `XHRRequest`


lawrence-forooghian:
We do not currently exercise our `FetchRequest` class in the tests; they all use XHR, the only exception being some of the tests for the modular variant the SDK. This might be hiding bugs in the `FetchRequest` implementation such as #1771. See if we could run the tests against `FetchRequest` as well as XHR, the same way we exercise all of the different transports in (some) tests.

PR #1773, opened 2024-05-21
Make `FetchRequest` not try and parse a 204 response body

### LINKED ISSUE #1771: `FetchRequest` doesn't properly handle a 204 response
It checks for the presence of a `Content-Type` response header and if present assumes that trying to parse the response will succeed:

https://github.com/ably/ably-js/blob/bc4f3ac4c06d015b867887807dc858c8f1882780/src/platform/web/lib/http/request/fetchrequest.ts#L72-L80

Realtime currently sends a 204 with a `Content-Type` header from the `POST rest.ably.io/push/publish` endpoint when using HTTP 1.1 (see [REA-1924](https://ably.atlassian.net/browse/REA-1924)). This bug hasn't been caught by the tests because I guess most of the time they use HTTP 2 in browsers I guess, and also we don't use `FetchRequest` except in the tests for the modular variant of the library (created #1772 for improving this situtation).

Easiest way to fix this would be to handle a 204 the same way as `XHRRequest`:

https://github.com/ably/ably-js/blob/bc4f3ac4c06d015b867887807dc858c8f1882780/src/platform/web/lib/http/request/xhrrequest.ts#L206-L209

[REA-1924]: https://ably.atlassian.net/browse/REA-1924?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4808) by [Unito](https://www.unito.io)

lawrence-forooghian:
It currently checks for the presence of a `Content-Type` response header and if present assumes that trying to parse the response will succeed.

Realtime currently sends a `204 No Content` with a `Content-Type` header from the `POST rest.ably.io/push/publish` endpoint when using HTTP 1.1 (see REA-1924).

(This bug hasn't been caught by the tests because I guess most of the time they use HTTP 2 in browsers, and also we don't use FetchRequest in the tests except briefly in those for the modular variant of the library. I’ve created #1772 for improving this situation).

The 204 handling approach is copied from that which we already have in `XHRRequest`.

Resolves #1771.

### DIFF
--- a/src/platform/web/lib/http/request/fetchrequest.ts
+++ b/src/platform/web/lib/http/request/fetchrequest.ts
@@ -69,6 +69,10 @@ export default async function fetchRequest(
 
       clearTimeout(timeout!);
 
+      if (res.status == 204) {
+        return { error: null, statusCode: res.status };
+      }
+
       const contentType = res.headers.get('Content-Type');
       let body;
       if (contentType && contentType.indexOf('application/x-msgpack') > -1) {


ttypic (approved):
LGTM

PR #1775, opened 2024-05-22
[REA-1702] feat: web push

owenpearson:
- adds a plugin which enables clients to be activated for receipt of push notifications.
- adds support for w3c push transport to web clients.
- adds push channel transport for testing push activation on all platforms.

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -594,6 +594,11 @@ export interface ClientOptions<Plugins = CorePlugins> extends AuthOptions {
    * @defaultValue 65536
    */
   maxMessageSize?: number;
+
+  /**
+   * A URL pointing to a service worker script which is used as the target for web push notifications.
+   */
+  pushServiceWorkerUrl?: string;
 }
 
 /**
@@ -604,6 +609,11 @@ export interface CorePlugins {
    * A plugin capable of decoding `vcdiff`-encoded messages. For more information on how to configure a channel to use delta encoding, see the [documentation for the `@ably-forks/vcdiff-decoder` package](https://github.com/ably-forks/vcdiff-decoder#usage).
    */
   vcdiff?: any;
+
+  /**
+   * A plugin which allows the client to be the target of push notifications.
+   */
+  Push?: unknown;
 }
 
 /**
@@ -1494,6 +1504,37 @@ export type recoverConnectionCallback = (
   callback: recoverConnectionCompletionCallback,
 ) => void;
 
+/**
+ * A standard callback format which is invoked upon completion of a task.
+ *
+ * @param err - An error object if the task failed.
+ * @param result - The result of the task, if any.
+ */
+type StandardCallback<T> = (err: ErrorInfo | null, result?: T) => void;
+
+/**
+ * A function passed to {@link Push.activate} in order to override the default implementation to register a device for push activation.
+ *
+ * @param device - A DeviceDetails object representing the local device
+ * @param callback - A callback to be invoked when the registration is complete
+ */
+export type RegisterCallback = (device: DeviceDetails, callback: StandardCallback<DeviceDetails>) => void;
+
+/**
+ * A function passed to {@link Push.activate} in order to override the default implementation to deregister a device for push activation.
+ *
+ * @param device - A DeviceDetails object representing the local device
+ * @param callback - A callback to be invoked when the deregistration is complete
+ */
+export type DeregisterCallback = (device: DeviceDetails, callback: StandardCallback<string>) => void;
+
+/**
+ * A callback which returns only an error, or null, when complete.
+ *
+ * @param error - The error if the task failed, or null not.
+ */
+export type ErrorCallback = (error: ErrorInfo | null) => void;
+
 // Internal Interfaces
 
 // To allow a uniform (callback) interface between on and once even in the
@@ -1928,6 +1969,39 @@ export declare interface RealtimePresence {
 }
 
 /**
+ * Enables devices to subscribe to push notifications for a channel.
+ */
+export declare interface PushChannel {
+  /**
+   * Subscribes the device to push notifications for the channel.
+   */
+  subscribeDevice(): Promise<void>;
+
+  /**
+   * Unsubscribes the device from receiving push notifications for the channel.
+   */
+  unsubscribeDevice(): Promise<void>;
+
+  /**
+   * Subscribes all devices associated with the current device's `clientId` to push notifications for the channel.
+   */
+  subscribeClient(): Promise<void>;
+
+  /**
+   * Unsubscribes all devices associated with the current device's `clientId` from receiving push notifications for the channel.
+   */
+  unsubscribeClient(): Promise<void>;
+
+  /**
+   * Retrieves all push subscriptions for the channel. Subscriptions can be filtered using a params object.
+   *
+   * @param params - An object containing key-value pairs to filter subscriptions by. Can contain `clientId`, `deviceId` or a combination of both, and a `limit` on the number of subscriptions returned, up to 1,000.
+   * @returns a {@link PaginatedResult} object containing an array of {@link PushChannelSubscription} objects.
+   */
+  listSubscriptions(params?: Record<string, string>): Promise<PaginatedResult<PushChannelSubscription>>;
+}
+
+/**
  * Enables messages to be published and historic messages to be retrieved for a channel.
  */
 export declare interface Channel {
@@ -1940,6 +2014,10 @@ export declare interface Channel {
    * A {@link Presence} object.
    */
   presence: Presence;
+  /**
+   * A {@link PushChannel} object.
+   */
+  push: PushChannel;
   /**
    * Retrieves a {@link PaginatedResult} object, containing an array of historical {@link InboundMessage} objects for the channel. If the channel is configured to persist messages, then messages can be retrieved from history for up to 72 hours in the past. If not, messages can only be retrieved from history for up to two minutes in the past.
    *
@@ -2540,6 +2618,21 @@ export declare interface Push {
    * A {@link PushAdmin} object.
    */
   admin: PushAdmin;
+
+  /**
+   * Activates the device for push notifications. Subsequently registers the device with Ably and stores the deviceIdentityToken in local storage.
+   *
+   * @param registerCallback - A function passed to override the default implementation to register the local device for push activation.
+   * @param updateFailedCallback - A callback to be invoked when the device registration failed to update.
+   */
+  activate(registerCallback?: RegisterCallback, updateFailedCallback?: ErrorCallback): Promise<void>;
+
+  /**
+   * Deactivates the device from receiving push notifications.
+   *
+   * @param deregisterCallback - A function passed to override the default implementation to deregister the local device for push activation.
+   */
+  deactivate(deregisterCallback: DeregisterCallback): Promise<void>;
 }
 
 /**
--- a/scripts/moduleReport.ts
+++ b/scripts/moduleReport.ts
@@ -6,7 +6,7 @@ import { gzip } from 'zlib';
 import Table from 'cli-table';
 
 // The maximum size we allow for a minimal useful Realtime bundle (i.e. one that can subscribe to a channel)
-const minimalUsefulRealtimeBundleSizeThresholdsKiB = { raw: 96, gzip: 29 };
+const minimalUsefulRealtimeBundleSizeThresholdsKiB = { raw: 98, gzip: 30 };
 
 const baseClientNames = ['BaseRest', 'BaseRealtime'];
 
--- a/scripts/moduleReport.ts
+++ b/scripts/moduleReport.ts
@@ -65,11 +65,17 @@ interface Output {
 }
 
 // Uses esbuild to create a bundle containing the named exports from 'ably/modular'
-function getBundleInfo(exports: string[]): BundleInfo {
-  const outfile = exports.join('');
+function getModularBundleInfo(exports: string[]): BundleInfo {
+  return getBundleInfo('./build/modular/index.mjs', exports);
+}
+
+// Uses esbuild to create a bundle containing the named exports from a given module
+function getBundleInfo(modulePath: string, exports?: string[]): BundleInfo {
+  const outfile = exports ? exports.join('') : 'all';
+  const exportTarget = exports ? `{ ${exports.join(', ')} }` : '*';
   const result = esbuild.buildSync({
     stdin: {
-      contents: `export { ${exports.join(', ')} } from './build/modular/index.mjs'`,
+      contents: `export ${exportTarget} from '${modulePath}'`,
       resolveDir: '.',
     },
     metafile: true,
@@ -78,6 +84,7 @@ function getBundleInfo(exports: string[]): BundleInfo {
     outfile,
     write: false,
     sourcemap: 'external',
+    external: ['ulid'],
   });
 
   const pathHasBase = (component: string) => {
@@ -97,8 +104,8 @@ function getBundleInfo(exports: string[]): BundleInfo {
 }
 
 // Gets the bundled size in bytes of an array of named exports from 'ably/modular'
-async function getImportSizes(exports: string[]): Promise<ByteSizes> {
-  const bundleInfo = getBundleInfo(exports);
+async function getModularImportSizes(exports: string[]): Promise<ByteSizes> {
+  const bundleInfo = getModularBundleInfo(exports);
 
   return {
     rawByteSize: bundleInfo.byteSize,
@@ -118,14 +125,14 @@ async function calculateAndCheckExportSizes(): Promise<Output> {
   const output: Output = { tableRows: [], errors: [] };
 
   for (const baseClient of baseClientNames) {
-    const baseClientSizes = await getImportSizes([baseClient]);
+    const baseClientSizes = await getModularImportSizes([baseClient]);
 
     // First output the size of the base client
     output.tableRows.push({ description: baseClient, sizes: baseClientSizes });
 
     // Then output the size of each export together with the base client
     for (const exportName of [...pluginNames, ...functions.map((functionData) => functionData.name)]) {
-      const sizes = await getImportSizes([baseClient, exportName]);
+      const sizes = await getModularImportSizes([baseClient, exportName]);
       output.tableRows.push({ description: `${baseClient} + ${exportName}`, sizes });
 
       if (!(baseClientSizes.rawByteSize < sizes.rawByteSize) && !(baseClient === 'BaseRest' && exportName === 'Rest')) {
@@ -146,13 +153,13 @@ async function calculateAndCheckFunctionSizes(): Promise<Output> {
     const { name: functionName, transitiveImports } = functionData;
 
     // First output the size of the function
-    const standaloneSizes = await getImportSizes([functionName]);
+    const standaloneSizes = await getModularImportSizes([functionName]);
     output.tableRows.push({ description: functionName, sizes: standaloneSizes });
 
     // Then output the size of the function together with the plugin we expect
     // it to transitively import
     if (transitiveImports.length > 0) {
-      const withTransitiveImportsSizes = await getImportSizes([functionName, ...transitiveImports]);
+      const withTransitiveImportsSizes = await getModularImportSizes([functionName, ...transitiveImports]);
       output.tableRows.push({
         description: `${functionName} + ${transitiveImports.join(' + ')}`,
         sizes: withTransitiveImportsSizes,
@@ -176,11 +183,27 @@ async function calculateAndCheckFunctionSizes(): Promise<Output> {
   return output;
 }
 
+async function calculatePushPluginSize(): Promise<Output> {
+  const output: Output = { tableRows: [], errors: [] };
+  const pushPluginBundleInfo = getBundleInfo('./build/push.js');
+  const sizes = {
+    rawByteSize: pushPluginBundleInfo.byteSize,
+    gzipEncodedByteSize: (await promisify(gzip)(pushPluginBundleInfo.code)).byteLength,
+  };
+
+  output.tableRows.push({
+    description: 'Push',
+    sizes: sizes,
+  });
+
+  return output;
+}
+
 async function calculateAndCheckMinimalUsefulRealtimeBundleSize(): Promise<Output> {
   const output: Output = { tableRows: [], errors: [] };
 
   const exports = ['BaseRealtime', 'FetchRequest', 'WebSocketTransport'];
-  const sizes = await getImportSizes(exports);
+  const sizes = await getModularImportSizes(exports);
 
   output.tableRows.push({ description: `Minimal useful Realtime (${exports.join(' + ')})`, sizes });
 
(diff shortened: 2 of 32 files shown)


src/plugins/push/pushactivation.ts
```diff
@@ -0,0 +1,571 @@
...
+        this.push.recipient = JSON.parse(Platform.Config.push.storage.get(persistKeys.pushRecipient) || 'null');
+      } else {
+        this.resetId();
+      }
+    }
+
+    persist() {
+      const config = this.rest.Platform.Config;
+      if (!config.push) {
+        throw new this.rest.ErrorInfo('hello', 1, 1);
```
ttypic, line 79: Probably it needs bette ErrorInfo description :)

src/plugins/push/pushactivation.ts
```diff
@@ -0,0 +1,571 @@
...
+  handling: boolean;
+  deactivatedCallback?: ErrCallback;
+  activatedCallback?: ErrCallback;
+  pushConfig: IPlatformPushConfig;
+  registerCallback?: RegisterCallback;
+  deregisterCallback?: DeregisterCallback;
+  updatedCallback?: ErrCallback;
+
+  // Used for testing
+  pushManager?: PushManager;
```
ttypic, line 137: maybe it's worth to add `_` at the begging just to make it clear that it's not a part of public API

src/plugins/push/pushactivation.ts
```diff
@@ -0,0 +1,571 @@
...
+// States
+const NotActivated = function (machine: ActivationStateMachine, event: ActivationEvent) {
+  if (event instanceof CalledDeactivate) {
+    machine.deactivatedCallback?.(null);
+    return NotActivated;
+  } else if (event instanceof CalledActivate) {
+    const device = machine.rest.device as LocalDevice;
+
+    if (device.deviceIdentityToken != null) {
+      // Already registered.
```
ttypic, line 413: According to the spec we should check `clientId` here [RSH3a2a1](https://sdk.ably.com/builds/ably/specification/main/features/#RSH3a2a1) and update token (not sure that it's valid for web push, but we have `WaitingForRegistrationUpdate` that makes me think it should be)

src/plugins/push/pushactivation.ts
```diff
@@ -0,0 +1,571 @@
...
+  } else if (event instanceof CalledDeactivate) {
+    machine.deregister();
+    return WaitingForDeregistration(WaitingForNewPushDeviceDetails);
+  } else if (event instanceof GotPushDeviceDetails) {
+    machine.updateRegistration();
+    return WaitingForRegistrationUpdate;
+  }
+};
+
+const WaitingForRegistrationUpdate = function (machine: ActivationStateMachine, event: ActivationEvent) {
```
ttypic, line 506: small thing but it's called `WaitingForRegistrationSync` in the [spec](https://sdk.ably.com/builds/ably/specification/main/features/#RSH3e)

ttypic (changes requested):
Looks good! very consistent to what we have in java and cocoa, added couple of small comments

src/plugins/push/pushactivation.ts
```diff
@@ -0,0 +1,605 @@
...
+    machine.callActivatedCallback(null);
+    return WaitingForNewPushDeviceDetails;
+  } else if (event instanceof CalledDeactivate) {
+    machine.deregister();
+    return WaitingForDeregistration(WaitingForNewPushDeviceDetails);
+  } else if (event instanceof GotPushDeviceDetails) {
+    machine.updateRegistration();
+    return WaitingForRegistrationSync;
+  }
+};
```
VeskeR, line 538: should `WaitingForNewPushDeviceDetails` also `return null` at the end if no event matched?

src/plugins/push/pushactivation.ts
```diff
@@ -0,0 +1,605 @@
...
+
+const WaitingForRegistrationSync = function (machine: ActivationStateMachine, event: ActivationEvent) {
+  if (event instanceof CalledActivate) {
```
VeskeR, line 541: is `unless the machine is in state WaitingForRegistrationSync as a result of a CalledActivate event` part of [RSH3e1](https://sdk.ably.com/builds/ably/specification/main/features/#RSH3e1) relevant here? Should we have an additional check for this?

src/plugins/push/pushactivation.ts
```diff
@@ -0,0 +1,605 @@
...
+    machine.callActivatedCallback(null);
+    return WaitingForRegistrationSync;
+  } else if (event instanceof RegistrationUpdated) {
```
VeskeR, line 544: I think `RegistrationUpdated` is called `RegistrationSynced` in [spec](https://sdk.ably.com/builds/ably/specification/main/features/#RSH3e2)?

src/plugins/push/pushactivation.ts
```diff
@@ -0,0 +1,605 @@
...
+    return WaitingForNewPushDeviceDetails;
+  } else if (event instanceof SyncRegistrationFailed) {
+    machine.callUpdateRegistrationFailedCallback(event.reason);
+    return AfterRegistrationUpdateFailed;
+  }
+  return null;
+};
+
+const AfterRegistrationUpdateFailed = function (machine: ActivationStateMachine, event: ActivationEvent) {
```
VeskeR, line 553: I think it's called `AfterRegistrationSyncFailed` in [spec](https://sdk.ably.com/builds/ably/specification/main/features/#RSH3f)?

src/plugins/push/pushactivation.ts
```diff
@@ -0,0 +1,605 @@
...
+    } else if (event instanceof DeregistrationFailed) {
+      machine.callDeactivatedCallback(event.reason);
+      return previousState;
+    }
+    return null;
+  };
+};
+
+function isPersistentState(state: ActivationState) {
+  return state.name == 'NotActivated' || state.name == 'WaitingForNewPushDeviceDetails';
```
VeskeR, line 604: I think this is gonna be a problem in the minified code, `.name` won't match expected function names (added `minified: true` to esbuild config for push plugin and all state names were gone).
can either convert state functions to actual classes (might be time consuming), or just add a custom property with hardcoded name to those state functions and compare using that.
can then change calls like `this.current.name` and `maybeNext.name` to use that new prop too.

`event.constructor.name` calls will suffer the similar fate

test/browser/push.test.js
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
+   * These tests don't work in CI for various reasons (below) but should work when running locally via `npm run test:webserver`, provided
+   * that you have enabled notification permissions for the origin serving the test ui.
+   *
+   * chromium, firefox - don't support push notifications in incognito
+   * webkit - doesn't have a way to launch programatically with notification permissions granted
+   */
+  if (Notification.permission === 'granted') {
+    describe.only('browser/push', function () {
```
VeskeR, line 27: missed `.only` or is it intentional given they won't run in CI?

test/rest/push.test.js
```diff
@@ -1,10 +1,33 @@
...
+) {
   var Utils = helper.Utils;
   var exports = {};
   var expect = chai.expect;
   var closeAndFinish = helper.closeAndFinish;
+  var displayError = helper.displayError;
+  var monitorConnection = helper.monitorConnection;
+  var testOnAllTransports = helper.testOnAllTransports;
+  var mixin = helper.Utils.mixin;
+  var echoServer = 'https://echo.ably.io';
```
VeskeR, line 19:
```
displayError
monitorConnection
testOnAllTransports
mixin
echoServer
```
not used so can remove

src/platform/web/lib/transport/push.ts
```diff
@@ -0,0 +1,84 @@
...
+      transportType: 'web',
+      targetUrl: btoa(endpoint),
+      encryptionKey: key,
+    };
+    device.persist();
+
+    machine.handleEvent(new GotPushDeviceDetails());
+  } catch (err) {
+    machine.handleEvent(
+      new GettingPushDeviceDetailsFailed(new ErrorInfo('failed to register service worker', 50000, 500, err as Error)),
```
VeskeR, line 81:
```suggestion
      new GettingPushDeviceDetailsFailed(new ErrorInfo('Failed to register service worker', 50000, 500, err as Error)),
```
nit: more consistent with the rest of codebase

test/support/push_channel_transport.js
```diff
@@ -0,0 +1,57 @@
+define(['ably'], function (Ably) {
+  var Defaults = Ably.Realtime.Platform.Defaults;
+  var ErrorInfo = Ably.ErrorInfo;
+  return (module.exports = {
+    getPushDeviceDetails: function (machine) {
+      var channel = machine.rest.options.pushRecipientChannel;
+      if (!channel) {
+        machine.handleEvent(
+          new machine.GettingPushDeviceDetailsFailed(
+            new ErrorInfo('missing ClientOptions.pushRecipientChannel', 40000, 400),
```
VeskeR, line 10:
```suggestion
            new ErrorInfo('Missing ClientOptions.pushRecipientChannel', 40000, 400),
```
nit: more consistent with the rest of codebase

test/support/push_channel_transport.js
```diff
@@ -0,0 +1,57 @@
...
+          ),
+        );
+        return;
+      }
+
+      var ablyKey = machine.rest.options.pushAblyKey || machine.rest.options.key;
+      if (!ablyKey) {
+        machine.handleEvent(
+          new machine.GettingPushDeviceDetailsFailed(new ErrorInfo('missing options.pushAblyKey', 40000, 400)),
```
VeskeR, line 19:
```suggestion
          new machine.GettingPushDeviceDetailsFailed(new ErrorInfo('Missing options.pushAblyKey', 40000, 400)),
```
nit: more consistent with the rest of codebase

VeskeR (changes requested):
Have a wall of comments but mostly they are structural suggestions.

Couple of things that I couldn't attach to code lines:

1. Should we add new Push plugin to `scripts/moduleReport.ts` `pluginNames` list? I imagine we would need to update `getBundleInfo` in there, as currently it always imports from `'./build/modular/index.mjs'`.
2. Per [RSH7](https://sdk.ably.com/builds/ably/specification/main/features/#RSH7) both `RestChannel` and `RealtimeChannel` should have `push` object. Why did we add `push` only to `RestChannel` here?


P.S. I'm really sorry for the amount of comments 😬 I wonder if we could've done it in separate PRs for web push integration branch.

src/common/lib/client/push.ts
```diff
@@ -6,14 +6,82 @@ import ErrorInfo from '../types/errorinfo';
...
   client: BaseClient;
   admin: Admin;
+  stateMachine?: ActivationStateMachine;
+  LocalDevice?: LocalDeviceConstructor;
 
   constructor(client: BaseClient) {
     this.client = client;
     this.admin = new Admin(client);
+    if (Platform.Config.push && client.options.plugins?.push) {
+      this.stateMachine = new client.options.plugins.push.ActivationStateMachine(client);
```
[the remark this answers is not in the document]
VeskeR: UPDATE: see https://github.com/ably/ably-js/pull/1775#discussion_r1639311873 comment for better approach

src/platform/web/lib/transport/push.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    const endpoint = subscription.endpoint;
+
+    const [p256dh, auth] = [subscription.getKey('p256dh'), subscription.getKey('auth')];
+
+    if (!p256dh || !auth) {
+      throw new ErrorInfo('Public key not found', 50000, 500);
+    }
+
+    const key = [p256dh, auth].map(toBase64Url).join(':');
```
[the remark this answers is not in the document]
paddybyers: should the `recipient` contain these as separate elements, eg:

```
recipient = {
  transportType: 'web',
  targetUrl: btoa(endpoint),
  encryptionKey: {
    p256dh: xxx,
    auth: yyyy
  },
};
```
?

src/common/lib/client/push.ts
```diff
@@ -6,14 +6,74 @@ import ErrorInfo from '../types/errorinfo';
...
+    await new Promise<void>((resolve, reject) => {
+      if (!this.client.options.plugins?.push) {
+        reject(Utils.createMissingPluginError('Push'));
+        return;
+      }
+      if (!this.stateMachine) {
+        reject(new Error('this platform is not supported as a target of push notifications'));
+        return;
+      }
+      this.stateMachine.activatedCallback = (err: ErrorInfo) => {
```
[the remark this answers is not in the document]
owenpearson: good spot 👍  i've fixed this now by throwing an exception if an activation/deactivation is in progress

src/common/lib/client/pushchannel.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  constructor(channel: RestChannel) {
+    this.channel = channel;
+    this.rest = channel.client;
+  }
+
+  async subscribeDevice() {
+    const rest = this.rest;
+    const device = rest.device as LocalDevice;
+    const format = rest.options.useBinaryProtocol ? Utils.Format.msgpack : Utils.Format.json,
```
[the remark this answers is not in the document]
owenpearson: i guess it's a matter of taste 😄 personally i think when the variable declarations include ternary conditions and type assertions then it's more readable on separate lines

src/platform/web/config.ts
```diff
@@ -79,6 +81,11 @@ const Config: IPlatformConfig = {
     return byteArray.buffer;
   },
   isWebworker: isWebWorkerContext(),
+  push: {
+    platform: DevicePlatform.Browser,
+    formFactor: DeviceFormFactor.Desktop,
```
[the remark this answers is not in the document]
owenpearson: Yeah it's intentional, the `platform` will always be browser since this is the config for the web platform (however will be different for react native). I looked at detecting form factor for web clients and decided that there wasn't a reliable + portable way to do it (and it's of questionable value anyway) so decided to just keep it as desktop for simplicity.

src/common/lib/client/push.ts
```diff
@@ -6,14 +6,82 @@ import ErrorInfo from '../types/errorinfo';
...
+  DeregisterCallback,
+  LocalDeviceConstructor,
+  RegisterCallback,
+} from 'plugins/push/pushactivation';
+import Platform from 'common/platform';
+import { ErrCallback } from 'common/types/utils';
 
 class Push {
   client: BaseClient;
   admin: Admin;
```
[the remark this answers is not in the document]
owenpearson: I would be in favour of this, but it would unfortunately be a breaking change so I don't think it's worth considering for this PR, definitely worth considering for a future major release though

src/common/lib/client/push.ts
```diff
@@ -6,14 +6,82 @@ import ErrorInfo from '../types/errorinfo';
 import PushChannelSubscription from '../types/pushchannelsubscription';
 import BaseClient from './baseclient';
 import Defaults from '../util/defaults';
+import {
```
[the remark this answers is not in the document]
owenpearson: good idea, i've used `import type` everywhere where we're importing between the plugin and the core now

src/platform/web/lib/transport/push.ts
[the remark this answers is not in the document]
owenpearson: yep, this is purely an artefact from the original push activation PR, so I agree we should tidy it up and i agree with the recipient schema listed above. We'll need a pqw change to support this so I'll wait until that's on sandbox before making the change here.

src/common/lib/client/pushchannel.ts
```diff
@@ -0,0 +1,117 @@
...
+      { clientId: clientId, channel: this.channel.name },
+      format,
+      true,
+    );
+  }
+
+  async listSubscriptions(params?: Record<string, string>) {
+    this.rest.Logger.logAction(
+      this.rest.logger,
+      this.rest.Logger.LOG_MICRO,
```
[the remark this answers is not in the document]
owenpearson: Yeah I thought about this and I don't think there's a worthwhile way to achieve log stripping for the plugin since there's no way to tell at import time whether the plugin is being used in a modular client or a default client. The only alternative I can think of is to offer a separate build of the plugin without logs but the size difference would be minimal so I think we should leave it as-is.

src/common/lib/client/pushchannel.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    return this.rest.push.admin.channelSubscriptions.list({ ...params, channel: this.channel.name });
+  }
+
+  private _getClientId() {
+    const clientId = this.rest.auth.clientId;
+    if (clientId) {
+      return clientId;
+    } else {
+      throw new this.rest.ErrorInfo('cannot subscribe from client without client ID', 50000, 500);
```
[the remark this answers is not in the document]
owenpearson: good idea, i've done just that

src/common/lib/client/pushchannel.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  async listSubscriptions(params?: Record<string, string>) {
+    this.rest.Logger.logAction(
+      this.rest.logger,
+      this.rest.Logger.LOG_MICRO,
+      'PushChannel.listSubscriptions()',
+      'channel = ' + this.channel.name,
+    );
+
+    return this.rest.push.admin.channelSubscriptions.list({ ...params, channel: this.channel.name });
```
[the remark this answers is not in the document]
owenpearson: The spec is correct here, i've updated the implementation + docstring

ably.d.ts
```diff
@@ -603,6 +608,11 @@ export interface CorePlugins {
    * A plugin capable of decoding `vcdiff`-encoded messages. For more information on how to configure a channel to use delta encoding, see the [documentation for the `@ably-forks/vcdiff-decoder` package](https://github.com/ably-forks/vcdiff-decoder#usage).
    */
   vcdiff?: any;
+
+  /**
+   * A plugin which allows the client to be the target of push notifications.
+   */
+  push?: any;
```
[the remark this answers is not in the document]
owenpearson: nice, i wasn't aware of this for modularplugins but i've done that now

src/plugins/push/pushactivation.ts
owenpearson: i figured converting them to actual classes was the better approach, done that now

src/plugins/push/pushactivation.ts
```diff
@@ -0,0 +1,605 @@
...
+      return WaitingForNewPushDeviceDetails;
+    }
+
+    if (device.push.recipient) {
+      machine.pendingEvents.push(new GotPushDeviceDetails());
+    } else {
+      if (machine.pushConfig.getPushDeviceDetails) {
+        machine.pushConfig.getPushDeviceDetails?.(machine);
+      } else if (machine.pushConfig.platform === DevicePlatform.Browser) {
+        machine.rest.options.plugins!.push.getW3CPushDeviceDetails(machine);
```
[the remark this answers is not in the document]
owenpearson: took a few different approaches to address each instance but i've gotten rid of all non-null assertions now

ably.d.ts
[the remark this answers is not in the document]
VeskeR:
```suggestion
  Push?: unknown;
```

need to update this

src/platform/web/config.ts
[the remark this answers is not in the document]
VeskeR: Do we even use `formFactor` somewhere? Is this part of the realtime implementation and it's a minor metadata for some analytics?

src/plugins/push/pushactivation.ts
```diff
@@ -0,0 +1,670 @@
...
+      });
+      return;
+    }
+
+    this.handling = true;
+    this.client.Logger.logAction(
+      this.client.logger,
+      this.client.Logger.LOG_MAJOR,
+      'Push.ActivationStateMachine.handleEvent()',
+      'handling event ' + event.constructor.name + ' from ' + this.current.name,
```
VeskeR, line 303: `event.constructor.name` - need to do the same thing you did for state classes in https://github.com/ably/ably-js/pull/1775#discussion_r1639386108, otherwise names will be scrambled in minified code. Have an explicit `name` prop set in a constructor

src/plugins/push/pushactivation.ts
```diff
@@ -0,0 +1,670 @@
...
+      machine.callActivatedCallback(event.reason);
+      return new NotActivated();
+    }
+    return null;
+  }
+}
+
+class WaitingForNewPushDeviceDetails extends ActivationState {
+  constructor() {
+    super('WaitingForPushDeviceDetails');
```
VeskeR, line 565:
```suggestion
    super('WaitingForNewPushDeviceDetails');
```

scripts/moduleReport.ts
```diff
@@ -264,14 +279,43 @@ async function checkBaseRealtimeFiles() {
     'src/platform/web/modular.ts',
   ]);
 
+  return checkBundleFiles(baseRealtimeBundleInfo, allowedFiles, thresholdBytes);
+}
+
+const pushPluginAllowedFiles = [];
```
VeskeR, line 285: probably can remove as it got replaced by `allowedFiles` below?

VeskeR (changes requested):
Great changes!
Really like the new classes for states! Looks much better and is easier to read.

Added couple more comments (and had to unresolve two previous comments I think), but we're really close now

src/plugins/push/pushactivation.ts
paddybyers: > Do we even use formFactor somewhere? Is this part of the realtime implementation and it's a minor metadata for some analytics?

No, it's not used anywhere - it's just informative for anyone who looks at the registration, eg via the push admin API

test/browser/push.test.js
```diff
@@ -0,0 +1,91 @@
...
+        if (worker) {
+          await worker.unregister();
+        }
+      });
+
+      afterEach(async function () {
+        await rest.push.deactivate();
+      });
+
+      it('push_activation_succeeds', async function () {
```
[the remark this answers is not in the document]
owenpearson: yeah i opted not to tag these ones since they test the same spec points as existing tests in `test/rest/push.test.js`, but using the web push transport rather than push channel transport, i'll add tags to them now though

src/platform/web/config.ts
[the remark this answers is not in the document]
VeskeR: Copying Paddy's reply from https://github.com/ably/ably-js/pull/1775#discussion_r1647220328:

> >Do we even use formFactor somewhere? Is this part of the realtime implementation and it's a minor metadata for some analytics?
> No, it's not used anywhere - it's just informative for anyone who looks at the registration, eg via the push admin API

src/plugins/push/pushactivation.ts
owenpearson: oooops

ably.d.ts
```diff
@@ -604,6 +609,11 @@ export interface CorePlugins {
    * A plugin capable of decoding `vcdiff`-encoded messages. For more information on how to configure a channel to use delta encoding, see the [documentation for the `@ably-forks/vcdiff-decoder` package](https://github.com/ably-forks/vcdiff-decoder#usage).
    */
   vcdiff?: any;
+
+  /**
+   * A plugin which allows the client to be the target of push notifications.
+   */
+  push?: unknown;
```
VeskeR, line 616:
```suggestion
  Push?: unknown;
```

should be uppercase `Push` I think, in the implementation we access it `plugins?.Push`  ([like here](https://github.com/ably/ably-js/pull/1775/files#diff-fe43cce9d2f61250a5486a4e011d587aa9e2daaaaa4742c8dda3e635481a00e0R35) and other places)

VeskeR (changes requested):

VeskeR (approved):
LGTM!

ttypic (approved):
LGTM

Issue #1780, opened 2024-06-03
Fix `Unable to enter presence channel while in suspended state` with `usePresence` after activating the page which has been inactive for a while


VeskeR:
When on the page which is using `usePresence` hook the error can be reproduced on mobile by minimizing the browser/on pc by going to sleep for couple of minutes and then opening the browser again.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4819) by [Unito](https://www.unito.io)

PR #1781, opened 2024-06-03
[ECO-4819] Fix `Unable to enter presence channel while in suspended state` error with `usePresence`

VeskeR:
Similar to the fix in https://github.com/ably/ably-js/pull/1761 with `Connection closed` errors, we should avoid entering presence for specific channel states.

Resolves https://github.com/ably/ably-js/issues/1780


ttypic (approved):
LGTM

PR #1787, opened 2024-06-11
logger: pass level to logHandlers when logging

AndyTWF:
This allows handlers to make choices depending on the level of the log being raised (e.g. sending the logs to crashlytics when a certain number of error-level logs occur).

Other ecosystems that do already have this have the level before the message, however in this case we'll put the level second in order to ensure backwards compatibility.

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -404,8 +404,9 @@ export interface ClientOptions<Plugins = CorePlugins> extends AuthOptions {
    * Controls the log output of the library. This is a function to handle each line of log output. If you do not set this value, then `console.log` will be used.
    *
    * @param msg - The log message emitted by the library.
+   * @param level - The level of the log. Values are one of: 0 (no logs), 1 (errors only), 2 (errors plus connection and channel state changes), 3 (high-level debug output), and 4 (full debug output).
    */
-  logHandler?: (msg: string) => void;
+  logHandler?: (msg: string, level: number) => void;
 
   /**
    * Enables a non-default Ably port to be specified. For development environments only. The default value is 80.
--- a/src/common/lib/util/logger.ts
+++ b/src/common/lib/util/logger.ts
@@ -38,7 +38,9 @@ function getHandler(logger: Function): Function {
             msg,
         );
       }
-    : logger;
+    : function (msg: string) {
+        logger(msg);
+      };
 }
 
 const getDefaultLoggers = (): [Function, Function] => {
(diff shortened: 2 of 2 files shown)


ably.d.ts
```diff
@@ -404,8 +404,9 @@ export interface ClientOptions<Plugins = CorePlugins> extends AuthOptions {
    * Controls the log output of the library. This is a function to handle each line of log output. If you do not set this value, then `console.log` will be used.
    *
    * @param msg - The log message emitted by the library.
+   * @param level - The level of the log.
```
owenpearson, line 407: we should copy the text from the logLevel docstring: 
```suggestion
   * @param level - The level of the log. Values are one of: 0 (no logs), 1 (errors only), 2 (errors plus connection and channel state changes), 3 (high-level debug output), and 4 (full debug output).
```

owenpearson (changes requested):
Looks good, just one nitpick:

owenpearson (approved):

src/common/lib/util/logger.ts
```diff
@@ -117,7 +117,7 @@ class Logger {
 
   private logAction(level: LogLevels, action: string, message?: string) {
     if (this.shouldLog(level)) {
-      (level === LogLevels.Error ? this.logErrorHandler : this.logHandler)('Ably: ' + action + ': ' + message);
+      (level === LogLevels.Error ? this.logErrorHandler : this.logHandler)('Ably: ' + action + ': ' + message, level);
```
ttypic, line 120: I think we also need to slightly change `getDefaultLoggers` function and ignore second argument for them. Otherwise I think default console logger will print log level at the end of log

AndyTWF: I checked this - it doesn't seem to log them!

ttypic: OK, you won 😄  because of weird `logTimestamps` that is always `true`. It never use default `console.log` as `logHandler` and always wrap it in high-level function that ignores all arguments except first. But if somebody make it `false` you'll see everything will fall apart. I would change `getHandler` function from:

```ts
function getHandler(logger: Function): Function {
  return Platform.Config.logTimestamps
    ? ...
    : logger;
}
```
to 
```ts
function getHandler(logger: Function): Function {
  return Platform.Config.logTimestamps
    ? ...
    : function (msg: unknown) {
        logger(msg);
      };
}
```

AndyTWF: :D done!

ttypic (approved):
Looks good to me!

PR #1793, opened 2024-06-17
ConnectionManager.ping(): rewrite to fix pings stacking if none ever succeed

SimonWoolf:
previously if the transport keeps becoming active but then gets disconnected before the ping succeeds, listeners would just keep stacking up

the whole previous implementation was just way over-complex now we no longer do transport upgrade. rewrote from scratch with a much simpler promise-based one.

### DIFF
--- a/src/common/lib/client/connection.ts
+++ b/src/common/lib/client/connection.ts
@@ -48,9 +48,7 @@ class Connection extends EventEmitter {
 
   async ping(): Promise<number> {
     Logger.logAction(this.logger, Logger.LOG_MINOR, 'Connection.ping()', '');
-    return new Promise((resolve, reject) => {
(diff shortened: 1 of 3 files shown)


SimonWoolf:
(this seems to have failed the linter github action because of a `npm audit` failure (not added by this pr, which did not change dependencies). I venture that this is not something that an action called "lint" should be checking; that's what dependabot is for)

ttypic (approved):
LGTM

owenpearson (approved):
Looks good 👍

w.r.t `npm audit --production` in CI, we opted for this a while ago because it helps us to be aware of vulnerabilities in prod deps faster (which seems to still work - we found out about it here before anyone saw the dependabot alert). The intended solution is for someone to fix it in the main branch and then rebase this PR, but I think it's fine to merge this PR with admin rights anyway as long as everything else is passing.

Issue #1795, opened 2024-06-19
Detach channel  in `ChannelProvider`


ttypic:
We should detach channel if there is no `ChannelProvider` associated with it

ianb-pomelo:
Is there a shorter term mitigation for this? We switch our channels somewhat regularly and without this our channels are never detached and have hit the 200 channel limit even though most are unused/unsubscribed (and are also billed for the channels)

ttypic:
It depends on what you want to achieve. Based on your business logic, you can invoke the `detach()` method when you no longer need the channel. Additionally, you can create a wrapper around the `ChannelProvider`:

```jsx
const DetachableChannelProvider = (channelProviderProps) => {
   const ablyClient = useAbly();
   const { channelName } = channelProviderProps;

  useEffect(() => {
      // it's better to explicitly attach to the channel in this case
      ablyClient.channels.get(channelName).attach();
      return () => ablyClient.channels.get(channelName).detach();
  }, [channelName])

  return <ChannelProvider {...channelProviderProps} />;
}

```

ianb-pomelo:
Thanks! I've tried to add this but occasionally get an error when switching channels quickly that I can't seem to resolve.

The error is
```
useChannel.js:56 Uncaught (in promise) Error: Channel detached
    at _RealtimeChannel.processMessage (ably.js:6240:81)
    at Channels2.processChannelMessage (ably.js:6657:19)
    at _ConnectionManager.processChannelMessage (ably.js:5627:34)
    at _ConnectionManager.processNextPendingChannelMessage (ably.js:5613:12)
    at _ConnectionManager.onChannelMessage (ably.js:5606:12)
    at WebSocketTransport.onProtocolMessage (ably.js:4042:32)
    at WebSocketTransport.onWsData (ably.js:7354:12)
    at wsConnection.onmessage (ably.js:7308:19)
```

which seems to be caused by attempting to re-subscribe to a channel that is detaching/detached. Is there anything I can do to handle this? It seems like the `handleChannelMount` is throwing but because it's in an effect I don't think I can handle it

ttypic:
Hey @ianb-pomelo,

I apologize for the delay. I missed the notification about your message. Could you please attach verbose logs or share a reproducible example? I’m having difficulty reproducing the issue you’re facing.

zameschua:
Hey @ttypic 

I'm guessing it's related to this issue #1594

Seems like the steps to reproduce and fix the bug are already in this other PR

PR #1796, opened 2024-06-19
fix: remove detach hack in `useChannel`

ttypic:
`channel.listeners.length` is always 1, so `detach` never invokes anyway. We should put `detach` into `ChannelProvider` see https://github.com/ably/ably-js/issues/1795


VeskeR (approved):
LGTM since it doesn't do anything rn anyway.

Is there an explanation somewhere available to why we wanted to do this in the first place? So we can correctly document the reasoning for https://github.com/ably/ably-js/issues/1795

ttypic:
@VeskeR, added reasoning for https://github.com/ably/ably-js/issues/1795

AndyTWF (approved):

PR #1800, opened 2024-06-20
Gather data about private API usage in test suite

lawrence-forooghian:
This PR collects data about the usage of private APIs in the test suite, and uses this data to generate some reports. We’ll use these reports to guide our work of converting the ably-js test suite into a unified test suite for all of our client libraries. See commit messages for more details.

Resolves [ECO-4821](https://ably.atlassian.net/browse/ECO-4821), resolves [ECO-4834](https://ably.atlassian.net/browse/ECO-4834).

[ECO-4821]: https://ably.atlassian.net/browse/ECO-4821?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ECO-4834]: https://ably.atlassian.net/browse/ECO-4834?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/processPrivateApiData/grouping.ts
+++ b/scripts/processPrivateApiData/grouping.ts
@@ -1,0 +1,76 @@
+import { PrivateApiUsageDto } from './dto';
+
+export type Group<Key, Value> = {
+  key: Key;
+  values: Value[];
+};
+
+export function grouped<Key, Value>(
+  values: Value[],
+  keyForValue: (value: Value) => Key,
+  areKeysEqual: (key1: Key, key2: Key) => boolean,
+) {
+  const result: Group<Key, Value>[] = [];
+
+  for (const value of values) {
+    const key = keyForValue(value);
+
+    let existingGroup = result.find((group) => areKeysEqual(group.key, key));
+
+    if (existingGroup === undefined) {
+      existingGroup = { key, values: [] };
+      result.push(existingGroup);
+    }
+
+    existingGroup.values.push(value);
+  }
+
+  return result;
+}
+
+/**
+ * Makes sure that each private API is only listed once in a given context.
+ */
+function dedupeUsages<Key>(contextGroups: Group<Key, PrivateApiUsageDto>[]) {
+  for (const contextGroup of contextGroups) {
+    const newUsages: typeof contextGroup.values = [];
+
+    for (const usage of contextGroup.values) {
+      const existing = newUsages.find((otherUsage) => otherUsage.privateAPIIdentifier === usage.privateAPIIdentifier);
+      if (existing === undefined) {
+        newUsages.push(usage);
+      }
+    }
+
+    contextGroup.values = newUsages;
+  }
+}
+
+export function groupedAndDeduped<Key>(
+  usages: PrivateApiUsageDto[],
+  keyForUsage: (usage: PrivateApiUsageDto) => Key,
+  areKeysEqual: (key1: Key, key2: Key) => boolean,
+) {
+  const result = grouped(usages, keyForUsage, areKeysEqual);
+  dedupeUsages(result);
+  return result;
+}
+
+/**
+ * Return value is sorted in decreasing order of usage of a given private API identifer
+ */
+export function groupedAndSortedByPrivateAPIIdentifier<Key>(
+  groupedByKey: Group<Key, PrivateApiUsageDto>[],
+): Group<string, Key>[] {
+  const flattened = groupedByKey.flatMap((group) => group.values.map((value) => ({ key: group.key, value })));
+
+  const groupedByPrivateAPIIdentifier = grouped(
+    flattened,
+    (value) => value.value.privateAPIIdentifier,
+    (id1, id2) => id1 === id2,
+  ).map((group) => ({ key: group.key, values: group.values.map((value) => value.key) }));
+
+  groupedByPrivateAPIIdentifier.sort((group1, group2) => group2.values.length - group1.values.length);
+
+  return groupedByPrivateAPIIdentifier;
+}
--- a/scripts/processPrivateApiData/runtimeContext.ts
+++ b/scripts/processPrivateApiData/runtimeContext.ts
@@ -1,0 +1,32 @@
+import { PrivateApiContextDto, PrivateApiUsageDto } from './dto';
+import { Group, groupedAndDeduped } from './grouping';
+import { joinComponents } from './utils';
+
+/**
+ * Used for determining whether two contexts are equal.
+ */
+export function runtimeContextIdentifier(context: RuntimeContext) {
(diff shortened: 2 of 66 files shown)


scripts/processPrivateApiData/grouping.ts
```diff
@@ -0,0 +1,76 @@
...
+  key: Key;
+  values: Value[];
+};
+
+export function grouped<Key, Value>(
+  values: Value[],
+  keyForValue: (value: Value) => Key,
+  areKeysEqual: (key1: Key, key2: Key) => boolean,
+) {
+  const result: Group<Key, Value>[] = [];
```
VeskeR, line 13: could we use `Map<Key, Value[]>` here to simplify the types a bit?

scripts/processPrivateApiData/runtimeContext.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Used for determining whether two contexts are equal.
+ */
+export function runtimeContextIdentifier(context: RuntimeContext) {
+  return joinComponents([
+    { key: 'type', value: context.type },
+    { key: 'file', value: context.file ?? 'null' },
+    { key: 'suite', value: context.suite?.join(',') ?? 'null' },
+    { key: 'title', value: context.type === 'definition' ? context.label : context.title },
+    { key: 'helperStack', value: 'helperStack' in context ? context.helperStack.join(',') : 'null' },
```
VeskeR, line 14: I think we should sort `helperStack` here before joining, since I believe we don't want the order in which we called `addingHelperFunction` to matter here.

scripts/processPrivateApiData/withoutPrivateAPIUsage.ts
```diff
@@ -0,0 +1,71 @@
...
+    if (
+      runtimeContext.file === test.context.file &&
+      mustSuiteHierarchyBeExecutedToRunTest(test, runtimeContext.suite)
+    ) {
+      // the hook must be executed to run this test
+      return true;
+    }
+  }
+
+  // otherwise, return true iff it’s the same test
```
VeskeR, line 39:
```suggestion
  // otherwise, return true if it’s the same test
```

VeskeR (changes requested):
This is an incredible work on private API usage! Couple of small comments below.

I really like the idea of test contexts and the design of `PrivateApiRecorder` class. I can see us expanding on this idea and adding something like `SpecItemCoverageRecorder` for aligning tests with the spec from https://github.com/ably/ably-js/pull/1806. This will give us better a better overview of which tests are covering specific spec item and how many times (given that we have things like `testOnAllTransports` we can repeat the test for the same spec item multiple times). And also we would be able to automate checking that each test case has an associated spec item (or tagged as `nospec`) by checking if it called a corresponding method from `SpecItemCoverageRecorder`.
Really looking forward to what we can do with this in the future

test/common/modules/shared_helper.js
```diff
@@ -10,299 +10,483 @@ define([
...
-   * Uses a callback to communicate the result of a `Promise`. The first argument passed to the callback will be either an error (when the promise is rejected) or `null` (when the promise is fulfilled). In the case where the promise is fulfilled, the resulting value will be passed to the callback as a second argument.
-   */
-  function whenPromiseSettles(promise, callback) {
-    promise
-      .then((result) => {
-        callback(null, result);
-      })
-      .catch((err) => {
-        callback(err);
+    async closeAndFinishAsync(realtime, err) {
```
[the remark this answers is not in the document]
lawrence-forooghian: Ah, yes, thanks!

test/common/modules/shared_helper.js
```diff
@@ -10,299 +10,483 @@ define([
...
+    }
+
+    /* testFn is assumed to be a function of realtimeOptions that returns a mocha test */
+    static testOnAllTransports(thisInDescribe, name, testFn, skip) {
+      const helper = this.forTestDefinition(thisInDescribe, name).addingHelperFunction('testOnAllTransports');
+      var itFn = skip ? it.skip : it;
+
+      function createTest(options) {
+        return function (done) {
+          this.test.helper = this.test.helper.withParameterisedTestTitle(name);
```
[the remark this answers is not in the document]
lawrence-forooghian: That would mean that the `parameterisedTestHelper` field in the JSON object emitted for this usage would end up being equal to the `title`, so it wouldn't give us any extra information. The aim with `parameterisedTestTitle`

The aim of calling `withParameterisedTestTitle` is to record which _piece of code_ this test was generated by, so that we can group all of its runtime usages into a single static usage when processing the data.
in my test

test/common/modules/private_api_recorder.js
```diff
@@ -0,0 +1,196 @@
...
+    'call.transport.send',
+    'delete.auth.authOptions.requestHeaders',
+    'deserialize.recoveryKey',
+    'listen.channel._allChannelChanges.attached',
+    'listen.channel._allChannelChanges.update',
+    'listen.connectionManager.connectiondetails',
+    'listen.connectionManager.transport.active',
+    'listen.connectionManager.transport.pending',
+    'listen.transport.disposed',
+    'new.Crypto.CipherParams', // This is in a bit of a grey area re whether it’s private API. The CipherParams class is indeed part of the public API in the spec, but it’s not clear whether you’re meant to be construct one directly, and furthermore the spec doesn’t describe it as being exposed as a property on Crypto.
```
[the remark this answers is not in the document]
lawrence-forooghian: Thanks1

scripts/processPrivateApiData/grouping.ts
lawrence-forooghian: Your question made me wonder why I had even created `Group` in the first place instead of just using `Map`! But I think it's because I wanted to use custom logic for determining whether two keys were equal, and I thought that using `Map` might make things confusing. Honestly, given that this is only a script, I don't think it's worth spending much time worrying about.

scripts/processPrivateApiData/runtimeContext.ts
lawrence-forooghian: I’m not sure I understand, but I don't think that sorting it would make any difference to the output. If you have `helperA` which calls `helperB` which calls `helperC`, it's unlikely that there’s going to be, for example, a situation in which `helperB` calls `helperC` which calls `helperA`.

lawrence-forooghian:
Resolving Andrii’s comments (he’s on holiday and said I could do so).

PR #1802, opened 2024-06-24
Fix RTL13a (stop auto-attaching from inappropriate states)

SimonWoolf:
https://sdk.ably.com/builds/ably/specification/main/features/#RTL13


ttypic (approved):
LGTM

PR #1806, opened 2024-06-26
[ECO-4757] Align ably-js tests with spec items

VeskeR:


lawrence-forooghian (commented):
This is a heroic effort, well done! I haven't looked at all of the annotations in detail, but here are a few thoughts on the approach:

- It would probably be good to add your above description of the various tags as documentation somewhere in the codebase, along with something indicating that they should be added for future tests.
- Is there a way of knowing whether a given spec item is currently fully tested or not? The presence of an `@spec` for a given spec item indicates that it is fully tested. But what about in the case where an item is fully tested but across multiple tests? In that case there will be multiple `@specpartial` for that spec item, but it's not clear whether there's a way of knowing whether the spec item is fully tested or not.
- I wonder whether there's any value in marking the React Hooks tests? Given that it's a completely separate product to the SDK, we already know that doesn't implement any of the spec points.

VeskeR:
> * It would probably be good to add your above description of the various tags as documentation somewhere in the codebase, along with something saying that these tags should be added for future tests.

Good idea! Added it to CONTRIBUTING.md for now, you can check the wording there.

> * Is there a way of knowing whether a given spec item is currently fully tested or not? The presence of an `@spec` for a given spec item indicates that it is fully tested. But what about in the case where an item is fully tested but across multiple tests? In that case there will be multiple `@specpartial` for that spec item, but it's not clear whether there's a way of knowing whether the spec item is fully tested or not.

Unfortunately, no. I was encountering situations like you described, where a spec item was technically fully covered, but only by a set of tests. I couldn't figure out a convenient and non-cumbersome way to mark such tests so it would be possible to deduce that the spec item is fully covered by the combination of those tests. I decided to leave it as `@specpartial` for now, and I believe it's better to refactor the spec items themselves. If we need a couple of different tests to cover a spec item, usually it's a sign that we can split the spec item into more granular ones. This will be part of [ECO-12](https://ably.atlassian.net/browse/ECO-12).

> * I wonder whether there's any value in marking the React Hooks tests? Given that it's a completely separate product to the SDK, we already know that doesn't implement any of the spec points.

Consistency + I wanted to be explicit about which spec items each test is covering (if any) and leave no room for interpretation. If we don't mark those tests, then a new person may wonder why they are not marked and if the spec for them actually exists somewhere. On the other hand, `@nospec` immediately indicates that there is simply no spec for them.

lawrence-forooghian:
> Good idea! Added it to CONTRIBUTING.md for now, you can check the wording there.

Cool, looks good.

> I decided to leave it as @specpartial for now, and I believe it's better to refactor the spec items themselves. If we need a couple of different tests to cover a spec item, usually it's a sign that we can split the spec item into more granular ones. This will be part of [ECO-12](https://ably.atlassian.net/browse/ECO-12).

I’m curious to see if / how this splitting will actually work in practice, but OK, let's wait and see.

> Consistency + I wanted to be explicit about which spec items each test is covering (if any) and leave no room for interpretation. If we don't mark those tests, then a new person may wonder why they are not marked and if the spec for them actually exists somewhere. On the other hand, @nospec immediately indicates that there is simply no spec for them.

Do you not think we could just address this with a line in the `CONTRIBUTING.md` notes to say that we don't mark the React tests? Doesn't block merging though; will approve and I’ll leave it to you to decide what to do.

lawrence-forooghian (approved):

VeskeR:
> Do you not think we could just address this with a line in the CONTRIBUTING.md notes to say that we don't mark the React tests? Doesn't block merging though; will approve and I’ll leave it to you to decide what to do.

I guess let's see what other people think too, it's a simple change I can do if more people prefer not to tag react hooks tests.
@ttypic @umair-ably wdyt?

ttypic (approved):
Looks amazing, I like it!

PR #1808, opened 2024-07-03
[ECO-4837] Update "Using Webpack" section in README for importing ably from static files

### LINKED ISSUE #1792: Update "Using WebPack" section in README about importing ably directly from static file
ably-js v2 uses `exports` field in its package.json, so bundlers that respect this field (like webpack v5) won't be able to resolve `import * as Ably from 'ably/build/ably.js'` which is currently suggested in README [here](https://github.com/ably/ably-js/blob/efb71d00c695f6e2576a9166e2d82b0c6a7b6fb5/README.md?plain=1#L87). It fails with `Error: Module not found: Error: Package path ./build/ably.js is not exported from package ...\node_modules\ably (see exports field in ...\node_modules\ably\package.json)`.

Instead the correct workaround is to use alias setting in webpack config like this:
```javascript
// webpack.config.js
const path = require('path');

module.exports = {
  module: {
    rules: [
      {
        resolve: {
          alias: {
            ably: path.resolve(__dirname, 'node_modules/ably/build/ably.js'),
          },
        },
      },
    ],
  },
};
```
See internal [slack thread](https://ably-real-time.slack.com/archives/CURL4U2FP/p1718124617006409) for more context.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4837) by [Unito](https://www.unito.io)

VeskeR:
Resolves #1792

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,9 +62,9 @@ var Ably = require('ably');
 
 For usage, jump to [Using the Realtime API](#using-the-realtime-api) or [Using the REST API](#using-the-rest-api).
 
-#### Serverside usage with webpack
+#### Serverside usage with Webpack
 
-If you are using a version older than 1.2.5 you will need to add 'ably' to `externals` in your webpack config to exclude it from webpack processing, and require and use it in as a external module using require('ably') as above.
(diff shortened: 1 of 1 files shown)


ttypic (approved):
LGTM

PR #1811, opened 2024-07-04
[ECO-4852] Gather spec items coverage information

VeskeR:
This PR is based on https://github.com/ably/ably-js/pull/1806, please review that one first.

This PR introduces new `specCoverageReport.ts` script to gather spec items coverage by ably-js tests (from docstring tags) and compare them with the [Ably features specification](https://sdk.ably.com/builds/ably/specification/main/features/).
Script fetches the latest spec from Ably specification repo using link to the textile file: https://raw.githubusercontent.com/ably/specification/main/textile/features.textile. Then parses textile file and docstrings in ably-js tests to output spec item coverage table.

This PR also adds new `Spec coverage report` workflow, which simply runs `npm run speccoveragereport` to output spec coverage report. See [job output here](https://github.com/ably/ably-js/actions/runs/9811467632/job/27093676079?pr=1811).

Here's the unformatted spec coverage data gathered based on test suite state for a1229af5325b28257b8ad35510e4d45b8bd80d94 commit: 
[spec-coverage-result.json](https://github.com/user-attachments/files/16097642/spec-coverage-result.json)

### DIFF
--- a/.github/workflows/spec-coverage-report.yml
+++ b/.github/workflows/spec-coverage-report.yml
@@ -1,0 +1,25 @@
+name: Spec coverage report
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+
+jobs:
+  spec-coverage-report:
+    runs-on: ubuntu-latest
+    permissions:
+      deployments: write
+      id-token: write
+    steps:
+      - uses: actions/checkout@v2
+      - name: Reconfigure git to use HTTP authentication
+        run: >
+          git config --global url."https://github.com/".insteadOf
+          ssh://git@github.com/
+      - name: Use Node.js 20.x
+        uses: actions/setup-node@v1
+        with:
+          node-version: 20.x
+      - run: npm ci
+      - run: npm run speccoveragereport
--- a/scripts/specCoverageReport.ts
+++ b/scripts/specCoverageReport.ts
@@ -35,6 +35,13 @@ interface TestSpecCoverage {
   testCode: string;
 }
 
+interface SpecItemCoverageStatus {
+  hasFullCoverage: boolean;
+  hasPartialCoverage: boolean;
+  testsContributingToFullCoverageCount: number;
+  testsContributingToPartialCoverageCount: number;
+}
+
 class SpecItemNode implements SpecItem {
   public level: number;
   public key: string;
@@ -42,7 +49,7 @@ class SpecItemNode implements SpecItem {
   public parentNode: SpecItemNode | null = null;
   public childNodes: SpecItemNode[] = [];
   public specCoverages: TestSpecCoverage[] = [];
-  private _coveragePercentage: number | null = null;
+  private _coverageStatus: SpecItemCoverageStatus | null = null;
 
   constructor(specItem: SpecItem) {
     this.level = specItem.level;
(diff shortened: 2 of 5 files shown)


scripts/specCoverageReport.ts
```diff
@@ -0,0 +1,424 @@
...
+  /**
+   * Spec item key covered by the test, if applicable based on {@link coverageType} value.
+   */
+  specKey?: string;
+  /**
+   * An optional comment for tagged spec item. Looks like this in test docstrings:
+   *
+   * `@spec ID123 - [comment]`
+   */
+  coverageDescription?: string;
```
lawrence-forooghian, line 34: Might be good to update the `CONTRIBUTING.md` guidance in #1806 to mention the ability to provide a comment alongside the tag?

scripts/specCoverageReport.ts
```diff
@@ -0,0 +1,424 @@
...
+  }
+
+  public isLeaf(): boolean {
+    return this.childNodes.length === 0;
+  }
+
+  public getCoveragePercentage(): number {
+    // general rules for coverage:
+    // if the spec item is covered by the test with @spec tag - it contributes a weigth of 1 (100%).
+    // if the spec item is covered by the test with @specpartial tag - it contributes a weigth of 0.5 (50%).
```
lawrence-forooghian, line 72: I’m not sure that this 50% rule is very intuitive; it seems a bit arbitrary and makes the data in the table seem more precise than it actually is. How about we instead have two columns indicating different coverage metrics:

- "Coverage (full)" — only `@spec` tags count towards this one
- "Coverage (partial)" — `@spec` or `@specpartial` tags count towards this one (i.e. if either of these tags exist for a spec point then the spec point is "covered" in this metric)

scripts/specCoverageReport.ts
```diff
@@ -0,0 +1,424 @@
...
+    if (this.specCoverages.findIndex((x) => x.coverageType === SpecCoverageType.specpartial) !== -1) {
+      return 0.5;
+    }
+
+    return 0;
+  }
+
+  private _getCoveragePercentageForParent(): number {
+    // here we use the next assumption:
+    // if all children spec items are covered by the test (average coverage is 1), then we can consider the parent covered too,
```
lawrence-forooghian, line 138: So, we're using information about the ably-js test suite (i.e. whether a given spec point has a test) to infer information about the spec (i.e. whether a given spec point needs a test)? I’m not sure that’s a good idea; it feels like the spec should be the source of truth.

I think that it'd be better if we could supply some metadata alongside the spec indicating which spec points are describing a testable behaviour, and which are either purely informational or purely a heading for their sub-points. Until then, I’d suggest that we perhaps not try and infer whether a given spec point needs to be explicitly tested or not.

lawrence-forooghian (commented):
I think this is a great start. I do have some concerns about how we're trying to apply heuristics in order to get around the fact that the spec, as currently written, doesn't allow us to gather great data about how much of it a given test suite implements. I think I’d prefer that we not try and hide these limitations, and then work on improving the spec.

scripts/specCoverageReport.ts
VeskeR: true, will add a format example for tags in the `CONTRIBUTING.md` in that PR

VeskeR: added guidance and examples to `CONTRIBUTING.md` in https://github.com/ably/ably-js/pull/1806/commits/25c7f5cde53955fc8c7d51fb927d268739107e32 (https://github.com/ably/ably-js/pull/1806 PR)

scripts/specCoverageReport.ts
VeskeR: Sounds reasonable. I agree that partial coverage is quite vague and doesn't necessarily mean that "50%" of the spec is covered by the partial test, so splitting them is probably better. Will change coverage calculation to use your suggestion

scripts/specCoverageReport.ts
VeskeR: > I’m not sure that’s a good idea; it feels like the spec should be the source of truth.

Definitely agree. I was trying to work with the fact that some spec item parents in the spec are only described and can be tested by their children spec items, but I can see the shortcomings of this approach. Test suite should no do such assumptions.

> I think that it'd be better if we could supply some metadata alongside the spec indicating which spec points are describing a testable behaviour, and which are either purely informational or purely a heading for their sub-points.

Yep, same idea was floating around regarding future improvements to the spec.
As you've suggested, until then, I'll change the report to expect that every spec item is currently needs to be tested, as there is no other indication available atm.

scripts/specCoverageReport.ts
VeskeR: Updated in https://github.com/ably/ably-js/pull/1811/commits/b1ab4b0c64fcec47cc5c798e498be17036d0ce7c

scripts/specCoverageReport.ts
VeskeR: Updated in https://github.com/ably/ably-js/pull/1811/commits/b1ab4b0c64fcec47cc5c798e498be17036d0ce7c

.github/workflows/spec-coverage-report.yml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+  spec-coverage-report:
+    runs-on: ubuntu-latest
+    permissions:
+      deployments: write
+      id-token: write
+    steps:
+      - uses: actions/checkout@v2
+      - name: Reconfigure git to use HTTP authentication
```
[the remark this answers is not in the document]
lawrence-forooghian: I imagine this is just a copy and paste from the other workflows. As for using HTTP authentication, there's not much explanation in the PR or commit that introduced it (https://github.com/ably/ably-js/pull/782).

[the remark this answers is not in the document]
ttypic: Thanks @lawrence-forooghian. No questions here then, worth to look at this separately

ttypic (approved):

PR #1815, opened 2024-07-16
Remove usage of `Defaults.protocolVersion` private API in tests

lawrence-forooghian:
I’m doing this as part of removing private API usage so that this test suite can be reused as a unified test suite for our client libraries, but it’s a change that I think actually makes sense even without this motivation; there’s no relation between the `request` parameters used by these tests and the protocol version baked into the library.


lawrence-forooghian:
(The test failures don’t seem related to the changes here; not sure if something is going on with sandbox.)

VeskeR (approved):

PR #1823, opened 2024-07-29
Upgrade Prettier and run on all supported files

lawrence-forooghian:
(Context: I was making some changes to the GitHub workflow files elsewhere and noticed that Git was showing me trailing whitespace, meaning that these files weren’t being automatically formatted. So thought I’d fix that.)


ttypic (approved):
Nice work, I would suggest merging without https://github.com/ably/ably-js/pull/1824

Issue #1844, opened 2024-08-12
Connection recovery not working


michalzaq12:
Ably ver: 2.3.1

**Description**
Realtime connection (websocket, only websocket transport provided) is not restored after network connection is lost. SDK tries to connect without success (network is already available). 

**My observation**
1. When a network problem occurs, `checkWsConnectivity()` is called, which of course fails (that's fine).
2. `checkWsConnectivity()` can by called only once (WHY?) , so `wsCheckResult` is always `false` https://github.com/ably/ably-js/blob/8ba8ec29f521b18ffd16bc4fcb33004c64a7d706/src/common/lib/transport/connectionmanager.ts#L1104
3.  `tryTransportWithFallbacks()` is called when trying to reconnect
4. `shouldContinue()` in `tryTransportWithFallbacks()` body always returns `false`, which call `transport.dispose()`

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4914) by [Unito](https://www.unito.io)

sacOO7:
@michalzaq12 reconnection works in following manner
1. Once client is disconnected, client checks if fallbacks can be used. Internet check is part of the same. 
2. When internet is available, all fallbacks ( 5 default ) are tried. If none of them succeed, connection goes into disconnected state again
3. New connection is retried again after `disconnectedRetryTimeout`

sacOO7:
@michalzaq12 Can you check confirm if connection recovery works after 15 seconds ?
Ideally, it should work as long as connection is in either `disconnected` or `suspended` state.
Only when connection goes into `closed` state, connection is not retried.

michalzaq12:
> Once client is disconnected, client checks if fallbacks can be used. Internet check is part of the same.
When internet is available, all fallbacks ( 5 default ) are tried. If none of them succeed, connection goes into disconnected state again
New connection is retried again after disconnectedRetryTimeout, value is 15 seconds.

I know exactly how it works. It doesn't change the fact that there's some bug in the implementation (regression from v1).  

> Can you check confirm if connection recovery works after 15 seconds ?

After 15 seconds, the SDK tries to reconnect but fails every time.

michalzaq12:
Is there any template for minimal reproducible repo?

sacOO7:
@michalzaq12 there's no template as such. It will be great if you can either post code here or best if create a separate repository with steps to reproduce the bug. That way, we can reproduce it given environment i.e. nodejs, browser etc

[...]

PR #1855, opened 2024-09-06
[ECO-4914] Fix websocket reconnection can get stuck in disconnected/connecting cycle

VeskeR:
This fixes a regression introduced in ably-js v2 in this [PR](https://github.com/ably/ably-js/pull/1645).
Under certain network conditions (internet down) a race condition could
occur in the connection manager. Specifically, websocket connectivity
check in the `startWebSocketSlowTimer` function needed to set the
`wsCheckResult` flag to `false` before the timer was cleared by other
code branches. If that happened, it then caused an endless loop of
websocket connection retries, even after the network conditions
stabilized. A websocket transport would successfully open but then be
immediately disposed due to the `wsCheckResult` flag still being set to
`false`, without any mechanism to reset it in this scenario.

Upon closer look, it makes sense that the websocket connection health
flags (`wsCheckResult` and `abandonedWebSocket`) should be reset when
attempting a new websocket connection. The previous solution did not
explicitly reset these flags in the `connectWs` function, leading to the
described race condition and endless loop. Thus, `wsCheckResult` and
`abandonedWebSocket` are now reset to their neutral values upon entering
the `connectWs` function.

The websocket reconnection test that was added now checks such condition. You can see it fail on all envrionments on a library version without the aforementioned fix: [node](https://github.com/ably/ably-js/actions/runs/10741768647/job/29792927662?pr=1855), [browsers](https://github.com/ably/ably-js/actions/runs/10741768651/job/29792928367?pr=1855)

Resolves https://github.com/ably/ably-js/issues/1844

## Summary by CodeRabbit

- **Tests**
	- Added a new test case to verify WebSocket reconnection behavior after connectivity failures, enhancing the reliability of the WebSocket transport mechanism.
- **Improvements**
	- Refined the handling of WebSocket connectivity checks to improve the reliability and clarity of connection state management.

### DIFF
--- a/src/common/lib/transport/connectionmanager.ts
+++ b/src/common/lib/transport/connectionmanager.ts
@@ -1101,27 +1101,25 @@ class ConnectionManager extends EventEmitter {
         'ConnectionManager WebSocket slow timer',
         'checking connectivity',
       );
-      if (this.wsCheckResult === null) {
-        this.checkWsConnectivity()
-          .then(() => {
-            Logger.logAction(
-              this.logger,
-              Logger.LOG_MINOR,
-              'ConnectionManager WebSocket slow timer',
-              'ws connectivity check succeeded',
-            );
-            this.wsCheckResult = true;
-          })
-          .catch(() => {
-            Logger.logAction(
-              this.logger,
-              Logger.LOG_MAJOR,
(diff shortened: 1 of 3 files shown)


ttypic (approved):
LGTM

PR #1879, opened 2024-09-29
Ignore instance id when comparing connections keys

amnonbc:
The connection key returned by frontdoor has the form

`<instance-id>!<hmac>-<handleID>`

where `<hmac>` is a hash of connectionID, appID, and clientID.

The test `page_refresh_with_recovery` checks that the connection ID which we get when a connection is recovered, 
it gets the same connectionID. It does this by stripping the handleID off the end of the connection key.
But in frontdoor, requests are handled by an arbitrary frontdoor instance, which is not necessarily the same instance
as the previous connection. So this PR ignores the instance id when performing the comparison.

Fixes https://ably.atlassian.net/browse/RAR-712

## Summary by CodeRabbit

- **New Features**
	- Introduced a function for improved handling of connection keys, enhancing the accuracy of connection comparisons.

- **Bug Fixes**
	- Enhanced logic for determining if two connection keys represent the same connection by utilizing a new extraction method.

### DIFF
--- a/test/browser/connection.test.js
+++ b/test/browser/connection.test.js
@@ -213,10 +213,24 @@ define(['shared_helper', 'chai'], function (Helper, chai) {
         });
       });
 
+      // strip instanceID and handleID from connectionKey */
+      function connectionHmac(key) {
+        /* connectionKey has the form <instanceID>!<hmac>-<handleID> */
+
+        /* remove the handleID from the end of key */
+        let k = key.split('-')[0];
+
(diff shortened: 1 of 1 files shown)


amnonbc:
@coderabbitai review

amnonbc:
@coderabbitai review

amnonbc:
@coderabbitai review

owenpearson (approved):
looks good!

PR #1884, opened 2024-10-04
[ECO-5002] Fix usePresence won't leave presence if unmount triggered during channel attaching

VeskeR:
This fixes a race condition issue, where a React app might trigger a component unmount while channel is still in the `attaching` state, thus `presence.leave` was never called for a channel and member would remain there until app reload.

Partially fixes https://ably.atlassian.net/browse/ECO-5002, see more context in the [internal slack thread](https://ably-real-time.slack.com/archives/CURL4U2FP/p1727284637623509).

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved logic for leaving presence, ensuring it only occurs under specific channel states ('attached' or 'attaching') and when the connection is active.
- **Documentation**
	- Updated comments for clarity on conditions affecting presence status.


ttypic (approved):
LGTM

PR #1887, opened 2024-10-09
[DTP-950] Handle initial state sync sequence

VeskeR:
This PR adds necessary realtime client changes to be able to handle new STATE_SYNC messages by the LiveObjects plugin. The pool initialization itself is implemented in https://github.com/ably/ably-js/pull/1890. I would suggest reviewing this PR commit by commit.
Bit flag indexes for new [bit flags](https://github.com/ably/ably-js/pull/1887/files#diff-cc90da29e902e3f8c2c5aee6e245c06684bb6b36e1452d2f5f22a3776aad91d0R55) are based on this: https://github.com/ably/realtime/blob/e3d86b386850b81f09745c3ad3360c976173e0b1/go/realtime/lib/channel/flags.go#L22

Tests for SYNC sequence are added in https://github.com/ably/ably-js/pull/1894.

Resolves [DTP-950](https://ably.atlassian.net/browse/DTP-950)

[DTP-950]: https://ably.atlassian.net/browse/DTP-950?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **New Features**
	- Introduced the `LiveObjectsPlugin` for enhanced state management and message handling.
	- New types and properties added for improved channel modes and options.
	- Enhanced message decoding capabilities with new methods for state operations.
	- New classes and methods for managing live object data and synchronization processes.
	- Updated handling of protocol messages to include live object data.

- **Bug Fixes**
	- Updated protocol message handling to ensure consistency across various platforms.

- **Documentation**
	- Updated documentation to reflect new interfaces and methods related to the `LiveObjectsPlugin`.

PR #1898, opened 2024-10-22
Minor LiveObjects fixes/refactoring

VeskeR:
Some minor fixes/changes suggested by @coderabbitai in https://github.com/ably/ably-js/pull/1887. Opening a separate PR to not drag merging that one for longer

## Summary by CodeRabbit

- **New Features**
	- Improved asynchronous handling in the decoding process for state messages, enhancing error management.
  
- **Bug Fixes**
	- Simplified logic for identifying extra object IDs in the LiveObjectsPool, improving readability and efficiency.
  
- **Refactor**
	- Optimized `isEmpty` and `reset` methods in the SyncLiveObjectsDataPool for better performance.
	- Removed the `arrSubtract` utility function to streamline codebase.

### DIFF
--- a/src/plugins/liveobjects/liveobjectspool.ts
+++ b/src/plugins/liveobjects/liveobjectspool.ts
@@ -27,7 +27,7 @@ export class LiveObjectsPool {
    */
   deleteExtraObjectIds(objectIds: string[]): void {
     const poolObjectIds = [...this._pool.keys()];
-    const extraObjectIds = this._client.Utils.arrSubtract(poolObjectIds, objectIds);
+    const extraObjectIds = poolObjectIds.filter((x) => !objectIds.includes(x));
 
     extraObjectIds.forEach((x) => this._pool.delete(x));
   }
--- a/src/plugins/liveobjects/statemessage.ts
+++ b/src/plugins/liveobjects/statemessage.ts
@@ -155,13 +155,13 @@ export class StateMessage {
 
     if (message.object?.map?.entries) {
       for (const entry of Object.values(message.object.map.entries)) {
-        decodeMapEntry(entry, inputContext, decodeDataFn);
+        await decodeMapEntry(entry, inputContext, decodeDataFn);
       }
     }
 
     if (message.operation?.map) {
       for (const entry of Object.values(message.operation.map)) {
-        decodeMapEntry(entry, inputContext, decodeDataFn);
+        await decodeMapEntry(entry, inputContext, decodeDataFn);
       }
     }
 
--- a/src/common/lib/util/utils.ts
+++ b/src/common/lib/util/utils.ts
@@ -165,15 +165,6 @@ export function arrIntersectOb<K extends string>(arr: Array<K>, ob: Partial<Reco
   return result;
 }
 
-export function arrSubtract<T>(arr1: Array<T>, arr2: Array<T>): Array<T> {
-  const result = [];
-  for (let i = 0; i < arr1.length; i++) {
-    const element = arr1[i];
-    if (arr2.indexOf(element) == -1) result.push(element);
-  }
-  return result;
-}
-
 export function arrDeleteValue<T>(arr: Array<T>, val: T): boolean {
   const idx = arr.indexOf(val);
   const res = idx != -1;
--- a/src/plugins/liveobjects/syncliveobjectsdatapool.ts
+++ b/src/plugins/liveobjects/syncliveobjectsdatapool.ts
@@ -47,11 +47,11 @@ export class SyncLiveObjectsDataPool {
   }
 
   isEmpty(): boolean {
-    return this.size() === 0;
+    return this._pool.size === 0;
   }
 
   reset(): void {
-    this._pool = new Map<string, AnyDataEntry>();
+    this._pool.clear();
   }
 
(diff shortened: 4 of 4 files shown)


mschristensen (approved):

src/plugins/liveobjects/liveobjectspool.ts
```diff
@@ -27,7 +27,7 @@ export class LiveObjectsPool {
    */
   deleteExtraObjectIds(objectIds: string[]): void {
     const poolObjectIds = [...this._pool.keys()];
-    const extraObjectIds = this._client.Utils.arrSubtract(poolObjectIds, objectIds);
+    const extraObjectIds = poolObjectIds.filter((x) => !objectIds.includes(x));
```
owenpearson, line 30: this is good and will be slightly more performant. would be good actually to change the util too

src/plugins/liveobjects/statemessage.ts
```diff
@@ -155,13 +155,13 @@ export class StateMessage {
 
     if (message.object?.map?.entries) {
       for (const entry of Object.values(message.object.map.entries)) {
-        decodeMapEntry(entry, inputContext, decodeDataFn);
+        await decodeMapEntry(entry, inputContext, decodeDataFn);
```
owenpearson, line 158: is it possible to decode multiple map entries in parallel? if so then it will surely be way more efficient to build an array of promises and use `Promise.all`

owenpearson (changes requested):

src/plugins/liveobjects/liveobjectspool.ts
VeskeR: `Utils.arrSubtract` is not used anywhere it seems, would you prefer to update it to have this .filter/.includes and keep it, or just remove `Utils.arrSubtract` at this point?

owenpearson: ah right, yeah i'd say just remove it 👍

src/plugins/liveobjects/statemessage.ts
VeskeR: `decodeMapEntry` ends up calling the same [decodeData](https://github.com/ably/ably-js/blob/fbf9d422195aad85b291d1f46db010a9853742e9/src/common/lib/types/message.ts#L166) function for Message as we do when processing regular [channel messages](https://github.com/ably/ably-js/blob/fbf9d422195aad85b291d1f46db010a9853742e9/src/common/lib/client/realtimechannel.ts#L701) and [presence messages](https://github.com/ably/ably-js/blob/fbf9d422195aad85b291d1f46db010a9853742e9/src/common/lib/client/realtimechannel.ts#L605), and they do decoding sequentially.
there is only one possible async call in this case, and it's [cipher.decrypt](https://github.com/ably/ably-js/blob/fbf9d422195aad85b291d1f46db010a9853742e9/src/common/lib/types/message.ts#L218)

VeskeR: it seems that we only support [AES in CBC mode](https://github.com/ably/ably-js/blob/fbf9d422195aad85b291d1f46db010a9853742e9/ably.d.ts#L764-L785), so I assume we can't decrypt concurrently?

owenpearson: i'm definitely not an expert on cryptography, but afaict, while CBC mode doesn't allow parallel encryption, parallel decryption is perfectly fine

owenpearson: i'd say we should change all 3 implementations to use `Promise.all`, worst that can happen is that the js engine processes them serially anyway.

either way, since it affects normal channel messages as well as liveobjects i'm happy for that to happen in another pr

VeskeR: > parallel decryption is perfectly fine

yea, that's what google tells me too.
also, our `decrypt` [implementation ](https://github.com/ably/ably-js/blob/fbf9d422195aad85b291d1f46db010a9853742e9/src/platform/web/lib/util/crypto.ts#L223-L233) (included web only, but node.js is equivalent) doesn't seem to even be dependant on the previous ciphertext we received, as the iv is always included at the start of the ciphertext.

VeskeR: at the same time, I don't really feel comfortable changing something like this as part of the LiveObjects work.
would you mine me creating a separate issue for this and asking people around if we can parallelize decryption in general? that would benefit regular and presence messages too.

owenpearson: yep of course, sgtm 👍

VeskeR: > either way, since it affects normal channel messages as well as liveobjects i'm happy for that to happen in another pr

yea, I would prefer that, will create an issue for that

VeskeR: created https://github.com/ably/ably-js/issues/1906

src/plugins/liveobjects/liveobjectspool.ts
VeskeR: removed in https://github.com/ably/ably-js/pull/1898/commits/b73a15cf34b876ba4c6f1375c5f425cea18f6dbc

owenpearson (approved):

PR #1903, opened 2024-10-22
Testing: handle case when connectionKey contains more than one '-'

amnonbc:
## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of connection states during online and offline transitions.
	- Enhanced error management for connection tests to ensure proper resource cleanup.

- **New Features**
	- Added comprehensive test cases for connection recovery and persistence of transport settings.
	- Refined event simulation for accurate representation of network connectivity changes.

### DIFF
--- a/test/browser/connection.test.js
+++ b/test/browser/connection.test.js
@@ -218,13 +218,16 @@ define(['shared_helper', 'chai'], function (Helper, chai) {
         /* connectionKey has the form <instanceID>!<hmac>-<handleID> */
 
         /* remove the handleID from the end of key */
-        let k = key.split('-')[0];
+        const lastIndex = key.lastIndexOf('-');
+        if (lastIndex !== -1) {
+          key = key.slice(0, lastIndex);
+        }
 
         /* skip the server instanceID if present, as reconnects may be routed to different frontends */
-        if (k.includes('!')) {
-          k = k.split('!')[1];
+        if (key.includes('!')) {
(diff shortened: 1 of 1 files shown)


test/browser/connection.test.js
```diff
@@ -218,13 +218,16 @@ define(['shared_helper', 'chai'], function (Helper, chai) {
         /* connectionKey has the form <instanceID>!<hmac>-<handleID> */
 
         /* remove the handleID from the end of key */
-        let k = key.split('-')[0];
+        const lastIndex = key.lastIndexOf('-');
+        if (lastIndex !== -1) {
+          key = key.slice(0, lastIndex);
```
VeskeR, line 223: I think `key` is never declared now?
should add `let key;` before if statement

amnonbc: line 217, it is declares as a parameter of function `connectionHmac`

VeskeR: right, it was outside of the visible diff, didn't see it

owenpearson: i do think we should revert back to `let k ... return k` - it's bad practice to mutate arguments like this

VeskeR: > i do think we should revert back to let k ... return k - it's bad practice to mutate arguments like this

agree

amnonbc:
Closing this. Superseded by https://github.com/ably/ably-js/pull/1939

PR #1917, opened 2024-11-05
[DTP-958, DTP-959] Liveobjects/subscriptions

VeskeR:
Implements direct object subscription for Live Objects:
- subscription callback is invoked when the live object data is updated via an incoming state operation
- subscription callback is invoked when the live object data is updated via a sync sequence (once sync sequence is applied to all objects)
- update object is passed to a callback function that describes a granular update made to the live object

Resolves [DTP-958](https://ably.atlassian.net/browse/DTP-958), [DTP-959](https://ably.atlassian.net/browse/DTP-959)


src/plugins/liveobjects/liveobject.ts
```diff
@@ -1,31 +1,80 @@
...
   }
 
+  subscribe(listener: (update: TUpdate) => void): SubscribeResponse {
+    this._eventEmitter.on(LiveObjectEvents.Updated, listener);
+
+    const unsubscribe = () => {
+      this._eventEmitter.off(LiveObjectEvents.Updated, listener);
+    };
+
+    return { unsubscribe };
```
owenpearson, line 60: this is an interesting pattern, i'm not against it but i wonder if it's a good idea to add it here without adding it elsewhere for `channel.subscribe`, `presence.subscribe`, `EventEmitter.on`, etc?

src/plugins/liveobjects/liveobject.ts
```diff
@@ -41,10 +90,14 @@ export abstract class LiveObject<T extends LiveObjectData = LiveObjectData> {
...
   /**
+   * Sets a new data reference for the LiveObject and returns an update object that describes the changes applied based on the object's previous value.
+   *
    * @internal
    */
-  setData(newDataRef: T): void {
+  setData(newDataRef: TData): TUpdate {
+    const update = this._updateFromDataDiff(newDataRef);
     this._dataRef = newDataRef;
+    return update;
```
owenpearson, line 100: very minor, but
```suggestion
    this._dataRef = newDataRef;
    return this._updateFromDataDiff(newDataRef);
```

owenpearson (approved):
looks great! a couple of non-essential comments:

src/plugins/liveobjects/liveobject.ts
```diff
@@ -1,31 +1,80 @@
 import type BaseClient from 'common/lib/client/baseclient';
+import type EventEmitter from 'common/lib/util/eventemitter';
```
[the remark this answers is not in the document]
VeskeR: it is not, EventEmitter class is exposed on a client object which is injected in our LiveObject class

src/plugins/liveobjects/livemap.ts
```diff
@@ -40,7 +42,11 @@ export interface LiveMapData extends LiveObjectData {
   data: Map<string, MapEntry>;
 }
 
-export class LiveMap extends LiveObject<LiveMapData> {
+export interface LiveMapUpdate extends LiveObjectUpdate {
+  update: { [keyName: string]: 'updated' | 'deleted' };
```
mschristensen, line 46: shall we call this update type `removed` instead of `deleted`, to match the internal name we use for the op? (I think we may have discussed this before but I can't remember the outcome)

src/plugins/liveobjects/livemap.ts
```diff
@@ -174,12 +188,63 @@ export class LiveMap extends LiveObject<LiveMapData> {
...
+        continue;
+      }
+      if (currentEntry.tombstone === false && newEntry.tombstone === true) {
+        // current prop is not tombstoned, but new is. it means prop was deleted
+        update.update[key] = 'deleted';
+        continue;
+      }
+
+      // if tombstones are matching, need to compare values with deep equals to see if it was changed
+      const valueChanged = !deepEqual(currentEntry.data, newEntry.data, { strict: true });
```
mschristensen, line 238: If objects are both tombstones, is that really an `update`?

mschristensen (commented):
Great work

src/plugins/liveobjects/livemap.ts
VeskeR: assuming MAP_REMOVE op doesn't send a `data` field for some reason, it is going to `undefined`, just as the local `data`. so `valueChanged = undefined !== undefined; // false`, there will be no update.
however, I can make it as an explicit case just to make sure it always behaves like that

src/plugins/liveobjects/livemap.ts
VeskeR: yea, we decided to keep parity with realtime, so `removed`.
this update object in the subscription api DR just had it still as `deleted`, so I used that.
I've updated the DR and will change it here to `removed ` too.

src/plugins/liveobjects/liveobject.ts
VeskeR: the order was intentional - `_updateFromDataDiff` uses the current data from `this._dataRef`, so you are supposed to call it before changin the `this._dataRef`.
however, I can see how this is easily missed and error-prone. I will change it to explicitly expect a `currentDataRef` parameter, so it is more intuitive how you need to call it

src/plugins/liveobjects/liveobject.ts
VeskeR: unfortunately it will be a breaking change for `channel.subscribe`, as it currently [returns the result](https://github.com/ably/ably-js/blob/18a255948c38d1e60715c8f5d6173369b57cb8d6/src/common/lib/client/realtimechannel.ts#L432) of the `channel.attach` call, a `ChannelStateChange` instance, if implicit attach is used (and it is by default).

we "could" add it to `presence.subsrcibe` and `EventEmitter.on` it seems, but we won't be able to keep parity in the API fully without this breaking change, and without specification changes.

VeskeR: I do agree that it may be not the best option to have this API only in one place, and Mike raised the same concerns about API parity here https://ably.atlassian.net/wiki/spaces/LOB/pages/3354787841/LODR-019+Realtime+Client+SDK+public+API+-+Subscriptions?atl_f=PAGETREE#:~:text=For%20a%20Counter%2C%20we%20would%20get%20an%20update%20when%20it%27s%20been%20incremented., which is why your regular `liveObject.unsubscribe(listener)` call is still supported, just like `channel.unsubscribe(listener)` (however, I do not allow calling `liveObject.unsubscribe()` without arguments, as it is just straight up buggy and should not be the API for channels/presence neither)

at the same time it just feels wrong to not include a really nice developer API just to make sure we keep parity with other inferior API

src/plugins/liveobjects/livemap.ts
VeskeR: changed in https://github.com/ably/ably-js/compare/e0fe4faf6bfba6398581965118e15bba0ba633c7..cf43cb204060344b6606f7c180d41ff5681b32a6

src/plugins/liveobjects/liveobject.ts
VeskeR: updated in https://github.com/ably/ably-js/compare/cf43cb204060344b6606f7c180d41ff5681b32a6..4bca1c797caa207de96f3ff767d916f2426489b8

src/plugins/liveobjects/liveobject.ts
owenpearson: ok good point about the breaking change for `channel.subscribe`. i'd be interested to know why calling `unsubscribe()` without arguments is buggy but i suppose that doesn't affect this PR so will resolve the comment for now

src/plugins/liveobjects/liveobject.ts
owenpearson: looks better now, thanks!

PR #1923, opened 2024-11-21
Mutable message field changes and Message/PresenceMessage.fromValues refactoring

SimonWoolf:
Implements https://github.com/ably/specification/pull/235 

Also ended up on a bit of a yakshave refactoring Message/PresenceMessage.fromValues and related methods, which just didn't make much sense to me.

Previously, fromValues() took an options object with a 'stringifyAction' bool property. If that was false, then it acted as a normal fromValues (used by eg turning a plain js object supplied by a user into a Message for publishing). If true, then it was effectively doing a partial fromDeserialized, parsing a wire protocol message into a message but without decoding the message.data.

I removed the options object and replaced it with an explicit separate Message.fromWireProtocol. (Which I can then add more stuff onto, like copying serial/version etc).

Also while doing that I came across various other things that seemed broken and fixed them:
- unnecessary type-assertions due to not having a wire-protocol type (which I added), plus some unnecessary use of any & unknown
- channeloptions being type-asserted as cipheroptions, then type-assert back to channeloptions
- cryptography support not being passed-in to history calls, so those wouldn't get decrypted even if the channel was encrypted (this is the bit I'm least sure of how it interacts with the tree-shakability work)
- `fromResponseBody` being mostly redundant to other methods, and relying on whether the `format` parameter was set to determine whether its argument was encoded or not -- refactored that away entirely

## Summary by CodeRabbit

- **New Features**
	- Enhanced handling of message and presence data with the introduction of `WireProtocolMessage` and `WireProtocolPresenceMessage` types.
	- Added methods for processing messages from wire protocol formats across various components, including `fromWireProtocol` methods.
	- Introduced a new method `presenceMessageFromWireProtocol` to the `PresenceMessagePlugin` interface.
	- Improved message versioning with the addition of a `version` property in message objects.

- **Bug Fixes**
	- Improved error handling in message processing and test cases to ensure robustness.

- **Documentation**
	- Updated method signatures and descriptions for clarity, particularly around message versioning and presence message handling.

- **Tests**
	- Refined tests to focus on the new `fromWireProtocol` methods, ensuring accurate message serialization and deserialization.

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -2516,9 +2516,8 @@ export interface PresenceMessageStatic {
    * Initialises a `PresenceMessage` from a `PresenceMessage`-like object.
    *
    * @param values - The values to intialise the `PresenceMessage` from.
-   * @param stringifyAction - Whether to convert the `action` field from a number to a string.
    */
-  fromValues(values: PresenceMessage | Record<string, unknown>, stringifyAction?: boolean): PresenceMessage;
+  fromValues(values: Partial<Pick<PresenceMessage, 'clientId' | 'data' | 'extras'>>): PresenceMessage;
 }
 
 /**
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -2332,6 +2332,7 @@ export interface Message {
   name?: string;
   /**
    * Timestamp of when the message was received by Ably, as milliseconds since the Unix epoch.
+   * (This is the timestamp of the current version of the message)
    */
   timestamp?: number;
   /**
@@ -2339,27 +2340,32 @@ export interface Message {
    */
   action?: MessageAction;
   /**
-   * This message's unique serial.
+   * This message's unique serial (an identifier that will be the same in all future
+   * updates of this message).
    */
   serial?: string;
   /**
-   * The serial of the message that this message is a reference to.
+   * If this message references another, the serial of that message.
    */
   refSerial?: string;
   /**
-   * The type of reference this message is, in relation to the message it references.
+   * If this message references another, the type of reference that is.
    */
   refType?: string;
   /**
-   * If an `update` operation was applied to this message, this will be the timestamp the update occurred.
+   * The timestamp of the very first version of a given message (will differ from
+   * createdAt only if the message has been updated or deleted).
    */
-  updatedAt?: number;
+  createdAt?: number;
   /**
-   * The serial of the operation that updated this message.
+   * The version of the message, lexicographically-comparable with other versions (that
+   * share the same serial) Will differ from the serial only if the message has been
+   * updated or deleted.
    */
-  updateSerial?: string;
+  version?: string;
   /**
-   * If this message resulted from an operation, this will contain the operation details.
+   * In the case of an updated or deleted message, this will contain metadata about the
+   * update or delete operation.
    */
   operation?: Operation;
 }
--- a/src/common/lib/client/realtimechannel.ts
+++ b/src/common/lib/client/realtimechannel.ts
@@ -603,7 +603,8 @@ class RealtimeChannel extends EventEmitter {
 
         const messages = message.messages as Array<Message>,
           firstMessage = messages[0],
-          lastMessage = messages[messages.length - 1];
+          lastMessage = messages[messages.length - 1],
+          channelSerial = message.channelSerial;
 
         if (
           firstMessage.extras &&
@@ -650,6 +651,16 @@ class RealtimeChannel extends EventEmitter {
         );
         if (unrecoverableError) {
           return;
+        }
+
+        for (let i = 0; i < messages.length; i++) {
+          const msg = messages[i];
+          if (channelSerial && !msg.version) {
+            msg.version = channelSerial + ':' + i.toString().padStart(3, '0');
+            // already done in fromWireProtocol -- but for realtime messages the source
+            // fields might be copied from the protocolmessage, so need to do it again
+            msg.expandFields();
+          }
         }
 
         this._lastPayload.messageId = lastMessage.id;
@@ -720,7 +731,7 @@ class RealtimeChannel extends EventEmitter {
 
       if (!msg.connectionId) msg.connectionId = connectionId;
       if (!msg.timestamp) msg.timestamp = timestamp;
-      if (!msg.id) msg.id = id + ':' + i;
+      if (id && !msg.id) msg.id = id + ':' + i;
     }
 
     return { unrecoverableError: false };
--- a/src/common/lib/client/restchannelmixin.ts
+++ b/src/common/lib/client/restchannelmixin.ts
@@ -2,7 +2,7 @@ import * as API from '../../../../ably';
 import RestChannel from './restchannel';
 import RealtimeChannel from './realtimechannel';
 import * as Utils from '../util/utils';
-import Message, { fromResponseBody as messageFromResponseBody } from '../types/message';
+import Message, { WireProtocolMessage, _fromEncodedArray } from '../types/message';
 import Defaults from '../util/defaults';
 import PaginatedResource, { PaginatedResult } from './paginatedresource';
 import Resource from './resource';
@@ -30,19 +30,16 @@ export class RestChannelMixin {
 
     Utils.mixin(headers, client.options.headers);
 
-    const options = channel.channelOptions;
     return new PaginatedResource(client, this.basePath(channel) + '/messages', headers, envelope, async function (
       body,
       headers,
       unpacked,
     ) {
-      return await messageFromResponseBody(
-        body as Message[],
-        options,
-        channel.logger,
-        client._MsgPack,
-        unpacked ? undefined : format,
-      );
+      const decoded: WireProtocolMessage[] = unpacked
+        ? (body as WireProtocolMessage[])
+        : Utils.decodeBody(body, client._MsgPack, format);
+
+      return _fromEncodedArray(decoded, channel);
     }).get(params as Record<string, unknown>);
   }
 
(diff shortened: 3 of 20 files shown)


src/common/lib/client/restpresencemixin.ts
```diff
@@ -24,18 +25,17 @@
     Utils.mixin(headers, client.options.headers);
 
     const options = presence.channel.channelOptions;
-    return new PaginatedResource(client, this.basePath(presence) + '/history', headers, envelope, async function (
-      body,
+    return new PaginatedResource(
+      client,
+      this.basePath(presence) + '/history',
```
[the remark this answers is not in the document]
SimonWoolf: It's not a static method. This is a mixin.

src/common/lib/util/utils.ts
```diff
@@ -467,3 +467,6 @@ export async function withTimeoutAsync<A>(promise: Promise<A>, timeout = 5000, e
   const e = new ErrorInfo(err, 50000, 500);
   return Promise.race([promise, new Promise<A>((_resolve, reject) => setTimeout(() => reject(e), timeout))]);
 }
+
+type NonFunctionKeyNames<A> = { [P in keyof A]: A[P] extends Function ? never : P }[keyof A];
+export type Properties<A> = Pick<A, NonFunctionKeyNames<A>>;
```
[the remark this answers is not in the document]
SimonWoolf: pointless change, `(...args: any[]) => any` is no less permissive than `Function`

SimonWoolf:
@coderabbitai pause

src/common/lib/client/realtimechannel.ts
```diff
@@ -622,6 +622,7 @@ class RealtimeChannel extends EventEmitter {
           firstMessage = messages[0],
           lastMessage = messages[messages.length - 1],
           id = message.id,
+          channelSerial = message.channelSerial,
```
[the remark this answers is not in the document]
maratal: Ok, docs [say](https://sdk.ably.com/builds/ably/specification/main/features/#TM2p) "over a realtime transport".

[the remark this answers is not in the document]
SimonWoolf: This is realtimechannel.ts; this code is specific to messages being received over a realtime transport, it is not run in the rest cases. Cf [TM2p](https://sdk.ably.com/builds/ably/specification/main/features/#TM2p) in the spec, which specifies "If a message received from Ably _over a realtime transport_ does not contain a version, ...".

splindsay-92 (approved):
LGTM!

ably.d.ts
```diff
@@ -2516,9 +2516,8 @@ export interface PresenceMessageStatic {
    * Initialises a `PresenceMessage` from a `PresenceMessage`-like object.
    *
    * @param values - The values to intialise the `PresenceMessage` from.
-   * @param stringifyAction - Whether to convert the `action` field from a number to a string.
```
lawrence-forooghian, line 2519: Is removing this not a breaking public API change?

src/common/lib/client/restchannelmixin.ts
```diff
@@ -36,13 +37,11 @@ export class RestChannelMixin {
...
-        options,
-        channel.logger,
-        client._MsgPack,
-        unpacked ? undefined : format,
-      );
+      const decoded: WireProtocolMessage[] = unpacked
+        ? (body as WireProtocolMessage[])
+        : Utils.decodeBody(body, client._MsgPack, format);
+
+      return fromEncodedArray(channel.logger, Platform.Crypto, decoded, options);
```
lawrence-forooghian, line 44: `Platform.Crypto` is not set in the modular variant of the SDK, so I don't think this will work. It sounds like you're saying that you observed an issue with history messages not getting decrypted; it might be worth raising a separate issue and adding some tests to demonstrate the problem?

ably.d.ts
```diff
@@ -2332,34 +2332,40 @@ export interface Message {
   name?: string;
   /**
    * Timestamp of when the message was received by Ably, as milliseconds since the Unix epoch.
+   * (This is the timestamp of the current version of the message)
```
lawrence-forooghian, line 2335: The docstrings for all of these new fields should have been added to https://github.com/ably/sdk-api-reference at the same time as they were added to the spec (that repo should be the source of all docstrings that get added to our SDKs); please could you add them there?

ably.d.ts
SimonWoolf: Technically, yes.

But this is an parameter that is only ever used by the library to decode presence messages sent over a websocket from ably. A user of the library would never have any reason to use it. Indeed a user of the library would never have any reason to know or care that we use numerical encoding in our internal wire protocol. It should never have been exposed in the SDK public API in the first place.

(This method appears to only have been added to support a very niche usecase of sending extras in a presence event, so is not widely used. Possibly not at all outside of the spaces sdk.)

src/common/lib/client/restchannelmixin.ts
SimonWoolf: shouldn't have implied there was an issue in the current ver, sorry -- this was a yakshave from removing fromResponseBody, which I wanted to do because relying on whether the format parameter was set to determine whether its argument was encoded or not struck me as very messy, and anyway the typings weren't at all right for that. I thought I could just substitute fromEncodedArray, which basically does the same thing. I'm not really familar with the modular variant implementation; having looked into it briefly it seems like I could just use client._Crypto, but might be neater to go back to having a separate form of the parser for when you have a channel object already so don't need to renormalize channel options. I'll do that.

SimonWoolf: done

and added a rest encrypted history test

ably.d.ts
VeskeR: @coderabbitai create a task to add mentioned requirement to the [CONTRIBUTING.md](https://github.com/ably/ably-js/blob/main/CONTRIBUTING.md)

VeskeR: please try again, but don't create it in jira, create an issue in this github repo

VeskeR: created https://github.com/ably/ably-js/issues/1929 to clarify about docstrings in [CONTRIBUTING.md](https://github.com/ably/ably-js/blob/main/CONTRIBUTING.md)

SimonWoolf: https://github.com/ably/sdk-api-reference/pull/44

SimonWoolf: I think that should be merged into the spec repo. I don't see why there's a separate repo just for docstrings.

src/common/lib/client/restchannelmixin.ts
lawrence-forooghian: Sorry, I’m still trying to unpick what's going on in c406bef8bc3b27ca5d4f92d540066384c3ae5751; specifically:

- what necessitated the added fetches of `Platform.Crypto`; as I mentioned above, in the modular variant of the library this is unlikely to have whatever effect that it is you're trying to achieve, but in order to tell you what to replace it with I need to understand better why they were added
- the part of the commit message that says "crypto not being passed-in to history calls, so not getting decrypted", which I’m reading as saying that there is some sort of an issue with history decryption. In order to satisfy us that there is _not_ an issue with history decryption, and to allow us to limit the scope of this commit to just being a refactor instead of refactor + maybe a fix to history, I’ve split out your added history test into a separate PR and added a similar test for the modular variant of the library (https://github.com/ably/ably-js/pull/1933)

SimonWoolf: > in order to tell you what to replace it with I need to understand better why they were added

The Platform.Crypto additions divided into three categories:
1. The ones in restpresence/restpresencemixin/restchannel/restchannelmixin, replacing the previous `fromResponseBody`. As described [above](https://github.com/ably/ably-js/pull/1923#discussion_r1861380400) in response to your last comment I removed those and replaced them with an underscore-prefixed form of `fromEncodedArray` for use when you already have a channel object.
2. the addition to defaultpresencemessage.ts. This states that it's used in the non-tree-shakable version of the library -- and also Platform.Crypto was already used in exactly the same way in `defaultmessage.ts`, for which `defaultpresencemessage.ts` is just the presence version. It just looked the tree-shakability work assumed that presence events could not have encrypted data, but of course this is false.
3. The addition to `web/modular/presencemessage.ts`. This was similar to the previous option -- it looks like the lib was assuming presence messages could not be encrypted -- but I made the mistake of copying the same fix I'd used in defaultpresencemessage here, when I should have looked at what was done in `web/module/message.ts`. I've now done that, and copied the approach there: 24f6cea4b361c29d3f448065eb1816182ffe1ca3

I've also updated the commit message to avoid misleading mentions of history, as mentioned history was not the problem that was in main (though I guess not decrypting presence messages is, at least in some versions).

lawrence-forooghian: > Platform.Crypto was already used in exactly the same way in defaultmessage.ts

Yep, it's fine for these `Default*` classes to access `Platform.Crypto`.

> it looks like the lib was assuming presence messages could not be encrypted

Yep, that was an oversight. Might be a good idea to add tests for your new `decodeEncryptedPresenceMessage*` functions? (you can copy the tests of the `decodeEncryptedMessage*` functions in `modular.test.js`, although I don't know if we have test fixtures for encrypted presence messages?)

Could we also please do a couple of things to make d863e91be85384a010b9ad915803d8009369c6d8 understandable for future ourselves, namely:

- remove the passing of `Platform.Crypto` in `src/platform/web/modular/realtimepresence.ts` (I know you undo this in 29d1ed9b7e51f5febf15d95c02520f36dfdcaf6c, but it's a bit confusing when looking at d863e91be85384a010b9ad915803d8009369c6d8 in isolation)
- drop the encrypted history test and the shared helper change that supports it, because it's not really clear how they're related to this commit now? we can get this test from #1933 instead.

SimonWoolf: Done all.

(no particular need for encrypted presence message specific fixtures, looks like the test just works from plain objects with `data` fields and compares the data field of the result, so the messages and presence messages versions are doing exactly the same thing)

lawrence-forooghian (approved):

SimonWoolf:
realtime/failure `invalid_cred_failure` test failure is from sandbox frontdoor change unrelated to this pr

PR #1932, opened 2024-12-05
[DTP-1104] Surface live objects to the end users only when they become valid

VeskeR:
Tried/thought about other implementation options that wouldn't work here:

- buffering `MAP_SET` operations at the main `LiveObjects` class level (not within a specific `LiveMap` instance): does not work because other operations for the object may arrive, updating its site timeserials. As a result, when we attempt to apply `MAP_SET`, it will be rejected due to an earlier site timeserial.

- creating a zero-value object for an object ID that does not yet exist and setting a key in a map to point to that zero-value object while postponing the update event on the map until the referenced object becomes valid. This also doesn't work: the value on the map has already changed (now returning `undefined` due to the invalid object), and it no longer returns the previous value (we effectively have overwritten it). However, we did not invoke the subscription callback for the user, so they are unaware that the value for the key in the map is now `undefined`. This could lead to all sorts of nasty bugs in users' applications. Calling subscription callback two times (first when it becomes `undefined`, second time when object is valid) is also not ideal as it provides poor DevEx

Resolves [DTP-1104](https://ably.atlassian.net/browse/DTP-1104)


mschristensen:
> buffering MAP_SET operations at the main LiveObjects class level (not within a specific LiveMap instance): does not work because other operations for the object may arrive, updating its site timeserials. As a result, when we attempt to apply MAP_SET, it will be rejected due to an earlier site timeserial.

I'm not sure that we should be special-casing buffering to `MAP_SET` ops only.

In a future model where the `CREATE` gives us some information that we need in order to know how to apply operations for an object, we would need to buffer all operations.

I think this also solves the above problem since all ops are buffered and will be applied in the regional order (i.e. the order they are received on the channel).

VeskeR:
Putting this on PR in draft as https://ably.atlassian.net/browse/DTP-1104 is put on hold, see https://ably.atlassian.net/wiki/spaces/LOB/pages/3602874370/LODR-030+Surfacing+of+non-valid+objects+to+end+users+in+the+Realtime+Client

PR #1934, opened 2024-12-06
[DTP-986] Handle `StateObject.tombstone` and `OBJECT_DELETE` messages

VeskeR:
- Tombstoned objects clear their underlying data by setting it to a zero-value: 0 for a counter object, empty map for a map.
- No state operations can be applied on a tombstoned object.
- Tombstoned objects are not surfaced to the end users.
- When deleted, object triggers a subscription callback with cleared data.

Resolves [DTP-986](https://ably.atlassian.net/browse/DTP-986)

[DTP-986]: https://ably.atlassian.net/browse/DTP-986?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Release Notes

- **New Features**
	- Introduced `OBJECT_DELETE` operation to enhance state management.
	- Added tombstone functionality to manage object deletion states.
	- Enhanced `LiveMap`, `LiveCounter`, and `LiveObject` classes to handle tombstoned states effectively.
	- Updated return types in `LiveMap` methods to include `undefined` for tombstoned entries.

- **Bug Fixes**
	- Improved error handling to prevent operations on tombstoned objects.

- **Documentation**
	- Updated method comments to clarify behavior regarding tombstoned objects.

- **Tests**
	- Added new test scenarios for handling object deletions and tombstoned states.

### DIFF
--- a/src/plugins/liveobjects/livemap.ts
+++ b/src/plugins/liveobjects/livemap.ts
@@ -78,6 +78,7 @@ export class LiveMap<T extends API.LiveMapType> extends LiveObject<LiveMapData, 
   /**
    * Returns the value associated with the specified key in the underlying Map object.
    *
+   * - If this map object is tombstoned (deleted), `undefined` is returned.
    * - If no entry is associated with the specified key, `undefined` is returned.
    * - If map entry is tombstoned (deleted), `undefined` is returned.
    * - If the value associated with the provided key is an objectId string of another Live Object, a reference to that Live Object
@@ -85,7 +86,11 @@ export class LiveMap<T extends API.LiveMapType> extends LiveObject<LiveMapData, 
    * - If the value is not an objectId, then that value is returned.
    */
   // force the key to be of type string as we only allow strings as key in a map
-  get<TKey extends keyof T & string>(key: TKey): T[TKey] extends StateValue ? T[TKey] : T[TKey] | undefined {
+  get<TKey extends keyof T & string>(key: TKey): T[TKey] | undefined {
+    if (this.isTombstoned()) {
+      return undefined as T[TKey];
+    }
+
     const element = this._dataRef.data.get(key);
 
     if (element === undefined) {
--- a/src/plugins/liveobjects/livemap.ts
+++ b/src/plugins/liveobjects/livemap.ts
@@ -77,13 +77,15 @@ export class LiveMap<T extends API.LiveMapType> extends LiveObject<LiveMapData, 
 
   /**
    * Returns the value associated with the specified key in the underlying Map object.
-   * If no element is associated with the specified key, undefined is returned.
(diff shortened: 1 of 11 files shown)


src/plugins/liveobjects/livemap.ts
```diff
@@ -77,13 +77,16 @@ export class LiveMap<T extends API.LiveMapType> extends LiveObject<LiveMapData,
...
   /**
    * Returns the value associated with the specified key in the underlying Map object.
-   * If no element is associated with the specified key, undefined is returned.
-   * If the value that is associated to the provided key is an objectId string of another Live Object,
-   * then you will get a reference to that Live Object if it exists in the local pool, or undefined otherwise.
-   * If the value is not an objectId, then you will get that value.
+   *
+   * - If no entry is associated with the specified key, `undefined` is returned.
+   * - If map entry is tombstoned (deleted), `undefined` is returned.
+   * - If this map object is tombstoned, `undefined` is returned.
```
mschristensen, line 83: It doesn't look like this method handled this case?

VeskeR:
```typescript
    if (refObject.isTombstoned()) {
      // tombstoned objects must not be surfaced to the end users
      return undefined as T[TKey];
    }
```
from the [lines below](https://github.com/ably/ably-js/pull/1934/files#diff-2c2d2b2e45367c0ba8c03880c978c1dc73ac8065b20b0887e36e6a56103740e7R114-R117).
probably just missed it among other changes?

VeskeR: oh, sorry

VeskeR: if "this map object"

VeskeR: when we tombstone an object, we remove all its data. so a tombstoned map doesn't have anything in its `_dataRef`, this way any element won't be present in the map.
so `If this map object is tombstoned, `undefined` is returned.` is an implicit consequence of tombstoning a map

VeskeR: I can add a comment to the method body to clarify this behavior

mschristensen: Yes please, I think that's worth calling out. Thanks

mschristensen (approved):

src/plugins/liveobjects/livemap.ts
VeskeR: updated instead to explicitly check for `.isTombstoned()` in a method https://github.com/ably/ably-js/compare/4ff364711ca85e8b5338588fa5df1f1f87e10b29..a7df3b6f8f3a428863bcd235f3fa5ce75b8542c2

PR #1939, opened 2025-01-07
test/browser/connection: fix hmac extraction

Zariel:
Relax the connectionKey hmac extraction and account for the possibilities of multiple - characters.

RAR-655

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved connection key processing logic in the connection handling mechanism.

### DIFF
--- a/test/browser/connection.test.js
+++ b/test/browser/connection.test.js
@@ -216,15 +216,12 @@ define(['shared_helper', 'chai'], function (Helper, chai) {
       // strip instanceID and handleID from connectionKey */
       function connectionHmac(key) {
         /* connectionKey has the form <instanceID>!<hmac>-<handleID> */
-
-        /* remove the handleID from the end of key */
-        let k = key.split('-')[0];
-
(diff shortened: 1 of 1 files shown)


Zariel:
supersedes #1903

SimonWoolf (changes requested):
why go to a lot of effort to parse out the connectionId from the connectionKey rather than just comparing connectionIds? trying to do the former is inherently flaky: the connectionkey format is a secret of the server and the sdk should not depend on it, either in code or tests. the library exposes the connectionId in `connection.id`, just use that

VeskeR (changes requested):
Agree with Simon on this, ideally we should not depend on some internal realtime knowledge that can change any time when there is an alternative.

It looks like @lawrence-forooghian  just opened a PR https://github.com/ably/ably-js/pull/1957 that fixes this issue the way Simon suggested - removes parsing of the connection key and compare connection id. I suggest we go with that PR and close this one.

lawrence-forooghian:
D'oh, completely missed the existence of this PR, sorry! Will merge #1957.

PR #1940, opened 2025-01-07
test/realtime/message: reduce subscribes to filtered channel flakes

Zariel:
The test waits to receive an end message on a channel then it runs the assertions, this can cause flakes because we dont know if the event loop had any other pending IO which could include delivering the 2 expected filtered messages. Run the assertions in a setImmediate so that it executes in the next event loop run allowing any queued messages to be delivered.

When running the test with logging the messages were being delivered but the handler was being called after the assertions had ran.

RAR-656

## Summary by CodeRabbit

- **Tests**
	- Improved reliability of message processing test by introducing an asynchronous approach to handle message assertions.
	- Resolved potential race conditions in test case for message filtering.


Zariel:
@VeskeR ping

VeskeR (approved):
Thank you!

PR #1941, opened 2025-01-08
Restructure and clean up message and presencemessage types 

SimonWoolf:
Motivation: I need to add a new AnnotationMessage message type, and the existing message classes were just... not nice to work with. I have:

- extract common functionality to a new BaseMessage file
- introduce proper `WireMessage` and `WirePresenceMessage` classes, such that:
  - you just call `Message.encode()` to get a `WireMessage`, and `WireMessage.decode()` to get a `Message`, and the type system will tell you which one you have
  - the `toJSON` override hack only applies to internal wire-protocol usesand doesn't interfere with the normal public Message/PresenceMessage instances
  - all necessary transformations between userfriendly message and wire protocol messages (encoding the data and the action)
    are now done by decode() and encode(), rather than (as before) the encoding happening in-place and the action encoding happening only on json-stringification
  - ProtocolMessage is now always a wire-protocol message and always holds WireMessages and WirePresence; all decoding is done by the channel
  - The PresencePlugin interface is now just the PresenceMessage and WirePresenceMessage types, rather than ad-hoc individual functions
  - The fromWireProtocol, encode, and decode functions in the DefaultMessage interface are no longer needed and can be removed
- Cleaned up and simplified the message decoding step in the channel, which was unnecessarily complicated

Also removed an unused field from RealtimeChannel, removed ATTACH_RESUME from the list of channel modes (it's not a channel mode, it should never have been in there), and made the agreed changes to the MessageAction enum.

This ended up being quite a bit of a yakshave. It isn't quite all the changes I saw and wanted to make, but I think this is a good place to stop.

## Release Notes

- **New Features**
	- Added support for message summaries with a new `MESSAGE_SUMMARY` action type.
	- Introduced new types and functions for improved message encryption and encoding.

- **Improvements**
	- Streamlined message type definitions across the Ably Realtime library.
	- Refined presence message handling with more consistent wire protocol representations.
	- Enhanced clarity and consistency in encryption test cases.

- **Breaking Changes**
	- Removed `ATTACH_RESUME` channel mode.
	- Deprecated certain message action types like `MESSAGE_UNSET`, `ANNOTATION_CREATE`, and `ANNOTATION_DELETE`.
	- Modified message and presence message creation methods to utilize new class methods.

- **Bug Fixes**
	- Improved type safety in message encoding and decoding.
	- Enhanced consistency in message action handling.

- **Refactoring**
	- Reorganized import statements and module structures.
	- Simplified message type definitions.
	- Moved common protocol message constants to a separate module.

### DIFF
--- a/src/common/lib/types/presencemessage.ts
+++ b/src/common/lib/types/presencemessage.ts
@@ -56,6 +56,11 @@ export async function _fromEncodedArray(
   );
 }
 
+// for tree-shakability
+export function fromValues(values: Properties<PresenceMessage>) {
+  return PresenceMessage.fromValues(values);
+}
+
 class PresenceMessage extends BaseMessage {
   action?: string;
 
--- a/src/platform/web/modular/presencemessage.ts
+++ b/src/platform/web/modular/presencemessage.ts
@@ -1,5 +1,5 @@
 import * as API from '../../../../ably';
-import PresenceMessage, { fromEncoded, fromEncodedArray } from '../../../common/lib/types/presencemessage';
+import { fromValues, fromEncoded, fromEncodedArray } from '../../../common/lib/types/presencemessage';
 import { Crypto } from './crypto';
 import Logger from '../../../common/lib/util/logger';
 
@@ -21,4 +21,4 @@ export const decodeEncryptedPresenceMessages = ((obj, options) => {
   return fromEncodedArray(Logger.defaultLogger, Crypto, obj, options);
 }) as API.PresenceMessageStatic['fromEncodedArray'];
 
-export const constructPresenceMessage = PresenceMessage.fromValues as API.PresenceMessageStatic['fromValues'];
+export const constructPresenceMessage = fromValues as API.PresenceMessageStatic['fromValues'];


SimonWoolf:
@coderabbitai pause

SimonWoolf:
@lawrence-forooghian could you help me understand why `Adding constructPresenceMessage to BaseRealtime does not increase the bundle size` is an error?

lawrence-forooghian:
> @lawrence-forooghian could you help me understand why `Adding constructPresenceMessage to BaseRealtime does not increase the bundle size` is an error?

Will take a look 👍

SimonWoolf:
aah ok I get it -- so the problem is that it can't prove that importing the whole module doesn't have side effects so it can't remove it, so I can keep the nice api by doing https://github.com/ably/ably-js/pull/1941/commits/2b867ad13b45b47283c33742bd549c1e32583667 and then it works again. cool, thanks

src/common/lib/client/realtimepresence.ts
```diff
@@ -435,7 +431,7 @@ class RealtimePresence extends EventEmitter {
   _synthesizeLeaves(items: PresenceMessage[]): void {
     const subscriptions = this.subscriptions;
     items.forEach(function (item) {
-      const presence = presenceMessageFromValues({
+      const presence = PresenceMessage.fromValues({
```
owenpearson, line 434: I think we switched to standalone functions rather than static methods because that makes them tree-shakable separately from the class, so for example if someone is using the `PresenceMessage` class but isn't using `presenceMessageFromValues` then the standalone function isn't imported. I guess this would only happen if someone is using rest presence, since we use `fromValues` internally here so anyone using realtime presence will be using it either way.

owenpearson: fwiw i think the minified size of that function is probably so small that it doesn't really matter anyway but just fyi

owenpearson (approved):
looks way better :) one minor comment about tree shaking but the rest lgtm

src/common/lib/types/presencemessage.ts
```diff
@@ -132,65 +87,65 @@ class PresenceMessage {
...
+    });
+    return encode(res, options);
+  }
+
+  static fromValues(values: Properties<PresenceMessage>): PresenceMessage {
+    return Object.assign(new PresenceMessage(), values);
+  }
+
+  static fromValuesArray(values: Properties<PresenceMessage>[]): PresenceMessage[] {
+    return values.map(PresenceMessage.fromValues);
```
ttypic, line 102: I believe that in JS, it’s better to use a lambda here. The map callback has three arguments, and although it’s unlikely that we’ll add a new optional argument to `fromValues`, it could happen, and this spot would be easy to overlook.
```suggestion
    return values.map(v => PresenceMessage.fromValues(v));
```

src/common/lib/client/realtimepresence.ts
SimonWoolf: Thanks, but.. yeah, I'm very disinclined to spend any brainpower worrying about that one-line function now being tree-shaken in that specific circumstance

fwiw it looks like the minimal bundle size after this pr is now down about 3/4 of a kilobyte, mostly from improved BaseMessage codesharing and removing some unnecessary decoding abstractions in realtimechannel. I'd argue that worrying overly much about making sure these sort of scraps are tree-shakable while ignoring the low-hanging fruit (neglected unnecessarily long code in non-tree-shakable parts of the lib) is misdirected energy

src/common/lib/types/presencemessage.ts
SimonWoolf: sure

SimonWoolf:
some browser test failures in `page_refresh_` but this appears to be a known flake due to multiple frontdoors in sandbox, don't reprod locally with one, so ignoring

PR #1942, opened 2025-01-13
CHORE: Add release information for v.3.0.0

splindsay-92:


CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 This contains only the most important and/or user-facing changes; for a full changelog, see the commit history.
 
+## [3.0.0](https://github.com/ably/ably-js/tree/3.0.0) (2025-01-13)
```
AndyTWF, line 5: If we're going for 3.0.0 just because of the changes to the `MESSAGE_ACTIONS` enum, I think we can probably just do this as a minor change. It's only used internally by Chat right now and is not documented for use elsewhere.

splindsay-92: @SimonWoolf Mentioned breaking changes in his commits, hence the 3.0.0 label. Looks like a change to the channel modes enum too.

SimonWoolf: No no, this doesn't need to be a new major ably-js version. Nothing is using MESSAGE_ACTIONS except the just-released chat alpha

SimonWoolf: and in any case to the extent it is a breaking change it's a breaking change in the wire protocol, not the public API. The public API for actions is the string, which isn't changing

splindsay-92: Correcting this and making a mental note to ignore coderabbits PR changelog

PR #1950, opened 2025-01-16
[DTP-1138] Add object-level write API for LiveMap/LiveCounter creation

VeskeR:
Resolves [DTP-1138](https://ably.atlassian.net/browse/DTP-1138)


src/plugins/liveobjects/liveobjects.ts
```diff
@@ -54,6 +54,46 @@ export class LiveObjects {
...
+   * @returns A promise which resolves upon receiving the ACK message for the published operation message. A promise is resolved with a zero-value object created in the local pool.
+   */
+  async createMap<T extends API.LiveMapType>(entries?: T): Promise<LiveMap<T>> {
+    const stateMessage = await LiveMap.createMapCreateMessage(this, entries);
+    const objectId = stateMessage.operation?.objectId!;
+
+    await this.publish([stateMessage]);
+
+    // return zero-value object so that users can interact and subscribe to it immediately
+    const object = this._liveObjectsPool.createZeroValueObjectIfNotExists(objectId);
```
mschristensen, line 73: should this store the initial value too? if the user accesses this object's value before it is actaully created via the echoed message, I think they would expect it to have the value they specified

src/plugins/liveobjects/statemessage.ts
```diff
@@ -232,6 +241,19 @@ export class StateMessage {
...
 
+  static encodeInitialValue(
+    utils: typeof Utils,
+    initialValue: Partial<StateOperation>,
+  ): {
+    encodedInitialValue: string;
+    format: Utils.Format;
+  } {
+    return {
+      encodedInitialValue: JSON.stringify(initialValue),
```
mschristensen, line 252: should we use the default encoding specified on the client?

VeskeR: yep, that's the plan. I think it's a `client.options.useBinaryProtocol` option to decide which encoding to use (`msgpack` if true).
For now it's hard coded to be always JSON as we're fixing the issue with `msgpack` encoding for the initial value on realtime.
Will update this part to depend on the client option once it's fixed and we know how `initialValue` field should look like when encoding is set to `msgpack`

src/plugins/liveobjects/liveobjects.ts
VeskeR: Hmm, it gets a bit complicated due to CRDT rules and site time serials.

First: site timeserials. We don't have a timeserial for a `CREATE` operation at the moment an object is created, so if we were to apply an initial value to it, we wouldn't be able to assign a site timeserial for that `CREATE` operation. So far, it just means that whenever we actually receive a `CREATE` op by the operation application rules, we will try to apply it to the object (as there is no existing site timeserial on the object that would prevent it).  

Ok, now second: so say we applied the initial value immediately, and now we received a `CREATE` op and there is not site timeserial to prevent application based on site timeserials. Objects have a flag `_createOperationIsMerged` to identify if the create operation was merged previously, to prevent double application (important for counters, as merging a `CREATE` op of a counter just adds the amount to the counter). So to guarantee that initial values for a counter are not double counted, we need to have the `_createOperationIsMerged` flag set to `true` when we assign an initial value on object creation during `liveObjects.create*` call. This means that the `CREATE` op won't be applied to objects when it is eventually received. 

Third: map objects have additional CRDT rules for their map entries where each map entry stores a separate timeserial. When a `CREATE` op for a map is received, the map entries in it will have a timeserial set to the timeserial of the `CREATE` op. But if we applied the initial value on object creation client-side and prevented additional `CREATE` op application via the `_createOperationIsMerged` flag, we won't set timeserials for map entries when the `CREATE` op is received. That being said, maybe missing timeserials on map entries after the creation is not that big of a deal, as since this object was just created, there can't exist an operation on that object that is CGO earlier than the time serial of the CREATE operation, as we only obtain an object ID after the CREATE op has happened. However, it would lead to this weird state for a map object, where its entries are missing timeserials, and it seems like something that can cause unexpected bugs in the future.

This can be solved if we introduce different behavior in handling a `_createOperationIsMerged` flag between live object types. So both objects, map and counter, are set with their initial value on creation client-side, but the counter won't allow the CREATE op to be applied (has `_createOperationIsMerged` set to `true`), and the map will allow it to be applied (will have `_createOperationIsMerged` set to `false`).

Alternatively, if we implement our idea with operation application on ACKs, that ACK message will have a timeserial for the `CREATE` operation. Since create operations are currently async and wait for an ACK before resolving a promise and returning an object to a user, it would be trivial to apply an actual initial value on the object with the corresponding timeserial upon receipt of that ACK message.

VeskeR: > That being said, maybe missing timeserials on map entries after the creation is not that big of a deal, as since this object was just created, there can't exist an operation on that object that is CGO earlier than the time serial of the CREATE operation, as we only obtain an object ID after the CREATE op has happened.

I can only think of a situation when there there is a substantial clock skew between realtime regions, and the `MAP_CREATE` operation for a new object in region A ends up having a timestamp `t1` which is greater than a following `MAP_SET` op `t2` for that object id in region B. So that `t1` > `t2` and by LWW rules `MAP_SET` should not be applied, but if we don't have timeserials on map entries client-side due to not applying a `CREATE` op we will wrongly apply that `MAP_SET` operation. Is this even technically possible in the context of cross-region operations?

VeskeR: added setting initial value on object creation in https://github.com/ably/ably-js/compare/8dc5396a07d0b3357a0a961f99e4fa89e7555ea7..c89a9dc90a8fa737e5dd9b7518f13cd2fbb96e86

VeskeR: fixed the race condition with the order of operations for this (discussed on the standup) in https://github.com/ably/ably-js/compare/c89a9dc90a8fa737e5dd9b7518f13cd2fbb96e86..d66a95f0b37a577a4800d7bda7545f135e73f142

mschristensen:
Changes look good, just waiting now on the msgpack changes in realtime to unblock this

src/plugins/liveobjects/statemessage.ts
VeskeR: Updated to support `json` and `msgpack` encoding for the `initialValue`.
The encoding is selected based on the client option `client.options.useBinaryProtocol` - basically it matches the encoding for the channel messages overall (if channel messages are msgpack encoded, then initial value will also be msgpack encoded).

See next commits:
https://github.com/ably/ably-js/pull/1950/commits/ae655f18a4669ff44e8e79e6028059d99fd63c02
https://github.com/ably/ably-js/pull/1950/commits/109fefb4e7eefe4bbc084703afc1f62030fc66c7
https://github.com/ably/ably-js/pull/1950/commits/6ea31dacf4fea78ffd0f175352e14e1a3ffff33f

src/plugins/liveobjects/liveobjects.ts
```diff
@@ -54,6 +54,66 @@ export class LiveObjects {
...
+
+    // we might have received CREATE operation already at this point (it might arrive before the ACK message for our publish message),
+    // so the object could exist in the local pool as it was added there during regular CREATE operation application.
+    // check if the object is there and return it in case it is. otherwise create a new object client-side
+    if (this._liveObjectsPool.get(objectId)) {
+      return this._liveObjectsPool.get(objectId) as LiveMap<T>;
+    }
+
+    // new map object can be created using locally constructed state operation, even though it is missing timeserials for map entries.
+    // CREATE operation is only applied once, and all map entries will have an "earliest possible" timeserial so that any subsequent operation can be applied to them.
```
mschristensen, line 80:
```suggestion
    // CREATE operation is idempotent, and all map entries will have an "earliest possible" timeserial so that any subsequent operation can be applied to them.
```

src/plugins/liveobjects/liveobjects.ts
```diff
@@ -54,6 +54,66 @@ export class LiveObjects {
...
+
+    // we might have received CREATE operation already at this point (it might arrive before the ACK message for our publish message),
+    // so the object could exist in the local pool as it was added there during regular CREATE operation application.
+    // check if the object is there and return it in case it is. otherwise create a new object client-side
+    if (this._liveObjectsPool.get(objectId)) {
+      return this._liveObjectsPool.get(objectId) as LiveCounter;
+    }
+
+    // new counter object can be created using locally constructed state operation.
+    // CREATE operation is only applied once, so the initial counter value won't be double counted when we eventually receive an echoed CREATE operation
```
mschristensen, line 110:
```suggestion
    // CREATE operation is idempotent, so the initial counter value won't be double counted when we eventually receive an echoed CREATE operation
```

mschristensen (approved):
Nice, LGTM

src/plugins/liveobjects/liveobjects.ts
VeskeR: `CREATE` ops are idempotent by design, but client can incorrectly handle setting the initial value here and essentially apply the `CREATE` op second time, hence I wasn't sure about the wording. will change it to "idempotent", but also repharse that client need to ensure it won't apply it again due to setting the initial value client-side

VeskeR: improved comments in https://github.com/ably/ably-js/pull/1950/commits/5cfbc9b130eb35e3ef7fc77008afc8bd6243edbd

src/plugins/liveobjects/liveobjects.ts
VeskeR: improved comments in https://github.com/ably/ably-js/pull/1950/commits/5cfbc9b130eb35e3ef7fc77008afc8bd6243edbd

Issue #1954, opened 2025-01-21
Return to a single enum for channel modes


lawrence-forooghian:
Split from #1952.

TODO description

PR #1955, opened 2025-01-21
Fix typings for `RealtimeChannel.modes`

### LINKED ISSUE #1952: case inconsistency between typings and runtime behaviour for ChannelModes
The typings for a ChannelMode are uppercase, eg `SUBSCRIBE`:
https://github.com/ably/ably-js/blob/af41744ca41f55f3ba526f980e7d9a44c508315b/ably.d.ts#L852-L869

But the runtime values in RealtimeChannel.modes are lowercase:

Image

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-5195) by [Unito](https://www.unito.io)

lawrence-forooghian:
For some reason, this property returns lowercased values. So, reflect this in the typings. This is the only non-breaking change we can make, since:

- if we change the return value of `RealtimeChannel.modes` to be uppercase, that will break things for users who are checking this return value against a list of lowercase values (having previously observed that this property returns lowercase values even though it contradicts the typings)

- if we change the `ChannelModes` type to be lowercase, then we break things for TypeScript users who are setting `modes` in their channel options

I have also changed the `ChannelModes` type to indicate that it accepts lowercased values; this allows a user to pass a `ResolvedChannelModes` when setting `modes` in their channel options.

In our next major release, we should remove this duplicate type and only allow a single case (probably lowercase, since that's what we use for all of the other enum values in this SDK). Have created #1954 for this.

Resolves #1952.

## Summary by CodeRabbit

- **New Features**
  - Enhanced channel mode definitions to support both uppercase and lowercase representations.
  - Introduced a new `ResolvedChannelMode` type for more flexible channel mode handling.

- **Improvements**
  - Updated type signatures to allow more versatile channel mode configurations.
  - Added lowercase variants of existing channel mode types.

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -853,29 +853,66 @@ declare namespace ChannelModes {
   /**
    * The client can publish messages.
    */
-  type PUBLISH = 'PUBLISH';
+  type PUBLISH = 'PUBLISH' | 'publish';
   /**
    * The client can subscribe to messages.
    */
-  type SUBSCRIBE = 'SUBSCRIBE';
+  type SUBSCRIBE = 'SUBSCRIBE' | 'subscribe';
   /**
    * The client can enter the presence set.
    */
-  type PRESENCE = 'PRESENCE';
+  type PRESENCE = 'PRESENCE' | 'presence';
   /**
    * The client can receive presence messages.
    */
-  type PRESENCE_SUBSCRIBE = 'PRESENCE_SUBSCRIBE';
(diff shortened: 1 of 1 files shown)


ably.d.ts
```diff
@@ -853,19 +853,19 @@ declare namespace ChannelModes {
   /**
```
SimonWoolf, line 853: Maybe worth leave a link to this pr or https://github.com/ably/ably-js/issues/1954 in the docstring comment for ChannelModes to explain the redundant values

SimonWoolf (approved):
:+1: with one comment

VeskeR (approved):

PR #1956, opened 2025-01-22
[DTP-948] Surface errors when user has not attached to a channel with state modes

VeskeR:
- `state_subscribe` mode is required for access/subscribe API
- `state_publish` mode is required for create/edit API

I think the existing subscribe mode checks are still needed in case the modes are updated via setOptions

Channel modes can change post channel attachment (if user calls `channel.setOptions`) at which point user may have already obtained a `root` and other live object instances. This means that every public live objects API method needs to check for the required modes set on a channel separately.

Resolves [DTP-948](https://ably.atlassian.net/browse/DTP-948)

[DTP-948]: https://ably.atlassian.net/browse/DTP-948?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **New Features**
	- Added state mode validation checks to LiveObjects, LiveMap, LiveCounter, and related classes.
	- Enhanced error handling for operations in different channel modes.

- **Bug Fixes**
	- Improved robustness of state management in live object operations.
	- Added precondition checks to prevent unauthorized state transitions.

- **Tests**
	- Updated test suite to cover new state mode validation scenarios.
	- Introduced new test functions for checking missing state modes.

- **Chores**
	- Added private API identifier for channel mode tracking.

### DIFF
--- a/src/plugins/liveobjects/batchcontext.ts
+++ b/src/plugins/liveobjects/batchcontext.ts
@@ -24,6 +24,7 @@ export class BatchContext {
   }
 
   getRoot<T extends API.LiveMapType = API.DefaultRoot>(): BatchContextLiveMap<T> {
+    this._liveObjects.throwIfMissingStateSubscribeMode();
     this.throwIfClosed();
     return this.getWrappedObject(ROOT_OBJECT_ID) as BatchContextLiveMap<T>;
   }
--- a/src/plugins/liveobjects/batchcontextlivecounter.ts
+++ b/src/plugins/liveobjects/batchcontextlivecounter.ts
@@ -15,17 +15,20 @@ export class BatchContextLiveCounter {
   }
 
   value(): number {
+    this._liveObjects.throwIfMissingStateSubscribeMode();
     this._batchContext.throwIfClosed();
     return this._counter.value();
   }
 
   increment(amount: number): void {
+    this._liveObjects.throwIfMissingStatePublishMode();
     this._batchContext.throwIfClosed();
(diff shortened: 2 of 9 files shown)


mschristensen (commented):
Looks good. One suggestion - since `state_subscribe` is the minimum mode needed to interact with live objects realtime client, should we throw if `state_subscribe` is not specified when accessing the `liveObjects` plugin on the channel? (I think the existing subscribe mode checks are still needed in case the modes are updated via `setOptions`)

VeskeR:
>  since `state_subscribe` is the minimum mode needed to interact with live objects realtime client, should we throw if state_subscribe is not specified when accessing the liveObjects plugin on the channel?

It's not really. We provide `createMap`/`createCounter` methods on liveObjects instance to create new objects, which do not technically require you to subscribe to a state (e.g. retrieve a root first). Now, I agree that currently you wouldn't be able to do much with those objects unless you retrieve a root object (thus need `state_subscribe`) to assign those objects to the root, but I can imagine in a future we might add something akin to the transient publishes we have for pub/sub sdk - some kind of a client-side operation that creates an object on a root (since root has a known object id), so there will be a useful functionality available for clients with only `state_publish` mode enabled.

> I think the existing subscribe mode checks are still needed in case the modes are updated via setOptions

that was my exact train of thought, channel modes can change during interaction with liveobjects if user calls `channel.setOptions`, so every public method needs to check it separately 100%.

mschristensen (approved):

PR #1957, opened 2025-01-22
Fix failing connection recovery tests

lawrence-forooghian:
It seems like the Realtime connection key format, which as the removed comment implies is not guaranteed to be stable, has changed on sandbox. Stop using our knowledge of it.

## Summary by CodeRabbit

- **Refactor**
	- Updated connection key handling in test cases
	- Simplified connection identification logic
	- Replaced connection key references with connection ID


VeskeR (approved):
Thanks!

PR #1958, opened 2025-01-23
[DTP-1034] Emit Live Objects lifecycle events

VeskeR:
Functionally complete but not tests yet as some of them depend on [DTP-1147](https://ably.atlassian.net/browse/DTP-1147). Worth revieweing this PR API side while I complete tests setup.

Resolves [DTP-1034](https://ably.atlassian.net/browse/DTP-1034)

[DTP-1034]: https://ably.atlassian.net/browse/DTP-1034?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DTP-1147]: https://ably.atlassian.net/browse/DTP-1147?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **New Features**
	- Enhanced event handling for LiveObjects and LiveObject classes.
	- Added lifecycle event management with new subscription methods.
	- Introduced more granular event control with `on`, `off`, and `offAll` methods.
	- New asynchronous utility functions for improved testing capabilities.

- **Improvements**
	- Updated event naming conventions to be more consistent.
	- Separated internal and public event emitters.
	- Added new events like `sync_start` and `sync_end`.
	- Unique object ID generation now includes timestamps for better differentiation.

### DIFF
--- a/src/plugins/liveobjects/liveobjects.ts
+++ b/src/plugins/liveobjects/liveobjects.ts
@@ -11,21 +11,34 @@ import { LiveObjectsPool, ROOT_OBJECT_ID } from './liveobjectspool';
 import { StateMessage, StateOperationAction } from './statemessage';
 import { SyncLiveObjectsDataPool } from './syncliveobjectsdatapool';
 
-enum LiveObjectsEvents {
-  SyncCompleted = 'SyncCompleted',
+export enum LiveObjectsEvent {
+  syncing = 'syncing',
+  synced = 'synced',
 }
+
+export enum LiveObjectsState {
+  initialized = 'initialized',
+  syncing = 'syncing',
+  synced = 'synced',
+}
+
+const StateToEventsMap: Record<LiveObjectsState, LiveObjectsEvent | undefined> = {
+  initialized: undefined,
+  syncing: LiveObjectsEvent.syncing,
+  synced: LiveObjectsEvent.synced,
+};
 
 type BatchCallback = (batchContext: BatchContext) => void;
 
 export class LiveObjects {
   private _client: BaseClient;
   private _channel: RealtimeChannel;
+  private _state: LiveObjectsState;
   // composition over inheritance since we cannot import class directly into plugin code.
   // instead we obtain a class type from the client
   private _eventEmitter: EventEmitter;
   private _liveObjectsPool: LiveObjectsPool;
   private _syncLiveObjectsDataPool: SyncLiveObjectsDataPool;
-  private _syncInProgress: boolean;
   private _currentSyncId: string | undefined;
   private _currentSyncCursor: string | undefined;
   private _bufferedStateOperations: StateMessage[];
@@ -36,10 +49,10 @@ export class LiveObjects {
   constructor(channel: RealtimeChannel) {
     this._channel = channel;
     this._client = channel.client;
+    this._state = LiveObjectsState.initialized;
     this._eventEmitter = new this._client.EventEmitter(this._client.logger);
     this._liveObjectsPool = new LiveObjectsPool(this);
     this._syncLiveObjectsDataPool = new SyncLiveObjectsDataPool(this);
-    this._syncInProgress = true;
     this._bufferedStateOperations = [];
   }
 
@@ -51,9 +64,9 @@ export class LiveObjects {
   async getRoot<T extends API.LiveMapType = API.DefaultRoot>(): Promise<LiveMap<T>> {
     this.throwIfMissingStateSubscribeMode();
 
-    // SYNC is currently in progress, wait for SYNC sequence to finish
-    if (this._syncInProgress) {
-      await this._eventEmitter.once(LiveObjectsEvents.SyncCompleted);
+    // if we're not synced yet, wait for SYNC sequence to finish before returning root
+    if (this._state !== LiveObjectsState.synced) {
+      await this._eventEmitter.once(LiveObjectsEvent.synced);
     }
 
     return this._liveObjectsPool.get(ROOT_OBJECT_ID) as LiveMap<T>;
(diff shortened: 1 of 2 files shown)


src/plugins/liveobjects/liveobjects.ts
```diff
@@ -11,18 +11,27 @@ import { LiveObjectsPool, ROOT_OBJECT_ID } from './liveobjectspool';
 import { StateMessage, StateOperationAction } from './statemessage';
 import { SyncLiveObjectsDataPool } from './syncliveobjectsdatapool';
 
-enum LiveObjectsEvents {
-  SyncCompleted = 'SyncCompleted',
+export enum LiveObjectsEvent {
+  sync_start = 'sync_start',
+  sync_end = 'sync_end',
```
mschristensen, line 16: I think these events should reflect the current "state" (in the state-machine sense) of the liveobjects instance, similar to how we do for [channels](https://ably.com/docs/channels/states#states) and [connections](https://ably.com/docs/connect/states?lang=javascript)

So I would suggest `initialized`, `syncing`, and `synced`, analogous to `initialized`, `connecting` and `connected` for connection state or `initialized`, `attaching` and `attached` for channels.

mschristensen (changes requested):
Looks great, left a suggestion to discuss

src/plugins/liveobjects/liveobjects.ts
VeskeR: sounds good, I think for `initialized` I will need to add another flag to the LiveObjects class to distinguish between sync is in progress (we received an ATTACHED with HAS_STATE flag) and when LiveObjects instance was just created. Currently there is only `_syncInProgress` flag that only covers `syncing` and `synced`.

Also, would you expect to **always** get a `syncing` event before getting a `synced` event? This is important because with the current implementation, we can be in a `synced` state and then reattach to a channel that doesn't have HAS_STATE flag - this essentially removes the state immediately and emits the `synced` event again, without emitting `syncing` event before that.
see here:
https://github.com/ably/ably-js/blob/210af2d7bb793f8ef7771c100ede3d755022e280/src/plugins/liveobjects/liveobjects.ts#L124-L130

mschristensen: What do we do for channels/connections? Do we enforce that you always get e.g. `connecting` before `connected`?

VeskeR: Small note from the standup: `HAS_STATE` flag on channel attachment is always `true` if `state_subscribe` mode is set, so this guarantees we can have `syncing` -> `synced` order, given `HAS_STATE` flag is set (`state_subscribe` mode is set for a channel)

Re order of `connecting`/`connected` events for channel/connection:

I couldn't find an explicit mention of `connecting` -> `connected` order, but I think it is guaranteed implicitly based on the next:

The `RealtimeChannel` will (potentially) emit a separate `update` event when it gets an `ATTACHED` message while being in an `attached` state: https://github.com/ably/ably-js/blob/471cdef3a52bc95282a570ed910b2ddccfd01436/src/common/lib/client/realtimechannel.ts#L524-L526
And more generally, if there is no special logic for an event, then `notifyEvent` method won't emit an event if current state of the channel matches the one it's attempting to notify: https://github.com/ably/ably-js/blob/471cdef3a52bc95282a570ed910b2ddccfd01436/src/common/lib/client/realtimechannel.ts#L718-L720
Spec point matches the behavior described above: https://sdk.ably.com/builds/ably/specification/main/features/#RTL12

The `Connection` (more precisely `ConnectionManager`) has pretty much the same behavior, where there is sometimes special logic to emit `update` event when connection configuration changes without an explicit reconnection: https://github.com/ably/ably-js/blob/471cdef3a52bc95282a570ed910b2ddccfd01436/src/common/lib/transport/connectionmanager.ts#L700-L717, and more generally it doesn't emit a state event if it is currently in the same state as attempting to notify: https://github.com/ably/ably-js/blob/471cdef3a52bc95282a570ed910b2ddccfd01436/src/common/lib/transport/connectionmanager.ts#L1217-L1218
And relevant spec point: https://sdk.ably.com/builds/ably/specification/main/features/#RTN24

based on this we can do the following:
- have `initialized`, `syncing`, `synced` and `update` events/states.
- we won't (generally) emit a state event if we're currently in it (can't receive two `syncing` events in a row).
- if we're currently in `synced` and:
- - receive a `STATE_SYNC` message that can be immediately applied (sync cursor is empty), we will emit an `update` event and stay in `synced`. otherwise we enter `syncing` and wait for sync sequence to complete as usual.
- - we reattach to a channel with no `HAS_STATE` flag - we emit `update` event and stay in `synced`
- also, I suggest we still allow to subscribe to events even if `state_subscribe` mode wasn't set (meaning we won't receive a `HAS_STATE` flag on attachment) to improve DX (say you attach to a channel without the mode, get liveobjects instance and subscribe to lifecycle events now, and only at some time in the future you reattach with correct modes. if we don't allow subscribing unless `state_subscribe` mode is set this no longer becomes possible). To maintain a consistent `syncing` -> `synced` order, when a channel is attached while in the `initialized` state 1. it will transition to `syncing`. 2. on the next event loop, it will emit `synced` since it is immediately synchronized due to the absence of a `HAS_STATE` flag.

mschristensen: Sounds good to me

VeskeR: updated implementation in:
https://github.com/ably/ably-js/pull/1958/commits/98d66e26f8270ddc422b8b9f016572cc8fefb3d9 - adds state to `LiveObjects` just like channel/connection
https://github.com/ably/ably-js/pull/1958/commits/cf6f5fd5cd44c8d43f7aef6da80683e6659f077e - emits state change events

as discussed on the standup  yesterday, I removed the `update` event as it will have different meaning here than in channel/connection (`update` here would simply mean that underlying data has changed, and we have live object subscriptions for that).

VeskeR: also should point out that it is not technically possible to receive a `initialized` event as it is set on channel/connection object construction at which point we can't event subscribe to events yet.
and no other state can transition back to `initialized`, so it's impossible to get it later too.

thus we don't emit `initialized` in liveobjects either.

mschristensen (approved):
Nice, LGTM

PR #1963, opened 2025-01-27
[DTP-1118] Apply `ConnectionDetails.maxMessageSize` limit to state messages

VeskeR:
Message size calculation algorithm is based on https://github.com/ably/realtime/blob/d0557b715a5dbace6dce456b14127eea75d6f019/go/realtime/lib/ablyrpc/state.go#L452, with changes:
- object ids are not added to the size ([DTP-1136](https://ably.atlassian.net/browse/DTP-1136)), except when `MAP_SET` operation is referencing another object (TBD)
- `createOp` in the object messages also contributes to the message size (TBD)

Message size calculation tests are based on https://github.com/ably/realtime/blob/d0557b715a5dbace6dce456b14127eea75d6f019/nodejs/realtime/test/unit/encoding.ts#L1055

Resolves [DTP-1118](https://ably.atlassian.net/browse/DTP-1118)


VeskeR:
@mschristensen Please let me know if these two changes are in-line with your idea for [DTP-1136](https://ably.atlassian.net/browse/DTP-1136):
> except when MAP_SET operation is referencing another object (TBD)
createOp in the object messages also contributes to the message size (TBD)

VeskeR:
Removed `except when MAP_SET operation is referencing another object` as discussed on the stand-up (it is now also not included in calculation)

src/plugins/liveobjects/statemessage.ts
```diff
@@ -445,4 +445,121 @@ export class StateMessage {
...
+
+    if (obj.map) {
+      size += this._getStateMapSize(obj.map);
+    }
+    if (obj.counter) {
+      size += this._getStateCounterSize(obj.counter);
+    }
+    if (obj.createOp) {
+      size += this._getStateOperationSize(obj.createOp);
+    }
```
zknill, line 496: I _think_ the createOp exists on the StateObject for the correct operation of the CRDT semantics and is not something that I'd expect the dev/customer to be setting.

This `getMessageSize()` is used on publish to liveObjects. But I can't think of the case where we'd be publishing a `StateObject` rather than a `StateOperation` (?) so perhaps it's neither here-nor-there.

VeskeR: > But I can't think of the case where we'd be publishing a StateObject rather than a StateOperation

Yeah, the client currently never publishes `StateObject` messages. However, we did discuss the possibility of this happening in the future with Yjs integration, where documents may need to be synced on attachment using a two-way `STATE_SYNC` with `StateObject`s set. But for now it's very much up in the air.

Even though the client doesn’t publish `StateObject` at the moment, the additional complexity of handling the calculation is minimal. It’s just a single method, which already reuses other methods required to calculate `StateOperation` anyway, and one `this._getStateObjectSize(this.object)` call just a [few lines above](https://github.com/ably/ably-js/pull/1963/files/2b600246896319a8ca2ccf5c55aecd47affaeeab#diff-48433d824e48fe4b83b91e608309501eb2b8505533bc3856f5120123d799e9deR457).

In return, we get parity with the realtime size calculation implementation - even the tests can be shared.

> I think the createOp exists on the StateObject for the correct operation of the CRDT semantics and is not something that I'd expect the dev/customer to be setting.

If `StateObject` size calculation is to satay, then we should definitely include `createOp`, as it carries user-provided data.

zknill (approved):

kaschula (approved):

PR #1965, opened 2025-01-31
[DTP-1096] Run LiveObjects tests using both `MsgPack` and `JSON` protocols, and on all transports

VeskeR:
Only the tests that interact with the realtime server are marked to run with both `MsgPack` and `JSON` protocols. Tests that rely on private `channel.processMessage` calls to inject forged state messages are not included.

Resolves [DTP-1096](https://ably.atlassian.net/browse/DTP-1096)

[DTP-1096]: https://ably.atlassian.net/browse/DTP-1096?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **Refactor**
	- Updated test helper methods across multiple test files.
	- Renamed `restTestOnJsonMsgpack` to `testOnJsonMsgpack`.
	- Renamed `testOnAllTransports` to `testOnAllTransportsAndProtocols`.
	- Modified test function signatures to support more flexible testing options.
	- Added `only` and `skip` methods for more granular test control.

- **Tests**
	- Enhanced test suite for live objects, history, requests, and status functionality.
	- Improved protocol and scenario testing capabilities.
	- Expanded testing scope to include both transports and protocols across various test cases.
	- Updated tests to ensure compatibility with new method signatures and parameters.
	- Refined existing test structures and descriptions without introducing new functionality.

### DIFF
--- a/test/common/modules/shared_helper.js
+++ b/test/common/modules/shared_helper.js
@@ -342,35 +342,49 @@ define([
       );
     }
 
-    /* testFn is assumed to be a function of realtimeOptions that returns a mocha test */
-    static testOnAllTransports(thisInDescribe, name, testFn, skip) {
-      const helper = this.forTestDefinition(thisInDescribe, name).addingHelperFunction('testOnAllTransports');
-      var itFn = skip ? it.skip : it;
+    /* testFn is assumed to be a function of realtimeOptions and channelName that returns a mocha test */
+    static testOnAllTransportsAndProtocols(thisInDescribe, testName, testFn, skip, only) {
+      const helper = SharedHelper.forTestDefinition(thisInDescribe, testName).addingHelperFunction(
+        'testOnAllTransportsAndProtocols',
+      );
+      const itFn = skip ? it.skip : only ? it.only : it;
 
-      function createTest(options) {
+      function createTest(options, channelName) {
         return function (done) {
-          this.test.helper = this.test.helper.withParameterisedTestTitle(name);
-          return testFn(options).apply(this, [done]);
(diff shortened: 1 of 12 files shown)


test/common/modules/shared_helper.js
lawrence-forooghian, line 1: I wonder if the reason that this was called `restTestOnJsonMsgpack` is because Realtime tests we instead use `testOnAllTransports`, which tests each transport in binary and text mode. Might it be appropriate to use `testOnAllTransports` for the LiveObjects tests too?

VeskeR: Agree, testing on all transports + protocols might catch more edge cases. I've changed live objects tests to use `testOnAllTransports` method.
+ some more refactoring for the `testOnAllTransports` method so it is a bit more clear that it tests all transports **and** protocols, and some QoL features (like support for callback and async based test functions).
see (more details in commit messages):
https://github.com/ably/ably-js/pull/1965/commits/48cb5efb751c4e7324cb7bfb9ba0e36f56c6ea76
https://github.com/ably/ably-js/pull/1965/commits/622039ce25203104c56e54a278a75b4519d5e33e

I would also like to keep refactoring changes made for `SharedHelper.restTestOnJsonMsgpack` in https://github.com/ably/ably-js/pull/1965/commits/b23d31c645d8a77a05dc740588abd81be1490750, as it generally improves the function.

lawrence-forooghian (approved):

PR #1969, opened 2025-02-13
[PUB-1039] Update Live Objects README section

VeskeR:
Describe the Live Objects create/edit/read API, live object types, root object, state channel modes, batch operations lifecycle events and user provided typings for Live Objects.

Resolves [PUB-1039](https://ably.atlassian.net/browse/PUB-1039)

[PUB-1039]: https://ably.atlassian.net/browse/PUB-1039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **Documentation**
  - Enhanced Ably client library documentation with refined guidance on push notifications and LiveObjects.
  - Corrected a typographical error in the push notifications text.
  - Renamed the "Live Objects" section to "LiveObjects" and added a new "Using the plugin" subsection with TypeScript code snippets.
  - Updated example code to TypeScript, offering clearer instructions for interacting with LiveObjects features such as LiveMap and LiveCounter.
  - Expanded explanations for LiveObjects types with detailed usage instructions and examples.
  - Standardized terminology by replacing "Live Objects" with "LiveObjects" in relevant comments and documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -584,13 +584,15 @@ const client = new Ably.Rest({
 
 The Push plugin is developed as part of the Ably client library, so it is available for the same versions as the Ably client library itself. It also means that it follows the same semantic versioning rules as they were defined for [the Ably client library](#for-browsers). For example, to lock into a major or minor version of the Push plugin, you can specify a specific version number such as https://cdn.ably.com/lib/push.umd.min-2.js for all v2._ versions, or https://cdn.ably.com/lib/push.umd.min-2.4.js for all v2.4._ versions, or you can lock into a single release with https://cdn.ably.com/lib/push.umd.min-2.4.0.js. Note you can load the non-minified version by omitting `.min` from the URL such as https://cdn.ably.com/lib/push.umd-2.js.
 
-For more information on publishing push notifcations over Ably, see the [Ably push documentation](https://ably.com/docs/push).
+For more information on publishing push notifications over Ably, see the [Ably push documentation](https://ably.com/docs/push).
 
-### Live Objects functionality
+### LiveObjects
 
-Live Objects functionality is supported for Realtime clients via the LiveObjects plugin. In order to use Live Objects, you must pass in the plugin via client options.
+#### Using the plugin
 
-```javascript
+LiveObjects functionality is supported for Realtime clients via the LiveObjects plugin. In order to use LiveObjects, you must pass in the plugin via client options.
+
+```typescript
 import * as Ably from 'ably';
 import LiveObjects from 'ably/liveobjects';
 
@@ -610,7 +612,7 @@ Alternatively, you can load the LiveObjects plugin directly in your HTML using `
 
 When loaded this way, the LiveObjects plugin will be available on the global object via the `AblyLiveObjectsPlugin` property, so you will need to pass it to the Ably instance as follows:
 
-```javascript
+```typescript
 const client = new Ably.Realtime({
   ...options,
   plugins: { LiveObjects: AblyLiveObjectsPlugin },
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -619,7 +621,283 @@ const client = new Ably.Realtime({
...
+
+```typescript
+const root = await liveObjects.getRoot();
+```
+
+The root object is always a `LiveMap` instance and serves as the starting point for storing and organizing Live Objects state.
+
+#### Live Object Types
+
+Live Objects support two primary data structures:
```
paddybyers, line 658: currently

README.md
```diff
@@ -619,7 +621,283 @@ const client = new Ably.Realtime({
...
+
+await counter.decrement(2);
+console.log(counter.value()); // output: 3
+```
+
+#### Subscribing to Updates
+
+Subscribing to updates on Live Objects allows you to receive changes made by other clients in realtime. Since multiple clients may modify the same Live Objects state, subscribing ensures that your application reacts to external updates as soon as they are received.
+
+Additionally, mutation methods such as `LiveMap.set`, `LiveCounter.increment`, and `LiveCounter.decrement` do not directly edit the current state of the object locally. Instead, they send the intended operation to the Ably system, and the change is applied to the local object only when the corresponding realtime operation is echoed back to the client. This means that the state you retrieve immediately after a mutation may not reflect the latest updates yet.
```
paddybyers, line 710: Can you clarify this? If you `await` a mutation, the local state does not have that mutation applied after completion of the operation?

VeskeR: Correct, the mutation operation might not be applied after we `await` the mutation call, as this only awaits the `ACK` message for the mutation operation. It will be applied only once the operation is echoed back to the client.  

We discussed this with Mike in this [DR](https://ably.atlassian.net/wiki/spaces/LOB/pages/3526950914/LODR-024+Realtime+Client+write-only+internal+spec?focusedCommentId=3539140626), and considered the possibility of applying the mutation client-side when we receive an `ACK` message (meaning it would be applied locally after `await`). It needed more thought (as we rely on applying messages from a given site in regional order), and it was put on hold for the time being.

README.md
```diff
@@ -619,7 +621,283 @@ const client = new Ably.Realtime({
...
+  plugins: { LiveObjects },
+});
+const channelOptions = { modes: ['state_subscribe', 'state_publish'] };
+const channel = client.channels.get('my_live_objects_channel', channelOptions);
+const liveObjects = channel.liveObjects;
+```
+
+The authentication token must include corresponding capabilities for the client to interact with Live Objects.
+
+#### Obtaining the Root Object
```
zknill, line 646: `Obtaining` is a really convoluted word. Where possible we should prefer the simplest possible language. How about "getting" or "accessing" ?

README.md
```diff
@@ -619,7 +621,283 @@ const client = new Ably.Realtime({
...
+
+The root object represents the top-level entry point for Live Objects state within a channel. It acts as a container for all nested data structures and must be retrieved in order to interact with the existing state.
+
+```typescript
+const root = await liveObjects.getRoot();
+```
+
+The root object is always a `LiveMap` instance and serves as the starting point for storing and organizing Live Objects state.
```
zknill, line 654:
```suggestion
The root object is a `LiveMap` instance and serves as the starting point for storing and organizing Live Objects state.
```

README.md
```diff
@@ -619,7 +621,283 @@ const client = new Ably.Realtime({
...
+
+#### Live Object Types
+
+Live Objects currently support two primary data structures:
```
zknill, line 658:
```suggestion
Live Objects currently supports two primary data structures; LiveMap and LiveCounter.
```

README.md
```diff
@@ -619,7 +621,283 @@ const client = new Ably.Realtime({
...
+console.log(counter.value()); // output: 3
+```
+
+#### Subscribing to Updates
+
+Subscribing to updates on Live Objects allows you to receive changes made by other clients in realtime. Since multiple clients may modify the same Live Objects state, subscribing ensures that your application reacts to external updates as soon as they are received.
+
+Additionally, mutation methods such as `LiveMap.set`, `LiveCounter.increment`, and `LiveCounter.decrement` do not directly edit the current state of the object locally. Instead, they send the intended operation to the Ably system, and the change is applied to the local object only when the corresponding realtime operation is echoed back to the client. This means that the state you retrieve immediately after a mutation may not reflect the latest updates yet.
+
+To ensure you react to state changes at the correct moment, you can subscribe to updates on all Live Objects as follows:
```
zknill, line 712:
```suggestion
You can subscribe to updates on all Live Objects using subscription listeners as follows:
```

I've added "subscription listeners" here because later way say you can de-register "subscription listeners" without mentioning them before. Obviously you can see from the code, but adding clarity here.

README.md
```diff
@@ -619,7 +621,283 @@ const client = new Ably.Realtime({
...
+
+// subscribe to updates on a LiveCounter
+const counter = await liveObjects.createCounter();
+counter.subscribe((update: LiveCounterUpdate) => {
+  console.log('LiveCounter updated:', update);
+});
+
+// perform operations on LiveMap and LiveCounter
+await root.set('name', 'Alice');
+// LiveMap updated: { update: { name: 'updated' } }
```
zknill, line 730: **question**: should the update event carry the new value as well as the key that was modified?

README.md
```diff
@@ -619,7 +621,283 @@ const client = new Ably.Realtime({
...
+  // Example: Show a loading indicator
+});
+
+liveObjects.on('synced', () => {
+  console.log('Live Objects state has been synced.');
+  // Example: Hide loading indicator
+});
+```
+
+**Object Deletion Events** - objects that have been orphaned for a long period (i.e., not connected to the state tree graph by being set as a key in a map related to the root) will eventually be deleted. Once a Live Object is deleted, it can no longer be interacted with. You should avoid accessing its data or try to update its value and should remove all references to the deleted object in your application.
```
zknill, line 825:
```suggestion
**Object Deletion Events** - objects that have been orphaned for a long period (i.e., not connected to the state tree graph by being set as a key in a map related to the root) will eventually be deleted. Once a Live Object is deleted, it can no longer be interacted with. You should avoid accessing its data or trying to update its value and you should remove all references to the deleted object in your application.
```

zknill (commented):
Nice! A few boring re-word suggestions for simplicity/clarity 👌

README.md
```diff
@@ -619,7 +621,283 @@ const client = new Ably.Realtime({
 
 The LiveObjects plugin is developed as part of the Ably client library, so it is available for the same versions as the Ably client library itself. It also means that it follows the same semantic versioning rules as they were defined for [the Ably client library](#for-browsers). For example, to lock into a major or minor version of the LiveObjects plugin, you can specify a specific version number such as https://cdn.ably.com/lib/liveobjects.umd.min-2.js for all v2._ versions, or https://cdn.ably.com/lib/liveobjects.umd.min-2.4.js for all v2.4._ versions, or you can lock into a single release with https://cdn.ably.com/lib/liveobjects.umd.min-2.4.0.js. Note you can load the non-minified version by omitting `.min` from the URL such as https://cdn.ably.com/lib/liveobjects.umd-2.js.
 
-For more information about Live Objects product, see the [Ably Live Objects documentation](https://ably.com/docs/products/liveobjects).
+For more information about the Live Objects product, see the [Ably Live Objects documentation](https://ably.com/docs/products/liveobjects).
+
+#### State Channel Modes
+
+To use Live Objects feature, clients must attach to a channel with the correct channel mode:
```
mschristensen, line 628: Are we calling it `Live Objects` or `LiveObjects`? I think we should use a consistent name, so probably `LiveObjects`

README.md
```diff
@@ -619,7 +621,283 @@ const client = new Ably.Realtime({
...
+console.log(root.get('name')); // output: undefined
+
+// different data types are supported
+await root.set('foo', 'str');
+await root.set('bar', 1);
+await root.set('baz', true);
+await root.set('qux', new Uint8Array([21, 31]));
+// as well as other live objects
+const counter = await liveObjects.createCounter();
+await root.get('quux', counter);
```
mschristensen, line 684: Should this be `root.set` ?

README.md
```diff
@@ -619,7 +621,283 @@ const client = new Ably.Realtime({
...
+// resulting structure:
+// root (LiveMap)
+// └── outerMap (LiveMap)
+//     └── map (LiveMap)
+//         └── counter (LiveCounter)
+```
+
+#### Batch Operations
+
+Batching allows multiple operations to be grouped into a single message that is sent to the Ably service.
```
mschristensen, line 787:
```suggestion
Batching allows multiple operations to be grouped into a single message that is sent to the Ably service. This allows batched operations to be applied atomically together.
```

README.md
mschristensen: Also maybe `related to the root` -> `accessible from the root map object`.

README.md
```diff
@@ -619,7 +621,283 @@ const client = new Ably.Realtime({
...
+};
+
+declare global {
+  export interface LiveObjectsTypes {
+    root: MyCustomRoot;
+  }
+}
+```
+
+This will enable IntelliSense support when interacting with the Live Objects API:
```
mschristensen, line 876: I think `IntelliSense` is a microsoft-specific concept, perhaps `code completion and editor hints` ?

mschristensen (changes requested):
Great stuff, thanks Andrii - just a couple of comments

README.md
VeskeR: Changed to "getting" as it is also used in "getting presence:"
https://github.com/ably/ably-js/compare/54a5125fcd43b43948ca982f7a1ed89c538cec50..3ca50035284754bf9cac21af0de744a0ae97ef85

README.md
```diff
@@ -619,7 +621,283 @@ const client = new Ably.Realtime({
...
+const channelOptions = { modes: ['state_subscribe', 'state_publish'] };
+const channel = client.channels.get('my_live_objects_channel', channelOptions);
+const liveObjects = channel.liveObjects;
+```
+
+The authentication token must include corresponding capabilities for the client to interact with Live Objects.
+
+#### Obtaining the Root Object
+
+The root object represents the top-level entry point for Live Objects state within a channel. It acts as a container for all nested data structures and must be retrieved in order to interact with the existing state.
```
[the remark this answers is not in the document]
VeskeR: updated 👍 
https://github.com/ably/ably-js/compare/54a5125fcd43b43948ca982f7a1ed89c538cec50..3ca50035284754bf9cac21af0de744a0ae97ef85

README.md
VeskeR: changed
https://github.com/ably/ably-js/compare/54a5125fcd43b43948ca982f7a1ed89c538cec50..3ca50035284754bf9cac21af0de744a0ae97ef85

README.md
VeskeR: changed
https://github.com/ably/ably-js/compare/54a5125fcd43b43948ca982f7a1ed89c538cec50..3ca50035284754bf9cac21af0de744a0ae97ef85

README.md
VeskeR: changed both
https://github.com/ably/ably-js/compare/54a5125fcd43b43948ca982f7a1ed89c538cec50..3ca50035284754bf9cac21af0de744a0ae97ef85

README.md
VeskeR: right, it's the `LiveObjects` _product_, but sometimes the readme explicitly refers to the _objects_ in LiveObjects, which are just `Live Objects`.
I've gone through all of the "liveobjects"/"live objects" uses to double check which they are referring to and update accordingly:
https://github.com/ably/ably-js/compare/54a5125fcd43b43948ca982f7a1ed89c538cec50..3ca50035284754bf9cac21af0de744a0ae97ef85
https://github.com/ably/ably-js/compare/3ca50035284754bf9cac21af0de744a0ae97ef85..2065e11c8b39e421b3289cad832930814d03a618

README.md
VeskeR: added
https://github.com/ably/ably-js/compare/54a5125fcd43b43948ca982f7a1ed89c538cec50..3ca50035284754bf9cac21af0de744a0ae97ef85

README.md
VeskeR: changed
https://github.com/ably/ably-js/compare/54a5125fcd43b43948ca982f7a1ed89c538cec50..3ca50035284754bf9cac21af0de744a0ae97ef85

README.md
VeskeR: the current API only exposes the name of the key that got updated/deleted. it is trivial at this point to just call `root.get(key)` to get the new value.

a bit more useful could be exposing both the previous and teh new value, just like we do with `ChannelStateChange`, but  currently it's not implemented

VeskeR: created https://ably.atlassian.net/browse/PUB-1207 to track this

README.md
lawrence-forooghian: The current behaviour is consistent with that of presence in the core SDKs, right? e.g. if you call `presence.get()` immediately after the completion of `presence.enter()`, you might not see yourself in the presence set.

README.md
```diff
@@ -619,7 +621,284 @@ const client = new Ably.Realtime({
...
+You can use `LiveMap` as follows:
+
+```typescript
+// root object is a LiveMap
+const root = await liveObjects.getRoot();
+
+// set a key on map
+await root.set('name', 'Alice');
+// get a value for a key
+console.log(root.get('name')); // output: 'Alice'
```
lawrence-forooghian, line 671: I get the point that we're trying to make in demonstrating this API, but it might be a bit misleading given that, as you say later, updates are not immediately reflected in the local state.

README.md
```diff
@@ -619,7 +621,284 @@ const client = new Ably.Realtime({
...
+
+Additionally, mutation methods such as `LiveMap.set`, `LiveCounter.increment`, and `LiveCounter.decrement` do not directly edit the current state of the object locally. Instead, they send the intended operation to the Ably system, and the change is applied to the local object only when the corresponding realtime operation is echoed back to the client. This means that the state you retrieve immediately after a mutation may not reflect the latest updates yet.
+
+You can subscribe to updates on all Live Objects using subscription listeners as follows:
+
+```typescript
+const root = await liveObjects.getRoot();
+
+// subscribe to updates on a LiveMap
+root.subscribe((update: LiveMapUpdate) => {
```
lawrence-forooghian, line 719: Maybe we could make it clearer that the user does not _need_ to look at the contents of the `update`; i.e. that they can just re-read the entire `LiveMap` or `LiveCounter` if they wish?

README.md
VeskeR: > The current behaviour is consistent with that of presence in the core SDKs, right? e.g. if you call presence.get() immediately after the completion of presence.enter(), you might not see yourself in the presence set.

I believe it matches the behavior of presence, yes. However, this is currently the case for LiveObjects due to technical limitations, as we do not yet have a way to apply operations client-side outside of the regional order. Eventually, we would like to apply the operation immediately after receiving acknowledgment (on receiving `ACK`) that the operation message was received by the server, or ideally, even optimistically before sending the message to the realtime system (which would enable a synchronous mutation API).

README.md
VeskeR: updated in https://github.com/ably/ably-js/compare/2065e11c8b39e421b3289cad832930814d03a618..2f5a07832c6227566cc4cd45cf6e96e096e90b3b, to be more explicit that you can just read the value in the callback directly.

README.md
VeskeR: Yeah, I wasn't sure about the best way to demonstrate the API here, but I agree that we should minimize confusion as much as possible.

See https://github.com/ably/ably-js/compare/2065e11c8b39e421b3289cad832930814d03a618..2f5a07832c6227566cc4cd45cf6e96e096e90b3b, where I've updated the snippets to focus on showcasing the methods (with synchronous methods first) without displaying expected output values. This should now give a more guidance-like feel rather than being a literal example of the expected behavior.

VeskeR: And did the same for `batch` method example:
https://github.com/ably/ably-js/compare/50e40ef4f3dcf227b17e75b9984f49c0e293c402..54c1c607d902aeba2de375f2d0b081f286ef710a

mschristensen (approved):
Nice LGTM :)

lawrence-forooghian (approved):

PR #1972, opened 2025-02-25
[PUB-1030] LiveObjects behavior under different channel states

VeskeR:
When the channel is `SUSPENDED`

- The local copy of the last known LiveObjects state is maintained, ensuring that the API for accessing objects continues to function and returns the latest known data.
- The getRoot() function will return the last known root object.
- Write API functions will throw an error, as the channel is in an invalid state (i.e., messages cannot be sent or queued while suspended).

When the channel is `DETACHED` or `FAILED`

- The local LiveObjects state copy is cleared (LiveObjects data update events are not emitted in this case (e.g., counter increments, map key set/delete), as the current state of LiveObjects data is unknown)
- All access and write API functions, including getRoot(), will throw an error indicating that the channel is in an invalid state.

Resolves [PUB-1030](https://ably.atlassian.net/browse/)

[PUB-1030]: https://ably.atlassian.net/browse/PUB-1030?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/plugins/liveobjects/batchcontext.ts
+++ b/src/plugins/liveobjects/batchcontext.ts
@@ -24,7 +24,7 @@ export class BatchContext {
   }
 
   getRoot<T extends API.LiveMapType = API.DefaultRoot>(): BatchContextLiveMap<T> {
-    this._liveObjects.throwIfMissingStateSubscribeMode();
+    this._liveObjects.throwIfInvalidAccessApiConfiguration();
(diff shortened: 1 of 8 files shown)


mschristensen (approved):
Nice, LGTM

PR #1973, opened 2025-02-25
Use endpoint as default connection option (ADR-119)

surminus:
This implements ADR-119[1], which specifies the client connection options to update requests to the endpoints implemented as part of ADR-042[2].

The endpoint may be one of the following:

* a routing policy name (such as `main`)
* a nonprod routing policy name (such as `nonprod:sandbox`)
* a FQDN such as `foo.example.com`

The endpoint option is not valid with any of `environment`, `restHost` or `realtimeHost`, but we still intend to support the legacy options.

If the client has been configured to use any of these legacy options, then they should continue to work in the same way, using the same primary and fallback hostnames.

If the client has not been explicitly configured, then the hostnames will change to the new `ably.net` domain when the package is upgraded.

## Notes on implementation

As per previous guidance, I have _not_ updated tests to use the endpoint client option, but since the hostnames are changing, then the assertions and default environment must be updated (from `sandbox` to `nonprod:sandbox`).

I have not added any explicit tests for the endpoint client option, but I can do this if we think it's necessary, although using `environment` is essentially the same as using `endpoint`.

See related Go (merged), Ruby and Python PRs:

* https://github.com/ably/ably-go/pull/679
* https://github.com/ably/ably-ruby/pull/441
* https://github.com/ably/ably-python/pull/590

[1] https://ably.atlassian.net/wiki/spaces/ENG/pages/3428810778/ADR-119+ClientOptions+for+new+DNS+structure
[2] https://ably.atlassian.net/wiki/spaces/ENG/pages/1791754276/ADR-042+DNS+Restructure

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -391,6 +391,11 @@ export interface ClientOptions<Plugins = CorePlugins> extends AuthOptions {
   echoMessages?: boolean;
 
   /**
+   * Set a routing policy or FQDN to connect to Ably. See [platform customization](https://ably.com/docs/platform-customization).
+   */
+  endpoint?: string;
+
+  /**
    * Enables a [custom environment](https://ably.com/docs/platform-customization) to be used with the Ably service.
    */
   environment?: string;
--- a/src/common/lib/util/defaults.ts
+++ b/src/common/lib/util/defaults.ts
@@ -13,6 +13,7 @@ import { ModularPlugins } from '../client/modularplugins';
 let agent = 'ably-js/' + version;
 
 type CompleteDefaults = IDefaults & {
+  ENDPOINT: string;
   ENVIRONMENT: string;
   REST_HOST: string;
   REALTIME_HOST: string;
@@ -40,7 +41,7 @@ type CompleteDefaults = IDefaults & {
   getHost(options: ClientOptions, host?: string | null, ws?: boolean): string;
   getPort(options: ClientOptions, tls?: boolean): number | undefined;
   getHttpScheme(options: ClientOptions): string;
-  environmentFallbackHosts(environment: string): string[];
+  getEndpointFallbackHosts(endpoint: string): string[];
   getFallbackHosts(options: NormalisedClientOptions): string[];
   getHosts(options: NormalisedClientOptions, ws?: boolean): string[];
   checkHost(host: string): void;
@@ -57,15 +58,16 @@ type CompleteDefaults = IDefaults & {
 };
 
 const Defaults = {
+  ENDPOINT: 'main',
   ENVIRONMENT: '',
   REST_HOST: 'rest.ably.io',
   REALTIME_HOST: 'realtime.ably.io',
   FALLBACK_HOSTS: [
-    'A.ably-realtime.com',
-    'B.ably-realtime.com',
-    'C.ably-realtime.com',
-    'D.ably-realtime.com',
-    'E.ably-realtime.com',
+    'main.a.fallback.ably-realtime.com',
+    'main.b.fallback.ably-realtime.com',
+    'main.c.fallback.ably-realtime.com',
+    'main.d.fallback.ably-realtime.com',
+    'main.e.fallback.ably-realtime.com',
   ],
   PORT: 80,
   TLS_PORT: 443,
@@ -94,7 +96,7 @@ const Defaults = {
   getHost,
   getPort,
   getHttpScheme,
-  environmentFallbackHosts,
+  getEndpointFallbackHosts,
   getFallbackHosts,
   getHosts,
   checkHost,
@@ -119,15 +121,30 @@ export function getHttpScheme(options: ClientOptions): string {
   return options.tls ? 'https://' : 'http://';
 }
 
-// construct environment fallback hosts as per RSC15i
-export function environmentFallbackHosts(environment: string): string[] {
-  return [
-    environment + '-a-fallback.ably-realtime.com',
-    environment + '-b-fallback.ably-realtime.com',
-    environment + '-c-fallback.ably-realtime.com',
-    environment + '-d-fallback.ably-realtime.com',
-    environment + '-e-fallback.ably-realtime.com',
-  ];
+export function getEndpointHostname(endpoint: string): string {
+  if (endpoint.includes('.') || endpoint.includes('::') || endpoint.includes('localhost')) {
+    return endpoint;
+  }
+
+  if (endpoint.startsWith('nonprod:')) {
+    const root = endpoint.replace('nonprod:', '');
+    return `${root}.realtime.ably-nonprod.net`;
+  }
+
+  return `${endpoint}.realtime.ably.net`;
+}
+
+export function getEndpointFallbackHosts(endpoint: string): string[] {
+  if (endpoint.startsWith('nonprod:')) {
+    const root = endpoint.replace('nonprod:', '');
+    return endpointFallbacks(root, 'ably-realtime-nonprod.com');
+  }
+
+  return endpointFallbacks(endpoint, 'ably-realtime.com');
+}
+
+export function endpointFallbacks(root: string, domain: string): string[] {
+  return ['a', 'b', 'c', 'd', 'e'].map((id) => `${root}.${id}.fallback.${domain}`);
 }
 
 export function getFallbackHosts(options: NormalisedClientOptions): string[] {
@@ -150,26 +167,6 @@ function checkHost(host: string): void {
   if (!host.length) {
     throw new ErrorInfo('host must not be zero-length', 40000, 400);
   }
-}
-
-function getRealtimeHost(options: ClientOptions, production: boolean, environment: string, logger: Logger): string {
-  if (options.realtimeHost) return options.realtimeHost;
-  /* prefer setting realtimeHost to restHost as a custom restHost typically indicates
-   * a development environment is being used that can't be inferred by the library */
-  if (options.restHost) {
-    Logger.logAction(
-      logger,
-      Logger.LOG_MINOR,
-      'Defaults.normaliseOptions',
-      'restHost is set to "' +
-        options.restHost +
-        '" but realtimeHost is not set, so setting realtimeHost to "' +
-        options.restHost +
-        '" too. If this is not what you want, please set realtimeHost explicitly.',
-    );
-    return options.restHost;
-  }
-  return production ? Defaults.REALTIME_HOST : environment + '-' + Defaults.REALTIME_HOST;
 }
 
 function getTimeouts(options: ClientOptions) {
@@ -262,16 +259,15 @@ export function normaliseOptions(
 
   if (!('queueMessages' in options)) options.queueMessages = true;
 
-  /* infer hosts and fallbacks based on the configured environment */
-  const environment = (options.environment && String(options.environment).toLowerCase()) || Defaults.ENVIRONMENT;
-  const production = !environment || environment === 'production';
+  /* infer hosts and fallbacks based on the specified endpoint */
+  const endpoint = options.endpoint || options.environment || Defaults.ENDPOINT;
 
   if (!options.fallbackHosts && !options.restHost && !options.realtimeHost && !options.port && !options.tlsPort) {
-    options.fallbackHosts = production ? Defaults.FALLBACK_HOSTS : environmentFallbackHosts(environment);
+    options.fallbackHosts = getEndpointFallbackHosts(endpoint);
   }
 
-  const restHost = options.restHost || (production ? Defaults.REST_HOST : environment + '-' + Defaults.REST_HOST);
-  const realtimeHost = getRealtimeHost(options, production, environment, loggerToUse);
+  const restHost = options.restHost || getEndpointHostname(endpoint);
+  const realtimeHost = options.realtimeHost || getEndpointHostname(endpoint);
 
   (options.fallbackHosts || []).concat(restHost, realtimeHost).forEach(checkHost);
 
--- a/test/common/globals/environment.js
+++ b/test/common/globals/environment.js
@@ -3,7 +3,7 @@
 define(function (require) {
   var defaultLogLevel = 4,
     environment = isBrowser ? window.__env__ || {} : process.env,
-    ablyEnvironment = environment.ABLY_ENV || 'sandbox',
+    ablyEnvironment = environment.ABLY_ENV || 'nonprod:sandbox',
     realtimeHost = environment.ABLY_REALTIME_HOST,
     restHost = environment.ABLY_REST_HOST,
     port = environment.ABLY_PORT || 80,
--- a/test/common/modules/testapp_manager.js
+++ b/test/common/modules/testapp_manager.js
@@ -3,7 +3,7 @@
 
 /* testapp module is responsible for setting up and tearing down apps in the test environment */
 define(['globals', 'ably'], function (ablyGlobals, ably) {
-  var restHost = ablyGlobals.restHost || prefixDomainWithEnvironment('rest.ably.io', ablyGlobals.environment),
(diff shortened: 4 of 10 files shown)


lawrence-forooghian:
@surminus as discussed on Slack, please could you make sure that this PR implements / tests the updates to REC1b2 from https://github.com/ably/specification/pull/302? Thanks!

ably.d.ts
```diff
@@ -390,6 +390,11 @@ export interface ClientOptions<Plugins = CorePlugins> extends AuthOptions {
    */
   echoMessages?: boolean;
 
+  /**
+   * Set a routing policy or FQDN to connect to Ably. See [platform customization](https://ably.com/docs/platform-customization).
+   */
+  endpoint?: string;
+
   /**
    * Enables a [custom environment](https://ably.com/docs/platform-customization) to be used with the Ably service.
```
lawrence-forooghian, line 399: I think we need to mark some things here as deprecated?

ably.d.ts
```diff
@@ -390,6 +390,11 @@ export interface ClientOptions<Plugins = CorePlugins> extends AuthOptions {
    */
   echoMessages?: boolean;
 
+  /**
+   * Set a routing policy or FQDN to connect to Ably. See [platform customization](https://ably.com/docs/platform-customization).
```
lawrence-forooghian, line 394: Just wanted to understand where this documentation has come from — did you write it yourself? We have a set of [docstrings](https://github.com/ably/specification/blob/main/textile/api-docstrings.md) that we use as the base for our documentation comments in all our SDKs. I don't see an entry for `endpoint` there though (I guess Paddy forgot to add one, which he should have done when adding this option, per [this guidance](https://github.com/ably/specification/blob/54e3a6af47608109ba811e621b3365daa3f52252/CONTRIBUTING.md?plain=1#L70)). Would you mind adding one there please?

ably.d.ts
lawrence-forooghian: (see e.g. 36162c4c4bfcbbbcc2e3276da8e197c95649a55c for an example)

ably.d.ts
surminus: https://github.com/ably/specification/pull/322

surminus:
@lawrence-forooghian I've pushed some changes including some new tests. One test is failing but seems unrelated to my changes?

src/common/lib/util/defaults.ts
```diff
@@ -119,15 +121,34 @@ export function getHttpScheme(options: ClientOptions): string {
...
+
+  if (endpoint.startsWith('nonprod:')) {
+    const root = endpoint.replace('nonprod:', '');
+    return `${root}.realtime.ably-nonprod.net`;
+  }
+
+  return `${endpoint}.realtime.ably.net`;
+}
+
+export function getEndpointFallbackHosts(endpoint: string): string[] {
```
lawrence-forooghian, line 137: Please can you annotate with the relevant spec points?

src/common/lib/util/defaults.ts
```diff
@@ -262,16 +263,15 @@ export function normaliseOptions(
...
+  const endpoint = options.endpoint || options.environment || Defaults.ENDPOINT;
 
   if (!options.fallbackHosts && !options.restHost && !options.realtimeHost && !options.port && !options.tlsPort) {
-    options.fallbackHosts = production ? Defaults.FALLBACK_HOSTS : environmentFallbackHosts(environment);
+    options.fallbackHosts = getEndpointFallbackHosts(endpoint);
   }
 
-  const restHost = options.restHost || (production ? Defaults.REST_HOST : environment + '-' + Defaults.REST_HOST);
-  const realtimeHost = getRealtimeHost(options, production, environment, loggerToUse);
+  const restHost = options.restHost || getEndpointHostname(endpoint);
```
lawrence-forooghian, line 273: I don't think that the spec has an internal concept of separate "rest host" and "realtime host" any more; isn't there just the "primary domain"? And then, REC1d specifies how the `restHost` and `realtimeHost` options, if specified, determine the primary domain.

src/common/lib/util/defaults.ts
```diff
@@ -119,15 +121,34 @@ export function getHttpScheme(options: ClientOptions): string {
...
-export function environmentFallbackHosts(environment: string): string[] {
-  return [
-    environment + '-a-fallback.ably-realtime.com',
-    environment + '-b-fallback.ably-realtime.com',
-    environment + '-c-fallback.ably-realtime.com',
-    environment + '-d-fallback.ably-realtime.com',
-    environment + '-e-fallback.ably-realtime.com',
-  ];
+export function getEndpointHostname(endpoint: string): string {
+  if (endpoint.includes('.') || endpoint.includes('::') || endpoint.includes('localhost')) {
```
lawrence-forooghian, line 125: The `localhost` check should be an equality check.

src/common/lib/util/defaults.ts
```diff
@@ -119,15 +121,34 @@ export function getHttpScheme(options: ClientOptions): string {
...
+  if (endpoint.startsWith('nonprod:')) {
+    const root = endpoint.replace('nonprod:', '');
+    return `${root}.realtime.ably-nonprod.net`;
+  }
+
+  return `${endpoint}.realtime.ably.net`;
+}
+
+export function getEndpointFallbackHosts(endpoint: string): string[] {
+  if (endpoint.includes('.') || endpoint.includes('::') || endpoint.includes('localhost')) {
```
lawrence-forooghian, line 138: The `localhost` check should be an equality check.

src/common/lib/util/defaults.ts
```diff
@@ -119,15 +121,34 @@ export function getHttpScheme(options: ClientOptions): string {
...
+    return [];
+  }
+
+  if (endpoint.startsWith('nonprod:')) {
+    const root = endpoint.replace('nonprod:', '');
```
lawrence-forooghian, line 143: Please can you use the "routing policy ID" terminology of the spec?

src/common/lib/util/defaults.ts
```diff
@@ -119,15 +121,34 @@ export function getHttpScheme(options: ClientOptions): string {
...
+    return endpointFallbacks(root, 'ably-realtime-nonprod.com');
+  }
+
+  return endpointFallbacks(endpoint, 'ably-realtime.com');
+}
+
+export function endpointFallbacks(root: string, domain: string): string[] {
```
lawrence-forooghian, line 150: Please can you use the "routing policy ID" terminology of the spec?

test/common/globals/environment.js
```diff
@@ -3,7 +3,7 @@
 define(function (require) {
   var defaultLogLevel = 4,
     environment = isBrowser ? window.__env__ || {} : process.env,
-    ablyEnvironment = environment.ABLY_ENV || 'sandbox',
+    ablyEnvironment = environment.ABLY_ENV || 'nonprod:sandbox',
```
lawrence-forooghian, line 6: Please can you update the documentation for this environment variable in `CONTRIBUTING.md`? Also, there's an `env` option described for web there — does that need updating to support an endpoint specifier too?

lawrence-forooghian: Also, I can't see any code that's been changed to make use of the new meaning of this option — I assumed that this option is ultimately used to generate a set of client options, so those client options would presumably need to be updated to use the `endpoint` option?

(And, maybe in that case it would also be clearer if we had a new option called `ABLY_ENDPOINT` that lets you set the `endpoint` client option, instead of overloading the meaning of "environment"?)

test/common/modules/testapp_manager.js
```diff
@@ -23,12 +23,12 @@ define(['globals', 'ably'], function (ablyGlobals, ably) {
     }
   }
 
-  function prefixDomainWithEnvironment(domain, environment) {
-    if (environment.toLowerCase() === 'production') {
-      return domain;
-    } else {
-      return environment + '-' + domain;
+  function setHostname(endpoint) {
```
lawrence-forooghian, line 26: The word "set" is a bit confusing here

lawrence-forooghian (commented):
Some initial comments, still having a look

surminus:
@lawrence-forooghian done (I think)

test/package/browser/template/playwright/index.tsx
```diff
@@ -9,7 +9,7 @@ beforeMount(async ({ App }) => {
 
   const client = new Ably.Realtime({
     key,
-    environment: 'sandbox',
+    environment: 'nonprod:sandbox',
```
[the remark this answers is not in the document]
lawrence-forooghian: I'm not expressing an opinion on which option is the right one to use, but `nonprod:sandbox` is not a valid value for the `environment` client option, no?

test/rest/defaults.test.js
```diff
@@ -35,49 +35,127 @@ define(['ably', 'chai'], function (Ably, chai) {
...
+      expect(Defaults.getHost(normalisedOptions, 'sandbox.realtime.ably-nonprod.net', true)).to.deep.equal(
+        'sandbox.realtime.ably-nonprod.net',
+      );
+
+      helper.recordPrivateApi('call.Defaults.getPort');
+      expect(Defaults.getPort(normalisedOptions)).to.equal(443);
+    });
+
+    /**
+     * @spec TO3k8
```
lawrence-forooghian, line 84: I notice that we have multiple `@spec` tags for a single spec point. Did you see the guidance about how to use these tags?

https://github.com/ably/ably-js/blob/f6fbe350c6c84f69811c0896e4999f94c2d80ef4/CONTRIBUTING.md?plain=1#L63-L66

Please could you check that you're using these tags correctly? It might be that `@specpartial` is more appropriate in some places, in which case it would be good to accompany them with an explanatory comment.

test/package/browser/template/playwright/index.tsx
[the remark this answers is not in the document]
surminus: It is, because `environment` is basically now just an alias for `endpoint`, and they behave the same way.

lawrence-forooghian:
Did you intentionally not implement the `connectivityCheckUrl` option (REC3)?

lawrence-forooghian:
I can't see implementations or tests for REC1b1 or REC1c1; please could you add?

surminus:
> Did you intentionally not implement the connectivityCheckUrl option (REC3)?

I don't think that URL has changed.

lawrence-forooghian:
I can't see mention of RTN17i or RSC25 — please could you add tests for them?

lawrence-forooghian:
> > Did you intentionally not implement the connectivityCheckUrl option (REC3)?
> 
> I don't think that URL has changed.

Ah, looks like it's something that already existed in ably-js but which has now been specified. What do you think about adding a spec point comment and checking there's a test for it whilst you're here?

surminus:
> What do you think about adding a spec point comment and checking there's a test for it whilst you're here?

If possible I'd like to limit the scope of this PR to as small as possible, because I already feel like I'm struggling to do the basics of the change I've already done :sweat_smile: This is the first time I've contributed to ably-js and consider myself a JavaScript novice, so if someone more experienced could handle adding missing tests for features contributed previously.

test/package/browser/template/playwright/index.tsx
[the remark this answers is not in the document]
surminus: But rather than try to amend this spec, I will just go through and update these tests to use `endpoint` instead.

src/common/lib/util/defaults.ts
surminus: I've pushed 65a30d5ebb05d571737c9f9723d61d457078508f, is that what you meant? I don't think I want to go and change `restHost` and `realtimeHost` anywhere else internally (or have the capacity to do that).

test/common/globals/environment.js
surminus: Should I change all instances of this from `environment` to `endpoint`? Where does this get used? I'm not sure if I understand what I would be changing here.

test/rest/defaults.test.js
surminus: I don't think I am. I copied and pasted these tests just with a different value for the `endpoint` client option. Is there a better way I should structure these tests?

surminus:
> I can't see mention of RTN17i or RSC25 — please could you add tests for them?

These seem unrelated to the change I made, even though they reference the primary domain. I don't think I would be confident writing tests for retries against the primary domain.

src/common/lib/util/defaults.ts
lawrence-forooghian: I think that if you don't feel comfortable making any further-reaching changes, then the change you've made is a good one.

test/common/globals/environment.js
lawrence-forooghian: I think we need to clear up the `environment` / `endpoint` confusion of https://github.com/ably/ably-js/pull/1973#discussion_r2114391667 first — currently, you're (I think) setting the `environment` option to a value that, per the current spec, only makes sense for the `endpoint` option.

lawrence-forooghian: > Update all tests to use `endpoint`

I had assumed that this was what we were going to do. It's what's written in https://ably.atlassian.net/wiki/spaces/ENG/pages/3428810778/ADR-119+ClientOptions+for+new+DNS+structure:

> SDKs, when updated to use target the new domains, should also have internal and test references to environment: ‘sandbox' updated to use the new option, ie endpoint: 'nonprod:sandbox';

surminus: Yeah, I did this in ably-go, and got told off (anyway, I will do it).

scripts/moduleReport.ts
```diff
@@ -6,7 +6,7 @@ import { gzip } from 'zlib';
 import Table from 'cli-table';
 
 // The maximum size we allow for a minimal useful Realtime bundle (i.e. one that can subscribe to a channel)
-const minimalUsefulRealtimeBundleSizeThresholdsKiB = { raw: 102, gzip: 31 };
+const minimalUsefulRealtimeBundleSizeThresholdsKiB = { raw: 103, gzip: 31 };
```
ttypic, line 9: Have to increase a little, bundle become 102.3kb

src/common/lib/util/defaults.ts
ttypic: Done

test/package/browser/template/playwright/index.tsx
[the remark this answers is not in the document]
ttypic: updated tests to use `endpoint`, I don't think we should treat `environment` as an alias for `endpoint` in the code and in the spec (it exists specifically for backward compatibility we don't want to accidentally introduce some new behavior to it). Tried to make it clear in the code, grouped all deprecated ways to define primary domain as a `primaryDomainFromLegacyOptions` variable

test/realtime/init.test.js
```diff
@@ -487,7 +494,10 @@ define(['ably', 'shared_helper', 'chai'], function (Ably, Helper, chai) {
           var hosts = new Ably.Rest._Http()._getHosts(realtime);
           /* restHost rather than realtimeHost as that's what connectionManager
            * knows about; converted to realtimeHost by the websocketTransport */
-          expect(hosts[0]).to.equal(realtime.options.realtimeHost, 'Check connected realtime host is the first option');
+          expect(hosts[0]).to.equal(
+            realtime.options.primaryDomain,
```
lawrence-forooghian, line 499: There are a few places where you're reading `options.primaryDomain`; please could you record this private API usage?

test/package/browser/template/playwright/index.tsx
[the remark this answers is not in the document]
lawrence-forooghian: Sounds like we need to make some changes in the Go implementation then?

lawrence-forooghian (approved):

PR #1988, opened 2025-03-23
PublishOptions type issue fix

mattheworiordan:
When using Ably with TypeScript, I cannot build because I am told that quickAck cannot be undefined based on the definition below, yet quickAck has a `?` which I assume means it does allow  boolean or undefined as a value. This is an incompatible change introduced in https://github.com/ably/ably-js/commit/02de82b07f8985cd0c2fa776af24b4ce841bc266 and is breaking TypeScript builds.

MO screenshot 2025-03-23 at 20 45 52

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -2287,7 +2287,7 @@ export type PublishOptions = {
    * Support any publish options that may be added serverside without needing
    * typings changes.
    */
-  [k: string]: string | number | boolean;
+  [k: string]: string | number | boolean | undefined;
 };
 
 /**


SimonWoolf (approved):

mattheworiordan:
Thanks for approving @SimonWoolf. I need a new release of ably-js with this fix before I can release my work as it's an NPM package, which cannot reference this branch I am afraid. Hoping we can get a release and bump shortly so I can push out my NPM package. 🤞

mschristensen (approved):

VeskeR:
> could you cut a release today?

Sure.

@mattheworiordan are you ok with me merging this PR?

VeskeR:
@mattheworiordan fix is published in [2.6.5](https://github.com/ably/ably-js/releases/tag/2.6.5) ably-js release. Can now install the latest version from npm.

PR #1991, opened 2025-03-26
Change Counter update object to have `amount` property instead of `inc`

VeskeR:
`amount` is more appropriate in this case as the value can be negative in case the counter was decremented.

Resolves [PUB-1540](https://ably.atlassian.net/browse/PUB-1540)

### DIFF
--- a/README.md
+++ b/README.md
@@ -747,11 +747,11 @@ await root.remove('name');
 
 await counter.increment(5);
 // LiveCounter new value: 5
-// LiveCounter update details: { update: { inc: 5 } }
+// LiveCounter update details: { update: { amount: 5 } }
 
 await counter.decrement(2);
 // LiveCounter new value: 3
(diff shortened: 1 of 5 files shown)


mschristensen (approved):

paddybyers:
> `amount` is more appropriate in this case

Not `value`? I'm not a fan of `amount`

VeskeR:
> > > `amount` is more appropriate in this case
> > 
> > 
> > Not `value`? I'm not a fan of `amount`
> 
>  its the amount by which the counter was incremented or decremented, not the current value of the counter

This.

We do use term "value" in the API when you need to get the, well, current value of the counter:
```typescript
export declare interface LiveCounter extends LiveObject<LiveCounterUpdate> {
  /**
   * Returns the current value of the counter.
   */
  value(): number;
}
```

PR #1995, opened 2025-04-09
[PUB-1094] Infer key name of the LiveMap in its direct subscription update object

VeskeR:
Resolves [PUB-1094](https://ably.atlassian.net/browse/PUB-1094)


mschristensen (approved):
Nice, LGTM

PR #1997, opened 2025-04-10
Release 2.6.6

vladvelici:
## Summary by CodeRabbit

- **New Features**
  - Introduced two channel modes enhancing annotation interactions.
  - Enhanced messaging with support for detailed message annotations and streamlined metadata handling.
  
- **Chores**
  - Updated the package version to 2.6.6.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "ably",
-  "version": "2.6.5",
+  "version": "2.6.6",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "ably",
-      "version": "2.6.5",
+      "version": "2.6.6",
       "license": "Apache-2.0",
       "dependencies": {
         "@ably/msgpack-js": "^0.4.0",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "ably",
   "description": "Realtime client library for Ably, the realtime messaging service",
-  "version": "2.6.5",
+  "version": "2.6.6",
   "license": "Apache-2.0",
   "bugs": {
     "url": "https://github.com/ably/ably-js/issues",
--- a/src/platform/react-hooks/src/AblyReactHooks.ts
+++ b/src/platform/react-hooks/src/AblyReactHooks.ts
@@ -12,7 +12,7 @@ export type ChannelNameAndOptions = {
 export type ChannelNameAndAblyId = Pick<ChannelNameAndOptions, 'channelName' | 'ablyId'>;
 export type ChannelParameters = string | ChannelNameAndOptions;
 
-export const version = '2.6.5';
+export const version = '2.6.6';
 
 /**
  * channel options for react-hooks


ttypic (approved):
LGTM

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 This contains only the most important and/or user-facing changes; for a full changelog, see the commit history.
 
+## [2.6.6](https://github.com/ably/ably-js/tree/2.6.6) (2025-04-10)
```
VeskeR, line 5: looks like message annotations is a new _feature_ that adds a bunch of new public API, I think it bump the minor version, so 2.7.0 release.

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
...
+
+- Adds `ANNOTATION_PUBLISH` and `ANNOTATION_SUBSCRIBE` channel modes.
+- Adds support for message annotations via `channel.annotations`.
+- The message action `meta.occupancy` is now renamed to `meta`. Similarly, `MessageActions.META` is now `MessageActions.META`.
```
VeskeR, line 9: >`MessageActions.META` is now `MessageActions.META`.

something is not right here? was it supposed to be
```
`MessageActions.META_OCCUPANCY` is now `MessageActions.META`
```
?

VeskeR: also, isn't `meta.occupancy` -> `meta` technically a breaking change? I can see it was changed in the [annotations PR](https://github.com/ably/ably-js/pull/1953/files#diff-576e9f53335394305aec3a0828db5ee69dd1df7f5cc554c7ba040304edca0817L2466-R2780), but would like to double-check that it won't break things for users if it was in use already

CHANGELOG.md
vladvelici: Will update the version

CHANGELOG.md
SimonWoolf: You're not wrong that this is technically a breaking change.

But the reason I wouldn't worry about it in this case is that `Message.action` was only introduced recently as part of the annotations/materialization work, which isn't yet part of a publicly released feature yet, and isn't added to our [main documentation](https://ably.com/docs/api/realtime-sdk/types#message). So it's just very unlikely that anyone is relying on the action yet. As you say, our inband occupancy events docs still all say to use the message name, which is not changing.

If that wasn't the case then we wouldn't have made this change.

VeskeR: > was only introduced recently as part of the annotations/materialization work, which isn't yet part of a publicly released feature yet, and isn't added to our [main documentation](https://ably.com/docs/api/realtime-sdk/types#message)

great, this is what I suspected happened here.
happy to approve the change in a non-major release then.

CHANGELOG.md
VeskeR: probably better to open a new PR with a new branch `release/2.7.0` so when it's merged we have a clean history in main (don't want to have a merged `release/2.6.6` branch that actually bumps the version to 2.7.0)

the other question about a potential breaking change is resolved, so feel free to open a 2.7.0 PR

VeskeR:
`publish and subscribe annotations` and `get all annotations rest request` tests are failing here with `Error: Aggregation method not allowed for anonymous client publish; allowed methods are: [multiple.v1 total.v1]` error. This wasn't the case for the annotations PR itself, [it passed there](https://github.com/ably/ably-js/actions/runs/14242369187/job/39931523998).
Has anything related to annotations and anonymous clients changed on the realtime side recently?

@SimonWoolf

vladvelici:
> Has anything related to annotations and anonymous clients changed on the realtime side recently?

It's this https://github.com/ably/realtime/pull/7290

vladvelici:
I'll fix the tests in a different branch and after that's done, follow up with a 2.7.0 PR. Closing this one.

PR #1998, opened 2025-04-11
Use identified connection in annotation tests

vladvelici:


vladvelici:
Replaced by https://github.com/ably/ably-js/pull/2001

PR #2013, opened 2025-04-17
[PUB-1634] push: get local channel subscriptions

### LINKED ISSUE #2003: Impossible to get own subscribe Push Channels for deviceId without `push-admin` capability?
It seems that the only way for a client to get the channels is itself subscribed in given them the `push-admin` permission.  At least that is what the doc states: https://ably.com/docs/api/realtime-sdk/push-admin#push-channel-subscriptions

Is this by design or is this a bug? 

If this is by design, that need that every consumer needs to implement a backend endpoint to get the client subscribed channels (per deviceId)?

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-5297) by [Unito](https://www.unito.io)

owenpearson:
- adds `Rest.device.channelSubscriptions` method which uses device auth to get channel subscriptions only for the local device.

NB: the newly added test currently fails with: `action not permitted` and will not pass until https://github.com/ably/realtime/pull/7314 is deployed in sandbox.

resolves #2003 

## Summary by CodeRabbit

- **New Features**
  - Added a method to allow users to list all push channel subscriptions currently active for their device.

- **Tests**
  - Introduced a new test to verify that device push channel subscriptions can be listed and validated.

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -3180,6 +3180,13 @@ export declare interface LocalDevice {
    * A unique device identity token that the device uses to authenticate itself with Ably.
    */
   deviceIdentityToken?: string;
+
+  /**
+   * Retrieves push subscriptions active for the local device.
+   *
+   * @returns a {@link PaginatedResult} object containing an array of {@link PushChannelSubscription} objects for each push channel subscription active for the local device.
+   */
+  listSubscriptions(): Promise<PaginatedResult<PushChannelSubscription>>;
 }
 
 /**
--- a/src/common/lib/client/rest.ts
+++ b/src/common/lib/client/rest.ts
@@ -17,6 +17,7 @@ import { useTokenAuth } from './auth';
 import { RestChannelMixin } from './restchannelmixin';
 import { RestPresenceMixin } from './restpresencemixin';
 import DeviceDetails from '../types/devicedetails';
+import PushChannelSubscription from '../types/pushchannelsubscription';
 
 type BatchResult<T> = API.BatchResult<T>;
 
@@ -44,7 +45,9 @@ export class Rest {
 
   // exposed for plugins but shouldn't be bundled with minimal realtime
   Resource = Resource;
+  PaginatedResource = PaginatedResource;
   DeviceDetails = DeviceDetails;
+  PushChannelSubscription = PushChannelSubscription;
 
   constructor(client: BaseClient) {
     this.client = client;
--- a/src/plugins/push/pushactivation.ts
+++ b/src/plugins/push/pushactivation.ts
@@ -4,8 +4,10 @@ import { ulid } from 'ulid';
 import type { ErrCallback, StandardCallback } from 'common/types/utils';
 import type ErrorInfo from 'common/lib/types/errorinfo';
 import DeviceDetails, { DevicePlatform, DevicePushDetails } from 'common/lib/types/devicedetails';
+import type PushChannelSubscription from 'common/lib/types/pushchannelsubscription';
 import { getW3CPushDeviceDetails } from './getW3CDeviceDetails';
 import type BaseClient from 'common/lib/client/baseclient';
+import type { PaginatedResult } from 'common/lib/client/paginatedresource';
 
 const persistKeys = {
   deviceId: 'ably.push.deviceId',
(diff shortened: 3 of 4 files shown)


distante:
Following. The linked PR does not exist so I do not know is this is still blocked.

ably.d.ts
```diff
@@ -3323,6 +3333,11 @@ export declare class Rest implements RestClient {
   ): Promise<BatchResult<BatchPublishSuccessResult | BatchPublishFailureResult>[]>;
   batchPresence(channels: string[]): Promise<BatchResult<BatchPresenceSuccessResult | BatchPresenceFailureResult>[]>;
   push: Push;
+
+  /**
+   * An optional {@link LocalDevice} object representing the local device if activated as a target for push notifications. This field is only accessible when using the `Push` plugin.
+   */
+  device?: LocalDevice;
```
VeskeR, line 3340: See [RSH8](https://sdk.ably.com/builds/ably/specification/main/features/#RSH8) - `device()` should be a method returning `LocalDevice`. This requires changing it to a method in [baseclient.ts](https://github.com/ably/ably-js/blob/b367509785f8bf8836de50b0c552e4743af4af69/src/common/lib/client/baseclient.ts#L128).

It should also be defined on base `RestClient`/`ReatlimeClient` interfaces, in both `Rest`/`Realtime` classes, and in `modular.d.ts` file in `BaseRest`/`BaseRealtime` classes.

Docstrings then only needs to be written on the interfaces, it will be inferred for classes.

ably.d.ts
```diff
@@ -3323,6 +3333,11 @@ export declare class Rest implements RestClient {
   ): Promise<BatchResult<BatchPublishSuccessResult | BatchPublishFailureResult>[]>;
   batchPresence(channels: string[]): Promise<BatchResult<BatchPresenceSuccessResult | BatchPresenceFailureResult>[]>;
   push: Push;
+
+  /**
+   * An optional {@link LocalDevice} object representing the local device if activated as a target for push notifications. This field is only accessible when using the `Push` plugin.
```
VeskeR, line 3338: Docstring should be taken from https://sdk.ably.com/builds/ably/specification/main/api-docstrings/, in this case:
```
Retrieves a {@link LocalDevice} object that represents the current state of the device as a target for push notifications.
```

ably.d.ts
```diff
@@ -3152,6 +3152,16 @@ export declare interface Push {
   deactivate(deregisterCallback?: DeregisterCallback): Promise<void>;
 }
 
+/**
+ * Represents the local device as activated as a target for push notifications.
```
VeskeR, line 3156:
```suggestion
 * Contains the device identity token and secret of a device. LocalDevice extends {@link DeviceDetails}.
```

Based on https://sdk.ably.com/builds/ably/specification/main/api-docstrings/

ably.d.ts
```diff
@@ -3152,6 +3152,16 @@ export declare interface Push {
...
+ */
+export declare interface LocalDevice {
```
VeskeR, line 3158: Based on [RSH8k](https://sdk.ably.com/builds/ably/specification/main/features/#RSH8k) and https://sdk.ably.com/builds/ably/specification/main/api-docstrings/:
Should extend `DeviceDetails` interface.

Should also have two properties exposed here:
```
/**
 * A unique device identity token used to communicate with APNS or FCM.
 */
deviceIdentityToken?: string;
/**
 * A unique device secret generated by the Ably SDK.
 */
deviceSecret: string;
```

VeskeR: Based on the https://sdk.ably.com/builds/ably/specification/main/features/#RSH8k and the rules for generating those tokens (and quickly looking at the code) it seems to me that `deviceIdentityToken` can be missing initially until device registartion has happened, and `deviceSecret` is always present.

Could you please confirm that @ttypic @umair-ably ?

VeskeR (changes requested):
LGTM function-wise.
Left some comments about the public API we expose to the clients, we should ensure it matches the existing spec

VeskeR:
Will also need to update the spec and API docstrings for the new method.
I'm happy to do it once this PR is finalized and merged.

ably.d.ts
VeskeR: addressed in https://github.com/ably/ably-js/pull/2020

ably.d.ts
VeskeR: addressed in https://github.com/ably/ably-js/pull/2020

ably.d.ts
VeskeR: addressed in https://github.com/ably/ably-js/pull/2020

ably.d.ts
VeskeR: addressed in https://github.com/ably/ably-js/pull/2020.

question re optionallity of `deviceIdentityToken` is still relevant ,moving the question to https://github.com/ably/ably-js/pull/2020

VeskeR:
Confirmed new and existing push tests are passing locally in Chrome by following the next steps (from [slack thread](https://ably-real-time.slack.com/archives/C04KNCEP4RH/p1744904875955309?thread_ts=1744728147.959019&cid=C04KNCEP4RH)):

- on this branch
- run `npm run test:webserver`
- open http://localhost:3000/mocha.html?grep=browser%2Fpush in any modern web browser to run only the web push tests.
- check that the tests pass. web-push relies on localstorage so if all of the push tests are failing, try deleting local storage from your browser in dev tools to see if that fixes it.

(make sure notifications are enabled for localhost)

VeskeR:
@owenpearson had to update the auth method used for device list subscriptions method to send `X-Ably-DeviceToken` header instead of the previous version with `Authentication: Bearer ...`. Based on the [discussion with Mike C](https://ably-real-time.slack.com/archives/C04KNCEP4RH/p1746022625266059?thread_ts=1744728147.959019&cid=C04KNCEP4RH), this seems to be the correct implementation

lawrence-forooghian:
Are we planning to add this to the spec?

VeskeR:
> Are we planning to add this to the spec?

yes, new spec point, docstring and also update the docs to explain the new method

VeskeR (approved):
(need to press approve as github for some reason blocks the merge as it thinks I have unresolved requested changes)

PR #2016, opened 2025-04-24
docs: add a deprecation notice in the README

ttypic:
- Deleted old section about spec version ably-js is targeting (it hasn't been updated and it's not very clear for SDK users anyway)
- Added a deprecation notice


VeskeR:
I does look scary to have a big **Deprecation Notice** badge at the top of the README for the latest version of the library.
My first impression is that the entire library is deprecated, which doesn’t inspire confidence in it.

README.md
```diff
@@ -6,9 +6,12 @@ _[Ably](https://ably.com) is the platform that powers synchronized digital exper
...
 
-This is a JavaScript client library for Ably Realtime.
-
-This library currently targets the [Ably client library features spec](https://www.ably.com/docs/client-lib-development-guide/features/) Version 1.2. You can jump to the '[Known Limitations](#known-limitations)' section to see the features this client library does not yet support or [view our client library SDKs feature support matrix](https://www.ably.com/download/sdk-feature-support-matrix) to see the list of all the available features.
+> [!IMPORTANT]
+>
+> **Deprecation Notice**
+>
+> Versions **1.2.36 and below are deprecated**.
+> Please update to version **1.2.37 or higher** to continue receiving updates, support and important fixes.
```
VeskeR, line 14: don't really like the phrasing here.
if a user updates to `1.2.50` they're not going to realistically receive any updates going forward, we're only committed to providing critical security updates for v1 version

m-hulbert, line 14:
```suggestion
> If you are running a deprecated version then update as soon as possible. Ably recommends updating to the latest version wherever possible. See the [deprecation policy](https://ably.com/docs/platform/deprecate) and [deprecation announcement](https://ably.com/docs/platform/deprecate/protocol-v1) for further info.
```

I'd suggest something like the above to encourage the latest version and direct people to the policy and support if needed.

paddybyers:
I tend to agree that having it right upfront here could be misinterpreted

PR #2020, opened 2025-04-30
RSH8: Add Push device public API

VeskeR:
Resolves comments re exposed public API from https://github.com/ably/ably-js/pull/2013.
See comments for more details.

## Summary by CodeRabbit

- **New Features**
  - Introduced a method to retrieve device information for push notifications, now available across REST and Realtime clients.

- **Refactor**
  - Changed device access from a property to a method call throughout the app for consistency.

- **Tests**
  - Updated tests to use the new device method for accessing device details.


ably.d.ts
```diff
@@ -3152,6 +3162,20 @@ export declare interface Push {
...
 }
 
+/**
+ * Contains the device identity token and secret of a device. LocalDevice extends {@link DeviceDetails}.
+ */
+export declare interface LocalDevice extends DeviceDetails {
+  /**
+   * A unique device identity token used to communicate with APNS or FCM.
+   */
+  deviceIdentityToken?: string;
```
VeskeR, line 3172: Based on the https://sdk.ably.com/builds/ably/specification/main/features/#RSH8k and the rules for generating those tokens (and by quickly looking at the code) it seems to me that `deviceIdentityToken` can be missing initially until device registartion has happened, and `deviceSecret` is always present.
Hence I marked `deviceIdentityToken` as optional here, and `deviceSecret` as always exist.

Could you please confirm that @ttypic @umair-ably ?

ttypic: Yes that's right, `deviceIdentityToken` we get after registration, secret we generate locally, it can be generated lazily when we need it in other SDKs, but it should always be present

ably.d.ts
```diff
@@ -1773,6 +1773,11 @@ export declare interface RestClient {
    * A {@link Push} object.
    */
   push: Push;
+  /**
+   * Retrieves a {@link LocalDevice} object that represents the current state of the device as a target for push notifications.
+   * @returns A {@link LocalDevice} object.
+   */
+  device(): LocalDevice;
```
ttypic, line 1780: What do you think about `get device()` instead? I think it more align with current public API

ably.d.ts
lawrence-forooghian: The spec says that they should be generated when `activate()` is called, https://sdk.ably.com/builds/ably/specification/main/features/#RSH3a2b, i.e. they are both nullable (we should update the IDL to state this, I can do that).

lawrence-forooghian: It would be good to be sure that we actually emit a type that conforms to the public API. If you change the `device()` getter in `BaseClient` to return `LocalDevice & API.LocalDevice` then you get a compilation error due to `formFactor` and `platform`, which unfortunately in the public API are locked to one of a fixed set (not convinced that's right, but let's not fight that now). But even if you change it to return `LocalDevice & Omit<API.LocalDevice, 'formFactor' | 'platform'>`, you still get it complaining about mismatch of the nullability of `deviceSecret`, so I'd suggest looking into that

ably.d.ts
```diff
@@ -3152,6 +3162,20 @@ export declare interface Push {
   deactivate(deregisterCallback?: DeregisterCallback): Promise<void>;
 }
 
+/**
+ * Contains the device identity token and secret of a device. LocalDevice extends {@link DeviceDetails}.
+ */
+export declare interface LocalDevice extends DeviceDetails {
+  /**
+   * A unique device identity token used to communicate with APNS or FCM.
```
lawrence-forooghian, line 3170: This comment doesn't seem right — APNS and FCM have nothing to do with a web `LocalDevice`, and also I don't think this properly describes what this token is for anyway; AFAIK it's for authenticating the device to Ably. Is this from the docstrings in the spec repo? Will take a look at them

ably.d.ts
lawrence-forooghian: >  i.e. they are both nullable (we should update the IDL to state this, I can do that)

https://github.com/ably/specification/pull/314

ably.d.ts
VeskeR: Well, the spec defines it as a _method_ https://sdk.ably.com/builds/ably/specification/main/features/#RSH8.

It's a method in ably-java https://github.com/ably/ably-java/blob/333dc5448a4b16648f9671ae67faf096ff8d5bd9/android/src/main/java/io/ably/lib/rest/AblyRest.java#L46, but a getter property in ably-cocoa https://github.com/ably/ably-cocoa/blob/ebe0e752e8d35ac9cceb0b4f2508675a4d519e15/Source/include/Ably/ARTRest.h#L72.

Is there a reason to deviate from the spec?

ably.d.ts
VeskeR: >  Is this from the docstrings in the spec repo?

Yes, it's from the docstrings in spec repo https://sdk.ably.com/builds/ably/specification/main/api-docstrings/.
And it's the same description as currently used in ably-java https://github.com/ably/ably-java/blob/333dc5448a4b16648f9671ae67faf096ff8d5bd9/android/src/main/java/io/ably/lib/push/LocalDevice.java#L26 and cocoa https://github.com/ably/ably-cocoa/blob/ebe0e752e8d35ac9cceb0b4f2508675a4d519e15/Source/include/Ably/ARTLocalDevice.h#L15 (although cocoa doesn't mention FCM). I think docstring predates the time we added Web Push support, so it only mentions APNS, FCM.

Do we need to change it to _include_ the Web Push API, or replace "APNS and FCM" with Web Push API?
cc @ttypic

ably.d.ts
lawrence-forooghian: Personally I'd say that a getter seems more intuitive here (there's no implicit action performed by fetching the device)

> Is there a reason to deviate from the spec?

I think it's the kind of deviation we frequently make for whatever feels more ergonomic in a given language

(no strong opinions here though, do whatever you feel best)

ably.d.ts
lawrence-forooghian: https://github.com/ably/specification/pull/315

lawrence-forooghian: > Do we need to change it to include the Web Push API, or replace "APNS and FCM" with Web Push API?

neither in my opinion

ably.d.ts
VeskeR: > The spec says that they should be generated when activate() is called, https://sdk.ably.com/builds/ably/specification/main/features/#RSH3a2b, i.e. they are both nullable (we should update the IDL to state this, I can do that).

Not sure about java and cocoa, but in js right now the `id` and `deviceSecret` will be generated when we load the persisted LocalDevice state if they do not exist, see https://github.com/ably/ably-js/blob/953563ba4afb07b6f54ca43822915bf8700f4889/src/plugins/push/pushactivation.ts#L73.
This essentially means that `deviceSecret` will always be present.

lawrence-forooghian: If we're able to make the type more restrictive in JS due to some implementation detail, go for it — but let's make sure that our internal and external types match up (as mentioned in https://github.com/ably/ably-js/pull/2020#discussion_r2068875354) so that if this stops being the case then we realise the public API is no longer correct

ably.d.ts
VeskeR: > there's no implicit action performed by fetching the device

There is actually. Calling `client.device()` delegates the call here https://github.com/ably/ably-js/blob/953563ba4afb07b6f54ca43822915bf8700f4889/src/plugins/push/pushactivation.ts#L56, which loads the device from the storage (localStorage) , does JSON parsing, or if it doesn't exist yet, generates id, deviceSecret and persists it on a storage.
Quite a lot of operations to hide behind the getter property.

ably.d.ts
VeskeR: so even if we ignore the enum clash with `platform` and `formFactor` and make `deviceSecret` optional, there is still is a mismatch for `id` and `push.recipient` fields that are actually coming from `DeviceDetails`, which isn't a part of this PR.
`DeviceDetails` class has all its fields optional in implementation https://github.com/ably/ably-js/blob/953563ba4afb07b6f54ca43822915bf8700f4889/src/common/lib/types/devicedetails.ts#L54-L61, but public API defined them as non-optional already: [id](https://github.com/ably/ably-js/blob/953563ba4afb07b6f54ca43822915bf8700f4889/ably.d.ts#L1223), and [push.recipient](https://github.com/ably/ably-js/blob/953563ba4afb07b6f54ca43822915bf8700f4889/ably.d.ts#L1280).

I don't want to spend time figuring out the implementation details regarding those fields right now, so we can do two things:
1. rely on the existing public API to be correct - that `id`, `push.recipient` are always present on `DeviceDetails` and by extension will be on `LocalDevice`, so just enforce typescript to think that `BaseClient.device(): API.LocalDevice` method has correct types.
2. change `id`, `push.recipient` fields in the public API to be optional to match the internal interface of `DeviceDetails`. Technically this is a breaking change.
3. don't extend the `LocalDevice` interface with `DeviceDetails` in public API for now. Only expose `deviceIdentityToken`
and `deviceSecret` which we can correctly decide if they're optional or not in https://github.com/ably/specification/pull/314

I'd prefer the third option based on our conversation.

lawrence-forooghian: Yeah, it's been a long-standing topic in the SDK team (@ttypic and I were just talking about it again now) that it doesn't make sense for `LocalDevice` to inherit from `DeviceDetails`. I am in favour of option 3. (Maybe with `id`, too?)

So, is the only reason that we've added the `LocalDevice` interface to allow us to have a place to put the `listSubscriptions()` API of #2013?

ably.d.ts
lawrence-forooghian: Fair point

ably.d.ts
lawrence-forooghian: Spec PR is merged

ably.d.ts
VeskeR: > So, is the only reason that we've added the LocalDevice interface to allow us to have a place to put the listSubscriptions() API of https://github.com/ably/ably-js/pull/2013?

Yes, correct.

>  (Maybe with id, too?)
uhh, `id` is one of those mismatch fields, just like `deviceSecret` it's not actually optional.
if you're going to update the spec in https://github.com/ably/specification/pull/314#discussion_r2068984210 to reflect that (it seems based on the comment there you will), then I can add a non-optional `id` field, yes

ably.d.ts
VeskeR: thanks, will update this PR

ably.d.ts
VeskeR: > that it doesn't make sense for LocalDevice to inherit from DeviceDetails

wasn't aware of this topic, it simplifies things a bit here for sure

lawrence-forooghian: ID is meant to be non-optional even on `DeviceDetails`, actually

lawrence-forooghian: But you're saying that our internal `LocalDevice.id` is non-nullable? if so that's fine, it can go in public interface right?

VeskeR: > But you're saying that our internal LocalDevice.id is non-nullable? if so that's fine, it can go in public interface right?

Internal `LocalDevice` extends internal `DeviceDetails`, which has optional `id` field https://github.com/ably/ably-js/blob/953563ba4afb07b6f54ca43822915bf8700f4889/src/common/lib/types/devicedetails.ts#L54, even though the publicly exposed `DeviceDetails` interface has a non-optional `id` field https://github.com/ably/ably-js/blob/953563ba4afb07b6f54ca43822915bf8700f4889/ably.d.ts#L1223.
This has caused one of the TS complaints if you do `LocalDevice & API.LocalDevice`.

Anyway, I'm updating the PR with the option 3, not extending `DeviceDetails` in public API. Will expose `id`, `deviceSecret` and `deviceIdentityToken` in public `LocalDevice` interface

lawrence-forooghian: Sorry, that was meant to be "(_if we are planning to go down the not-inheriting-from-`DeviceDetails` route then_) if so that's fine, it can go in public interface right?"

ably.d.ts
VeskeR: updated to the new docstring in https://github.com/ably/ably-js/compare/7c5c0240a88caac42da73a9f8e777b6a16984833..6d1eab03c74ff0dbabc8231017b20977d52281dc

ably.d.ts
VeskeR: going to leave as a method then

lawrence-forooghian (approved):

ttypic (approved):
LGTM

PR #2021, opened 2025-05-01
Refactor internal `PaginatedResult` API to have non-optional methods

VeskeR:
This changes condition-based function properties in the internal `PaginatedResult` to regular class methods.

The spec [#TG5](https://sdk.ably.com/builds/ably/specification/main/features/#TG5) does not specify if the link to the first page (or current) can be missing in the response and what should be done in such
case, so the new implementation is the best-effort attempt to bring the internal `PaginatedResult` API in line with the public API, while keeping the same function behavior as before.

The new implementation will result in a thrown error if there was no link to the first or current page, just like it would before since those methods would not exist on the class. Essentially this changes only the type of the error thrown: for example, instead of "TypeError: page.current is not a function" the client will get "Error: Missing link to the current page of results".

Users of the client library should not have expected the `.first()` and `.current()` methods to be optional before, as the public definition for `PaginatedResult` in ably.d.ts declares them as regular methods, thus this should not be a breaking change.

This makes internal `PaginatedResult` match the public one in ably.d.ts. This enables union types between internal and public APIs for `PaginatedResult`, which will be needed for the new Push device listSubscriptions API in https://github.com/ably/ably-js/pull/2013.

## Summary by CodeRabbit

- **Refactor**
  - Improved the reliability and consistency of paginated navigation controls, ensuring clearer behavior when navigating between pages of results.
  - Enhanced error handling for pagination actions, providing more informative feedback when navigation links are unavailable.

### DIFF
--- a/src/common/lib/client/paginatedresource.ts
+++ b/src/common/lib/client/paginatedresource.ts
@@ -1,7 +1,7 @@
 import * as Utils from '../util/utils';
 import Logger from '../util/logger';
 import Resource, { ResourceResult } from './resource';
-import { IPartialErrorInfo } from '../types/errorinfo';
+import ErrorInfo, { IPartialErrorInfo } from '../types/errorinfo';
 import BaseClient from './baseclient';
 import { RequestBody, ResponseHeaders } from 'common/types/http';
 import HttpStatusCodes from '../../constants/HttpStatusCodes';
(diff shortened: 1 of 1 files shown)


src/common/lib/client/paginatedresource.ts
```diff
@@ -134,43 +134,52 @@ class PaginatedResource {
...
     this.items = items;
+    this._relParams = relParams;
+  }
+
+  async first(): Promise<PaginatedResult<T>> {
+    if (this.hasFirst()) {
+      return this.get(this._relParams!.first);
+    }
+
+    throw new ErrorInfo('Missing link to the first page of results', 50000, 500);
```
ttypic, line 150: I don't think it should be 500 or any 5xx, we wrap client.request response into paginated result, and server doesn't guarantee that all requests are paginated, e.g. /time is not paginated. It's better to throw 4xx instead

src/common/lib/client/paginatedresource.ts
```diff
@@ -134,43 +134,52 @@ class PaginatedResource {
...
-      };
-      this.isLast = () => {
-        return !this.hasNext?.();
-      };
+  async next(): Promise<PaginatedResult<T> | null> {
+    if (this.hasNext()) {
+      return this.get(this._relParams!.next);
     }
+
+    return null;
```
ttypic, line 166: I prefer consistency across methods: hasFirst()/first(), hasCurrent()/current(), hasNext()/next() 

So if we throw exception if hasFirst()/hasCurrent() are false, I think it will be fair to expect same thing from hasNext(). WDYT?

VeskeR: I've slightly updated the commit message description to clarify that bit:

> The spec [#TG5](https://sdk.ably.com/builds/ably/specification/main/features/#TG5) does not specify if the link to the first page (or current) can be missing in the response and what should be done in such
case, so the new implementation is the best-effort attempt to bring the internal PaginatedResult API in line with the public API, while keeping the same function behavior as before.
>
> The new implementation will result in a thrown error if there was no link to the first or current page, just like it would before since those methods would not exist on the class.

Essentially I tried to stick to the spec. The spec for the #next [TG4](https://sdk.ably.com/builds/ably/specification/main/features/#TG4) explicitly states that we should return a `null` if there is no next page. This was the previous behavior in ably-js, and remains the same in this PR.
The spec for `#current` does not exist at all (but the method exists here and in ably-java at least), and for `#first` it does not specify what to do even there is no page.
I wasn't able to find the consistent answer to this by looking at other SDKs:
- ably-java seems to return `null` in all cases if first/current/next page does not exist https://github.com/ably/ably-java/blob/333dc5448a4b16648f9671ae67faf096ff8d5bd9/lib/src/main/java/io/ably/lib/http/BasePaginatedQuery.java#L103-L135
- ably-dotnet for some reason does not have an implementation for `PaginatedResult`, only for `HttpPaginatedResponse`, which is governed by another spec and states that in situations like this it should return an empty `HttpPaginatedResponse` object https://github.com/ably/ably-dotnet/blob/ad86379466676ef6ed9eaa43dbe53dc3abcc1280/src/IO.Ably.Shared/HttpPaginatedResponse.cs#L125-L133
- ably-go - can't really make sense of it, but it seems that it doesn't make any additional checks and just makes a request and hopes the link exist (so I guess it will result in an error if it doesnt?) https://github.com/ably/ably-go/blob/4596ce4cda88298e84684d0bf6ac5b3845bfdab6/ably/paginated_result.go#L164-L165
- ably-cocoa - even harder to read than go for me, but I _think_ it also just makes a request and will error if page does not exist https://github.com/ably/ably-cocoa/blob/ebe0e752e8d35ac9cceb0b4f2508675a4d519e15/Source/ARTPaginatedResult.m#L121

Basically couldn't find the consistent answer and decided to just keep the same behavior as before: .next() returns `null` per spec, `.first()` and `.current()` will result in an error if called and no page exist (previously method would not exist at all and result in the TypeError).
Additionally, I wasn't comfortable with changing the behavior for `.first()` and `.current()` to now optionally return `null`, as this might be a breaking change in the public API

src/common/lib/client/paginatedresource.ts
VeskeR: Will check how `client.request('/time')` looks right now, it may help answer the `null`/error on missing page [discussion ](https://github.com/ably/ably-js/pull/2021#discussion_r2070006431) below too.

VeskeR: will change to "No link to the first page of results, 40400, 404"

VeskeR: On a side note re `res.request()` method. It wraps the response in `HttpPaginatedResponse` class, which is by the spec [#HP2](https://sdk.ably.com/builds/ably/specification/main/features/#HP2) should override the `next()` and `first()` calls so they return new empty `HttpPaginatedResponse` instance.
This is not the case currently in ably-js, as the internal `HttpPaginatedResponse` simply extends the `PaginatedResult` and suffers from the same `null`/missing link problems.
It feels like the whole paginated response handling in ably-js is a bit outdated. Created for this https://github.com/ably/ably-js/issues/2023

VeskeR: changed to 404 in https://github.com/ably/ably-js/compare/be5e4572867d06398503316819ec5a20009fccd3..45f51b2b6fc1a6e9d3750ca603527411c0996a62

ttypic (approved):

PR #2037, opened 2025-05-23
EDU-1889: Improve pub/sub java script Readme

franrob-projects:
This PR streamlines and refocuses the content of README.md to improve readability, simplify navigation, and make ongoing maintenance easier.

`EDU-1889: Rewrites introduction and overview`

> Replaces marketing-heavy intro with a concise overview of the Ably Pub/Sub SDK. 

`EDU-1889: Rewrites supported platforms section`

> Reorganizes platform support information into a clearer, more concise format. Merges browser compatibility support into a table.

`EDU-1889: Adds getting started`

> Introduces a new "Getting Started" section with quick links for JavaScript, React, and LiveObjects.

`EDU-1889: Removed as covered in docs`

> Removes **install** content already maintained in our docs.

`EDU-1889: Condenses and moves Webpack instructions`

> - Simplifies Webpack setup steps. Consolidates browser and server usage and separates modular and default bundling.
> - Note: Aware that server-side rendering for webpeck is not documented elsewhere.

`EDU-1889: Moves Modular variant instructions`

> Relocates the modular variant details into its own section after install instructions. 

`EDU-1889: Removes outdated TypeScript and NativeScript examples`

> Removes legacy code blocks and GIF demos. 

`EDU-1889: Removes Realtime and REST API usage as covered in docs`

> Removes usage examples for Realtime and REST APIs to prevent (already present) drift from docs. 

`EDU-1889: Removes Push activation as covered in docs`

> Removes usage examples for Push to prevent drift from docs. 

`EDU-1889: Removes LiveObjects as covered in docs`

> - Removes LiveObjects content to prevent a future drift from docs.
> - Link added under getting started to LiveObjects.

`EDU-1889: Removes Delta plugin as covered in docs`

> - Removes Delta plugin as covered in docs.
> - Potential ticket required to update [Delta](https://ably.com/docs/channels/options/deltas#subscribe) section in the docs

`EDU-1889: Moves Contributing and removes Credits`
> - Moves the Contributing section further up the doc
> - Removes unneeded

`EDU-1889: Adds Releases section`
> Introduces a Releases section linking to CHANGELOG and changelog.ably.com for version history and upgrade notes.

`EDU-1889: Condenses Support section`
> Streamlines troubleshooting info and adds direct links to help resources.

**Note:** I have removed the following badges for now. I want to discuss keeping them if needed, removing permanently, or adding new ones!
 
image
 
image


owenpearson (commented):
My initial thoughts on this:

- Visually, this seems like a bit of a downgrade compared to the previous version - the badges and emphasised text looked kind of nice. Admittedly it didn't look great before but now it's pretty much entirely plain unstyled text which I think looks a bit dull. Compare to something like https://github.com/vitejs/vite, which has a big logo, badges, quoted text and emojis right at the start - it looks a lot more flashy, and I think that's something that's worth pursuing.
- I like how you've made collapsible sections for the modular variant docs! we should do the same for the support issues too (chrome extensions, etc)
- The delta plugin is documented on the website but the documentation there is a bit lacking compared to what we have here - for example it doesn't say how to actually install the plugin. At a minimum we should update the website docs to link out to https://github.com/ably-forks/vcdiff-decoder#usage.
- The NPM badge is very common for javascript repos and I find quite useful as a quick way to get to the npmjs.com page for a package, so I would prefer to keep that. I'm happy to remove the features CI badge, but I think badges are generally nice looking and can be useful so I think removing them all seems like a bad idea to me.
- I am happy to remove the bulk of the usage examples from the README but I do think at a minimum we should have a simple installation section and some very basic usage snippets. Personally, I really hate having to go to a company website just to find the most basic API usage or what `npm install` command I need to run - we don't have to have every rest endpoint or whatever but just an `npm install ably` command and a simple pub/sub example here would be less annoying for developers like myself while also making the README more visually appealing.
- The 'supported platforms' section I think could do with some more thought - most of the browser versions are very old, and the only build tool we mention is webpack (while most developers nowadays are using vite, esbuild, etc) so even if it's still true it kinda looks like we haven't updated it in years. I appreciate that this can be changed in a separate PR, but now seems like as good time to do it as ever 🙂 . Here I would suggest just advertising LTS support for web browsers, dropping any mentions of node 16 (we can still link to the package.json 'engines' field which mentions node 16 anyway), and reconsidering how we document usage with build tools. These build tools only really need to know what modules and ES version the package uses so if we document these then users can figure it out themselves and figure it out using the documentation for their build tools. @VeskeR @ttypic wdyt?

README.md
```diff
@@ -1,117 +1,103 @@
...
 
-_[Ably](https://ably.com) is the platform that powers synchronized digital experiences in realtime. Whether attending an event in a virtual venue, receiving realtime financial information, or monitoring live car performance data – consumers simply expect realtime digital experiences as standard. Ably provides a suite of APIs to build, extend, and deliver powerful digital experiences in realtime for more than 250 million devices across 80 countries each month. Organizations like Bloomberg, HubSpot, Verizon, and Hopin depend on Ably’s platform to offload the growing complexity of business-critical realtime data synchronization at global scale. For more information, see the [Ably documentation](https://ably.com/docs)._
+Ably Pub/Sub provides flexible APIs that deliver features such as pub-sub messaging, message history, presence, and push notifications. Utilizing Ably’s realtime messaging platform, applications benefit from its highly performant, reliable, and scalable infrastructure.
 
-[![npm version](https://img.shields.io/npm/v/ably.svg?style=flat)](https://img.shields.io/npm/v/ably.svg?style=flat)
```
ttypic, line 7: Let's keep npm badge at the top of the page under the title

README.md
```diff
@@ -1,117 +1,103 @@
...
 
-## Installation
+The Ably Pub/Sub SDK includes support for Webpack compiling browsers.
 
-### Node.js
+<details>
+<summary>Webpack installation details</summary>
 
-    npm install ably --save
+If you are using a version older than 1.2.5 you will need to add `ably` to `externals` in your Webpack config to exclude it from Webpack processing, and require and use it as an external module using `require('ably')` as above.
```
ttypic: I would delete this paragraph, since version 1.2.5 is considered deprecated

ttypic (commented):
Everything looks good, except badge, I think it's useful to see current npm version

franrob-projects:
> * Visually, this seems like a bit of a downgrade compared to the previous version - the badges and emphasised text looked kind of nice. Admittedly it didn't look great before but now it's pretty much entirely plain unstyled text which I think looks a bit dull. Compare to something like https://github.com/vitejs/vite, which has a big logo, badges, quoted text and emojis right at the start - it looks a lot more flashy, and I think that's something that's worth pursuing.

- Graphic headers have been added here for visual appeal, as they have with all of the SDK repos
- Only relevant badges have been added. The same badges types have been applied to all of the SDK repos (version and license).

> * I like how you've made collapsible sections for the modular variant docs! we should do the same for the support issues too (chrome extensions, etc)

- Collapsible sections have now been added to all longer pieces of info.

> * The delta plugin is documented on the website but the documentation there is a bit lacking compared to what we have here - for example it doesn't say how to actually install the plugin. At a minimum we should update the website docs to link out to https://github.com/ably-forks/vcdiff-decoder#usage.

- The JS Delta docs have been updated with information referenced mainly in [vcdiff-decoder documentation](https://github.com/ably-forks/vcdiff-decoder#usage). The Delta section removed for this repo was not entirely different from what was already in the doc.

> * The NPM badge is very common for javascript repos and I find quite useful as a quick way to get to the npmjs.com page for a package, so I would prefer to keep that. I'm happy to remove the features CI badge, but I think badges are generally nice looking and can be useful so I think removing them all seems like a bad idea to me.

- Only relevant badges have been added. The same badges types have been applied to all of the SDK repos (version and license).

> * I am happy to remove the bulk of the usage examples from the README but I do think at a minimum we should have a simple installation section and some very basic usage snippets. Personally, I really hate having to go to a company website just to find the most basic API usage or what `npm install` command I need to run - we don't have to have every rest endpoint or whatever but just an `npm install ably` command and a simple pub/sub example here would be less annoying for developers like myself while also making the README more visually appealing.

- Adds refined installation and usage sections.

> * The 'supported platforms' section I think could do with some more thought - most of the browser versions are very old, and the only build tool we mention is webpack (while most developers nowadays are using vite, esbuild, etc) so even if it's still true it kinda looks like we haven't updated it in years. I appreciate that this can be changed in a separate PR, but now seems like as good time to do it as ever 🙂 . Here I would suggest just advertising LTS support for web browsers, dropping any mentions of node 16 (we can still link to the package.json 'engines' field which mentions node 16 anyway), and reconsidering how we document usage with build tools. These build tools only really need to know what modules and ES version the package uses so if we document these then users can figure it out themselves and figure it out using the documentation for their build tools. @VeskeR @ttypic wdyt?

- On the supported platforms section, LTS support has been added to all browsers.
- For Node.js, `>=16.x or later` has been dropped.

README.md
```diff
@@ -1,121 +1,144 @@
...
+The following browser versions are supported:
 
-We regression-test the library against a selection of Node.js versions, which will change over time. We will always support and test against current LTS Node.js versions, and optionally some older versions that are still supported by upstream dependencies. We reserve the right to drop support for non-LTS versions in a non-major release. We will update the `engines` field in [package.json](./package.json) whenever we change the Node.js versions supported by the project. Please refer to [the test-node GitHub workflow](./.github/workflows/test-node.yml) for the set of versions that currently undergo CI testing.
+| Browser | Support |
+|---------|---------|
+| Chrome | Long term support. |
+| Firefox | Long term support. |
+| Edge | Long term support. |
+| Safari | Long term support. |
+| Opera | Long term support. |
```
ttypic: I would simply write that we support all modern browsers. This table looks a bit odd

ttypic (approved):
Looks good for me, added small comment about browser support table

README.md
```diff
@@ -1,121 +1,134 @@
...
+// or, for ES6/TypeScript:
+import * as Ably from 'ably';
 ```
 
-The Ably client library follows [Semantic Versioning](http://semver.org/). To lock into a major or minor version of the client library, you can specify a specific version number such as https://cdn.ably.com/lib/ably.min-2.js for all v2._ versions, or https://cdn.ably.com/lib/ably.min-2.6.js for all v2.6._ versions, or you can lock into a single release with https://cdn.ably.com/lib/ably.min-2.6.3.js. Note you can load the non-minified version by omitting `.min` from the URL such as https://cdn.ably.com/lib/ably-2.js. See https://github.com/ably/ably-js/tags for a list of tagged releases.
+With `target: 'browser'`, Webpack uses the browser-compatible CommonJS build by default.
+
+If needed (for example with custom targets):
+
+- Webpack 5: To resolve Ably correctly, add this alias to your webpack configuration:
```
owenpearson: not needed in this PR but someone should revisit this webpack section, i don't believe you would actually need to do any of this since ably-js v2

README.md
```diff
@@ -1,121 +1,134 @@
...
 
-(This applies to using Webpack to compile for a browser; for Node.js, see [Serverside usage with Webpack](#serverside-usage-with-webpack))
+## Usage
 
-Webpack will search your `node_modules` folder by default, so if you include `ably` in your `package.json` file, when running Webpack the following will allow you to `require('ably')` (or if using typescript or ES6 modules, `import * as Ably from 'ably';`). If your Webpack target is set to 'browser', this will automatically use the browser commonjs distribution.
+The following code connects to Ably's realtime messaging service, subscribes to a channel to receive messages, and publishes a test message to that same channel.
 
-If that doesn't work for some reason (e.g. you are using a custom Webpack target), you can use one of the solutions below depending on your Webpack version:
+```javascript
+const realtimeClient = new Ably.Realtime({ key: '', clientId: 'a-client' });
```
owenpearson: I think the string examples we have here look a little odd, would suggest the following:
```suggestion
const realtimeClient = new Ably.Realtime({ key: 'your-ably-api-key', clientId: 'me' });
```

README.md
```diff
@@ -1,121 +1,134 @@
...
-            alias: {
-              ably: path.resolve(__dirname, 'node_modules/ably/build/ably.js'),
-            },
-          },
-        },
-      ],
-    },
-  };
-  ```
+const channel = realtimeClient.channels.get('a-channel');
```
owenpearson:
```suggestion
const channel = realtimeClient.channels.get('test-channel');
```

README.md
```diff
@@ -1,121 +1,134 @@
...
+
+await channel.subscribe((message) => {
+  console.log(`Received message: ${message.data}`);
+});
+
+await channel.publish('example', 'a-message');
```
owenpearson:
```suggestion
await channel.publish('test-event', 'hello world');
```

README.md
```diff
@@ -1,121 +1,134 @@
...
+```sh
+npm install ably
+```
 
-and require as:
+Run the following to instantiate a client:
 
 ```javascript
-var Ably = require('ably');
+const Ably = require('ably');
```
owenpearson: Definitely think we should default to ES6 imports at this point:
```suggestion
import * as Ably from 'ably';
```

README.md
```diff
@@ -1,121 +1,134 @@
...
+> [!IMPORTANT]
+> SDK versions < 1.2.36 will be [deprecated](https://ably.com/docs/platform/deprecate/protocol-v1) from November 1, 2025.
 
-For complete API documentation, see the [Ably documentation](https://www.ably.com/docs).
+---
 
 ## Installation
 
-### Node.js
+The NodeJS SDK is available as an [NPM module](https://www.npmjs.com/package/ably). To get started with your project, install the package:
```
owenpearson: generally most users will use NPM regardless of whether they're using nodejs or not, so would just say:
```suggestion
The JavaScript SDK is available as an [NPM module](https://www.npmjs.com/package/ably). To get started with your project, install the package:
```

README.md
```diff
@@ -1,121 +1,134 @@
...
-Webpack will search your `node_modules` folder by default, so if you include `ably` in your `package.json` file, when running Webpack the following will allow you to `require('ably')` (or if using typescript or ES6 modules, `import * as Ably from 'ably';`). If your Webpack target is set to 'browser', this will automatically use the browser commonjs distribution.
+The following code connects to Ably's realtime messaging service, subscribes to a channel to receive messages, and publishes a test message to that same channel.
 
-If that doesn't work for some reason (e.g. you are using a custom Webpack target), you can use one of the solutions below depending on your Webpack version:
+```javascript
+const realtimeClient = new Ably.Realtime({ key: '', clientId: 'a-client' });
 
-- for Webpack 5: you can use alias setting in the Webpack config like this:
+await realtimeClient.connection.once('connected');
+  console.log(`Connected to Ably`);
```
owenpearson: there's a tab at the start of this line
```suggestion
console.log(`Connected to Ably`);
```

README.md
owenpearson: Also I think this would look a bit nicer with comments, eg:

```typescript
// Create a new realtime client
const realtimeClient = new Ably.Realtime({ key: 'your-ably-api-key', clientId: 'me' });

// Wait for the client to become connected
await realtimeClient.connection.once('connected');
console.log('Connected to Ably');

// Get a reference to the 'test-channel' channel
const channel = realtimeClient.channels.get('test-channel');

// Subscribe a listener to fire on all events on the channel
await channel.subscribe((message) => {
  console.log(`Received message: ${message.data}`);
});

// Publish a test message to the channel
await channel.publish('test-event', 'Hello, world!');
```

README.md
owenpearson: Also the string is using backticks but isn't using any template string features
```suggestion
  console.log('Connected to Ably');
```

owenpearson (changes requested):
It's looking good overall now! A few minor comments:

PR #2043, opened 2025-06-19
Change the name for the `Objects` class to `RealtimeObjects`

VeskeR:
This aligns `Objects` with the naming convention for other realtime features in the SDK and matches the spec https://github.com/ably/specification/pull/332.

This is a public API breaking change, so PR will be merged into the integration branch for Objects API improvements that may introduce other API changes

## Summary by CodeRabbit

- **Refactor**
  - Public API and types renamed from "Objects" to "RealtimeObjects" across the library and plugin surface.

- **Documentation**
  - Docs, inline references, and property/docs text updated to use "RealtimeObjects".

- **Tests**
  - Test suites and assertions updated to target "RealtimeObjects" instead of "Objects".

- **Chores**
  - Build/config checks and private API tracking updated to reflect the renamed public surface.


sacOO7 (approved):
lgtm

PR #2044, opened 2025-06-20
Objects spec tags: ObjectMessage and sync sequence

VeskeR:
Based on the spec added in:
https://github.com/ably/specification/pull/298
https://github.com/ably/specification/pull/327
https://github.com/ably/specification/pull/328
https://github.com/ably/specification/pull/332
https://github.com/ably/specification/pull/335
https://github.com/ably/specification/pull/333

## Summary by CodeRabbit

- **Documentation**
  - Added detailed specification and inline comments throughout various modules to improve code clarity and traceability to formal specifications.
  - Enhanced documentation for classes, interfaces, enums, and methods related to live objects, maps, counters, object messages, and synchronization logic.
  - Updated and corrected spec identifiers in documentation comments for improved accuracy.
  - No changes to application behavior or user-facing functionality.

### DIFF
--- a/src/common/lib/client/realtimechannel.ts
+++ b/src/common/lib/client/realtimechannel.ts
@@ -149,11 +149,12 @@ class RealtimeChannel extends EventEmitter {
     return this._push;
   }
 
+  /** @spec RTL27 */
   get objects() {
     if (!this._objects) {
-      Utils.throwMissingPluginError('Objects');
+      Utils.throwMissingPluginError('Objects'); // RTL27b
     }
-    return this._objects;
+    return this._objects; // RTL27a
   }
 
   invalidStateError(): ErrorInfo {
--- a/src/common/lib/client/realtimechannel.ts
+++ b/src/common/lib/client/realtimechannel.ts
@@ -614,6 +614,7 @@ class RealtimeChannel extends EventEmitter {
         break;
       }
 
+      // RTL1
       // OBJECT and OBJECT_SYNC message processing share most of the logic, so group them together
       case actions.OBJECT:
       case actions.OBJECT_SYNC: {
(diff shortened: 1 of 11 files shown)


src/common/lib/client/realtimechannel.ts
```diff
@@ -149,11 +149,12 @@ class RealtimeChannel extends EventEmitter {
     return this._push;
   }
 
+  /** @spec RTL27 */
```
lawrence-forooghian, line 152: I think that this notation is new to the codebase (elsewhere we just mention spec point identifiers without any special notation). Is there any particular reasoning behind the notation?

lawrence-forooghian (commented):
I am assuming that the spec point identifiers are correct; I don't intend to check them!

Are you planning to add `@spec` etc tags to the tests?

sacOO7 (commented):
Currently trying to go through each spec point and relevant ably-js code. Will review/approve by end of this week.

src/common/lib/client/realtimechannel.ts
VeskeR: It felt a bit odd to just add the spec item as inline text, and we've used `@spec` notation in tests when spec point is part of the property/method docsting, so I thought I'd use it here.
Essentially something like this didn't look right:
```
  /**
   * RTO1 - When called without a type variable, we return a default root type which is based on globally defined interface for Objects feature.
   * A user can provide an explicit type for the getRoot method to explicitly set the type structure on this particular channel.
   * This is useful when working with multiple channels with different underlying data structure.
   * RTO1 (or here)
   */
  async getRoot<T extends API.LiveMapType = API.DefaultRoot>(): Promise<LiveMap<T>> {
```

while a dedicated docstring tag highlights the spec item and makes it easier to read IMO

lawrence-forooghian: Alrighty 👍

lawrence-forooghian (approved):

PR #2051, opened 2025-06-26
[PUB-1672] Refactor encoding/decoding for `ObjectMessage`

VeskeR:
Use `WireObjectMessage` to represent the message received from the server which can be decoded into an `ObjectMessage`. `ObjectMessage` can then be encoded into a `WireObjectMessage`.
This matches the style used for messages/presencemessages/annotationmessages which was introduced in https://github.com/ably/ably-js/pull/1941 (see [Simon's comment](https://github.com/ably/ably-js/pull/1941#issuecomment-2580956182) to see changes in the PR, as it bugged out and "Files changed" doesn't work for some reason)

## Summary by CodeRabbit

* **New Features**
  * Improved support for wire-format object messages, enabling more robust encoding and decoding of data exchanged with the server.

* **Refactor**
  * Enhanced type safety and clarity by introducing generic typing for object operations, states, and map entries.
  * Separated raw (wire-format) and decoded object message representations for clearer data handling.
  * Updated internal message handling to use encoded wire-format messages for transmission.
  * Streamlined message encoding and decoding processes, including base64 handling for JSON protocol.

* **Style**
  * Refined type annotations across various components for improved code maintainability and correctness.

### DIFF
--- a/src/plugins/objects/objectmessage.ts
+++ b/src/plugins/objects/objectmessage.ts
@@ -1,6 +1,5 @@
 import type BaseClient from 'common/lib/client/baseclient';
 import type { MessageEncoding } from 'common/lib/types/basemessage';
-import type Logger from 'common/lib/util/logger';
 import type * as Utils from 'common/lib/util/utils';
 import type { Bufferlike } from 'common/platform';
 
@@ -9,7 +8,7 @@ export type EncodeInitialValueFunction = (
   encoding?: string | null,
 ) => { data: any; encoding?: string | null };
 
-export type EncodeObjectDataFunction = (data: ObjectData) => ObjectData;
+export type EncodeObjectDataFunction = (data: ObjectData | WireObjectData) => WireObjectData;
 
 /** @spec OOP2 */
 export enum ObjectOperationAction {
@@ -27,7 +26,7 @@ export enum ObjectsMapSemantics {
 }
 
 /**
- * An ObjectData represents a value in an object on a channel.
+ * An ObjectData represents a value in an object on a channel decoded from {@link WireObjectData}.
  * @spec OD1
  */
 export interface ObjectData {
@@ -47,14 +46,34 @@ export interface ObjectData {
 }
 
 /**
+ * A WireObjectData represents a value in an object on a channel received from the server.
+ * @spec OD1
+ */
+export interface WireObjectData {
+  /** A reference to another object, used to support composable object structures. */
+  objectId?: string; // OD2a
+
+  /** May be set by the client to indicate that value in `string` field have an encoding. */
+  encoding?: string; // OD2b
+  /** A primitive boolean leaf value in the object graph. Only one value field can be set. */
+  boolean?: boolean; // OD2c
+  /** A primitive binary leaf value in the object graph. Only one value field can be set. Represented as a Base64-encoded string in JSON protocol */
+  bytes?: Bufferlike | string; // OD2d
+  /** A primitive number leaf value in the object graph. Only one value field can be set. */
+  number?: number; // OD2e
+  /** A primitive string leaf value in the object graph. Only one value field can be set. */
+  string?: string; // OD2f
+}
+
+/**
  * An ObjectsMapOp describes an operation to be applied to a Map object.
  * @spec OMO1
  */
-export interface ObjectsMapOp {
+export interface ObjectsMapOp<TData> {
   /** The key of the map entry to which the operation should be applied. */
   key: string; // OMO2a
   /** The data that the map entry should contain if the operation is a MAP_SET operation. */
-  data?: ObjectData; // OMO2b
+  data?: TData; // OMO2b
 }
 
 /**
@@ -70,7 +89,7 @@ export interface ObjectsCounterOp {
  * An ObjectsMapEntry represents the value at a given key in a Map object.
  * @spec OME1
  */
-export interface ObjectsMapEntry {
+export interface ObjectsMapEntry<TData> {
   /** Indicates whether the map entry has been removed. */
   tombstone?: boolean; // OME2a
   /**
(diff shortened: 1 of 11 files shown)


src/plugins/objects/objectmessage.ts
```diff
@@ -159,24 +178,179 @@ export interface ObjectState {
...
+  result += ']';
+  return result;
+}
+
 /**
- * An individual object message to be sent or received via the Ably Realtime service.
+ * Deep copy public properties of object message.
+ * Buffer values won't be correctly copied, use the original message instead if you need buffer values.
+ */
+function copyMsg(
```
mschristensen, line 338: **suggestion**: would it be possible to define a buffer util to correctly clone a buffer for each of the platform implementations (e.g. Buffer [from](https://nodejs.org/api/buffer.html#static-method-bufferfromarray) or ArrayBuffer [slice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/slice)) and use that here? Seems like this could be quite bug prone otherwise.

mschristensen (approved):
Looks good to me

src/plugins/objects/objectmessage.ts
```diff
@@ -46,15 +45,35 @@ export interface ObjectData {
...
+  /** A primitive string leaf value in the object graph. Only one value field can be set. */
+  string?: string; // OD2f
+}
+
 /**
  * An ObjectsMapOp describes an operation to be applied to a Map object.
  * @spec OMO1
  */
-export interface ObjectsMapOp {
+export interface ObjectsMapOp<T> {
```
lawrence-forooghian: Can we make the name of this generic type parameter more clearly communicate its purpose — e.g. `Data`? Ditto throughout.

src/plugins/objects/objectmessage.ts
```diff
@@ -706,4 +750,63 @@ export class ObjectMessage {
...
+    } catch (error) {
+      client.Logger.logAction(
+        client.logger,
+        client.Logger.LOG_ERROR,
+        'WireObjectMessage._decodeObjectData()',
+        this._utils.inspectError(error),
+      );
+      return {
+        ...objectData,
+      } as ObjectData;
```
lawrence-forooghian, line 823: There seem to be quite a few type assertions (`x as y`) in this file. It's not clear to me at a glance whether they're needed or whether they're just a shortcut to allow us to write less code. Do you have any thoughts on this? The less we can bypass compiler checking, the better.

src/plugins/objects/objectmessage.ts
```diff
@@ -159,24 +178,179 @@ export interface ObjectState {
...
+function encode(
+  message: Utils.Properties<ObjectMessage> | Utils.Properties<WireObjectMessage>,
+  utils: typeof Utils,
+  messageEncoding: typeof MessageEncoding,
+  encodeObjectDataFn: EncodeObjectDataFunction,
+  encodeInitialValueFn: EncodeInitialValueFunction,
+): WireObjectMessage {
+  // deep copy the message to avoid mutating the original one.
+  // buffer values won't be correctly copied, so we will need to use the original message when encoding.
+  const result: WireObjectMessage = Object.assign(new WireObjectMessage(utils, messageEncoding), copyMsg(message));
```
lawrence-forooghian: I have a similar question re this as for the type assertions that I asked about elsewhere — it feels like we're bypassing some type checking that the compiler could do for us in order to allow us to write less code; is this the case?

src/plugins/objects/objectmessage.ts
VeskeR: type assertions used in this file:
1. in the `.reduce()` call for the initial value: `.reduce(() => {...}, {} as something)`. I believe this is a usual practice to use type assertions for initial values in `.reduce()` calls, as otherwise TS complier can't infer the type of the object. The other option would be to use the template argument `.reduce<something>(() => {...}, {})`, but it's essentially the same thing
2. type assertion for `initialValue` in `encodeInitialValue`. I've added the comment that explains why we do that there: https://github.com/ably/ably-js/pull/2051/files#diff-f4703a5f5cfd6019b6628a3042165921467f8e35630916f96e45127031ebac5aR277-R283. tldr: to write less code in this case.
3. `decodedBytes = objectData.bytes as Bufferlike` in `_decodeObjectData`. This one has a [comment](https://github.com/ably/ably-js/pull/2051/files#diff-f4703a5f5cfd6019b6628a3042165921467f8e35630916f96e45127031ebac5aR796) already. `bytes` field received from realtime is either a string when json protocol is used, or bytes and msgpack is used. So we just cast to the appropriate type based on the protocol used.
4.  `{...objectData} as ObjectData` - added a [small comment](https://github.com/ably/ably-js/pull/2051/files#diff-f4703a5f5cfd6019b6628a3042165921467f8e35630916f96e45127031ebac5aR815), but essentially if data decoding has failed, we follow the same concept used for regular messages - we return the value as is without finishing all the decoding steps. Regular messages conveniently use `any` type for `data` field, so it's never a problem there, but ObjectMessages have stricter types with their typed properties, so we need to cast WireObjectData to ObjectData as type for the `bytes` property is mismatched. I think this can be improved in [another PR](https://github.com/ably/ably-js/pull/2052/files#diff-f4703a5f5cfd6019b6628a3042165921467f8e35630916f96e45127031ebac5aR771) which adds support for JSON object value for LiveMap and changes `ObjectData` interface to just have a single `ObjectData.value` field - we can just set the undecoded value to the `ObjectData.value` field and remove the `as ObjectData` casting

src/plugins/objects/objectmessage.ts
VeskeR: oh, `const result: WireObjectMessage` type assertion is actually not needed here (and in one more place that does Object.assign). The correct type is already inferred from the Object.assign call. I've just removed the assertion from those two places

src/plugins/objects/objectmessage.ts
VeskeR: Sure, did in https://github.com/ably/ably-js/compare/516accab8890cb73c6881e4af3929f5ee089835b..5c25961f11b9bccc6b0b4c435886130091b87c65

src/plugins/objects/objectmessage.ts
VeskeR: I had a go at adding the `BufferUtils.copy` method and then updating the `copyMsg` function to also copy buffer values. I think it actually made things more complicated. Now we would need to iterate over every map entry and map operation in the `object` and `operation` fields within the `copyMsg` function as well (which we’re already doing in the `decode` and `encode` calls anyway and it's not going away). Additionally, the function becomes dependent on the core library's `Platform` class to get access to `BufferUtils`, which would need to be passed around a lot and added to the `ObjectMessage` and `WireObjectMessage` constructors.

Essentially, this just introduced more complexity and additional lines of code without providing much benefit. The `copyMsg` function isn’t even exported - it's used only within this file in two specific cases - and it already clearly states that buffer values are not copied. I don’t think there’s much value in trying to handle buffer copying here.

I've added a bit more detais to the function's jsdoc to clarify what's going on: https://github.com/ably/ably-js/compare/7d2e10530771b99cc5861864639dc176b006df03..3a332be1bad7f7afa04a07bb78fceb6b4e6c8bd8

lawrence-forooghian (approved):

Issue #2057, opened 2025-07-09
Ably Real-time Connection Fails on Vercel Deployment but Works Locally


0xdecapitator:
**Issue Summary**

I'm experiencing a persistent issue where my Ably real-time connection works perfectly during local development but consistently fails when deployed to Vercel. I've created a minimal reproduction of the setup to isolate and demonstrate the problem.

**Link to Minimal Reproduction**

You can find the complete source code and setup instructions in this repository:
[**github-repo: ably-realtime-debug**](https://github.com/0xdecapitator/ably-realtime-debug)

The repository includes a `/debug` page with automated connection tests to help diagnose the issue.

**Steps to Reproduce**

The issue can be observed by following these steps:

1.  **Clone the repository and install dependencies:**
    ```bash
    git clone [your-repo-link]
    cd [repo-name]
    bun install
    ```
2.  **Set up environment variables:**
    - Create a `.env.local` file.
    - Add your `ABLY_API_KEY`.
    ```env
    ABLY_API_KEY=your_ably_api_key
    ```
3.  **Run Locally (Works as expected ✓):**
    - Run `bun dev`.
    - Open multiple browser tabs and join the same room.
    - Observe that presence updates, position synchronization, and other real-time features work flawlessly across all tabs.

4.  **Deploy to Vercel (Fails ✗):**
    - Deploy the application to Vercel using the `vercel` CLI or by linking the repository.
    - **Ensure the `ABLY_API_KEY` is added to the project's environment variables in the Vercel dashboard.**
    - Access the deployed application.
    - Repeat the same testing steps as the local environment.
    - Observe connection failures, missing real-time updates, and stalled presence information.

**Expected Behavior**

When deployed to Vercel, the application should function identically to the local version:
* Players can join any of the 5 game rooms.
* Real-time position updates are synchronized instantly.
* The presence set correctly reflects all players in a room.
* The connection manager pools and reuses connections efficiently.

**Actual Behavior**

On the Vercel deployment, one or more of the following issues occur:
* WebSocket connections fail to establish, sometimes without falling back to XHR polling.
* Real-time messages (like position updates) are not received by other clients.
* Presence updates are delayed or missed entirely.
* The debug page and console logs show connection state errors.

**Technical Architecture**

The minimal reproduction uses the following architecture:
* **Connection Management**: A custom `AblyConnectionManager` handles connection pooling with a maximum of 20 clients to avoid hitting Ably's soft limit.
[...]

PR #2059, opened 2025-07-14
fix: minimization issue with Next.js 15+

ttypic:
Related to https://github.com/ably/ably-js/issues/2057

SWC that Next.js uses for minimization, inlines functions inside `XHRRequest` class, that results in losing `this` context and broken code.

Unfortunately there is no easy way to fix this on the Next.js 15+ side:

> Starting with v15, minification cannot be customized using `next.config.js`. 

This PR fixes a single immediate issue. Another problem with other minimizers may appear soon. I’ll create a RFC with code practices we should adopt to avoid this in the future.

## Summary by CodeRabbit

* **Refactor**
  * Improved the handling of HTTP request state changes for better consistency and reliability.

### DIFF
--- a/src/platform/web/lib/http/request/xhrrequest.ts
+++ b/src/platform/web/lib/http/request/xhrrequest.ts
@@ -183,7 +183,7 @@ class XHRRequest extends EventEmitter implements IXHRRequest {
       Logger.logAction(this.logger, Logger.LOG_ERROR, 'Request.on' + errorEvent.type + '()', errorMessage);
       this.complete(new PartialErrorInfo(errorMessage, code, statusCode));
     };
-    xhr.onerror = function (errorEvent) {
+    xhr.onerror = (errorEvent) => {
       errorHandler(errorEvent, 'XHR error occurred', null, 400);
     };
     xhr.onabort = (errorEvent) => {
@@ -193,7 +193,7 @@ class XHRRequest extends EventEmitter implements IXHRRequest {
         errorHandler(errorEvent, 'Request cancelled', null, 400);
       }
     };
-    xhr.ontimeout = function (errorEvent) {
+    xhr.ontimeout = (errorEvent) => {
       errorHandler(errorEvent, 'Request timed out', null, 408);
     };
 
@@ -306,7 +306,7 @@ class XHRRequest extends EventEmitter implements IXHRRequest {
       });
     };
 
-    xhr.onreadystatechange = function () {
+    xhr.onreadystatechange = () => {
       const readyState = xhr.readyState;
       if (readyState < 3) return;
       if (xhr.status !== 0) {


src/platform/web/lib/http/request/xhrrequest.ts
```diff
@@ -306,7 +306,7 @@ class XHRRequest extends EventEmitter implements IXHRRequest {
       });
     };
 
-    xhr.onreadystatechange = function () {
+    xhr.onreadystatechange = () => {
```
VeskeR, line 309: I'm not sure I understand what the immediate problem is here. Sure, using arrow functions instead of function expressions is generally better, but I don't immediately see what's failing. The PR description mentions:
> that results in losing `this` context

but `onreadystatechange` does not call `this`, it only calls other functions that were defined as arrow functions so `this` should be preserved for them.

VeskeR, line 309: also, if function expressions are a problem in SWC minimization, shouldn't we also change `xhr.onerror` and `xhr.ontimeout` assignments in this file to also use arrow functions just to be safe?

ttypic: when it got minimized and `onResponse` got inlined it starts using `this`

VeskeR: oh, based on the RFC you just posted https://ably.atlassian.net/wiki/spaces/SDK/pages/4172283949/RFC+Avoid+Mixing+Regular+and+Arrow+Functions+in+JS+SDKs, I can imagine that the problem is exactly because function expressions and arrow functions are mixed here:
> My tests show that the problem appears only when regular function is mixing with arrow function. Arrow function got inlined and this context starts missing

So the fact that we call arrow functions like `onResponse` from function expression at `onreadystatechange` is what causing the trouble. Would like to see the result of minification anyway to get a clearer picture

VeskeR: > when it god minimized and onResponse got inlined it starts using this

yea, got it

VeskeR: let's update the rest of function expressions in this file too then, as per comment below

ttypic: I looked into minimized code and 

```js
a.onerror=function(e){u(e,"XHR error occurred",null,400)}
``` 
and
```js
a.ontimeout=function(e){u(e,"Request timed out",null,408)}
```
it's nothing to inline here. On other hand, here is `onreadystatechange`:

```js
a.onreadystatechange=function(){let i=a.readyState;i<3||0!==a.status&&(void 0===t&&(t=a.status,(()=>{var s;if(clearTimeout(r),n=t<400,204==t)return this.complete(null,null,null,null,t)...
```

To fix this properly, we need to decide on a strategy for all function declarations across all SDKs. I think it would be good to release this fix first so SDK users don’t have to wait. This definitely doesn’t introduce any new problems but does fix some existing ones.

VeskeR: > I looked into minimized code and
a.onerror=function(e){u(e,"XHR error occurred",null,400)}
and
a.ontimeout=function(e){u(e,"Request timed out",null,408)}

huh, so there are multiple places which call `errorHandler` so it doesn't get inlined and as a result `xhr.onerror` and `xhr.ontimeout` don't break

VeskeR: I mean, changing `xhr.onerror` and `xhr.ontimeout` to arrow functions is trivial but if you'd like to keep this PR specific and fix the immediate problem with `xhr.onreadystatechange` then sure. Will approve the PR as it fixes the problem

VeskeR (approved):

PR #2070, opened 2025-08-15
Web: use Promise.resolve().then(...) for nextTick in the web distribution rather than setTimeout(..., 0)

SimonWoolf:
`setImmediate` is not supported in any modern browsers.

`setTimeout(..., 0)` has issues with Chrome background tab throttling, which gives a minimum timer duration, leading to races.

`Promise.resolve().then(...)` is closer to process.nextTick() in node than setImmediate in that it adds it to the task queue in the current tick rather than the next one, but that's fine, after all we use nextTick in node.


VeskeR (approved):
LGTM.

I'd suggest we also change the react native config to use Promise.resolve().then() at 
https://github.com/ably/ably-js/blob/c923b2e66904604af80e0e76690351861fe469e1/src/platform/react-native/config.ts#L21-L23

SimonWoolf:
Updated the pr to use queueMicrotask where available. (fallback calculated outside rather than inside the closure of course, that coderabbit code is terrible). I'm not worried about speed (it's all microtasks), but it's neater than the promise way, which is ultimately a hack.

Also not worried about semantics of uncaught exception vs unhandled rejection, if either happens the lib would be equally failing the guarantees it should be upholding.

(Kept the promise fallback just to be on the safe side; evergreen browsers all support queueMicrotask but we want to release this in a patch release so worth making sure it doesn't break the lib for anyone).

@AlexanderArvidsson thanks for your comments and for reporting the bug!

PR #2078, opened 2025-09-09
Add clipped field to annotation summary types

vladvelici:
Add `clipped` and `totalClientIds` fields to summary types. https://ably.atlassian.net/browse/CHA-1100

https://ably.atlassian.net/browse/CHA-1178

## Summary by CodeRabbit

- **New Features**
  - Summaries now include explicit clipping indicators and a total distinct-client count so truncated results are clearly reported.

- **Behavior**
  - Summary entries default clipping indicators to false when omitted, ensuring consistent summary output for reactions and votes.

- **Documentation**
  - Field descriptions clarified to explain clipping semantics and total distinct clientIds.

- **Tests**
  - Added unit tests verifying clipped defaults and preservation of explicit clipped values.

- **Chores**
  - Relaxed minimal Realtime bundle size thresholds.

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -3147,6 +3147,9 @@ export interface SummaryClientIdList {
   /** A list of the clientIds of all clients who have published an annotation with this name (or
    * type, depending on context). */
   clientIds: string[];
+  /** Whether the list of clientIds has been clipped due to exceeding the maximum number of
+   * clients. */
+  clipped?: boolean;
 }
 
 /** The per-name value for the multiple.v1 aggregation method. */
@@ -3160,6 +3163,11 @@ export interface SummaryClientIdCounts {
   /** The sum of the counts from all unidentified clients who have published an annotation with this
    * name, and so who are not included in the clientIds list */
   totalUnidentified: number;
+  /** Whether the list of clientIds has been clipped due to exceeding the maximum number of
+   * clients. */
+  clipped?: boolean;
+  /** The total number of distinct clientIds in the map (equal to length of map if clipped is false). */
+  totalClientIds: number;
 }
 
 /** The summary entry for aggregated annotations that use the total.v1
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -3128,7 +3128,7 @@ export interface SummaryDistinctValues {
 
 /** The summary entry for aggregated annotations that use the unique.v1
  * aggregation method. */
-interface SummaryUniqueValues {
+export interface SummaryUniqueValues {
   [key: string]: SummaryClientIdList;
 }
 
@@ -3149,7 +3149,7 @@ export interface SummaryClientIdList {
   clientIds: string[];
   /** Whether the list of clientIds has been clipped due to exceeding the maximum number of
    * clients. */
-  clipped?: boolean;
+  clipped: boolean;
 }
 
 /** The per-name value for the multiple.v1 aggregation method. */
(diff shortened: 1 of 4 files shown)


ably.d.ts
```diff
@@ -3145,6 +3145,9 @@ export interface SummaryClientIdList {
   /** A list of the clientIds of all clients who have published an annotation with this name (or
    * type, depending on context). */
   clientIds: string[];
+  /** Whether the list of clientIds has been clipped due to exceeding the maximum number of
+   * clients. */
+  clipped?: boolean;
```
AndyTWF: Is this likely to become non-nullable in the future? Would it be worth making it non-nullable and then having it just forced to `false` in `expandFields()`?

vladvelici: It's `omitempty` and that's a good thing I'd say, so not sure we'll ever explicitly set clipped: false?

AndyTWF: I think this will be mainly driven by what other languages will do and how much faff it introduces there - in languages like Go and C# it would require null checks to do a simple `if (clipped)`.

vladvelici: I don't imagine in Go we will make it
```
clipped *bool `json:"omitempty"`
```
but rather
```
clipped bool `json:"omitempty"`
```

For languages where it matters there's no point to have a third "this wasn't set" state, just use `false`, but in JS this maps to the JSON representation and all checks are identical (`if (x.clipped) {}` is the same even if clipped field is optional).

AndyTWF: I went away and thought about it for a bit and my gut feeling is that the field should match its semantics. There's no current case for unset here (we always know if its clipped) so I think it should be a non-nullable boolean.

SimonWoolf: I agree with Andy. We should use a nullable type in the client API if and only if it makes semantic sense for it to be nullable, i.e. it can be unset. And whichever way we decide, we should be consistent about it between SDKs.

So a `Message` might or might not have a clientId: if it was published by an unidentified client, it just won't have one. So it is right for that field to be nullable. But in the case of 'clipped', there's no semantic 'unset' distinct from false, if the summary isn't clipped it should just say false.

> in JS this maps to the JSON representation

This is not a strong consideration for the question of what the SDK type should be.There are plenty of fields we don't populate on the wire to save space but are still populated (and typed as such) by the SDK when a message arrives, like message.timestamp/id/connectionId (populated by the sdk from the ChannelMessage), or the version from the serial for CREATEs, etc. The SDK is expected to unpack the wire representation a bit to present a nicer API.

vladvelici: updated it to be explicitly set

src/common/lib/types/message.ts
```diff
@@ -121,6 +121,22 @@ class Message extends BaseMessage {
         this.createdAt = this.timestamp;
       }
     }
+    if (this.summary) {
```
AndyTWF: If we don't have a spec entry for this behaviour, we should add one (and add a comment here)

test/realtime/message.test.js
```diff
@@ -1330,6 +1330,105 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, Helper, async
         expect(updateMessage.createdAt).to.equal(undefined);
         expect(updateMessage.serial).to.equal(undefined);
       });
+
+      it('should ensure clipped field is set to false where not explicitly provided', async function () {
```
AndyTWF, line 1419: Ditto re a spec comment

src/common/lib/types/message.ts
vladvelici: done

test/realtime/message.test.js
vladvelici: done

PR #2072, opened 2025-08-20
[PUB-2015] Surface `clientId` who submitted a LiveObject operation in subscribe callbacks

VeskeR:
Resolves [PUB-2015](https://ably.atlassian.net/browse/PUB-2015)

[PUB-2015]: https://ably.atlassian.net/browse/PUB-2015?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Subscription updates for live objects, maps, and counters now include an optional clientId identifying the client that made the change; initial create updates carry clientId when applicable.

* **Refactor**
  * Object handlers now receive full message context so clientId is consistently propagated through create and mutation flows.

* **Tests**
  * End-to-end tests added/updated to verify subscription updates include the publishing client’s clientId for create and mutation scenarios.

PR #2084, opened 2025-09-18
[PUB-2558] Surface `connectionId` of a client who submitted a LiveObject operation in subscribe callbacks

VeskeR:
Based on the similar implementation for `clientId` from https://github.com/ably/ably-js/pull/2072

Resolves [PUB-2558](https://ably.atlassian.net/browse/PUB-2558)

[PUB-2558]: https://ably.atlassian.net/browse/PUB-2558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- New Features
  - Subscription updates for LiveObject, LiveMap, and LiveCounter now include an optional connectionId alongside clientId so apps can identify the originating connection.
  - Public types updated to expose connectionId in update payloads; backward compatible.

- Tests
  - Updated tests to assert presence and correctness of connectionId in subscription updates and operation flows.

### DIFF
--- a/test/common/modules/objects_helper.js
+++ b/test/common/modules/objects_helper.js
@@ -241,7 +241,7 @@ define(['ably', 'shared_helper', 'objects'], function (Ably, Helper, ObjectsPlug
     }
 
     objectOperationMessage(opts) {
-      const { channelName, serial, serialTimestamp, siteCode, state, clientId } = opts;
+      const { channelName, serial, serialTimestamp, siteCode, state, clientId, connectionId } = opts;
 
       state?.forEach((objectMessage, i) => {
         objectMessage.serial = serial;
@@ -254,6 +254,7 @@ define(['ably', 'shared_helper', 'objects'], function (Ably, Helper, ObjectsPlug
         action: 19, // OBJECT
         channel: channelName,
         channelSerial: serial,
+        connectionId,
         state: state ?? [],
       };
     }
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -2738,6 +2738,8 @@ export declare interface LiveObjectUpdate {
   update: any;
   /** The client ID of the client that published this update. */
   clientId?: string;
+  /** The connection ID of the client that published this update. */
+  connectionId?: string;
 }
 
 /**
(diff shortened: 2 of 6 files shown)


test/common/modules/objects_helper.js
```diff
@@ -254,6 +254,7 @@ define(['ably', 'shared_helper', 'objects'], function (Ably, Helper, ObjectsPlug
         action: 19, // OBJECT
         channel: channelName,
         channelSerial: serial,
+        connectionId,
```
mschristensen, line 257: Why is this defined at the top level for the `connectionId` but on each state message for the `clientId`?

VeskeR: This is intended to match the current realtime behavior.
Top level here is for ProtocolMessage, and currently realtime returns a protocol message for object messages like this:
```sh
[ProtocolMessage; action=OBJECT; id=SpzJHflHUH:9; channel=my_liveobjects_channel_4; channelSerial=01758186708862-000@108Uw-GWQBti3i59680125; connectionId=SpzJHflHUH; timestamp=1758186708862; state=[ [WireObjectMessage; clientId=test-client-id; operation={"action":2,"objectId":"root","mapOp":{"key":"image"}}; serial=01758186708862-000@108Uw-GWQBti3i59680125:000; serialTimestamp=1758186708862; siteCode=108] ]]; connectionId = SpzJHflHUH
```
if you look closely, `connectionId` is not actually exposed on object messages, but only on a ProtocolMessage. It looks like this is an expected behavior and SDKs usually handle populating this and other fields (`id`, `timestamp` and `connectionId`) for a regular/presence/annotation/object message from the parent ProtocolMessage, see https://github.com/ably/ably-js/blob/bbbd814dd0d0df2abb02c7382c5bb7dfad20fa05/src/common/lib/types/basemessage.ts#L327-L364.

So the test mimics this behavior and sets `connectionId` on a ProtocolMessage and expects the library to populate it on object messages

PR #2091, opened 2025-09-24
[PUB-2057] Implement base PathObject class with access and mutation methods

VeskeR:
Adds new path-based types to ably.d.ts.

The previous LiveObject types are temporarily marked as *Deprecated in ably.d.ts until they are fully converted to the new type system in the following PRs.
The `RealtimeObject` entrypoint temporarily adds a `getPathObject` method to get a PathObject instance for a root and preserves the existing `.get()` method which returns a LiveMap instance directly so that corresponding tests do not fail yet.

Resolves [PUB-2057](https://ably.atlassian.net/browse/PUB-2057)


src/plugins/objects/pathobject.ts
```diff
@@ -0,0 +1,249 @@
...
+   * Get a PathObject at the specified path relative to this object
+   */
+  at<U extends Value = Value>(path: string): PathObject<U> {
+    if (typeof path !== 'string') {
+      throw new this._client.ErrorInfo(`Path must be a string: ${path}`, 40003, 400);
+    }
+    // split path on unescaped dots
+    const pathAsArray = path.split(/(?<!\\)\./).map((segment) => segment.replace(/\\\./g, '.'));
+    return new DefaultPathObject<U>(this._realtimeObject, this._root, pathAsArray, this) as unknown as PathObject<U>;
+  }
```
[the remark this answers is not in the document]
VeskeR: nice find, will fix

src/plugins/objects/pathobject.ts
```diff
@@ -0,0 +1,249 @@
...
+    path: string[],
+    parent?: DefaultPathObject<any>,
+  ) {
+    this._client = this._realtimeObject.getClient();
+    // copy parent path array
+    this._path = [...(parent?._path ?? []), ...path];
+  }
+
+  /**
+   * Returns the fully-qualified string path that this PathObject represents
```
mschristensen: I think worth including a note in the comment here that map keys with dots in them will be escaped

src/plugins/objects/pathobject.ts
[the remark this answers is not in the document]
mschristensen: @VeskeR do you plan to fix this?

src/plugins/objects/pathobject.ts
```diff
@@ -0,0 +1,249 @@
...
+        typeof resolved === 'number' ||
+        typeof resolved === 'boolean' ||
+        typeof resolved === 'object' ||
+        resolved === null
+      ) {
+        // primitive type - return it
+        return resolved as U;
+      } else {
+        // unknown type - return undefined
+        return undefined;
```
mschristensen, line 135: Should we log a warn here? I think this would indicate a library bug, so if a customer reported this warn log we would know about it

src/plugins/objects/pathobject.ts
```diff
@@ -0,0 +1,249 @@
...
+      }
+    } catch (error) {
+      if (this._client.Utils.isErrorInfoOrPartialErrorInfo(error)) {
```
mschristensen: I don't see anything in this statement which determines that the error info is indeed a path resolution error. Should we add a specific error code for path resultion errors?

src/plugins/objects/pathobject.ts
```diff
@@ -0,0 +1,249 @@
...
+      // otherwise rethrow unexpected errors
+      throw error;
+    }
+  }
+
+  /**
+   * Returns an iterator of [key, value] pairs for LiveMap entries
+   */
+  *entries<U extends Record<string, Value>>(): IterableIterator<[keyof U, PathObject<U[keyof U]>]> {
+    const resolved = this._resolvePath(this._path);
```
mschristensen: I think we should treat path resolution errors consistently in `value()` and this and other methods. I think we should either consistently throw if the path doesn't resolve, or else we should consistently catch path resolution errors and return undefined. What do you think would provide the best experience?

src/plugins/objects/pathobject.ts
mschristensen: And a separate error code for a path resolving to an unexpected object type, e.g. calling `keys()` on a path that doesn't resolve to a `LiveMap`

mschristensen (commented):
Looks great, thanks. Some small comments

src/plugins/objects/pathobject.ts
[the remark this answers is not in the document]
VeskeR: @mschristensen fixed in https://github.com/ably/ably-js/compare/fb75485f7abe73530854aef7e72c3cdef78d9744..6954bce8746912f659ed32b5ff0e7292cadee937, we split path manually now

src/plugins/objects/pathobject.ts
VeskeR: clarified about escaped dots both here and in the public type definitions, see https://github.com/ably/ably-js/compare/fb75485f7abe73530854aef7e72c3cdef78d9744..6954bce8746912f659ed32b5ff0e7292cadee937

src/plugins/objects/pathobject.ts
VeskeR: added a log at major level, see https://github.com/ably/ably-js/compare/fb75485f7abe73530854aef7e72c3cdef78d9744..6954bce8746912f659ed32b5ff0e7292cadee937

src/plugins/objects/pathobject.ts
VeskeR: found existing error codes that fit nicely here:
https://faqs.ably.com/error-code-92005 - path resolution has failed, no entry at path.
https://faqs.ably.com/error-code-92007 - operation not processable at path.

updated to use them  in https://github.com/ably/ably-js/compare/fb75485f7abe73530854aef7e72c3cdef78d9744..6954bce8746912f659ed32b5ff0e7292cadee937.

also created https://ably.atlassian.net/browse/PUB-3305 to update the docs for those error codes to include explanation for realtime client (currently they only talk about REST API).

src/plugins/objects/pathobject.ts
VeskeR: changed the path resolution approach in https://github.com/ably/ably-js/compare/fb75485f7abe73530854aef7e72c3cdef78d9744..6954bce8746912f659ed32b5ff0e7292cadee937.

matches what we have discussed previously on the call. Idea in general:
- mutation methods (which usually return a promise) always throw to indicate that operation has failed and change was not applied.
- access methods silently resolve path resolution errors and return "empty" result based on the context. for `.value()` and `.size()` this means returning `undefined` as we couldn't resolve the value/size, for iterator methods we return empty iterator (as a convenience so developers don't need to check for nullish values every time they call an iterator method)

PR #2096, opened 2025-10-01
[PUB-2060] Implement LiveMap and LiveCounter creation via value types

VeskeR:
Exposes `LiveMap.create()` and `LiveCounter.create()` static methods in the `ably/objects` export.

Resolves [PUB-2060](https://ably.atlassian.net/browse/PUB-2060)


ably.d.ts
```diff
@@ -2417,7 +2417,7 @@ export type Primitive =
  * Unique symbol for nominal typing within TypeScript's structural type system.
  * This prevents structural compatibility between LiveObject types.
  */
-declare const __livetype: unique symbol;
+export declare const __livetype: unique symbol;
```
mschristensen, line 2385: I don't know that this should be exported publicly, perhaps we can define it alongside the dependent types in livemapvaluetype.ts and livecountervaluetype.ts then we import and use it here?

objects.d.ts
```diff
@@ -1,9 +1,49 @@
...
 /* eslint-enable no-unused-vars, @typescript-eslint/no-unused-vars */
 
+/**
+ * Static utilities related to LiveMap instances.
+ */
+export class LiveMap {
+  /**
+   * Creates a {@link LiveMapType | LiveMap} value type that can be passed to mutation methods
+   * (such as {@link LiveMapOperations.set}) to create a new LiveMap object
+   * in the object tree on a channel.
```
mschristensen: I think we should avoid references to the "object tree" now that we conceptualise it as a single nested object on the channel. Perhaps this could be "to assign a new LiveMap to the channel object"?

src/plugins/objects/livecountervaluetype.ts
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Properties of this class are immutable after construction and the instance
+ * will be frozen to prevent mutation.
+ */
+export class LiveCounterValueType implements API.LiveCounter {
+  declare readonly [API.__livetype]: 'LiveCounter'; // type-only, unique symbol to satisfy branded interfaces, no JS emitted
+  private readonly _livetype = 'LiveCounter'; // use a runtime property to provide a reliable cross-bundle type identification instead of `instanceof` operator
+  private readonly _count: number;
+
+  private constructor(count: number) {
```
mschristensen, line 25: Can we make this optional and default to 0?

src/plugins/objects/livemapvaluetype.ts
```diff
@@ -0,0 +1,170 @@
...
+        } as ObjectOperation<ObjectData>,
+      },
+      client.Utils,
+      client.MessageEncoding,
+    );
+
+    return {
+      mapCreateMsg,
+      nestedObjectsCreateMsgs,
+    };
```
mschristensen, line 116: Why do we re-implement all this logic here rather than call `LiveMap.createMapCreateMessage()` in `livemap.ts`, like we do for the counter value type?

src/plugins/objects/livecountervaluetype.ts
VeskeR: this a private constructor, public API creation is via `LiveCounter.create()` call, and count argument is optional there and is 0 by default, see https://github.com/ably/ably-js/pull/2096/files#diff-e06de53ebadf741a5d16720d3ed0b9c0266aac62475b916683808ef881ed6734R42

objects.d.ts
VeskeR: changed wording in https://github.com/ably/ably-js/compare/e88c6f3b9b65070b2bcd2a9bd8b208412270638c..dda76be4b774b877d692375afdedf2cd6194ae1a

(I did find it more hard to write docs with a new tree-less concept indeed. Some sentences just ask for "object tree" or similar)

src/plugins/objects/livemapvaluetype.ts
VeskeR: LiveMap creation from a value type is more involved than the previous implementation from `LiveMap.createMapCreateMessage()`  as we now also need to recursively create object messages for any nested value types (this wasn't the case before, as you've referenced other objects directly by their instance, so we already had an objectId).

That being said, I realized that `LiveMap.createMapCreateMessage()` and `LiveCounter.createCounterCreateMessage()` are actually only used as part of the now obsolete `RealtimeObject.createMap()` and `.createCounter()` API. Since we're going to remove that API anyway, I went ahead and did just that in https://github.com/ably/ably-js/pull/2096/commits/dda76be4b774b877d692375afdedf2cd6194ae1a (both internally and publicly).
As a result, we don't have an almost identical MAP_CREATE object creation logic anymore, there is only one for value type creation.

The tests were already updated to use the new value type constructors so it works just fine.

ably.d.ts
VeskeR: Uhh, this one is tricky.

Since our current public type definitions aren't automatically generated by TypeScript from the implementation and are instead maintained manually via `ably.d.ts`, simply importing from the plugin folder in the src won't work. When we package the build, that file isn't included in the output, so type resolution fails because it can't find the `__livetype` symbol - there's no file to import it from.

We *could* add a `src/plugins/objects/types.d.ts` file to the `files` list in `package.json`, but that’s a bit ugly.
A better option would be to add a root-level `types.d.ts` file (alongside `ably.d.ts`) that contains the `__livetype` symbol. That way, it can be imported both by the LiveObjects implementation and by `ably.d.ts`, without being exposed to developers - since our `exports` in `package.json` explicitly direct to either `ably.d.ts` for the core library or `objects.d.ts` for LiveObjects.

I don’t feel strongly about leaving it exported as is, or adding a separate `types.d.ts` to the package build.
If we do decide to add it, I'll tag Eugene to confirm there isn't a specific reason not to.

src/plugins/objects/livecountervaluetype.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  static create(initialCount: number = 0): API.LiveCounter {
+    // We can't directly import the ErrorInfo class from the core library into the plugin (as this would bloat the plugin size),
+    // and, since we're in a user-facing static method, we can't expect a user to pass a client library instance, as this would make the API ugly.
+    // Since we can't use ErrorInfo here, we won't do any validation at this step; instead, validation will happen in the mutation methods
+    // when we try to create this object.
+
+    return new LiveCounterValueType(initialCount);
+  }
```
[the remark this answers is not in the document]
VeskeR: this is intentional, updated tests to instead expect validation to happen when object value type is assigned to the object tree (on .set() call)

ably.d.ts
mschristensen: Thanks for the explanation - in that case, it's fine to leave it as is from my perspective.

VeskeR:
`Test NPM package` test is failing due to us introducing named exports in this PR, and our current plugin build setup doesn't suport named exports (incorrect type definitions exports and build process). This is fixed separately in https://github.com/ably/ably-js/pull/2106 (both for Objects and Push plugins).
I tested locally that with with a fix from https://github.com/ably/ably-js/pull/2106 named exports from this PR work as expected.
Merging this as is as all other tests are passing.

PR #2097, opened 2025-10-03
[PUB-2061] Add path based subscriptions support for LiveObjects

VeskeR:
Implements:
- LiveObject parent tracking
- Path-based LiveObject event emission
- PathObject subscriptions path matching with deep subscriptions
- Path event emission for LiveMap key updates
- Full ObjectMessage (user-facing type) argument for subscription callbacks

Resolves [PUB-2061](https://ably.atlassian.net/browse/PUB-2061)

[PUB-2061]: https://ably.atlassian.net/browse/PUB-2061?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **New Features**
  - Path- and instance-based subscriptions with depth/bubbling controls, richer event payloads including standardized object-operation messages and user-facing message serialization, unsubscribe handlers, root-object access and iterable object listing, and tombstone notifications.

- **Bug Fixes / Improvements**
  - Consistent update-type markers and payload shapes, consistent map/counter update behavior, parent-reference tracking and rebuilds to ensure correct path notifications, and more stable user-facing message formatting.

- **Documentation**
  - Clarified subscribe semantics, event payload shapes, depth and bubbling behavior.

- **Tests**
  - Extensive subscription tests covering depth, bubbling, unsubscribe semantics, replacements, and error resilience.

### DIFF
--- a/src/plugins/objects/livecounter.ts
+++ b/src/plugins/objects/livecounter.ts
@@ -127,7 +127,7 @@ export class LiveCounter extends LiveObject<LiveCounterData, LiveCounterUpdate> 
       return;
     }
 
-    let update: LiveCounterUpdate | LiveObjectUpdateNoop = { noop: true };
+    let update: LiveCounterUpdate | LiveObjectUpdateNoop;
     switch (op.action) {
       case ObjectOperationAction.COUNTER_CREATE:
         update = this._applyCounterCreate(op, msg);
@@ -144,7 +144,7 @@ export class LiveCounter extends LiveObject<LiveCounterData, LiveCounterUpdate> 
         break;
 
       case ObjectOperationAction.OBJECT_DELETE:
-        this._applyObjectDelete(msg);
+        update = this._applyObjectDelete(msg);
         break;
 
       default:
@@ -205,23 +205,23 @@ export class LiveCounter extends LiveObject<LiveCounterData, LiveCounterUpdate> 
     }
 
     const previousDataRef = this._dataRef;
+    let update: LiveCounterUpdate;
     if (objectState.tombstone) {
       // tombstone this object and ignore the data from the object state message
-      this.tombstone(objectMessage);
+      update = this.tombstone(objectMessage);
     } else {
-      // override data for this object with data from the object state
+      // otherwise override data for this object with data from the object state
       this._createOperationIsMerged = false; // RTLC6b
       this._dataRef = { data: objectState.counter?.count ?? 0 }; // RTLC6c
       // RTLC6d
       if (!this._client.Utils.isNil(objectState.createOp)) {
         this._mergeInitialDataFromCreateOperation(objectState.createOp, objectMessage);
       }
-    }
 
-    // if object got tombstoned, the update object will include all data that got cleared.
-    // otherwise it is a diff between previous value and new value from object state.
-    const update = this._updateFromDataDiff(previousDataRef, this._dataRef);
-    update.objectMessage = objectMessage;
+      // update will contain the diff between previous value and new value from object state
+      update = this._updateFromDataDiff(previousDataRef, this._dataRef);
+      update.objectMessage = objectMessage;
+    }
 
     return update;
   }
--- a/src/plugins/objects/liveobject.ts
+++ b/src/plugins/objects/liveobject.ts
@@ -22,6 +22,8 @@ export interface LiveObjectUpdate {
   update: any;
   /** Object message that caused an update to an object, if available */
   objectMessage?: ObjectMessage;
+  /** Indicates whether this update is a result of a tombstone (delete) operation. */
+  tombstone?: boolean;
 }
 
 export interface LiveObjectUpdateNoop {
@@ -139,6 +141,11 @@ export abstract class LiveObject<
 
     this._notifyInstanceSubscriptions(update);
     this._notifyPathSubscriptions(update);
+
+    if (update.tombstone) {
+      // deregister all listeners if update was a result of a tombstone operation
+      this._subscriptions.off();
+    }
   }
 
   /**
@@ -146,7 +153,7 @@ export abstract class LiveObject<
    *
    * @internal
    */
-  tombstone(objectMessage: ObjectMessage): void {
+  tombstone(objectMessage: ObjectMessage): TUpdate {
     this._tombstone = true;
     if (objectMessage.serialTimestamp != null) {
       this._tombstonedAt = objectMessage.serialTimestamp;
@@ -161,12 +168,11 @@ export abstract class LiveObject<
     }
     const update = this.clearData();
     update.objectMessage = objectMessage;
+    update.tombstone = true;
 
     this._lifecycleEvents.emit(LiveObjectLifecycleEvent.deleted);
 
-    // notify subscribers about the delete operation and then deregister all listeners
-    this.notifyUpdated(update);
-    this._subscriptions.off();
+    return update;
   }
 
   /**
@@ -307,8 +313,8 @@ export abstract class LiveObject<
     return !siteSerial || opSerial > siteSerial;
   }
 
-  protected _applyObjectDelete(objectMessage: ObjectMessage): void {
-    this.tombstone(objectMessage);
+  protected _applyObjectDelete(objectMessage: ObjectMessage): TUpdate {
+    return this.tombstone(objectMessage);
   }
 
   private _notifyInstanceSubscriptions(update: TUpdate): void {
--- a/src/plugins/objects/livemap.ts
+++ b/src/plugins/objects/livemap.ts
@@ -379,7 +379,7 @@ export class LiveMap<T extends API.LiveMapType> extends LiveObject<LiveMapData, 
       return;
     }
 
-    let update: LiveMapUpdate<T> | LiveObjectUpdateNoop = { noop: true };
+    let update: LiveMapUpdate<T> | LiveObjectUpdateNoop;
     switch (op.action) {
       case ObjectOperationAction.MAP_CREATE:
         update = this._applyMapCreate(op, msg);
@@ -406,7 +406,7 @@ export class LiveMap<T extends API.LiveMapType> extends LiveObject<LiveMapData, 
         break;
 
       case ObjectOperationAction.OBJECT_DELETE:
-        this._applyObjectDelete(msg);
+        update = this._applyObjectDelete(msg);
         break;
 
       default:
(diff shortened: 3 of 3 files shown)


ably.d.ts
```diff
@@ -2468,6 +2475,29 @@ interface PathObjectBase<_T extends Value> {
    * @experimental
    */
   compact(): any;
+
+  /**
+   * Registers a listener that is called each time the object or a primitive value at this path is updated.
+   *
+   * The provided listener receives an object message that carried an operation.
```
mschristensen:
```suggestion
   * The provided listener receives an {@link ObjectMessage} that carried the operation that led to the change, as well as an {@link PathObject} representing the path that was updated.
```

ably.d.ts
```diff
@@ -2838,6 +2868,25 @@ interface InstanceBase<_T extends Value> {
    * @experimental
    */
   compact(): any;
+
+  /**
+   * Registers a listener that is called each time this instance is updated.
+   *
+   * The provided listener receives an object message that carried an operation.
```
mschristensen:
```suggestion
   * The provided listener receives an {@link ObjectMessage} that carried the operation that led to the change, as well as an {@link Instance} providing a reference to the instance that was updated.
```

ably.d.ts
```diff
@@ -3001,6 +3050,93 @@ export type Instance<T extends Value> = [T] extends [LiveMap<infer T>]
       ? PrimitiveInstance<T>
       : AnyInstance<T>;
 
+/**
+ * The event object passed to a {@link PathObject} subscription listener.
+ */
+export type PathObjectSubscriptionEvent = {
+  /** The {@link PathObject} representing the updated path. */
+  object: PathObject;
+  /** The object message describing the operation that triggered the update, if available. */
```
mschristensen:
```suggestion
  /** The {@link ObjectMessage} describing the operation that triggered the update, if available. */
```

ably.d.ts
```diff
@@ -3001,6 +3050,93 @@ export type Instance<T extends Value> = [T] extends [LiveMap<infer T>]
...
+  depth?: number;
+}
+
+/**
+ * The event object passed to an {@link Instance} subscription listener.
+ */
+export type InstanceSubscriptionEvent<T extends Value> = {
+  /** The {@link Instance} representing the updated object. */
+  object: Instance<T>;
+  /** The object message describing the operation that triggered the update, if available. */
```
mschristensen:
```suggestion
  /** The {@link ObjectMessage} describing the operation that triggered the update, if available. */
```

ably.d.ts
```diff
@@ -3001,6 +3050,93 @@ export type Instance<T extends Value> = [T] extends [LiveMap<infer T>]
...
+    /**
+     * A set of key–value pair headers included with this object message.
+     */
+    headers?: Record<string, string>;
+    [key: string]: any;
+  };
+  /**
+   * The operation payload of the object message.
+   */
+  payload: any;
```
mschristensen: Can we ensure this is fully typed? The shape of operations is known so we should be able to type it. Happy to ticket this separately.

src/plugins/objects/liveobject.ts
```diff
@@ -82,6 +95,19 @@ export abstract class LiveObject<
     return { unsubscribe };
   }
 
+  // TODO: replace obsolete .subscribe call with this one when we completely remove previous API and switch to path-based one
```
mschristensen: Can we make these changes now and remove the old subscription approach? This solution depends on the old subscription approach with the old subscription event payload, which is a subset of the data we actually want to surface. Instead of doing that, can we bubble up the underlying ObjectMessage (with a transformation step in the middle if needed) through to the path or instance object subscription, instead of translating the old subscription event into the new?

src/plugins/objects/livecounter.ts
```diff
@@ -214,7 +215,7 @@ export class LiveCounter extends LiveObject<LiveCounterData, LiveCounterUpdate>
         break;
 
       case ObjectOperationAction.OBJECT_DELETE:
-        update = this._applyObjectDelete(msg);
+        this._applyObjectDelete(msg);
```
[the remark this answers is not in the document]
VeskeR: _applyObjectDelete no longer returns an update object

ably.d.ts
VeskeR: updated in https://github.com/ably/ably-js/compare/6d712ac3446ded3e5af18c0daf0e15730ce97291..69b0b165899b957088fadbeae6b5ce14341b14f6

ably.d.ts
VeskeR: updated in https://github.com/ably/ably-js/compare/6d712ac3446ded3e5af18c0daf0e15730ce97291..69b0b165899b957088fadbeae6b5ce14341b14f6

ably.d.ts
VeskeR: updated in https://github.com/ably/ably-js/compare/6d712ac3446ded3e5af18c0daf0e15730ce97291..69b0b165899b957088fadbeae6b5ce14341b14f6

ably.d.ts
VeskeR: updated in https://github.com/ably/ably-js/compare/6d712ac3446ded3e5af18c0daf0e15730ce97291..69b0b165899b957088fadbeae6b5ce14341b14f6

ably.d.ts
VeskeR: As agreed on the call, I've removed `ObjectMessage.payload` and added `ObjectMessage.operation` instead.
If `ObjectMessage` is present in the subscription event, it always represents an operation on an object (not a sync message). We do not expose object sync messages, just trigger subscription listeners for them.

See: https://github.com/ably/ably-js/pull/2097/commits/54f8ae298fd57fea21e4e792703d629c6f32f88d

test/realtime/objects.test.js
```diff
@@ -4444,6 +4444,534 @@ define(['ably', 'shared_helper', 'chai', 'objects', 'objects_helper'], function
...
+            await counterUpdatedPromise;
+
+            // If object at the subscribed path is replaced, that should also trigger the subscription
+            keyUpdatedPromise = waitForMapKeyUpdate(root, 'nested');
+            await entryPathObject.set('nested', LiveMap.create());
+            await keyUpdatedPromise;
+
+            await subscriptionPromise;
+          },
+        },
```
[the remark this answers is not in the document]
VeskeR: keys are correct, path is for the object that was updated, not its key

test/realtime/objects.test.js
```diff
@@ -4444,6 +4444,534 @@ define(['ably', 'shared_helper', 'chai', 'objects', 'objects_helper'], function
...
+            await entryPathObject.get('nested').set('nestedKey', 'foo');
+            await keyUpdatedPromise;
+
+            // Now make a direct change to the root object, should trigger the callback
+            keyUpdatedPromise = waitForMapKeyUpdate(root, 'directKey');
+            await entryPathObject.set('directKey', 'bar');
+            await keyUpdatedPromise;
+
+            await Promise.all([subscriptionDepthOnePromise, subscriptionDepthTwoPromise]);
+          },
```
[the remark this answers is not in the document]
VeskeR: current implementation is correct

src/plugins/objects/liveobject.ts
VeskeR: The now obsolete LiveObject subscription API was removed in https://github.com/ably/ably-js/pull/2097/commits/6024f4bf8085a736955c2a04a5e10d0fa426ab24.
Only PathObject and Instance subscriptions API is now exposed.
Note that internally I decided to preserve the concept of the "update" object which results in the event to be sent to the path object or instance subscriptions. This "update" object now carries the ObjectMessage directly, and some kind of a diff that explains what was changed on an object. Even though we don't expose delta publicly at the moment, the internal diff object is still useful for two reasons:
1. The diff object is useful when emitting events for direct key changes in collections as a result of object sync sequence -  when a map was changed as a result of sync sequence, the ObjectMessage is not sufficient enough to identify the keys that were changed, we need to know specific keys that got updated. The previous implementation already did that, so I kept it as is internally and resused it to emit PathObject subscription events for direct key changes.
2. When we eventually implement delta for subscription events, we can reuse existing internals that calculated diffs for objects, so no point to remove it now just to re-add it later.

src/plugins/objects/livecounter.ts
```diff
@@ -126,7 +127,7 @@ export class LiveCounter extends LiveObject<LiveCounterData, LiveCounterUpdate>
       return;
     }
 
-    let update: LiveCounterUpdate | LiveObjectUpdateNoop;
+    let update: LiveCounterUpdate | LiveObjectUpdateNoop = { noop: true };
```
mschristensen: This logic is a bit hard to follow - we call `notifyUpdate` for non-`OBJECT_DELETE` ops here, but treat the `OBJECT_DELETE` op as a no-op update; however we do in fact call `notifyUpdate` with an appropriate update for the delete inside `tombstone()` on the `LiveObject`.

Could we instead make all `notifyUpdate` calls from a single place, i.e. here? We should be able to deregister the subscription as required here as well rather than on the parent class.

ably.d.ts
```diff
@@ -3312,49 +3573,10 @@ export declare interface LiveCounterDeprecated extends LiveObjectDeprecated<Live
...
  */
-export declare interface LiveObjectDeprecated<TUpdate extends LiveObjectUpdate = LiveObjectUpdate> {
-  /**
-   * Registers a listener that is called each time this LiveObject is updated.
-   *
-   * @param listener - An event listener function that is called with an update object whenever this LiveObject is updated.
-   * @returns A {@link SubscribeResponse} object that allows the provided listener to be deregistered from future updates.
-   * @experimental
-   */
-  subscribe(listener: LiveObjectUpdateCallback<TUpdate>): SubscribeResponse;
```
mschristensen, line 3341: We should remove the typedoc references to these methods elsewhere in this file

VeskeR: removed in https://github.com/ably/ably-js/compare/07ec19d9d6421cb135c33169f7bba7bdde370ef4..5406c870419a3072695029c4e2f1d7b965ae6626

src/plugins/objects/livecounter.ts
VeskeR: Reverted changes to handling of the `OBJECT_DELETE` update object in https://github.com/ably/ably-js/compare/5406c870419a3072695029c4e2f1d7b965ae6626..3a940c6a6c4d1b388d1c154726ff699296eb8558.

It is now handled just like other operations. `OBJECT_DELETE` operation / tombstone operation returns an update that also indicates that it is a result of a tombstone, and then `notifyUpdated` handles all subscriptions and deregestering in one place.

PR #2098, opened 2025-10-03
[PUB-2064] Implement .compact() method for PathObject and Instance types for LiveObjects

VeskeR:
Resolves [PUB-2064](https://ably.atlassian.net/browse/PUB-2064)


src/plugins/objects/pathobject.ts
```diff
@@ -44,10 +45,25 @@ export class DefaultPathObject<T extends Value = Value> implements AnyPathObject
...
+    try {
+      const resolved = this._resolvePath(this._path);
+
+      if (resolved instanceof LiveMap) {
+        return resolved.compact() as CompactedValue<U>;
+      }
+
+      return this.value() as CompactedValue<U>;
+    } catch (error) {
+      if (this._client.Utils.isErrorInfoOrPartialErrorInfo(error)) {
```
mschristensen: This should assert on error code `92005` after https://github.com/ably/ably-js/pull/2092 is merged

src/plugins/objects/livemap.ts
```diff
@@ -640,6 +641,34 @@ export class LiveMap<T extends API.LiveMapType> extends LiveObject<LiveMapData,
...
+        continue;
+      }
+
+      if (value instanceof LiveCounter) {
+        result[key] = value.value();
+        continue;
+      }
+
+      // Other values return as is
+      result[key] = value;
```
mschristensen, line 576: I think buffer values should be base64 encoded strings; I think the object returned from `compact()` should be a trivially JSON-encodable object

src/plugins/objects/instance.ts
```diff
@@ -77,21 +81,14 @@ export class DefaultInstance<T extends Value> implements AnyInstance<T> {
       // primitive type - return it
       return this._value as unknown as U;
     } else {
-      this._client.Logger.logAction(
-        this._client.logger,
-        this._client.Logger.LOG_MAJOR,
-        'DefaultInstance.value()',
-        `unexpected value type for instance, resolving to undefined; value=${this._value}`,
-      );
       // unknown type - return undefined
```
mschristensen, line 97: Can we log the type of `this_value` here? We should never encounter this case so if we do we should have a log to help us understand the bug

src/plugins/objects/pathobject.ts
VeskeR: updated in https://github.com/ably/ably-js/compare/ab567497741434f6e0ffc88254996573ee10e76b..f3eee80162ec8890bab371727b6a25e5508c1cac

src/plugins/objects/instance.ts
VeskeR: added type to the log message in https://github.com/ably/ably-js/compare/f3eee80162ec8890bab371727b6a25e5508c1cac..b1d0234e23c1482b87a6bed42e19be15d6953bfc

src/plugins/objects/livemap.ts
VeskeR: made binary values return base64 encoded strings in https://github.com/ably/ably-js/pull/2098/commits/94be648505106b2c73a23692e2953de8224e1170

PR #2100, opened 2025-10-08
fix: implement RTC17 clientId

AndyTWF:
We currently have the RTC17 clientId declared in `ably.d.ts` but it's not implemented anywhere, so it always returns undefined. This change adds a getter to `BaseRealtime`.


src/common/lib/client/baserealtime.ts
```diff
@@ -83,6 +83,10 @@ class BaseRealtime extends BaseClient {
     return this._channels;
   }
 
+  get clientId() {
+    return this.auth.clientId;
```
lawrence-forooghian, line 88: Can you put a spec point comment here please?

AndyTWF: Done!

PR #2108, opened 2025-11-11
Remove non-path-based API from LiveObjects

VeskeR:
This marks full transition to path-based API for LiveObjects. Included in this PR:
- resolve TODOs related to path-based API
- update all LiveObjects tests to use path-based API
- remove publicly exposed LiveObject, LiveMap and LiveCounter types from ably.d.ts
- remove lifecycle subscriptions API for liveobjects (API and implementation). OBJECT_DELETE events are handled by regular subscription events.

## Summary by CodeRabbit

* **New Features**
  * Added Subscription and StatusSubscription interfaces for simpler subscribe/unsubscribe handling.
  * Introduced JSON type aliases for clearer JSON value typing.

* **Refactor**
  * Simplified object/value typing to a flexible record/value model.
  * Reworked object retrieval to return path-wrapped objects for consistent access patterns.
  * Streamlined map value handling to a unified value-centric model.

* **Chores**
  * Removed deprecated lifecycle event APIs and legacy types.

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -1658,11 +1658,6 @@ export type EventCallback<T> = (event: T) => void;
 export type ObjectsEventCallback = () => void;
 
 /**
- * The callback used for the lifecycle events emitted by {@link LiveObjectDeprecated}.
- */
-export type LiveObjectLifecycleEventCallback = () => void;
-
-/**
  * A function passed to the {@link BatchOperations.batch | batch} method to group multiple Objects operations into a single channel message.
  *
  * The function must be synchronous.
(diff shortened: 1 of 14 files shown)


test/realtime/objects.test.js
```diff
@@ -172,15 +172,15 @@ define(['ably', 'shared_helper', 'chai', 'objects', 'objects_helper'], function
...
+    const entryPathObject = await channel.object.get();
     const entryInstance = entryPathObject.instance();
 
     await Promise.all(
       expectedKeys.map((key) => (entryInstance.get(key) ? undefined : waitForMapKeyUpdate(entryInstance, key))),
     );
   }
 
-  describe('realtime/objects', function () {
+  describe.only('realtime/objects', function () {
```
mschristensen: We should remove this `.only`

ably.d.ts
```diff
@@ -2430,6 +2468,31 @@ export type CompactedValue<T extends Value> =
                       ? T
                       : any;
 
+declare global {
```
mschristensen, line 2471: Unrelated to this PR specifically, but:

I had a thought about this default typings declaration. I don't know this this is very useful to be honest, since you likely use multiple channels in your app, and each channel probably needs a different schema. Therefore I am tempted to remove this in the new version of the API and rely only on the type parameter approach.

Let's discuss

mschristensen (approved):
Small issue to fix on the test otherwise looks great

test/realtime/objects.test.js
VeskeR: oops, removed in https://github.com/ably/ably-js/compare/e7d734e2cb0b35c7fc032cc4a602afc920d19cf3..70540ffafc5effd252df2d6b256012a65a6409ad

ably.d.ts
VeskeR: Seeing in our demos how easy it is to provide an explicit type parameter the first time you call `object.get<MyObject>` I'd agree, seems like there is no actual need to have a global interface.
Also explaning how to use global type is harder than just saying "provide type for your object as a type parameter to the .get() call", so it's a win-win

VeskeR: done in https://github.com/ably/ably-js/pull/2113

PR #2109, opened 2025-11-14
[AIT-27, PUB-1197] LiveObjects REST client

VeskeR:
Resolves [PUB-1197](https://ably.atlassian.net/browse/PUB-1197), [AIT-27](https://ably.atlassian.net/browse/AIT-27)

[PUB-1197]: https://ably.atlassian.net/browse/PUB-1197?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added REST API support for LiveObjects plugin, enabling access via `RestChannel.object`
  * Introduced `RestObject` with `get()` and `publish()` methods for REST-based live object operations
  * Extended LiveObjects plugin to support REST channels alongside realtime channels

* **Tests**
  * Added comprehensive test coverage for REST LiveObjects operations

[AIT-27]: https://ably.atlassian.net/browse/AIT-27?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/liveobjects.d.ts
+++ b/liveobjects.d.ts
@@ -12,11 +12,12 @@ import {
   EventCallback,
   RealtimeChannel,
   RealtimeClient,
+  RestClient,
   StatusSubscription,
   Subscription,
   __livetype,
 } from './ably';
-import { BaseRealtime } from './modular';
+import { BaseRealtime, BaseRest, Rest } from './modular';
 /* eslint-enable no-unused-vars, @typescript-eslint/no-unused-vars */
 
 /**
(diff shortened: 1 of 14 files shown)


src/plugins/objects/restobject.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  /**
+   * Read object data. Defaults to { compact: true } and entrypoint=root when no id is provided.
+   * Returns undefined if provided objectId/path does not resolve to an object.
+   */
+  async get(params?: Omit<GetObjectParams, 'compact'> & { compact?: true }): Promise<RestCompactObjectData | undefined>;
+  async get(params: Omit<GetObjectParams, 'compact'> & { compact: false }): Promise<RestLiveObject | undefined>;
+  async get(params?: GetObjectParams): Promise<RestCompactObjectData | RestLiveObject | undefined> {
+    const client = this.channel.client;
+    const format = client.options.useBinaryProtocol ? client.Utils.Format.msgpack : client.Utils.Format.json;
```
mschristensen: We don't currently support msgpack format REST responses on the realtime side, so I think we should always use JSON here for now

mschristensen (commented):
LGTM, although I dont think these tests run with the `useBinaryProtocol` option set to `true`, given https://github.com/ably/ably-js/blob/5f7807f67872c66a13b82d1e443f814e6ddc7e27/test/common/modules/objects_helper.js#L33

Can we try running these tests with msgpack to see if it works?  I did find this ticket which may indicate an issue with msgpack over rest currently: https://ably.atlassian.net/browse/PUB-1835

src/plugins/liveobjects/restobject.ts
```diff
@@ -0,0 +1,184 @@
...
+          path: op.path,
+          data: Object.entries(op.entries).reduce(
+            (acc, [key, value]) => {
+              acc[key] = this._encodePrimitive(value, format);
+              return acc;
+            },
+            {} as Record<string, any>,
+          ),
+        };
+      case 'map.set':
```
kaschula: Hello, These operation keys in their current format won't pass the rest validation. The current format can be found here in the docs https://ably.com/docs/api/liveobjects-rest#/paths/~1channels~1%7BchannelName%7D~1objects/post

For example `map.set` would have to be `MAP_SET`.

test/rest/objects.test.js
```diff
@@ -0,0 +1,480 @@
...
+
+      describe('RestObject.publish()', () => {
+        /** @nospec */
+        it('should publish single MAP_SET operation', async function () {
+          const helper = this.test.helper;
+          const client = RestWithObjects(helper);
+          const channel = client.channels.get('publish-map-set');
+
+          const operation = {
+            operation: 'map.set',
```
kaschula: This request will fail because the map set operation requires a key. If the intention of this request is to set a key of `testKey` on the root then it needs to be:

```
const operation = {
        operation: 'map.set',
        path: '',
        key: 'testKey',
        value: 'testValue',
    };
```

kaschula:
Something I noticed when testing this branch for Msgpack requests. The SDK currently isn't setting the correct content type header of `application/msgpack` it was being set to JSON still

test/rest/objects.test.js
VeskeR: fixed in https://github.com/ably/ably-js/pull/2109/commits/173b03c0358c1d1a23a970bd65667a604ab2ce38, thanks

src/plugins/liveobjects/restobject.ts
VeskeR: fixed in https://github.com/ably/ably-js/compare/568d0a0b060088a08c8db11b1b9bc4fb8c1b0dca..879347fb8c1ce499a871d718619a6f84987d6c62

liveobjects.d.ts
```diff
@@ -1832,3 +2091,17 @@ declare module 'ably' {
...
   }
 }
+
+/**
+ * Module augmentation to add the `object` property to `RestChannel` when
+ * importing from 'ably/liveobjects'. This ensures all LiveObjects types come from
+ * the same module (CJS or ESM), avoiding type incompatibility issues.
+ */
+declare module 'ably' {
+  interface RestChannel {
```
lawrence-forooghian: The type is called `Channel`, not `RestChannel`, so currently none of the REST API is accessible from TypeScript. I'd suggest extending the `test/package/browser/template/src/index-liveobjects.ts` test to cover REST too, to catch issues like this.

liveobjects.d.ts
```diff
@@ -1793,22 +2051,23 @@ export class LiveCounter {
...
  *
  * ```javascript
  * import { Realtime } from 'ably';
  * import { LiveObjects } from 'ably/liveobjects';
  * const realtime = new Realtime({ ...options, plugins: { LiveObjects } });
  * ```
  *
- * The LiveObjects plugin can also be used with a {@link BaseRealtime} client:
+ * The LiveObjects plugin can also be used with a {@link BaseRest} or {@link BaseRealtime} client,
+ * with the additional requirement that you must also use the {@link Rest} plugin
```
lawrence-forooghian: > `with the additional requirement that you must also use the {@link Rest} plugin`

Not sure what this is trying to express — `BaseRest` already includes `Rest`, and `BaseRealtime` has no way of using the LiveObjects REST API.

liveobjects.d.ts
```diff
@@ -67,7 +68,256 @@ export type ObjectsEventCallback = () => void;
...
+   */
+  objectIds: string[];
+}
+
+/**
+ * Compact representation of object data returned from {@link RestObject.get} when `compact=true` (default).
+ * Provides a JSON-like view with primitive values, nested objects, and object references for handling cycles.
+ */
+export type RestCompactObjectData =
+  | string
```
lawrence-forooghian: Can we use `JsonScalar` here?

liveobjects.d.ts
```diff
@@ -67,7 +68,256 @@ export type ObjectsEventCallback = () => void;
...
+  | number
+  | boolean
+  | null
+  | JsonArray
+  | JsonObject
+  | { objectId: string } // cyclic references handling
```
lawrence-forooghian: I just wanted to understand what we're trying to achieve here — as I understand it, this doesn't provide any sort of actual ability for the user to narrow a type since the `JsonObject` type permits an `objectId` key whose corresponding value isn't a `string`, as does `{ [key: string]: RestCompactObjectData }`.

If it's not contributing anything at a type level then the significance of `objectId` might be better off inside the documentation comment?

(I do remember having a similar conversation with @SimonWoolf somewhere about a similar situation in some other type the public API and he had a justification — but I don't remember what it was or what the justification was!)

liveobjects.d.ts
```diff
@@ -67,7 +68,256 @@ export type ObjectsEventCallback = () => void;
...
+
+/**
+ * Operation to set a key to a specified value in an existing map object.
+ * Can target the map by either object ID or path.
+ */
+export type OperationMapSet = AnyTargetOperationBase & {
+  operation: ObjectOperationActions.MAP_SET;
+  key: string;
+  value: PrimitiveOrObjectReference;
+  encoding?: string;
```
lawrence-forooghian: What is `encoding`? It doesn't seem to be exercised in the tests anywhere and it's not clear what it does.

liveobjects.d.ts
```diff
@@ -67,7 +68,256 @@ export type ObjectsEventCallback = () => void;
...
+export interface RestObjectMapEntry {
+  /** The value associated with this map entry. */
+  data: RestLiveObject | RestObjectData;
+}
+
+/**
+ * Represents a value in an object on a channel.
+ * Either a primitive data or a reference to another object.
+ */
+export interface RestObjectData {
```
lawrence-forooghian: As written, there's no guarantee that _any_ of these properties is populated; is that intentional?

e.g. if we had something like

```typescript
  export type RestObjectData =
    | { objectId: string }
    | { number: number }
    | { boolean: boolean }
    | { string: string }
    | { bytes: Buffer | ArrayBuffer }
    | { json: JsonArray | JsonObject };
```

then if you had an `if` statement that checked all the keys the compiler would know that at least one of them would actually find something. (This is not necessarily useful if you know the shape of the data you expect to receive, but it is useful if you, say, want to write an algorithm for converting an `RestObjectData` into some alternative representation of such a type. If we wish for it not to be _possible_ to write such an algorithm — e.g. because we wish to have the flexibility to add new fields here in the future — then that's a different matter and the type as written is good.)

Edit: Ah, I see in the decoding implementation that we do indeed seem to anticipate receiving unknown data types in the future and want to handle them gracefully, returning `{}`.

liveobjects.d.ts
```diff
@@ -124,6 +374,14 @@ export type Primitive =
   | JsonArray
   | JsonObject;
 
+/**
+ * Primitive types and object references that may appear in `data` fields
+ * of response bodies when using the Objects REST API.
```
lawrence-forooghian: They're also used for requests too — `OperationMapSet.value` and `OperationMapCreate.entries`.

src/plugins/liveobjects/restobject.ts
```diff
@@ -0,0 +1,352 @@
...
+      number: number;
+    };
+  };
+}
+
+export type WireAnyRestLiveObject = {
+  objectId: string;
+};
+
+export class RestObject {
```
lawrence-forooghian: It would be good for this to declare that it implements the public `RestObject` type, to ensure we conform to the public contract.

src/plugins/liveobjects/restobject.ts
```diff
@@ -0,0 +1,352 @@
...
+          },
+        };
+
+      case 'map.set':
+        return {
+          id: op.id,
+          extras: op.extras,
+          objectId: op.objectId,
+          path: op.path,
+          mapSet: {
```
lawrence-forooghian: I'm curious why we chose a public API for the representing REST object operations that is different to that which is sent over the wire, especially when the public API for representing inbound operations in the Realtime case (as of fdc33cbf5c2c09a6a80adf2fc5565e835c03bbe2) _is_ consistent with the REST wire structure.

e.g. 

- Realtime public API: `{ action: 'map.set', mapSet: { key: 'foo', value: { string: 'bar' } } }`
- REST public API: `{ operation: 'map.set', key: 'foo', value: 'bar' }` — but a nested `mapSet` gets sent over the wire

src/plugins/liveobjects/restobject.ts
```diff
@@ -0,0 +1,352 @@
...
+
+    // live counter or JSON object
+    return obj;
+  }
+
+  private _decodeWireRestLiveMap(obj: WireRestLiveMap, format: Utils.Format): RestLiveObject {
+    return {
+      objectId: obj.objectId,
+      map: {
+        semantics: (mapSemantics[obj.map.semantics] ?? 'unknown') as ObjectsMapSemantics,
```
lawrence-forooghian: I know that in general we fudge this kind of thing in ably-js because TypeScript allows us to, but _really_ the right thing to do here is either:

- represent this situation in the public API
- throw an error

test/rest/liveobjects.test.js
```diff
@@ -0,0 +1,732 @@
...
+
+        /** @nospec */
+        Helper.testOnJsonMsgpack(
+          'should handle different data types in MAP_CREATE operation',
+          async function (options, channelName, helper) {
+            const client = RestWithObjects(helper, options);
+            const channel = client.channels.get(channelName);
+
+            // first test primitive data types
+            const entries = primitiveKeyData.reduce((acc, keyData) => {
```
lawrence-forooghian: This logic (for converting the `primitiveKeyData` values into something that we can use inside the operation that we pass to `object.publish()`) is duplicated. Can we DRY it up and add a bit of documentation?

test/rest/liveobjects.test.js
```diff
@@ -0,0 +1,732 @@
...
+    } else if (data.json != null) {
+      const expectedObject = JSON.parse(data.json);
+      expect(obj.map.entries[key].data.json).to.deep.equal(expectedObject, msg);
+    } else if (data.string != null) {
+      expect(obj.map.entries[key].data.string).to.deep.equal(data.string, msg);
+    } else if (data.number != null) {
+      expect(obj.map.entries[key].data.number).to.deep.equal(data.number, msg);
+    } else if (data.boolean != null) {
+      expect(obj.map.entries[key].data.boolean).to.deep.equal(data.boolean, msg);
+    }
```
lawrence-forooghian, line 105: nit: It seems like if none of these `if`s match then it's a logic error — might be worth throwing an error

test/rest/liveobjects.test.js
```diff
@@ -0,0 +1,732 @@
...
+    return helper.AblyRest({ ...options, plugins: { LiveObjects: LiveObjectsPlugin } });
+  }
+
+  function expectInstanceOf(object, className, msg) {
+    // esbuild changes the name for classes with static method to include an underscore as prefix.
+    // so LiveMap becomes _LiveMap. we account for it here.
+    expect(object.constructor.name).to.match(new RegExp(`_?${className}`), msg);
+  }
+
+  function checkKeyDataOnExpandedObject({ helper, keyData, obj, msg }) {
```
lawrence-forooghian: A bit of documentation here would be helpful — there are so many different ways of representing data that has something to do with a LiveMap, and we don't have the type system to help us understand what this method aims to do.

test/rest/liveobjects.test.js
```diff
@@ -0,0 +1,732 @@
...
+
+            // check referenced objects
+            expect(obj.map.entries.emptyCounter.data.counter.data.number).to.equal(0);
+            expect(obj.map.entries.initialValueCounter.data.counter.data.number).to.equal(10);
+            expect(obj.map.entries.emptyMap.data.map.entries).to.deep.equal({});
+          },
+        );
+      });
+
+      const primitiveKeyData = [
```
lawrence-forooghian: I think a comment would be useful here to explain what these elements represent (an abstract representation of a `RestObjectData`)

src/plugins/liveobjects/restobject.ts
```diff
@@ -0,0 +1,352 @@
...
+          path: op.path,
+          counterInc: { number: op.amount },
+        };
+
+      default:
+        throw new this._channel.client.ErrorInfo('Unsupported publish operation action: ' + operation, 40003, 400);
+    }
+  }
+
+  private _encodePrimitive(value: PrimitiveOrObjectReference, format: Utils.Format): any {
```
lawrence-forooghian: Can this not return `WireRestObjectData` instead of `any`?

src/plugins/liveobjects/restobject.ts
```diff
@@ -0,0 +1,352 @@
...
+        params ?? {},
+        envelope,
+        true,
+      );
+
+      let decodedBody: RestCompactObjectData | WireRestLiveObject | undefined;
+      if (format) {
+        decodedBody = client.Utils.decodeBody(response.body, client._MsgPack, format);
+      } else {
+        decodedBody = response.body;
```
lawrence-forooghian: This branch looks like dead code, at least from the types?

src/plugins/liveobjects/restobject.ts
```diff
@@ -0,0 +1,352 @@
...
+      }
+
+      if (decodedBody == undefined) {
+        return undefined;
+      }
+
+      // non-object primitive values
+      if (typeof decodedBody !== 'object') {
```
lawrence-forooghian: Might be worth having a test that `get()` works with leaf values?

src/plugins/liveobjects/restobject.ts
```diff
@@ -0,0 +1,352 @@
...
+      }
+
+      // live counter or JSON object
+      if (!this._isWireLiveMap(decodedBody)) {
+        return decodedBody;
+      }
+
+      return this._decodeWireLiveObject(decodedBody, format);
+    } catch (error) {
+      if (this._channel.client.Utils.isErrorInfoOrPartialErrorInfo(error) && error.code === 40400) {
```
lawrence-forooghian: I guess this means that the REST API treats both "path resolved to nothing" and "object ID doesn't exist" as equivalent? The latter _feels_ like an error that should propagate to the user, but I don't have any background on this decision.

liveobjects.d.ts
VeskeR: fixed and added package test in https://github.com/ably/ably-js/pull/2109/changes/b6dac5a5e05a5bd924baed1d064b64c40d7f751c

liveobjects.d.ts
VeskeR: To be honest, I just copied the phrasing from the Push plugin without giving it much thought. On second look, you're correct , the second part doesn't make sense for LiveObjects. Just removed it and added an explicit example how you would create BaseRest with LiveObjects plugin in https://github.com/ably/ably-js/pull/2109/changes/b6dac5a5e05a5bd924baed1d064b64c40d7f751c

liveobjects.d.ts
VeskeR: we can, probably slightly nicer as compact result is essentially JSON-like (apart from buffers). updated in https://github.com/ably/ably-js/pull/2109/changes/0deb2e405ce137facb3aa94df65642b570ea807e as part of wider API changes. (the type is called RestObjectGetCompactResult now)

liveobjects.d.ts
VeskeR: Agree not much point to keep it there. Removed from the union type and added a comment to the .get method instead, see https://github.com/ably/ably-js/pull/2109/changes/0deb2e405ce137facb3aa94df65642b570ea807e

liveobjects.d.ts
VeskeR: Sorry, a leftover from the previous api where we had a plan for it. Removed it in https://github.com/ably/ably-js/pull/2109/changes/0deb2e405ce137facb3aa94df65642b570ea807e

liveobjects.d.ts
VeskeR: Yeah, we cannot guarantee that we won't add anything else in the future, so probably not a good idea to restrict the type. There is now a comment near object data decoding that describes that we pass through unrecognized values https://github.com/ably/ably-js/pull/2109/changes/d022ab52160a51b8e8c60c7ca22de7c77f08780a#diff-cce6d163053922a04519676693e06c7a3eeeab840982c1ccd0b76999818fa65eR187-R189

liveobjects.d.ts
VeskeR: Removed the type in favour of a stricter `PublishObjectData` which has more verbose and explicit typing to forbid setting any other value fields, see https://github.com/ably/ably-js/pull/2109/changes/d022ab52160a51b8e8c60c7ca22de7c77f08780a.

src/plugins/liveobjects/restobject.ts
VeskeR: Done in https://github.com/ably/ably-js/pull/2109/changes/d022ab52160a51b8e8c60c7ca22de7c77f08780a

src/plugins/liveobjects/restobject.ts
VeskeR: Aligned API in https://github.com/ably/ably-js/pull/2109/changes/d022ab52160a51b8e8c60c7ca22de7c77f08780a. It now correctly reflects the existing REST API, which is mostly consistent with the realtime protocol. The main difference is that there is no `action` in REST API, there is an `id` for idempotent publishing (although this eventually will be supported in realtime too), and there is a `path` parameter for path based REST operations.

src/plugins/liveobjects/restobject.ts
VeskeR: Added to public types in https://github.com/ably/ably-js/pull/2109/changes/9520d356ddccf9107786de7deaf00f40e09a31b5.
Rest implementation now uses shared map semantics encoding/decoding helper function in https://github.com/ably/ably-js/pull/2109/changes/d022ab52160a51b8e8c60c7ca22de7c77f08780a

test/rest/liveobjects.test.js
VeskeR: Made numerous changes to tests, it looks quite sleek right now IMO. See https://github.com/ably/ably-js/pull/2109/changes/d022ab52160a51b8e8c60c7ca22de7c77f08780a.

test/rest/liveobjects.test.js
VeskeR: Reworded some helper functions and added some comments, see https://github.com/ably/ably-js/pull/2109/changes/d022ab52160a51b8e8c60c7ca22de7c77f08780a

test/rest/liveobjects.test.js
VeskeR: this is now shared between rest/realtime tests in a helper, with comments. see https://github.com/ably/ably-js/pull/2109/changes/c0e7a375cb781b7536f2b0e34022b7063b3831d3

src/plugins/liveobjects/restobject.ts
VeskeR: Scrapped the whole encoding/decoding logic from Rest implementation as it was mostly a duplicate of what we're doing for realtime in ObjectMessage.
Refactored ObjectMessage a bit in https://github.com/ably/ably-js/pull/2109/changes/a1efc2f4da95971c69a74fc0c4aed1386c3f61b0 to expose some helper functions that are now used by Rest in https://github.com/ably/ably-js/pull/2109/changes/d022ab52160a51b8e8c60c7ca22de7c77f08780a. The whole code flow is much cleaner now

src/plugins/liveobjects/restobject.ts
VeskeR: Sorry, it was meant to be if format is msgpack, then msgpack decode. Fixed in https://github.com/ably/ably-js/pull/2109/changes/d022ab52160a51b8e8c60c7ca22de7c77f08780a

src/plugins/liveobjects/restobject.ts
VeskeR: Handling of return types is now done in a clearer way in _decodeNonCompactResult https://github.com/ably/ably-js/pull/2109/changes/d022ab52160a51b8e8c60c7ca22de7c77f08780a#diff-cce6d163053922a04519676693e06c7a3eeeab840982c1ccd0b76999818fa65eR177-R190, and there is now a bunch of tests for things like `'"compact: false" returns decoded object data for string leaf',` for compact/non-compact views. See https://github.com/ably/ably-js/pull/2109/changes/d022ab52160a51b8e8c60c7ca22de7c77f08780a

src/plugins/liveobjects/restobject.ts
VeskeR: > I guess this means that the REST API treats both "path resolved to nothing" and "object ID doesn't exist" as equivalent?

Tested this, yes. Non-existed objectid/path results in the same error unfortunately.

> The latter feels like an error that should propagate to the user, but I don't have any background on this decision.

Don't have strong opinion here, but I think the original idea to return `undefined` was not explicilty required and was just a copy-over from realtime, where most of the access API returns `undefined`. After thinking about it, I decided to remove undefined from the types and try/catch block. Now both errors will propagate to the user. Changed this and added two test in https://github.com/ably/ably-js/pull/2109/changes/d022ab52160a51b8e8c60c7ca22de7c77f08780a

PS: The error from REST API looks like: `Error: no object found at path ${path} from object ID ${objectId}`

test/rest/liveobjects.test.js
VeskeR: Fixed in https://github.com/ably/ably-js/pull/2109/commits/a063708786c7828aaec69d08a8b942fcd375dfec

src/plugins/liveobjects/restobject.ts
VeskeR: Also added the same for public OperationAction enum while at it https://github.com/ably/ably-js/pull/2109/changes/8f90fe9f69ebb6f862bd1ddef98d10dbcae78b1e, same reasoning (will leave this one as a separate commit, before REST SDK commit)

src/plugins/liveobjects/objectmessage.ts
```diff
@@ -481,6 +481,13 @@ export function decodeWireObjectData(
       return { string: wireData.string };
     }
 
+    // unrecognized ObjectData shape - pass through as-is so we don't lose any data
```
lawrence-forooghian, line 492: There's no behavioural change in c15c06f — seems like it's still returning `{}`. Also, I'm not convinced by the intent of the commit — i.e. passing through raw wire data as if it's been decoded. What if the new fields have some decoding rule that we're now skipping? Is it really better to return non-decoded data than to return nothing?

test/rest/liveobjects.test.js
lawrence-forooghian: It seems like 5cd4e089 fixes something that was broken in c0e7a375? If so, it should probably be squashed at the end too

lawrence-forooghian: Ditto the helper changes from 2d93846a can be squashed into c0e7a37 since they just undo something that was introduced there, I think?

liveobjects.d.ts
```diff
@@ -67,7 +68,444 @@ export type ObjectsEventCallback = () => void;
...
+   * (`bytes` decoded to `Buffer`/`ArrayBuffer`, `json` decoded to native objects/arrays).
+   * If the path resolves to a leaf value in a map, returns the decoded {@link RestObjectData | ObjectData} directly.
+   *
+   * Cyclic references are included as `{ objectId: string }` rather than including the same object instance
+   * in the result more than once.
+   *
+   * @param params - Parameters specifying the object to fetch with `compact: false`.
+   * @returns A promise that resolves to the object data in full format.
+   */
+  get(params: RestObjectGetFullParams): Promise<RestObjectGetResult>;
```
lawrence-forooghian: Looking at this now, would `RestObjectGetFullResult` not be a better type name, for consistency
with the params type names and `RestObjectGetCompactResult`?

liveobjects.d.ts
```diff
@@ -67,7 +68,444 @@ export type ObjectsEventCallback = () => void;
...
+   * Returns a {@link RestObjectGetCompactResult} representing the logical structure of your data as a JSON-like value.
+   * {@link LiveMap} instances appear as JSON objects with their entries, and {@link LiveCounter} instances appear
+   * as numbers. Binary values appear as base64 strings (JSON protocol) or `Buffer`/`ArrayBuffer` (binary protocol).
+   * JSON-typed values remain as their JSON-encoded string representation.
+   *
+   * Cyclic references are included as `{ objectId: string }` rather than including the same object instance
+   * in the result more than once.
+   *
+   * @param params - Optional parameters to specify the object to fetch.
+   * @returns A promise that resolves to the object data in compact format.
```
lawrence-forooghian: I think these all need the standard throws-`ErrorInfo` e.g. "Upon failure, the promise will be rejected with an {@link ErrorInfo} object which explains the error."

src/plugins/liveobjects/objectmessage.ts
VeskeR: > There's no behavioural change in https://github.com/ably/ably-js/commit/c15c06fefb3ace52c77cf9bd133c98f7bbe006ed — seems like it's still returning {}

Oops, must've lost during changes. Fixed in https://github.com/ably/ably-js/pull/2109/commits/d334f8d89b2637fb2f72bb333a640be3e7b78e50

> Also, I'm not convinced by the intent of the commit — i.e. passing through raw wire data as if it's been decoded. What if the new fields have some decoding rule that we're now skipping? Is it really better to return non-decoded data than to return nothing?

I believe this is better, and it's not entirely different from what we do if said decoding of data actually fails. We have an established procedure if decoding fails to just return the data as is, as seen for [message](https://github.com/ably/ably-js/blob/0035db0ebcec105ae510bcbbe5a605c0edde9f12/src/common/lib/types/message.ts#L209-L234), [presence](https://github.com/ably/ably-js/blob/0035db0ebcec105ae510bcbbe5a605c0edde9f12/src/common/lib/types/presencemessage.ts#L133-L145), [annotation ](https://github.com/ably/ably-js/blob/0035db0ebcec105ae510bcbbe5a605c0edde9f12/src/common/lib/types/annotation.ts#L106-L118) decodings. The same approach was adopted for [object message data decoding](https://github.com/ably/ably-js/blob/0035db0ebcec105ae510bcbbe5a605c0edde9f12/src/plugins/liveobjects/objectmessage.ts#L799-L810). This change only extends this pattern to the unexpected typed fields on `ObjectData` - which you could argue is a decoding failure, hence should be just logged and returned as is.

liveobjects.d.ts
VeskeR: (let's see how efficiently Claude can do that)

liveobjects.d.ts
VeskeR: Suuure. I didn't have `RestObjectGetFullResult` / `RestObjectGetCompactResult` and was using `Omit<RestObjectGetParams, 'compact> & { compact: true }` type, but typedoc was complaining about undocumented `compact` field in those (sigh), so had to change to named interfaces. Will update

liveobjects.d.ts
VeskeR: https://github.com/ably/ably-js/pull/2109/commits/98ec148136c4d543cc61b02a5816c6631727715f

liveobjects.d.ts
VeskeR: https://github.com/ably/ably-js/pull/2109/commits/98ec148136c4d543cc61b02a5816c6631727715f

src/plugins/liveobjects/objectmessage.ts
lawrence-forooghian: OK, cool, I don't have very strong feelings about it. And I assume dd34f8d will be squashed into c1506f?

VeskeR: > And I assume dd34f8d will be squashed into c1506f?

Yes, once we're done with changes I'll clean up the commit history

src/plugins/liveobjects/restobject.ts
```diff
@@ -0,0 +1,231 @@
...
+          format,
+        );
+
+    const compact = params?.compact ?? true;
+    if (compact) {
+      // Compact mode: return as-is. Values are JSON-like; bytes appear as base64 strings
+      // (JSON protocol) or Buffer/ArrayBuffer (binary protocol). We cannot deterministically
+      // decode values since we can't tell string vs JSON-encoded string.
+      return decoded as RestObjectGetCompactResult;
+    }
```
lawrence-forooghian, line 132: Could we change the naming and comment here to reflect the fact that this is also used for decoding map entries, not just top-level response bodies?

src/plugins/liveobjects/restobject.ts
```diff
@@ -0,0 +1,231 @@
...
+  private _constructWireOperations(op: AnyRestObjectOperation, format: Utils.Format): WireRestObjectOperation {
+    const { id, path, mapCreate, ...rest } = op;
+
+    // Build the operation fields for encoding. If mapCreate is present, convert semantics
+    // from public string to internal enum before passing to the encoding pipeline.
+    const operationFields: Partial<ObjectOperation<ObjectData>> = mapCreate
+      ? {
+          ...rest,
+          mapCreate: { ...mapCreate, semantics: encodeMapSemantics(mapCreate.semantics, this._channel.client) },
+        }
```
lawrence-forooghian, line 274: Can we update the comment for this method to indicate that it's also used for REST now (the comment just mentions the creation of `initialValue`)?

test/rest/liveobjects.test.js
```diff
@@ -0,0 +1,892 @@
...
+      expect(entryData.boolean, label).to.equal(jsonData.boolean);
+    } else {
+      throw new Error(
+        `checkFullObjectKeyData: unrecognized jsonData shape for key "${key}": ${JSON.stringify(jsonData)}`,
+      );
+    }
+  }
+
+  /**
+   * Checks that the compact (from "compact: false" responses) value for a key matches the expected value,
```
lawrence-forooghian: should be `compact: true`?

test/rest/liveobjects.test.js
```diff
@@ -0,0 +1,892 @@
...
+   * accounting for protocol-dependent bytes handling.
+   */
+  function checkCompactKeyData({ helper, keyData, value, msg }) {
```
lawrence-forooghian: for understandability could we make this consistent with `checkFullObjectKeyData` and have it also accept the container and extract the value for that key?

test/rest/liveobjects.test.js
```diff
@@ -0,0 +1,892 @@
...
+    const label = msg || `compact value for "${keyData.key}"`;
+
+    if (keyData.jsonData.bytes != null) {
+      // bytes values appear as base64 strings (JSON protocol) or buffers (binary protocol)
+      if (BufferUtils.isBuffer(value)) {
+        helper.recordPrivateApi('call.BufferUtils.base64Decode');
+        helper.recordPrivateApi('call.BufferUtils.areBuffersEqual');
+        expect(BufferUtils.areBuffersEqual(value, BufferUtils.base64Decode(keyData.compactValue)), label).to.be.true;
```
lawrence-forooghian: The comment on `primitiveKeyData` that it contains the "expected values when read back in compact format via REST API" isn't quite true — it's the expected for compact _JSON_ format, right?

also can we have comments on these two `check*ObjectKeyData` functions explaining what `keyData` is (i.e. a member of `primitiveKeyData` from helper), for understandability?

test/rest/liveobjects.test.js
```diff
@@ -0,0 +1,892 @@
...
+                compact: false,
+              });
+              expect(result.counter).to.exist;
+              expect(result.counter.data.number).to.equal(20);
+            },
+          },
+
+          {
+            jsonMsgpack: true,
+            description: '"compact: true" returns all data types correctly',
```
lawrence-forooghian: maybe "returns all data types correctly for entries within a map path" to make it clear how it's different to the tests that follow? (ditto for the compact equivalent)

liveobjects.d.ts
```diff
@@ -67,7 +68,444 @@ export type ObjectsEventCallback = () => void;
...
+        }
+      >;
+    };
+  };
+
+/**
+ * Operation to create a new map object with a client-generated object ID and initial entries.
+ */
+export type RestObjectOperationMapCreateWithObjectId = RestObjectOperationBase &
+  TargetByObjectId & {
```
lawrence-forooghian: Do we have some documentation we can link to on how the user is meant to construct the object ID in this case in order for it to be accepted by the server? It doesn't seem like they'll figure this out themselves.

liveobjects.d.ts
```diff
@@ -67,7 +68,444 @@ export type ObjectsEventCallback = () => void;
...
+   * @returns A promise which, upon success, will be fulfilled with the object data in the requested format. Upon failure, the promise will be rejected with an {@link ErrorInfo} object which explains the error.
+   */
+  get(params?: RestObjectGetParams): Promise<RestObjectGetCompactResult | RestObjectGetFullResult>;
+
+  /**
+   * Publishes one or more operations to modify objects on the channel.
+   * When an array is provided, all operations are published as an atomic batch.
+   *
+   * Operations can target objects using `objectId`, `path`, or neither (for create operations that create standalone objects):
+   * - `objectId`): The unique identifier of the object instance to create or update.
```
lawrence-forooghian: stray `)`

src/plugins/liveobjects/restobject.ts
VeskeR: For some reason some comments are for the wrong lines, but I assume you refer to `_decodeNonCompactResult` here. Sure, will think of something

src/plugins/liveobjects/restobject.ts
VeskeR: Also wrong line, I assume it's about `encodePartialObjectOperationForWire`. Will do

test/rest/liveobjects.test.js
VeskeR: Yeah, will fix a typo

test/rest/liveobjects.test.js
VeskeR: Compact responses could be raw unwrapped values, without a container.
Think `object.get({ path: 'counter' })` <- you will get just a number. Hence, this check accepts just a `value`

test/rest/liveobjects.test.js
VeskeR: Sure, will update comments

test/rest/liveobjects.test.js
VeskeR: Sounds good, I wasn't sure about the name for this test either after we added "type-path" specific ones below. Will change

liveobjects.d.ts
VeskeR: Part of the "Using the REST API/SDK" docs at https://ably.com/docs/liveobjects/rest-api-usage#client-generated-ids.
That said, there is a comment from Mike in REST SDK docs https://github.com/ably/docs/pull/3258#issuecomment-4053558952 suggesting:
> I think we should provide a helper method for constructing an object ID given an initial value arg passed like e.g. JSON.stringify()`

I think it is reasonable to do for the dedicated REST SDK. Just give one more reason for people to actually usse SDK instead of the raw API.
So expect some changes regarding `*MapCreateWithObjectId` ops

liveobjects.d.ts
VeskeR: 👍

src/plugins/liveobjects/restobject.ts
lawrence-forooghian: Sorry about the wrong lines — was trying to submit comments via my editor, clearly hasn't worked out too well

test/rest/liveobjects.test.js
lawrence-forooghian: The only time it's used in the tests though it's used to check the entries of a map though, no?

src/plugins/liveobjects/restobject.ts
VeskeR: Updated in https://github.com/ably/ably-js/pull/2109/changes/1b07db642742a84e736bc818b90914dc012a5c88

src/plugins/liveobjects/restobject.ts
VeskeR: Updated in https://github.com/ably/ably-js/pull/2109/changes/1b07db642742a84e736bc818b90914dc012a5c88

test/rest/liveobjects.test.js
VeskeR: fixed in https://github.com/ably/ably-js/pull/2109/changes/1b07db642742a84e736bc818b90914dc012a5c88

test/rest/liveobjects.test.js
VeskeR: changed in https://github.com/ably/ably-js/pull/2109/changes/1b07db642742a84e736bc818b90914dc012a5c88

test/rest/liveobjects.test.js
VeskeR: updated in https://github.com/ably/ably-js/pull/2109/changes/1b07db642742a84e736bc818b90914dc012a5c88

test/rest/liveobjects.test.js
VeskeR: updated in https://github.com/ably/ably-js/pull/2109/changes/1b07db642742a84e736bc818b90914dc012a5c88

liveobjects.d.ts
VeskeR: fixed in https://github.com/ably/ably-js/pull/2109/changes/1b07db642742a84e736bc818b90914dc012a5c88

liveobjects.d.ts
VeskeR: See the new `RestObject.generateObjectId` helper method in https://github.com/ably/ably-js/pull/2109/commits/2d973c6cf7b2db82ed79d3544a2323c053ddac38, added based on Mike's suggestion in docs PR https://github.com/ably/docs/pull/3258#issuecomment-4053558952.
`objectId` for RestObjectOperationMapCreateWithObjectId (and counter) now tells the user to use that helper method.

lawrence-forooghian (commented):
LGTM — happy to resolve / approve once commits sorted

VeskeR:
> LGTM — happy to resolve / approve once commits sorted

Cleaned up commit history

PR #2119, opened 2025-12-03
[AIT-119] Implicit attach on channel.object.get()

VeskeR:
Resolves [AIT-119](https://ably.atlassian.net/browse/AIT-119)

[AIT-119]: https://ably.atlassian.net/browse/AIT-119?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Fetching realtime objects on unattached channels now implicitly attaches and waits for synchronization; clearer errors when subscription mode is disallowed.

* **Behavior Changes**
  * Presence sync wait converted to a Promise-based API (no callback).
  * Presence retrieval returns current state immediately when not waiting, and raises a clear error if suspended while a sync is requested.
  * Channel operations now ensure attachment before proceeding.

* **Tests**
  * Added tests for implicit attach-and-sync and expanded channel-mode error assertions.

### DIFF
--- a/src/plugins/objects/realtimeobject.ts
+++ b/src/plugins/objects/realtimeobject.ts
@@ -80,7 +80,10 @@ export class RealtimeObject {
    * This is useful when working with multiple channels with different underlying data structure.
    */
   async get<T extends Record<string, API.Value>>(): Promise<API.PathObject<API.LiveMap<T>>> {
-    this.throwIfInvalidAccessApiConfiguration(); // RTO1a, RTO1b
+    this._throwIfMissingChannelMode('object_subscribe');
+
+    // implicit attach before proceeding
+    await this._channel.ensureAttached();
 
     // if we're not synced yet, wait for sync sequence to finish before returning root
     if (this._state !== ObjectsState.synced) {
--- a/test/realtime/objects.test.js
+++ b/test/realtime/objects.test.js
@@ -429,6 +429,34 @@ define(['ably', 'shared_helper', 'chai', 'objects', 'objects_helper'], function 
         }, client);
       });
 
+      /** @nospec */
+      it('RealtimeObject.get() on unattached channel implicitly attaches and waits for sync', async function () {
+        const helper = this.test.helper;
+        const client = RealtimeWithObjects(helper);
+
+        await helper.monitorConnectionThenCloseAndFinishAsync(async () => {
+          const channel = client.channels.get('channel', channelOptionsWithObjects());
+          expect(channel.state).to.equal('initialized', 'Channel should be in initialized state');
+
+          // Set up a timeout to catch if get() hangs
+          const timeoutPromise = new Promise((_, reject) => {
+            setTimeout(() => reject(new Error('RealtimeObject.get() timed out')), 10000);
+          });
+
+          // Call get() on unattached channel - this should automatically attach and resolve
+          const getPromise = channel.object.get();
+
+          // Race between get() and timeout - get() should win by implicitly attaching and syncing state
+          const entryPathObject = await Promise.race([getPromise, timeoutPromise]);
+
+          // Channel should now be attached, and root object returned
+          expect(channel.state).to.equal('attached', 'Channel should be attached after RealtimeObject.get() call');
+
+          expectInstanceOf(entryPathObject, 'DefaultPathObject', 'entrypoint should be of DefaultPathObject type');
+          expect(entryPathObject.instance().id).to.equal('root', 'entrypoint should have an object id "root"');
+        }, client);
+      });
+
       function checkKeyDataOnPathObject({ helper, key, keyData, pathObject, msg }) {
         if (keyData.data.bytes != null) {
           helper.recordPrivateApi('call.BufferUtils.base64Decode');
@@ -7089,8 +7117,6 @@ define(['ably', 'shared_helper', 'chai', 'objects', 'objects_helper'], function 
       });
 
       const expectAccessApiToThrow = async ({ realtimeObject, map, counter, errorMsg }) => {
-        await expectToThrowAsync(async () => realtimeObject.get(), errorMsg);
-
         expect(() => counter.value()).to.throw(errorMsg);
 
         expect(() => map.get('key')).to.throw(errorMsg);
@@ -7156,6 +7182,7 @@ define(['ably', 'shared_helper', 'chai', 'objects', 'objects_helper'], function 
               expectBatchContextWriteApiToThrow({ ctx, errorMsg: '"object_publish" channel mode' });
             });
 
+            await expectToThrowAsync(async () => realtimeObject.get(), '"object_subscribe" channel mode');
             await expectAccessApiToThrow({ realtimeObject, map, counter, errorMsg: '"object_subscribe" channel mode' });
             await expectWriteApiToThrow({ entryInstance, map, counter, errorMsg: '"object_publish" channel mode' });
           },
@@ -7178,6 +7205,7 @@ define(['ably', 'shared_helper', 'chai', 'objects', 'objects_helper'], function 
               expectBatchContextWriteApiToThrow({ ctx, errorMsg: '"object_publish" channel mode' });
             });
 
+            await expectToThrowAsync(async () => realtimeObject.get(), '"object_subscribe" channel mode');
             await expectAccessApiToThrow({ realtimeObject, map, counter, errorMsg: '"object_subscribe" channel mode' });
             await expectWriteApiToThrow({ entryInstance, map, counter, errorMsg: '"object_publish" channel mode' });
           },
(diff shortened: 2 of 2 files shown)


src/plugins/objects/realtimeobject.ts
```diff
@@ -80,11 +80,24 @@ export class RealtimeObject {
    * This is useful when working with multiple channels with different underlying data structure.
    */
   async get<T extends Record<string, API.Value>>(): Promise<API.PathObject<API.LiveMap<T>>> {
-    this.throwIfInvalidAccessApiConfiguration(); // RTO1a, RTO1b
+    this._throwIfMissingChannelMode('object_subscribe');
+
+    let syncedPromise;
+    // grab the synced promise now before attaching to avoid race condition
```
lawrence-forooghian: Could you please elaborate on what this is trying to achieve? What's the sequence of events that this is accounting for?

src/plugins/objects/realtimeobject.ts
```diff
@@ -486,4 +499,20 @@ export class RealtimeObject {
...
+      case 'attached':
+      case 'suspended':
+        break;
+      case 'initialized':
+      case 'detached':
+      case 'detaching':
+      case 'attaching':
+        await this._channel.attach();
+        break;
+      default:
```
lawrence-forooghian: I think this is easier to understand if it explicitly says `FAILED`.

src/plugins/objects/realtimeobject.ts
```diff
@@ -486,4 +499,20 @@ export class RealtimeObject {
       );
     }
   }
+
+  private async _waitAttached(): Promise<void> {
```
lawrence-forooghian: I don't think the name is an accurate description of what this does; it is more than just a "wait".

src/plugins/objects/realtimeobject.ts
```diff
@@ -80,11 +80,24 @@ export class RealtimeObject {
...
+    this._throwIfMissingChannelMode('object_subscribe');
+
+    let syncedPromise;
+    // grab the synced promise now before attaching to avoid race condition
+    if (this._state !== ObjectsState.synced) {
+      syncedPromise = this._eventEmitterInternal.once(ObjectsEvent.synced);
+    }
+
+    // implicit attach before proceeding
+    await this._waitAttached();
```
lawrence-forooghian: Presence has additional handling for `SUSPENDED`; do we not want that?

src/plugins/objects/realtimeobject.ts
lawrence-forooghian: Also if the intention is that the logic be the same as presence, can we put this logic somewhere shared?

lawrence-forooghian: It also seems like this logic (which looks copied from presence) doesn't align with that of [RTP11b](https://sdk.ably.com/builds/ably/specification/main/features/#RTP11b):

> Implicitly attaches the RealtimeChannel if the channel is in the INITIALIZED state. However, if the channel is in or enters the DETACHED or FAILED state before the operation succeeds, it will result in an error

— we're attaching in more states than just `INITIALIZED`

all the more reason to make this logic shared with presence, so that if / when we align it with the spec we just have to do it once?

src/plugins/objects/realtimeobject.ts
VeskeR: Huh, doesn't really makes sense as it is now, I agree. I think  I had a WIP version that could've had a race condition, but looking at this right now it can't happen.
Removed this unnecessary promise creation in https://github.com/ably/ably-js/compare/168862755af2cab184d382e724638a458b66ee1a..1d1738a2db36a3db178d6e9804ec772a40e97b89

src/plugins/objects/realtimeobject.ts
VeskeR: sure ,added `failed` in https://github.com/ably/ably-js/compare/1d1738a2db36a3db178d6e9804ec772a40e97b89..e54f2829f8e99a108cb4dc3ba557174acf31c36e

src/plugins/objects/realtimeobject.ts
VeskeR: Contrary to presence, LiveObjects preserves the latest known local state in `suspended` state and returns that to the user. Only Write API throws errors when in `suspended`. As such, we don't throw for access API when in `suspended` and can just return the root object.
Based on DR https://ably.atlassian.net/wiki/spaces/LOB/pages/3784933378/LODR-032+Realtime+Client+behavior+under+channel+states

src/plugins/objects/realtimeobject.ts
VeskeR: Changed the name to `ensureAttached` and made it shared between presence and objects (added new method to RealtimeChannel), see https://github.com/ably/ably-js/pull/2119/commits/644223b2d438123a667afed796cee043b77c1398

Not really sure about [RTP11b](https://sdk.ably.com/builds/ably/specification/main/features/#RTP11b). Don't see any obvious comments in ably-js to indicate why behavior here is different. Not really something I'd want to look into right now though

src/plugins/objects/realtimeobject.ts
lawrence-forooghian: Can we get rid of the `default`? If it's not actually covering any cases then I think it's confusing to the reader.

src/common/lib/client/realtimechannel.ts
```diff
@@ -1016,6 +1016,23 @@ class RealtimeChannel extends EventEmitter {
   async status(): Promise<API.ChannelDetails> {
     return this.client.rest.channelMixin.status(this);
   }
+
+  async ensureAttached(): Promise<void> {
```
lawrence-forooghian, line 1028: Could we have a documentation comment here, please? It's quite a generic name but it's intended for a pretty specific scenario (especially once we align this further with the spec).

src/plugins/objects/realtimeobject.ts
lawrence-forooghian: Ah fair enough. From what I can see the DR doesn't explain why we chose to deviate from presence for `SUSPENDED`; do you remember why we did?

src/plugins/objects/realtimeobject.ts
lawrence-forooghian: Yeah I think it's fine to just keep the implementation consistent with presence for this, but the intention is that the behaviour be consistent with that of presence, right? i.e. that's what we'll write in the spec

lawrence-forooghian (commented):
Please could you either restructure or squash the commits? currently we have a commit where we add some duplicated logic and then another where we remove what we just added

src/plugins/objects/realtimeobject.ts
VeskeR: I'd rather keep it in (very unlikely) case we introduce new channel states. Otherwise it will function as if the new state is equal to `attached`

lawrence-forooghian: I'd argue that if we add a new channel state then there's no getting around auditing all the places in the codebase that we check the channel state. But if having a `default` is the pattern we're using (and I guess that TypeScript doesn't give us exhaustive checking for channel state) then sure

src/plugins/objects/realtimeobject.ts
VeskeR: > but the intention is that the behaviour be consistent with that of presence, right

yes, the implicit attach behavior should be consistent with presence

src/plugins/objects/realtimeobject.ts
VeskeR: uh, I think we just decided that for now we don't need `waitForSync`-like behavior for LiveObjects, as currently it functions identical to presence if  you set `waitForSync=false`. Essentially you still have the snapshot of the last know state of objects on a channel and access methods work, you just can't publish any operations. If a developer would like to hide that data too, they can observe channel state changes and react to `suspended` directly.

src/plugins/objects/realtimeobject.ts
VeskeR: > TypeScript doesn't give us exhaustive checking for channel state

yeah, not something you have there by default, though there is a way to have an type-safe exhaustive enum check with some clever use of `never` but it will have marginal benefit here as all other places just don't do that and will need to checked explicitly if we add a new state anyway.

src/common/lib/client/realtimechannel.ts
VeskeR: Added comment in https://github.com/ably/ably-js/compare/650545d991ad244e6e3e76be12b8258ee1a43b72..2f065e4f340191ba3cc6bfe51b9487b003c73913

PR #2121, opened 2025-12-04
Remove deferral of Objects `SYNCED` events, and add sync event tests

lawrence-forooghian:
This does the following:

- removes the deferral of `SYNCED` events to the next event loop cycle (see commit message)
- adds tests for the `SYNCING` and `SYNCED` events based on the currently-implemented behaviour; will use this as a basis for https://github.com/ably/ably-liveobjects-swift-plugin/issues/80

## Summary by CodeRabbit

* **Refactor**
  * Improved object synchronization efficiency by streamlining internal state management, eliminating deferred event signaling in favor of immediate execution for enhanced responsiveness.

* **Tests**
  * Added comprehensive test suite for object synchronization events, validating sequences across multiple channel attachment scenarios and edge cases to ensure reliability.

### DIFF
--- a/test/realtime/objects.test.js
+++ b/test/realtime/objects.test.js
@@ -5548,12 +5548,17 @@ define(['ably', 'shared_helper', 'chai', 'objects', 'objects_helper'], function 
         },
 
         {
-          description: 'ATTACHED with HAS_OBJECTS false once SYNCED should not provoke further events',
+          description: 'ATTACHED with HAS_OBJECTS false once SYNCED emits SYNCING and then SYNCED',
           channelEvents: [
             { type: 'attached', hasObjects: false },
             { type: 'attached', hasObjects: false },
           ],
-          expectedSyncEvents: ['syncing', 'synced'],
+          expectedSyncEvents: [
+            'syncing',
+            'synced', // The initial SYNCED
+            'syncing',
+            'synced', // From the subsequent ATTACHED
+          ],
         },
 
         {
--- a/src/plugins/objects/objects.ts
+++ b/src/plugins/objects/objects.ts
@@ -268,11 +268,7 @@ export class Objects {
     );
 
     // RTO4a
-    if (hasObjects || this._state == ObjectsState.initialized) {
-      // should always start a new sync sequence if we're in the initialized state, no matter the HAS_OBJECTS flag value.
-      // this guarantees we emit both "syncing" -> "synced" events in that order.
-      this._startNewSync();
-    }
+    this._startNewSync();
 
     // RTO4b
     if (!hasObjects) {


test/realtime/objects.test.js
```diff
@@ -5518,5 +5518,203 @@ define(['ably', 'shared_helper', 'chai', 'objects', 'objects_helper'], function
...
+            { type: 'attached', hasObjects: true },
+            { type: 'objectSync', channelSerial: 'foo:1' },
+            { type: 'objectSync', channelSerial: 'foo:2' },
+            { type: 'objectSync', channelSerial: 'bar:1' },
+          ],
+          expectedSyncEvents: ['syncing'],
+        },
+      ];
+
+      forScenarios(this, syncEventsScenarios, async function (helper, scenario) {
```
VeskeR: I'd suggest to always use `clientOptions` and `channelName` provided passed from `forScenarios` anyway. Client options are used for testing on protocols and transports and channel name includes description to make sure we don't accidentally share state between tests.
Even though not really relevant for these tests right now, it's a good idea to have it consistent in case we decide to change how tests work and need that functionality

test/realtime/objects.test.js
```diff
@@ -5518,5 +5518,203 @@ define(['ably', 'shared_helper', 'chai', 'objects', 'objects_helper'], function
...
+
+        {
+          description:
+            'The first ATTACHED should always provoke a SYNCING even when HAS_OBJECTS is false, so that the SYNCED is preceded by SYNCING',
+          channelEvents: [{ type: 'attached', hasObjects: false }],
+          expectedSyncEvents: ['syncing', 'synced'],
+        },
+
+        {
+          description: 'ATTACHED with HAS_OBJECTS false once SYNCED should not provoke further events',
```
VeskeR: Hmm, I wonder if this should be expected behavior.
When we receive HAS_OBJECTS false, it means that state is empty on a channel, and we will clear anything we currently have locally. This means that we do technically modify local data, just as if we have received HAS_OBJECTS=true that changed something.
And for subsequent HAS_OBJECTS=true we decided to emit `syncing -> synced` again, so maybe we should too for HAS_OBJECTS=false?

VeskeR (changes requested):

test/realtime/objects.test.js
lawrence-forooghian: Sure, have added a commit for that. I don't really have opinions on it (as mentioned yesterday, I'm not sure it makes sense to emit SYNCING in _any_ of the cases where the sync data is updated in a one-shot fashion, but as we discussed we don't really have a great idea of how we intend for this API to be used, so agree that we should at least be consistent)

test/realtime/objects.test.js
lawrence-forooghian: Thanks for pointing that out, done.

PR #2124, opened 2025-12-09
[AIT-31] LiveObjects Path-based API migration guide

VeskeR:
Related to [AIT-31](https://ably.atlassian.net/browse/AIT-31)

[AIT-31]: https://ably.atlassian.net/browse/AIT-31?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive LiveObjects v2 migration guide covering PathObject-based APIs, runtime path resolution, migration steps, TypeScript updates, new patterns, and examples.

* **Bug Fixes**
  * Clarified subscription event descriptions to reflect the actual path where an object changed.

* **New Features**
  * Subscription API now returns a subscription handle and includes an explicit unregister method for safer listener management.

### DIFF
--- a/docs/migration-guides/v2/liveobjects.md
+++ b/docs/migration-guides/v2/liveobjects.md
@@ -14,7 +14,7 @@ Here's how to migrate your LiveObjects usage to the new PathObject-based API int
 
 1. [Understand `PathObject`](#understand-pathobject).
 2. [Update to v2.16 or later and handle breaking changes](#update-to-v216-or-later-and-handle-breaking-changes).
-3. (Optional) [Take advantage of new Objects features that v2.16 introduces](#take-advantage-of-new-objects-features-that-v216-introduces).
+3. (Optional) [Take advantage of new LiveObjects features that v2.16 introduces](#take-advantage-of-new-liveobjects-features-that-v216-introduces).
 4. (Optional) Check out [common migration patterns](#common-migration-patterns) for a quick reference.
 
 ## Understand `PathObject`
@@ -43,6 +43,40 @@ The changes below are split into:
 - [changes that only affect TypeScript users](#only-typescript-users)
 
 ### General changes
+
+#### Update LiveObjects plugin import
+
+The LiveObjects plugin import has changed in several ways:
+
+1. The import path has changed from `'ably/objects'` to `'ably/liveobjects'`
+2. The plugin is now a named export instead of a default export
+3. The plugin name has changed from `Objects` to `LiveObjects`, which also affects the key used in the `plugins` client option
+
+**Before:**
+
+```typescript
+import * as Ably from 'ably';
+import Objects from 'ably/objects';
+
+const client = new Ably.Realtime({
+  key: 'your-api-key',
+  plugins: { Objects },
+});
+```
+
+**After:**
+
+```typescript
+import * as Ably from 'ably';
+import { LiveObjects } from 'ably/liveobjects';
+
+const client = new Ably.Realtime({
+  key: 'your-api-key',
+  plugins: { LiveObjects },
+});
+```
+
+**Note:** If you're using the UMD bundle via a `<script>` tag, the global variable name is now `AblyLiveObjectsPlugin` instead of `AblyObjectsPlugin`.
 
 #### Update the entrypoint: `channel.objects` → `channel.object`
 
@@ -172,7 +206,7 @@ await root.set('user', map);
 **After:**
 
 ```typescript
-import { LiveCounter, LiveMap } from 'ably/objects';
+import { LiveCounter, LiveMap } from 'ably/liveobjects';
 
 const myObject = await channel.object.get();
 await myObject.set('visits', LiveCounter.create(0));
@@ -340,6 +374,37 @@ subscription1.unsubscribe();
 subscription2.unsubscribe();
 ```
 
+#### Replace `offAll()` with individual listener management
+
+The `offAll()` method has been removed from the `RealtimeObject` status event API. Instead, deregister listeners individually using either the subscription object returned by `.on()`, or by calling `.off(event, callback)` with the callback reference.
+
+**Before:**
+
+```typescript
+const channelObjects = channel.objects;
+
+channelObjects.on('synced', () => console.log('Synced 1'));
+channelObjects.on('synced', () => console.log('Synced 2'));
+
+// Unregister all listeners at once
+channelObjects.offAll();
+```
+
+**After:**
+
+```typescript
+const channelObject = channel.object;
+
+// Option 1: Use the subscription object returned by .on()
+const subscription1 = channelObject.on('synced', () => console.log('Synced 1'));
+subscription1.off();
+
+// Option 2: Use .off(event, callback) with a callback reference
+const onSynced2 = () => console.log('Synced 2');
+channelObject.on('synced', onSynced2);
+channelObject.off('synced', onSynced2);
+```
+
 #### Change usage of `objects.batch()` to `PathObject.batch()`/`Instance.batch()`
 
 The batch API, previously available at `channel.objects.batch()`, is now available as a `.batch()` method on any `PathObject` or `Instance` instead. It now supports object creation inside a batch function.
@@ -406,7 +471,7 @@ In most scenarios, using `PathObject` is recommended as it provides path-based o
 
 1. **Subscribe to a specific instance regardless of its location**: Instance subscriptions follow the object even if it moves within the hierarchy or is stored in different map keys.
 
-2. **Get the underlying object ID for REST API operations**: Each LiveMap and LiveCounter has a unique object ID (accessible via the `.id` property) that can be used with the Objects REST API.
+2. **Get the underlying object ID for REST API operations**: Each LiveMap and LiveCounter has a unique object ID (accessible via the `.id` property) that can be used with the LiveObjects REST API.
 
 **Before:**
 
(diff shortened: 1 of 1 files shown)


docs/migration-guides/v2/liveobjects.md
```diff
@@ -0,0 +1,614 @@
...
+The key difference with `PathObject` is that **operations resolve the path at runtime** when the method is called. This means:
+
+- **Obtaining a PathObject never fails** - even if nothing exists at that path yet:
+  ```typescript
+  const shape = myObject.get('shape'); // Always succeeds, even if 'shape' doesn't exist
+  ```
+
+- **Access methods return empty defaults** when the path doesn't resolve to an appropriate object at runtime:
+  ```typescript
+  // If 'visits' doesn't exist or isn't a LiveCounter
```
lawrence-forooghian:
```suggestion
  // If 'visits' doesn't exist or isn't a primitive or a LiveCounter
```

I think?

docs/migration-guides/v2/liveobjects.md
```diff
@@ -0,0 +1,614 @@
...
+const myObject = await channel.object.get();
+
+// Option 1: Use PathObject for path-based operations (recommended)
+await myObject.get('players').get('player1').set('score', 100);
+
+// Option 2: Get the explicit instance
+const player = myObject.get('players').get('player1').instance();
+// player is a LiveMap | undefined
+if (player) {
+  console.log(player.id()); // map:...
```
lawrence-forooghian: I don't know whether this applies to any other methods (I know that some can throw errors so it makes sense for them to be methods) but `id` seems like it should just be a property and not a method?

docs/migration-guides/v2/liveobjects.md
```diff
@@ -0,0 +1,614 @@
...
+const shape = myObject.get('shape').instance();
+if (shape) {
+  const compact = shape.compact(); // Same result
+}
+```
+
+The `.compact()` method automatically handles:
+
+- Nested LiveMaps and LiveCounters
+- Primitive values. Binary data is encoded as base64 strings
```
lawrence-forooghian: Is the expectation that the user has sufficient knowledge of their own schema to be able to distinguish between a string and base64-encoded binary data? Is there a reason we can't use actual binary data types?

docs/migration-guides/v2/liveobjects.md
VeskeR: you're right, fixed in https://github.com/ably/ably-js/compare/066b2cfa580af27ea287d08fcf62b947c9f385ba..8e7b06da2c3e77569d5338deaa9277adcfa854e5

docs/migration-guides/v2/liveobjects.md
VeskeR: `Instance` object can wrap a primitive value as well as a LiveObject. In this case, `id()` returns `undefined` as this is not a valid operation on a primitive value (doesn't have an ID). As such, this was left as method

docs/migration-guides/v2/liveobjects.md
VeskeR: The intention for a `compact()` method is to provide an easily JSON-encodable object to the user (for easy compares, logs, etc.). For binary data this means converting it to base64 strings.

lawrence-forooghian: So what you're saying is that the output of `compact()` is only intended for human consumption, and that it's not intended as an unambiguous machine-readable representation of an object hierarchy?

docs/migration-guides/v2/liveobjects.md
lawrence-forooghian: I'm not sure I understand the connection; why does it need to be a method in order to return `undefined` for primitive `Instance`s?

docs/migration-guides/v2/liveobjects.md
VeskeR: Based on a team call discussion yesterday, we may want to change the way `compact()` works and/or introduce a new method to handle JSON-encodable/machine-readable representations without any overlaps between them.
Created a separate issue for this: https://ably.atlassian.net/browse/AIT-193

docs/migration-guides/v2/liveobjects.md
```diff
@@ -0,0 +1,775 @@
...
+
+  // Subscribe to this instance, tracking it wherever it moves
+  player.subscribe(({ object, message }) => {
+    // Notified about changes to this specific player instance
+    // even if it's moved to a different key (e.g., from 'player1' to 'player2')
+  });
+
+  await player.set('score', 100);
+}
+```
```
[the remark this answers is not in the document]
VeskeR: fixed

docs/migration-guides/v2/liveobjects.md
```diff
@@ -0,0 +1,775 @@
...
+  // Assume 'visits' is a LiveCounter, not a LiveMap
+  const visits = myObject.get('visits').instance(); // Returns LiveCounter instance
+
+  // Calling LiveMap-specific methods returns empty defaults
+  for (const [key, value] of visits.entries()) {
+    // Empty iterator - loop body never executes
+  }
+
+  const size = visits.size(); // Returns undefined
+  ```
```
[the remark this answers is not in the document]
VeskeR: fixed

docs/migration-guides/v2/liveobjects.md
VeskeR: Oh, sorry, I see. Yeah, technically it can be a getter that checks the type of the underlying instance. We don't really do any  computations to have it as a method, but I think we just didn't consider to make it just a prop? The whole API surface for LiveObjects is methods. Even stuff like `.size()` is a method (but rightly so, it iterates through entries and performs additional checks whether to count them in or not, as opposed to just returning the number of entries in the internal map), so we didn't really think much of introducing properties to the API.

Do you feel like this is something we should consider for ably-js specifically, or can be left up to the specific SDK to decide how they want to expose id based on what is most idiomatic?

lawrence-forooghian: I think that individual SDKs can choose, will leave it to you to decide based on what we do elsewhere in JS.

docs/migration-guides/v2/liveobjects.md
lawrence-forooghian: I'll resolve this comment since the migration guide reflects the current state of things; we can update it later if we change the API

docs/migration-guides/v2/liveobjects.md
```diff
@@ -0,0 +1,775 @@
...
+    }),
+  }),
+);
+
+const compactRepresentation = myObject.get('gameState').compact();
+// Returns:
+// {
+//   "playerName": "Alice",
+//   "score": 100,                    // LiveCounter compacted to number
+//   "avatar": "AAAAAAAAAAA=",        // ArrayBuffer converted to base64 string
```
mschristensen: This needs updating after our discussion and the changes you're about to make

mschristensen (approved):

docs/migration-guides/v2/liveobjects.md
VeskeR: Had a look at other types we have ,more specifically classes with behavior and unique identifiers.
Found two examples: `Connection` and `LocalDevice` describe a type that is more than just a "data structure" and expose `id` as a prop, not a method.
As such, changed `id` for liveobjects API to also be a prop (getter in this instance as we need to do some checks), see: https://github.com/ably/ably-js/pull/2128.
Updated the guide to reflect the change, see: https://github.com/ably/ably-js/compare/8a291ea8eee55ea0441dafd2245786b7afe87548..8b901e01c96df6b6a5077e05e0286424932c10e6

docs/migration-guides/v2/liveobjects.md
VeskeR: Split `compact` to `compact` and `compactJson` in https://github.com/ably/ably-js/pull/2129, also updated this migration guide to reflect those changes

docs/migration-guides/v2/liveobjects.md
VeskeR: Compact method changes in question: https://github.com/ably/ably-js/pull/2129

I've updated the migration guide to reflect those changes, see: https://github.com/ably/ably-js/compare/8a291ea8eee55ea0441dafd2245786b7afe87548..8b901e01c96df6b6a5077e05e0286424932c10e6

src/plugins/liveobjects/realtimeobject.ts
```diff
@@ -32,10 +32,6 @@ const StateToEventsMap: Record<ObjectsState, ObjectsEvent | undefined> = {
 
 export type ObjectsEventCallback = () => void;
 
-export interface OnObjectsEventResponse {
```
ttypic: should these changes be part of this PR?

VeskeR: This diff specifically is an internal type duplicate and [this commit](https://github.com/ably/ably-js/commit/0e2063f583beec321c067eb363c306eab0073634) just removed it from the internal codebase.
We had the same named public type that was changed to `StatusSubscription`, and it's being used by the codebase directly now.
And the public type change is mentioned here:
https://github.com/ably/ably-js/pull/2124/changes#diff-f16182659ccbfe39d69f5b8ce319af1a0cc65725070ec11b6cabc76a40d8b787R631

VeskeR: oh, sorry, if you're asking why this change is in this PR at all - I noticed the dupe when writing the migration guide and didn't bother to open a separate PR such a small change that doesn't change anything anyway

VeskeR: moved to https://github.com/ably/ably-js/pull/2139

PR #2126, opened 2025-12-15
Readme tweak: re-add CDN

SimonWoolf:
I don't know why this was removed. This seem quite important. Plenty of people still just write scripts and pull in the hosted lib. (And LLMs are still recommending the v1 cdn url, because they can't see the v2 one!).

Also remove redundant usage line, and add note about how to instantiate the rest client.

## Summary by CodeRabbit

* **Documentation**
  * Replaced import-based JavaScript SDK initialization examples with an HTML CDN example for v2.
  * Added guidance showing REST client usage with an Ably.Rest(...) note and inline comment in the JS snippet to clarify the REST alternative.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,11 +59,10 @@ The JavaScript SDK is available as an [NPM module](https://www.npmjs.com/package
 npm install ably
 ```
 
-Run the following to instantiate a client:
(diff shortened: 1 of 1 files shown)


owenpearson (approved):
FYI it was removed in #2037, but I agree it's good to have this in the readme

PR #2127, opened 2025-12-15
Implement message append, protocol v5 with publish/update return values, and update/delete over realtime

SimonWoolf:
AIT-94

## Summary by CodeRabbit

* **New Features**
  * Publish now returns per-message serials; update/delete/append return version info.
  * Message append operation added.
  * Default protocol baseline bumped to v5.

* **Behavior / Bug Fixes**
  * Acknowledgements now deliver per-message publish responses to callers.
  * Publish/update/delete/append APIs consistently return result objects with serial/version details.

* **Tests**
  * Added realtime and REST tests for publish/update/delete/append (serial/version validation).
  * Removed an old bundling test.

### DIFF
--- a/test/rest/updates-deletes.test.js
+++ b/test/rest/updates-deletes.test.js
@@ -140,7 +140,7 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, Helper, async
     /**
      * Test deleteMessage with operation metadata
      *
-     * @spec RSL13a
+     * @spec RSL15a
      */
     it('Should delete a message (with operation metadata)', async function () {
       const helper = this.test.helper;


ably.d.ts
```diff
@@ -2600,18 +2626,39 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
    * @param operation - An optional {@link MessageOperation} object containing metadata about the update operation.
    * @param params - Optional parameters sent as part of the query string.
-   * @returns A promise which on success will be fulfilled, and on failure, rejected with an {@link ErrorInfo} object which explains the error.
+   * @returns A promise which, upon success, will be fulfilled with an {@link UpdateDeleteResult} object containing the serial of the new version of the message. Upon failure, the promise will be rejected with an {@link ErrorInfo} object which explains the error.
    */
-  updateMessage(message: Message, operation?: MessageOperation, params?: Record<string, any>): Promise<void>;
+  updateMessage(
+    message: Message,
+    operation?: MessageOperation,
+    params?: Record<string, any>,
```
lawrence-forooghian: The realtime variants of these methods still accept a `params` although they don't in the spec and they are presumably ignored.

cc @ttypic is the removal of these `params` from the Realtime variant a breaking change in Java? in ably-cocoa we can introduce an overload that doesn't take `params` and then document that the variant that takes `params` will ignore them in the realtime case (not great). We haven't marked these APIs as experimental or anything so can't _really_ just remove the params-accepting variant

SimonWoolf: > in ably-cocoa we can introduce an overload that doesn't take params and then document that the variant that takes params will ignore them in the realtime case

What'd be the point?

Like: right now there are no supported per-message params. So there is no user-visible behaviour difference between doing something with the params, in either rest or realtime apis, and not. But if at some point we discover a need for a per-message param, it's pretty likely we'll want to support it in both rest and realtime apis, no?

Right now there's no spec'd standard for conveying params in a realtime publish, only a REST publish, but the asymmetry is obviously weird and annoying. I'm tempted to say that we should just have params in the api and put in the spec that if the user provides params the sdk should encode them in a ProtocolMessage.params map<string, string>. We don't currently do anything with that field serverside for a MESSAGE protocolmessage, but if we need to we can just.. start, with no protocol change

lawrence-forooghian: Sure, my preference is to avoid any sort of breaking change to the realtime API — in that case can we revert the API change in the spec PR and reintroduce the `params` (I don't hugely care what we do with them — spec can just say to ignore them for now if we like)?

SimonWoolf: spec change https://github.com/ably/specification/pull/410 , implemented in https://github.com/ably/ably-js/pull/2127/commits/77e76e0f7d8e1388729c20dbdf3bfcb6c17e2338

PR #2150, opened 2026-01-16
[AIT-285] Fix rules for clearing buffered object operations during sync sequence

VeskeR:
Buffered object operations must be cleared on every ATTACHED message,
not at the start of a new OBJECT_SYNC sequence. If HAS_OBJECTS is set
on ATTACHED, the server will deliver a sync sequence following the
attachment, guaranteeing that the objects in that sequence include at
least all operations up to the attach point. If HAS_OBJECTS is not set,
the client performs an implicit sync sequence by transitioning to
SYNCING and immediately clearing local state.

Read more in LODR-058 [1] and the specification change [2].

Resolves [AIT-285](https://ably.atlassian.net/browse/AIT-285)

[1] https://ably.atlassian.net/wiki/x/NQAEHgE
[2] https://github.com/ably/specification/pull/416

[AIT-285]: https://ably.atlassian.net/browse/AIT-285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Objects now always re-synchronize when a channel becomes attached, ensuring object state is refreshed even on resume.
  * Buffered object operations are cleared at attach/start of sync and reliably applied then cleared at end of sync to avoid stale updates.

* **Tests**
  * Consolidated and expanded tests for attached scenarios (resumed vs non‑resumed), HAS_OBJECTS permutations, buffering, and sync sequencing.

### DIFF
--- a/src/common/lib/client/realtimechannel.ts
+++ b/src/common/lib/client/realtimechannel.ts
@@ -579,11 +579,15 @@ class RealtimeChannel extends EventEmitter {
             if (this._presence) {
               this._presence.onAttached(hasPresence);
             }
-            // the Objects tree needs to be re-synced
-            if (this._object) {
-              this._object.onAttached(hasObjects);
-            }
           }
+
+          // Must always resync the Objects tree after an ATTACHED.
+          // Whether there are objects on the channel and whether an OBJECT_SYNC sequence will follow
+          // is determined by the HAS_OBJECTS flag.
+          if (this._object) {
+            this._object.onAttached(hasObjects);
+          }
+
           const change = new ChannelStateChange(this.state, this.state, resumed, hasBacklog, message.error);
           if (!resumed || this.channelOptions.updateOnAttached) {
             this.emit('update', change);
--- a/src/plugins/liveobjects/realtimeobject.ts
+++ b/src/plugins/liveobjects/realtimeobject.ts
@@ -153,6 +153,13 @@ export class RealtimeObject {
   /**
    * @internal
    * @spec RTO5
+   *
+   * Note on server-initiated resync: if the realtime server needs to force a resync, it is expected
+   * to send an ATTACHED message before the new OBJECT_SYNC sequence. However, if an OBJECT_SYNC
+   * is received after a previously completed sync without a preceding ATTACHED,
+   * we handle it on a best-effort basis: enter the SYNCING state and start buffering OBJECT messages
+   * from that point. Since the buffer is cleared at the end of each completed sync sequence, receiving
+   * an OBJECT_SYNC while in the SYNCED state means we start with an empty buffer.
    */
   handleObjectSyncMessages(objectMessages: ObjectMessage[], syncChannelSerial: string | null | undefined): void {
     const { syncId, syncCursor } = this._parseSyncChannelSerial(syncChannelSerial); // RTO5a
@@ -200,13 +207,19 @@ export class RealtimeObject {
       `channel=${this._channel.name}, hasObjects=${hasObjects}`,
     );
 
-    // RTO4a
+    // Regardless of whether HAS_OBJECTS is set, the client must drop any previously buffered object operations
+    // and start a new sync sequence. If HAS_OBJECTS is set, the realtime server will deliver a sync sequence
+    // following the ATTACHED, guaranteeing that the objects in that sequence include at least all operations
+    // up to the point of attachment.
+    // RTO4d
+    this._bufferedObjectOperations = [];
+    // RTO4c
     this._startNewSync();
 
     // RTO4b
     if (!hasObjects) {
-      // if no HAS_OBJECTS flag received on attach, we can end sync sequence immediately and treat it as no objects on a channel.
-      // reset the objects pool to its initial state, and emit update events so subscribers to root object get notified about changes.
+      // If no HAS_OBJECTS flag was received on attach, end the sync sequence immediately and treat it as no objects on the channel.
+      // Reset the objects pool to its initial state and emit update events so subscribers to the root object are notified of changes.
       this._objectsPool.resetToInitialPool(true); // RTO4b1, RTO4b2
       this._syncObjectsDataPool.clear(); // RTO4b3
       this._endSync(); // RTO4b4
@@ -386,8 +399,6 @@ export class RealtimeObject {
   }
 
   private _startNewSync(syncId?: string, syncCursor?: string): void {
-    // need to discard all buffered object operation messages on new sync start
-    this._bufferedObjectOperations = [];
     this._syncObjectsDataPool.clear();
     this._currentSyncId = syncId;
     this._currentSyncCursor = syncCursor;
@@ -397,11 +408,11 @@ export class RealtimeObject {
   /** @spec RTO5c */
   private _endSync(): void {
     this._applySync();
-    // should apply buffered object operations after we applied the sync.
-    // can use regular object messages application logic
+    // Apply buffered object operations after the sync has been applied.
+    // Uses the regular object message application logic.
     this._applyObjectMessages(this._bufferedObjectOperations, ObjectsOperationSource.channel); // RTO5c6
 
-    this._bufferedObjectOperations = [];
+    this._bufferedObjectOperations = []; // RTO5c5
     this._syncObjectsDataPool.clear(); // RTO5c4
     this._currentSyncId = undefined; // RTO5c3
     this._currentSyncCursor = undefined; // RTO5c3
--- a/test/realtime/liveobjects.test.js
+++ b/test/realtime/liveobjects.test.js
@@ -164,9 +164,9 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
   }
 
   /**
-   * Helper function to inject an ATTACHED protocol message with or without HAS_OBJECTS flag
+   * Helper function to inject an ATTACHED protocol message
    */
-  async function injectAttachedMessage(helper, channel, hasObjects) {
+  async function injectAttachedMessage(helper, channel, flags) {
     helper.recordPrivateApi('call.connectionManager.activeProtocol.getTransport');
     helper.recordPrivateApi('call.transport.onProtocolMessage');
     helper.recordPrivateApi('call.makeProtocolMessageFromDeserialized');
@@ -174,7 +174,7 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
     const pm = createPM({
       action: 11, // ATTACHED
       channel: channel.name,
-      flags: hasObjects ? 1 << 7 : 0, // HAS_OBJECTS flag is bit 7
+      flags,
     });
     await transport.onProtocolMessage(pm);
   }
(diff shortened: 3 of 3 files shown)


src/plugins/liveobjects/realtimeobject.ts
```diff
@@ -182,14 +189,19 @@ export class RealtimeObject {
...
       this._client.Logger.LOG_MINOR,
-      'RealtimeObject.onAttached()',
+      'RealtimeObject.onNonResumeAttached()',
       `channel=${this._channel.name}, hasObjects=${hasObjects}`,
     );
 
+    // RTO4d
+    // Clear buffered object operations on attach (RESUMED=false or absent).
+    // The realtime server caches the object state at the moment of channel attachment,
+    // so we must start fresh and buffer all incoming OBJECT messages from this point.
```
mschristensen: I think a more accurate description would be something like: "The realtime server will deliver a sync sequence following a non-resume attached, and guarantees that the objects sent in that sequence will at least include all operations up to the point of attachment."

src/common/lib/client/realtimechannel.ts
```diff
@@ -564,7 +564,7 @@ class RealtimeChannel extends EventEmitter {
             }
             // the Objects tree needs to be re-synced
             if (this._object) {
-              this._object.onAttached(hasObjects);
+              this._object.onNonResumeAttached(hasObjects);
```
lawrence-forooghian: Why is all of the on-`ATTACHED` behaviour now gated behind the absence of the `RESUMED` flag? The RTO4 behaviours in general are not dependent on whether it's `RESUMED` or not.

lawrence-forooghian: Hmm, looking at it again I see that it was _always_ gated behind that flag. But it's not consistent with what the spec says, so we need to decide the correct behaviour and ensure spec and JS are consistent.

lawrence-forooghian: Thought about this a bit more. Was trying to understand how we populate the local objects state when we resume after connection recovery ([internal Slack thread](https://ably-real-time.slack.com/archives/C012DA3MCB0/p1769091059666059)). Simon said:

> for presence, in protocol v2 we always do a full sync after resume. maybe we might change that at some point, but right now that's what we do. I assume liveobjects does similarly

I think that if this _is_ the case for LiveObjects, then the JS behaviour of not becoming `SYNCING` when `resumed == true` is probably not correct, since you could have a scenario like the following (I'm no longer specifically thinking about connection recovery here; in this scenario the client is already `SYNCED`):

Why become SYNCING on resume drawio

i.e. this appears to be a scenario in which the assumption of RTO5f in https://github.com/ably/specification/pull/416 (specifically, that the preceding `ATTACHED` has `resumed == false`) is not true.

@mschristensen are you able to clarify the Realtime behaviour here please?

mschristensen: The current behaviour in realtime is the same as presence, i.e. we send a sync sequence unconditionally after a resume https://github.com/ably/realtime/blob/5a7663451577fea09446ffec626aef4f0dfc65d5/go/realtime/lib/channel/attachment.go#L916

So it looks like we should change this in realtime so that the client can know whether or not to buffer ops post ATTACHED. Ticket: https://ably.atlassian.net/browse/AIT-326

lawrence-forooghian: If we were to make that change, how would a client populate its local objects state after connection recovery (i.e. it resumes an attachment but without having any local objects data yet)?

lawrence-forooghian: We discussed this in standup today; conclusions:

- currently, if the `HAS_OBJECTS` flag is set on `ATTACHED`, this indicates that Realtime _will_ send a sync sequence; thus, the LiveObjects client's condition for entering `SYNCING` and buffering ops should be to do so when `HAS_OBJECTS` is set, as opposed to checking `RESUMED`
	- note that spec points RTO4 and RTO4a _do_ actually already mention the unconditional nature of the state sync
- we'd like to consider introducing a mechanism for Realtime to _conditionally_ send a sync sequence (for efficiency); have captured this in [AIT-330](https://ably.atlassian.net/browse/AIT-330)

src/plugins/liveobjects/realtimeobject.ts
VeskeR: updated

src/common/lib/client/realtimechannel.ts
VeskeR: Summarized our discussions and the agreed approach in the DR: https://ably.atlassian.net/wiki/x/NQAEHgE.

Updated this PR accordingly.

src/plugins/liveobjects/realtimeobject.ts
```diff
@@ -200,13 +207,19 @@ export class RealtimeObject {
       `channel=${this._channel.name}, hasObjects=${hasObjects}`,
     );
 
+    // Regardless of whether HAS_OBJECTS is set, the client must drop any previously buffered object operations
+    // and start a new sync sequence. If HAS_OBJECTS is set, the realtime server will deliver a sync sequence
+    // following the ATTACHED, guaranteeing that the objects in that sequence include at least all operations
+    // up to the point of attachment.
+    // RTO4d
+    this._bufferedObjectOperations = [];
     // RTO4a
```
lawrence-forooghian: Is this not RTO4c?

test/realtime/liveobjects.test.js
```diff
@@ -2754,69 +2804,201 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
...
+            expect(entryInstance.get('baz').value()).to.equal(
+              'qux',
+              'Check root has data from operations received after ATTACHED',
             );
+          },
+        },
 
-            // start new sync with new sequence id
+        {
+          description: 'buffered object operation messages are discarded on ATTACHED with RESUMED flag set',
```
lawrence-forooghian: I think this test has the potential to confuse a reader who isn't familiar with the changes that we're making in this PR, since there's nothing in the spec or code that mentions the `RESUMED` flag any more. I'd suggest either dropping it or adding a comment that explains why it exists (i.e. give some historical context, or contrast it with presence).

test/realtime/liveobjects.test.js
```diff
@@ -681,6 +681,56 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
...
+            // client should have emitted SYNCING and remain waiting for the OBJECT_SYNC sequence
+            expect(receivedEvents).to.deep.equal(
+              ['syncing'],
+              'Check client emitted SYNCING and is waiting for sync sequence after ATTACHED with HAS_OBJECTS',
+            );
+          },
+        },
+
+        {
+          description: 'on ATTACHED without HAS_OBJECTS, client transitions to SYNCED and clears local state',
```
lawrence-forooghian: This seems similar to https://github.com/ably/ably-js/blob/0035db0e/test/realtime/liveobjects.test.js#L8917-L8922; do we need this new one?

(The new one also asserts that the state is cleared but presumably we have a test for that already elsewhere?)

test/realtime/liveobjects.test.js
```diff
@@ -681,6 +681,56 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
       ];
 
       const objectSyncSequenceScenarios = [
+        {
+          description: 'on ATTACHED with HAS_OBJECTS=true, client remains in SYNCING and waits for the sync sequence',
```
lawrence-forooghian: This seems similar to https://github.com/ably/ably-js/blob/0035db0e/test/realtime/liveobjects.test.js#L8950-L8954; do we need this new one?

test/realtime/liveobjects.test.js
```diff
@@ -2754,69 +2804,201 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
...
+            // check root has data from operations received after ATTACHED
+            expect(entryInstance.get('baz').value()).to.equal(
+              'qux',
+              'Check root has data from operations received after RESUMED ATTACHED',
+            );
+          },
+        },
+
+        {
+          description: 'buffered object operation messages are NOT discarded on new OBJECT_SYNC sequence',
```
lawrence-forooghian, line 2885: Similar comment to https://github.com/ably/ably-js/pull/2150/changes#r2860141405; we should make it clear to a new reader why this test exists, or drop it

src/plugins/liveobjects/realtimeobject.ts
VeskeR: Not part of this PR, but sure, can change now. I wouldn't go much deeper into refactoring though, as I expect most of the `_startNewSync`-related code to change as a result of https://ably.atlassian.net/wiki/x/AYDeHwE

test/realtime/liveobjects.test.js
VeskeR: I'll add a comment to explain the context. Also clarify the name/test body a bit - it's not only the ATTACHED message with RESUMED flag, it's when we're _already_ attached, and we receive an ATTACHED message with RESUME=true (SDK used to not clear buffered operations, but it must)

test/realtime/liveobjects.test.js
VeskeR: Agree, will remove this one. I'd like to add `await new Promise((res) => nextTick(res));` line to the sync tests though, to ensure all protocol messages/sync messages are processed and events are emitted before checking the events array

test/realtime/liveobjects.test.js
VeskeR: > The new one also asserts that the state is cleared but presumably we have a test for that already elsewhere?

We don't have a test for this actually. I'll remove the `SYNCED` part, and only test for the cleared state. Will also add it in a separate commit to fix a test coverage gap

VeskeR: see https://github.com/ably/ably-js/pull/2150/commits/dbdb15cac03c5f16ef8d0a81dcb3c66b629e79ed

test/realtime/liveobjects.test.js
VeskeR: see https://github.com/ably/ably-js/pull/2150/commits/fb9f7fd04146c84f92fc356c7af443ab5fcef717

test/realtime/liveobjects.test.js
VeskeR: updated in https://github.com/ably/ably-js/pull/2150/commits/6e175ba08b9d5c55ceb4348900708d6fb3099f74, see compare https://github.com/ably/ably-js/compare/059be029dc4f62fd927ce91c4d1df7549e54582e..6e175ba08b9d5c55ceb4348900708d6fb3099f74

test/realtime/liveobjects.test.js
VeskeR: updated in https://github.com/ably/ably-js/pull/2150/commits/6e175ba08b9d5c55ceb4348900708d6fb3099f74, see compare https://github.com/ably/ably-js/compare/059be029dc4f62fd927ce91c4d1df7549e54582e..6e175ba08b9d5c55ceb4348900708d6fb3099f74

src/plugins/liveobjects/realtimeobject.ts
VeskeR: updated in https://github.com/ably/ably-js/pull/2150/commits/6e175ba08b9d5c55ceb4348900708d6fb3099f74, see compare https://github.com/ably/ably-js/compare/059be029dc4f62fd927ce91c4d1df7549e54582e..6e175ba08b9d5c55ceb4348900708d6fb3099f74

PR #2152, opened 2026-01-23
[AIT-38] Add support for partial sync of objects - protocol v6

VeskeR:
This PR is based on https://github.com/ably/ably-js/pull/2159, please review that one first.

Spec for this https://github.com/ably/specification/pull/413, DR https://ably.atlassian.net/wiki/x/AQBxCQE

Resolves [AIT-38](https://ably.atlassian.net/browse/AIT-38).

This also removes unnecessary `SyncObjectsDataPool` class and simplifies a sync logic a bit by moving it to `RealtimeObject`.

[AIT-38]: https://ably.atlassian.net/browse/AIT-38?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * New synchronization pool implementation to accumulate and merge object sync messages during sync.

* **Refactor**
  * Internal replacement of the previous sync data pool with the new pool implementation (no public API changes).

* **Bug Fixes**
  * Unsupported object shapes are now logged and skipped instead of causing errors.
  * Improved merging for partial OBJECT_SYNC sequences, preserving map entries across messages.

* **Tests**
  * Added OBJECT_SYNC tests covering multi-message builds, partial merges, and unknown-type resilience.

### DIFF
--- a/src/plugins/liveobjects/realtimeobject.ts
+++ b/src/plugins/liveobjects/realtimeobject.ts
@@ -163,11 +163,11 @@ export class RealtimeObject {
     }
 
     // RTO5a3 - continue current sync sequence
-    this._syncObjectsPool.applyObjectSyncMessages(objectMessages); // RTO5b
+    this._syncObjectsPool.applyObjectSyncMessages(objectMessages); // RTO5f
 
     // RTO5a4 - if this is the last (or only) message in a sequence of sync updates, end the sync
     if (!syncCursor) {
-      this._endSync();
+      this._endSync(); // RTO5c
     }
   }
 
@@ -443,34 +443,34 @@ export class RealtimeObject {
     }[] = [];
 
     // RTO5c1
-    for (const [objectId, entry] of this._syncObjectsPool.entries()) {
+    for (const [objectId, objectMessage] of this._syncObjectsPool.entries()) {
       receivedObjectIds.add(objectId);
       const existingObject = this._objectsPool.get(objectId);
 
       // RTO5c1a
       if (existingObject) {
-        const update = existingObject.overrideWithObjectState(entry.objectMessage); // RTO5c1a1
+        const update = existingObject.overrideWithObjectState(objectMessage); // RTO5c1a1
         // store updates to call subscription callbacks for all of them once the sync sequence is completed.
         // this will ensure that clients get notified about the changes only once everything has been applied.
         existingObjectUpdates.push({ object: existingObject, update });
         continue;
       }
 
-      // RTO5c1b,
+      // RTO5c1b
       let newObject: LiveObject;
-      // assign to a variable so TS doesn't complain about 'never' type in the default case
-      const objectType = entry.objectType;
-      switch (objectType) {
-        case 'LiveCounter':
-          newObject = LiveCounter.fromObjectState(this, entry.objectMessage); // RTO5c1b1a
-          break;
-
-        case 'LiveMap':
-          newObject = LiveMap.fromObjectState(this, entry.objectMessage); // RTO5c1b1b
-          break;
-
-        default:
-          throw new this._client.ErrorInfo(`Unknown LiveObject type: ${objectType}`, 50000, 500); // RTO5c1b1c
+      if (objectMessage.object?.counter) {
+        newObject = LiveCounter.fromObjectState(this, objectMessage); // RTO5c1b1a
+      } else if (objectMessage.object?.map) {
+        newObject = LiveMap.fromObjectState(this, objectMessage); // RTO5c1b1b
+      } else {
+        // RTO5c1b1c
+        this._client.Logger.logAction(
+          this._client.logger,
+          this._client.Logger.LOG_MAJOR,
+          'RealtimeObject._applySync()',
+          `received unsupported object state message during OBJECT_SYNC, expected 'counter' or 'map' to be present, skipping message; message id: ${objectMessage.id}, channel: ${this._channel.name}`,
+        );
+        continue;
       }
 
       this._objectsPool.set(objectId, newObject); // RTO5c1b1
--- a/src/plugins/liveobjects/syncobjectspool.ts
+++ b/src/plugins/liveobjects/syncobjectspool.ts
@@ -3,34 +3,19 @@ import type RealtimeChannel from 'common/lib/client/realtimechannel';
 import { ObjectMessage } from './objectmessage';
 import { RealtimeObject } from './realtimeobject';
 
-export interface LiveObjectDataEntry {
-  objectMessage: ObjectMessage;
-  objectType: 'LiveMap' | 'LiveCounter';
-}
-
-export interface LiveCounterDataEntry extends LiveObjectDataEntry {
-  objectType: 'LiveCounter';
-}
-
-export interface LiveMapDataEntry extends LiveObjectDataEntry {
-  objectType: 'LiveMap';
-}
-
-export type AnyDataEntry = LiveCounterDataEntry | LiveMapDataEntry;
-
 /**
  * @internal
- * @spec RTO5b
  */
 export class SyncObjectsPool {
   private _client: BaseClient;
   private _channel: RealtimeChannel;
-  private _pool: Map<string, AnyDataEntry>;
+  /** Used to accumulate object state during a sync sequence, keyed by object ID */
+  private _pool: Map<string, ObjectMessage>;
 
   constructor(private _realtimeObject: RealtimeObject) {
     this._client = this._realtimeObject.getClient();
     this._channel = this._realtimeObject.getChannel();
-    this._pool = new Map<string, AnyDataEntry>();
+    this._pool = new Map<string, ObjectMessage>();
   }
 
   entries() {
@@ -49,6 +34,7 @@ export class SyncObjectsPool {
     this._pool.clear();
   }
 
+  /** @spec RTO5f */
   applyObjectSyncMessages(objectMessages: ObjectMessage[]): void {
     for (const objectMessage of objectMessages) {
       if (!objectMessage.object) {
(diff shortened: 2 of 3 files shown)


src/plugins/liveobjects/realtimeobject.ts
```diff
@@ -370,7 +369,57 @@ export class RealtimeObject {
...
+    }
+  }
+
+  /**
+   * Merges map entries from a partial sync message into an existing entry.
+   * Other fields on the ObjectState envelope are identical across all partial messages
+   * for the same object, so only the entries need to be merged.
+   * @spec RTO5b2b1
+   */
+  private _mergeMapSyncState(existingEntry: ObjectMessage, newObjectMessage: ObjectMessage): void {
```
lawrence-forooghian: Question on spec re which `ObjectMessage`'s `serialTimestamp` property should be kept: https://github.com/ably/specification/pull/413#discussion_r2741637177

src/plugins/liveobjects/syncobjectsdatapool.ts
lawrence-forooghian, line 1: What was your motivation for removing this class? I'm a bit surprised that you did it as part of this work since it seems that it would have been an ideal place to put the new logic for merging `ObjectMessage`s into the pool, instead of expanding `RealtimeObject`.

src/plugins/liveobjects/realtimeobject.ts
```diff
@@ -370,7 +369,57 @@ export class RealtimeObject {
...
+    const existingObjectState = existingEntry.object!;
+    const newObjectState = newObjectMessage.object!;
+
+    if (!existingObjectState.map!.entries) {
+      existingObjectState.map!.entries = {};
+    }
+
+    if (newObjectState.map?.entries) {
+      // During partial sync, no two messages contain the same map key,
```
lawrence-forooghian: Worth mentioning in the spec, I'd say

test/realtime/liveobjects.test.js
```diff
@@ -1132,6 +1178,161 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
             ).to.be.true;
           },
         },
+
+        {
+          description: 'partial OBJECT_SYNC builds object tree across multiple messages',
```
lawrence-forooghian: I feel I must have missed something with this first test — in what sense is it testing a partial object sync? There doesn't seem to be any object whose state is split across multiple messages, which I thought was the defining characteristic of a partial sync.

test/realtime/liveobjects.test.js
```diff
@@ -1132,6 +1178,161 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
...
+              syncSerial: 'serial:cursor2',
+              state: [
+                objectsHelper.mapObject({
+                  objectId: mapId,
+                  siteTimeserials: { aaa: lexicoTimeserial('aaa', 0, 0) },
+                  // initialEntries are the same across all partial messages
+                  initialEntries: {
+                    initialKey: { timeserial: lexicoTimeserial('aaa', 0, 0), data: { string: 'initial' } },
+                  },
+                  // materialisedEntries are merged across partial messages
```
lawrence-forooghian: - the usage of "are" here is inconsistent — in the first one it describes a characteristic of the data sent by the server (makes sense), and in the second it describes an expected behaviour of the SDK (this note would be more appropriate against the assertions I think)
- I think that the injection of the final three messages would be easier to read if it were done in a loop in which the only thing that varies is the `syncSerial` and the `materialisedEntries`

src/plugins/liveobjects/realtimeobject.ts
VeskeR: answered in spec PR https://github.com/ably/specification/pull/413#discussion_r2827409682:

> If a tombstoned object is received over a sync sequence, the realtime guarantees the invariant that there will only ever be a single ObjectMessage for such an object.
The reason is that 1. tombstoned objects can't be un-tombstoned - that operation is terminal - and 2. an object's state is cleared upon tombstoning, which means there is nothing to paginate across multiple messages during a partial sync sequence, as the effective size of a tombstoned object is 0.
I think I'll add an explicit spec point for serialTimestamp for tombstoned objects with a small note.
See [slack discussion](https://ably-real-time.slack.com/archives/C09SY1AQGK0/p1771419359076359)

VeskeR: I'll re-check if we have to have something explicit in the implementation to get the correct `serialTimestamp` for tombstoned objects.

src/plugins/liveobjects/syncobjectsdatapool.ts
VeskeR: yea, now that we actually have something to do on the object sync (merging of partial state) we can keep that class.
I'll rename it to `SyncObjectsPool` though to match the spec

test/realtime/liveobjects.test.js
VeskeR: You're right. It's not about the **partial** sync here, it's about having an explicit test to check that we can successfully build the expected state out of the multiple sync messages arrived separately - essentially a prerequisite for partial sync. Will update the name

src/plugins/liveobjects/realtimeobject.ts
VeskeR: Nothing we need to do in the implementation, tombstoning during sync sequence and using the correct `serialTimestamp` is already covered by the end of the sync sequence procedure, see [RTLC6f](https://sdk.ably.com/builds/ably/specification/main/objects-features/#RTLC6f), [RTLM6f](https://sdk.ably.com/builds/ably/specification/main/objects-features/#RTLM6f) and [RTLO4e](https://sdk.ably.com/builds/ably/specification/main/objects-features/#RTLO4e)

src/plugins/liveobjects/syncobjectsdatapool.ts
VeskeR: done in https://github.com/ably/ably-js/compare/7c87463bd9f22014ecea505d7530575123bc4288..a8afac712bba2e60d0a3691b916ab43d41879536

test/realtime/liveobjects.test.js
VeskeR: Updated in https://github.com/ably/ably-js/compare/7c87463bd9f22014ecea505d7530575123bc4288..a8afac712bba2e60d0a3691b916ab43d41879536

test/realtime/liveobjects.test.js
VeskeR: Updated in https://github.com/ably/ably-js/compare/7c87463bd9f22014ecea505d7530575123bc4288..a8afac712bba2e60d0a3691b916ab43d41879536

src/plugins/liveobjects/realtimeobject.ts
VeskeR: Updated the spec in https://github.com/ably/specification/pull/413/changes/f6722adfb3343ef992d37c8e1288f922a507d864

src/plugins/liveobjects/realtimeobject.ts
VeskeR: Added to the spec in https://github.com/ably/specification/pull/413/changes/f6722adfb3343ef992d37c8e1288f922a507d864

src/plugins/liveobjects/syncobjectspool.ts
```diff
@@ -56,43 +42,71 @@ export class SyncObjectsPool {
...
+      // A tombstone flag on any partial message takes precedence over previously accumulated entries
+      this._pool.set(existingObjectState.objectId, newObjectMessage);
+      return;
+    }
 
-    return newEntry;
+    // Other fields on the ObjectState envelope (such as siteTimeserials) and the map envelope
+    // (such as semantics) are identical across all partial messages for the same object,
+    // so only the entries need to be merged.
+    if (!existingObjectState.map!.entries) {
```
lawrence-forooghian, line 110: minor: I'd suggest putting this inside the next `if` to make it clear that it enables the merging of the new entries (but keep the preceding comment where it is — it doesn't really specifically describe the initialization of `existingObjectState.map!.entries`)

src/plugins/liveobjects/syncobjectspool.ts
```diff
@@ -56,43 +42,71 @@ export class SyncObjectsPool {
...
+      existingObjectState.map!.entries = {};
+    }
+
+    if (newObjectState.map?.entries) {
```
lawrence-forooghian: Shouldn't the force-unwrapping and optional chaining be the other way round on `existingObjectState.map!.entries` and `newObjectState.map?.entries`? i.e. the caller has verified that `newObjectState.map` exists already

test/realtime/liveobjects.test.js
```diff
@@ -704,6 +704,194 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
...
+
+            expect(entryPathObject.get('stringKey').value()).to.equal('hello', 'Check root has correct string value');
+            expect(entryPathObject.get('counter').value()).to.equal(15, 'Check counter has correct aggregated value');
+            expect(entryPathObject.get('map').get('foo').value()).to.equal('bar', 'Check map has initial entries');
+            expect(entryPathObject.get('map').get('baz').value()).to.equal('qux', 'Check map has materialised entries');
+          },
+        },
+
+        {
+          description: 'OBJECT_SYNC does not break when receiving an unknown object type',
```
lawrence-forooghian, line 944: This test seems like it's a fix for JS not conforming to the existing RTO5c1b1c. Can we have it as a standalone commit in that case?

test/realtime/liveobjects.test.js
lawrence-forooghian: This seems like it's a gap in the existing test coverage in that case — best as a preceding commit?

VeskeR: extracted two tests to a separate commit https://github.com/ably/ably-js/pull/2152/changes/5eb70a260a802d660ac61d8c8a61ca9e0d1d6401, both fill the gap in test coverage for existing behavior

src/plugins/liveobjects/syncobjectspool.ts
VeskeR: the next `if` no longer exist after changes as a result of https://github.com/ably/ably-js/pull/2152#discussion_r2841794892 and spec update in https://github.com/ably/specification/pull/428

src/plugins/liveobjects/syncobjectspool.ts
VeskeR: Both `existingObjectState.map` and `newObjectState.map` are guaranteed to exist by earlier checks in `applyObjectSyncMessages`, so this `if` can be removed.
I also slightly reordered the unsupported data type checks ([spec](https://github.com/ably/specification/pull/428) and [in this PR](https://github.com/ably/ably-js/pull/2152/commits/4f71d2018f28ecddcb6d425ac1cd95585e7cec74) - will squash once approved) to avoid adding unsupported object types to the object pool.

test/realtime/liveobjects.test.js
VeskeR: Even though ably-js implementation of _applying_ the object sync did not conform to RTO5c1b1c (which is fixed in this PR), the expected behavior of said spec point was covered in https://github.com/ably/ably-js/blob/0035db0ebcec105ae510bcbbe5a605c0edde9f12/src/plugins/liveobjects/syncobjectsdatapool.ts#L71-L76 already - during _receiving_ object sync messages. So this test actually just fills the gap in the test coverage ,and would pass on its own.
Extracted it to a separate commit https://github.com/ably/ably-js/pull/2152/commits/5eb70a260a802d660ac61d8c8a61ca9e0d1d6401 (just like the one about multiple message above)

PR #2153, opened 2026-01-23
fix: non-monotonic msgSerial on stale connection reset

owenpearson:
Fixes a case where js clients may send non-monotonic msgSerials on a connection

The specific issue was that we weren't always resetting `sendAttempted` on pending messages when resetting the connection.

The flow which caused non-monotonic msgSerials was:

  1. Client is connected, sends message with msgSerial=5 (sendAttempted=true, in Protocol.messageQueue)
  2. Transport disconnects before ACK recieved
  3. deactivateTransport(): message is requeued to queuedMessages (sendAttempted=true, keeps msgSerial=5)
  4. State = 'disconnected' (queueEvents=true, so messages are NOT failed)
  5. Time passes... (longer than connectionStateTtl + maxIdleInterval)
  6. Client tries to reconnect, calls startConnect()
  7. checkConnectionStateFreshness() detects stale connection, calls clearConnection(): msgSerial=0, connectionKey=undefined
  8. Since connectionKey is undefined, mode='clean' (not resume)
  9. Client connects, gets new connection
  10. sendQueuedMessages(): message has sendAttempted=true, so it's sent with its old previously attempted serial (msgSerial=5)
  11. User publishes new message: gets msgSerial=0 (from the reset counter)
  12. Server sees messages published with msgSerials 5, 0 - non-monotonic

nb: this doesn't affect suspended connections since we fail queued messages when a connection becomes suspended.

## Summary by CodeRabbit

* **Bug Fixes**
  * Clearing a stale connection now properly resets queued message state so messages receive fresh identifiers on reconnect.

* **Tests**
  * Added a test that verifies connection reset clears message queue state and identifier counters.
  * Updated test tracking to recognize additional connection-manager events.

### DIFF
--- a/src/common/lib/transport/connectionmanager.ts
+++ b/src/common/lib/transport/connectionmanager.ts
@@ -846,6 +846,9 @@ class ConnectionManager extends EventEmitter {
     this.realtime.connection.id = this.connectionId = undefined;
     this.realtime.connection.key = this.connectionKey = undefined;
     this.msgSerial = 0;
+    // RTN19a2: On a new connection, previous msgSerials are meaningless.
+    // Reset sendAttempted so queued messages get new serials.
+    this.queuedMessages.resetSendAttempted();
     this.unpersistConnection();
   }
 


test/realtime/connection.test.js
```diff
@@ -354,6 +354,82 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, Helper, async
...
+              var originalMsgSerial = pendingMessages[0].message.msgSerial;
+              expect(originalMsgSerial).to.equal(0, 'Message should have msgSerial=0');
+
+              helper.recordPrivateApi('call.connectionManager.queuePendingMessages');
+              connectionManager.queuePendingMessages(pendingMessages);
+
+              // Simulate what checkConnectionStateFreshness does when connection is stale:
+              // It calls clearConnection() which should reset both msgSerial AND sendAttempted
+              helper.recordPrivateApi('call.connectionManager.clearConnection');
+              connectionManager.clearConnection();
```
VeskeR: I wonder if we could reduce the number of assumptions this test makes about the exact internal methods that are called when the state transitions to `disconnected` and then to `suspended`. Could we replicate the state transitions by using an invalid host and configuring the timeouts accordingly, as in [this failure test](https://github.com/ably/ably-js/blob/a06036cf5f7905274ff9478da0139af06232088c/test/realtime/failure.test.js#L130-L160)? We would then expect that, upon entering `suspended` (as a result of a stale connection), `msgSerial` is reset and pending messages have `sendAttempted` set to false.

owenpearson: yeah fair point, I'm not sure if using an invalid host would work because we need the client to become connected or else it won't bother queueing the message at all, we actually need to become disconnected while the message is pending. i've updated the test now so should be a bit more realistic now, but still uses quite a few internal apis, i don't see an obvious way around this tbh.

VeskeR: The new version looks much better, thanks! Approved

PR #2155, opened 2026-01-26
[AIT-280] Apply operations on ACK

lawrence-forooghian:
**Note: This PR is based on top of #2167; please review that one first.**

Based on https://github.com/ably/specification/pull/419 at `d809334`. Implementation and tests are Claude-generated from the spec; I've reviewed them and given plenty of feedback, but largely resisted the temptation to tweak things that aren't quite how I'd write them but which are still correct.

The only behaviour here that's not in the spec is to also apply-on-ACK for batch operations (the batch API isn't in the spec yet).

Summary of decisions re modifications to existing tests (written by Claude):

- Removed redundant `waitFor*` calls after SDK operations (`map.set()`, `counter.increment()`, etc.) - with apply-on-ACK, values are available immediately after the operation promise resolves
- Kept `waitFor*` calls after REST operations (`objectsHelper.operationRequest()`, `objectsHelper.createAndSetOnMap()`) - these still require waiting for the echo to arrive over Realtime
- Added explanatory comment to `applyOperationsScenarios` noting that those tests cover operations received over Realtime (via REST), and pointing to the new "Apply on ACK" section for tests of locally-applied operations

Docs PR: https://github.com/ably/docs/pull/3161/

## Summary by CodeRabbit

* **New Features**
  * Operations now apply locally once the server ACKs them, improving local consistency

* **Improvements**
  * Publish, state and presence calls now surface explicit ACK results
  * More deterministic error handling for message delivery and attach/detach flows
  * Internal operation sources tracked to avoid double-apply and better serial handling

* **Documentation**
  * Clarified batching semantics and ACK/apply timing

* **Tests**
  * Added interceptors and extensive tests for apply-on-ACK, echo/ACK sequencing, and sync scenarios

### DIFF
--- a/src/common/lib/client/realtimechannel.ts
+++ b/src/common/lib/client/realtimechannel.ts
@@ -536,13 +536,13 @@ class RealtimeChannel extends EventEmitter {
     await this.sendAndAwaitAck(msg);
   }
 
-  async sendState(objectMessages: WireObjectMessage[]): Promise<void> {
+  async sendState(objectMessages: WireObjectMessage[]): Promise<API.PublishResult> {
     const msg = protocolMessageFromValues({
       action: actions.OBJECT,
       channel: this.name,
       state: objectMessages,
     });
-    await this.sendAndAwaitAck(msg);
+    return this.sendAndAwaitAck(msg);
   }
 
   // Access to this method is synchronised by ConnectionManager#processChannelMessage, in order to synchronise access to the state stored in _decodingContext.
--- a/src/plugins/liveobjects/livecounter.ts
+++ b/src/plugins/liveobjects/livecounter.ts
@@ -2,7 +2,7 @@ import { __livetype } from '../../../ably';
 import { LiveCounter as PublicLiveCounter } from '../../../liveobjects';
 import { LiveObject, LiveObjectData, LiveObjectUpdate, LiveObjectUpdateNoop } from './liveobject';
 import { ObjectData, ObjectMessage, ObjectOperation, ObjectOperationAction, ObjectsCounterOp } from './objectmessage';
-import { RealtimeObject } from './realtimeobject';
+import { ObjectsOperationSource, RealtimeObject } from './realtimeobject';
 
 export interface LiveCounterData extends LiveObjectData {
   data: number; // RTLC3
@@ -72,15 +72,15 @@ export class LiveCounter extends LiveObject<LiveCounterData, LiveCounterUpdate> 
   /**
    * Send a COUNTER_INC operation to the realtime system to increment a value on this LiveCounter object.
    *
-   * This does not modify the underlying data of this LiveCounter object. Instead, the change will be applied when
-   * the published COUNTER_INC operation is echoed back to the client and applied to the object following the regular
-   * operation application procedure.
+   * The change will be applied locally when the ACK is received from Realtime.
    *
-   * @returns A promise which resolves upon receiving the ACK message for the published operation message.
+   * @returns A promise which resolves upon receiving the ACK message for the published operation message
+   * and applying the operation locally.
+   * @spec RTLC12
    */
   async increment(amount: number): Promise<void> {
     const msg = LiveCounter.createCounterIncMessage(this._realtimeObject, this.getObjectId(), amount);
-    return this._realtimeObject.publish([msg]);
+    return this._realtimeObject.publishAndApply([msg]);
   }
 
   /**
@@ -98,8 +98,9 @@ export class LiveCounter extends LiveObject<LiveCounterData, LiveCounterUpdate> 
 
   /**
    * @internal
+   * @spec RTLC7
    */
-  applyOperation(op: ObjectOperation<ObjectData>, msg: ObjectMessage): void {
+  applyOperation(op: ObjectOperation<ObjectData>, msg: ObjectMessage, source: ObjectsOperationSource): boolean {
     if (op.objectId !== this.getObjectId()) {
       throw new this._client.ErrorInfo(
         `Cannot apply object operation with objectId=${op.objectId}, to this LiveCounter with objectId=${this.getObjectId()}`,
@@ -117,15 +118,19 @@ export class LiveCounter extends LiveObject<LiveCounterData, LiveCounterUpdate> 
         'LiveCounter.applyOperation()',
         `skipping ${op.action} op: op serial ${opSerial.toString()} <= site serial ${this._siteTimeserials[opSiteCode]?.toString()}; objectId=${this.getObjectId()}`,
       );
-      return;
+      return false; // RTLC7b
     }
-    // should update stored site serial immediately. doesn't matter if we successfully apply the op,
-    // as it's important to mark that the op was processed by the object
-    this._siteTimeserials[opSiteCode] = opSerial;
+
+    // RTLC7c
+    if (source === ObjectsOperationSource.channel) {
+      // should update stored site serial immediately. doesn't matter if we successfully apply the op,
+      // as it's important to mark that the op was processed by the object
+      this._siteTimeserials[opSiteCode] = opSerial;
+    }
 
     if (this.isTombstoned()) {
       // this object is tombstoned so the operation cannot be applied
-      return;
+      return false; // RTLC7e
     }
 
     let update: LiveCounterUpdate | LiveObjectUpdateNoop;
@@ -155,6 +160,7 @@ export class LiveCounter extends LiveObject<LiveCounterData, LiveCounterUpdate> 
     }
 
     this.notifyUpdated(update);
+    return true; // RTLC7d1b, RTLC7d2b, RTLC7d4b
   }
 
   /**
--- a/src/plugins/liveobjects/livemap.ts
+++ b/src/plugins/liveobjects/livemap.ts
@@ -22,7 +22,7 @@ import {
   ObjectsMapOp,
   ObjectsMapSemantics,
 } from './objectmessage';
-import { RealtimeObject } from './realtimeobject';
+import { ObjectsOperationSource, RealtimeObject } from './realtimeobject';
 
 export interface ObjectIdObjectData {
   /** A reference to another object, used to support composable object structures. */
@@ -265,38 +265,39 @@ export class LiveMap<T extends Record<string, Value> = Record<string, Value>>
   /**
    * Send a MAP_SET operation to the realtime system to set a key on this LiveMap object to a specified value.
    *
-   * This does not modify the underlying data of this LiveMap object. Instead, the change will be applied when
-   * the published MAP_SET operation is echoed back to the client and applied to the object following the regular
-   * operation application procedure.
+   * The change will be applied locally when the ACK is received from Realtime.
    *
-   * @returns A promise which resolves upon receiving the ACK message for the published operation message.
+   * @returns A promise which resolves upon receiving the ACK message for the published operation message
+   * and applying the operation locally.
+   * @spec RTLM20
    */
   async set<TKey extends keyof T & string>(
     key: TKey,
     value: T[TKey] | LiveCounterValueType | LiveMapValueType,
   ): Promise<void> {
     const msgs = await LiveMap.createMapSetMessage(this._realtimeObject, this.getObjectId(), key, value);
-    return this._realtimeObject.publish(msgs);
+    return this._realtimeObject.publishAndApply(msgs);
   }
 
   /**
    * Send a MAP_REMOVE operation to the realtime system to tombstone a key on this LiveMap object.
    *
-   * This does not modify the underlying data of this LiveMap object. Instead, the change will be applied when
-   * the published MAP_REMOVE operation is echoed back to the client and applied to the object following the regular
-   * operation application procedure.
+   * The change will be applied locally when the ACK is received from Realtime.
    *
-   * @returns A promise which resolves upon receiving the ACK message for the published operation message.
+   * @returns A promise which resolves upon receiving the ACK message for the published operation message
+   * and applying the operation locally.
+   * @spec RTLM21
    */
   async remove<TKey extends keyof T & string>(key: TKey): Promise<void> {
     const msg = LiveMap.createMapRemoveMessage(this._realtimeObject, this.getObjectId(), key);
-    return this._realtimeObject.publish([msg]);
+    return this._realtimeObject.publishAndApply([msg]);
   }
 
   /**
    * @internal
+   * @spec RTLM15
    */
-  applyOperation(op: ObjectOperation<ObjectData>, msg: ObjectMessage): void {
+  applyOperation(op: ObjectOperation<ObjectData>, msg: ObjectMessage, source: ObjectsOperationSource): boolean {
     if (op.objectId !== this.getObjectId()) {
       throw new this._client.ErrorInfo(
         `Cannot apply object operation with objectId=${op.objectId}, to this LiveMap with objectId=${this.getObjectId()}`,
@@ -314,15 +315,19 @@ export class LiveMap<T extends Record<string, Value> = Record<string, Value>>
         'LiveMap.applyOperation()',
         `skipping ${op.action} op: op serial ${opSerial.toString()} <= site serial ${this._siteTimeserials[opSiteCode]?.toString()}; objectId=${this.getObjectId()}`,
       );
-      return;
+      return false; // RTLM15b
     }
-    // should update stored site serial immediately. doesn't matter if we successfully apply the op,
-    // as it's important to mark that the op was processed by the object
-    this._siteTimeserials[opSiteCode] = opSerial;
+
+    // RTLM15c
+    if (source === ObjectsOperationSource.channel) {
+      // should update stored site serial immediately. doesn't matter if we successfully apply the op,
+      // as it's important to mark that the op was processed by the object
+      this._siteTimeserials[opSiteCode] = opSerial;
+    }
 
     if (this.isTombstoned()) {
       // this object is tombstoned so the operation cannot be applied
-      return;
+      return false; // RTLM15e
     }
 
     let update: LiveMapUpdate<T> | LiveObjectUpdateNoop;
@@ -360,6 +365,7 @@ export class LiveMap<T extends Record<string, Value> = Record<string, Value>>
     }
 
     this.notifyUpdated(update);
+    return true; // RTLM15d1b, RTLM15d2b, RTLM15d3b, RTLM15d5b
   }
 
   /**
--- a/src/plugins/liveobjects/realtimeobject.ts
+++ b/src/plugins/liveobjects/realtimeobject.ts
@@ -1,6 +1,7 @@
 import type BaseClient from 'common/lib/client/baseclient';
 import type RealtimeChannel from 'common/lib/client/realtimechannel';
 import type EventEmitter from 'common/lib/util/eventemitter';
+import type * as API from '../../../ably';
 import type { ChannelState, StatusSubscription } from '../../../ably';
 import type * as ObjectsApi from '../../../liveobjects';
 import { DEFAULTS } from './defaults';
@@ -16,6 +17,12 @@ import { SyncObjectsDataPool } from './syncobjectsdatapool';
 export enum ObjectsEvent {
   syncing = 'syncing',
   synced = 'synced',
+}
+
+/** @spec RTO22 */
+export enum ObjectsOperationSource {
+  local = 'local',
+  channel = 'channel',
 }
 
 export enum ObjectsState {
@@ -48,6 +55,7 @@ export class RealtimeObject {
   private _currentSyncId: string | undefined;
   private _currentSyncCursor: string | undefined;
   private _bufferedObjectOperations: ObjectMessage[];
+  private _appliedOnAckSerials: Set<string>; // RTO7b
   private _pathObjectSubscriptionRegister: PathObjectSubscriptionRegister;
 
   // Used by tests
@@ -62,6 +70,7 @@ export class RealtimeObject {
     this._objectsPool = new ObjectsPool(this);
     this._syncObjectsDataPool = new SyncObjectsDataPool(this);
     this._bufferedObjectOperations = [];
+    this._appliedOnAckSerials = new Set(); // RTO7b1
     this._pathObjectSubscriptionRegister = new PathObjectSubscriptionRegister(this);
     // use server-provided objectsGCGracePeriod if available, and subscribe to new connectionDetails that can be emitted as part of the RTN24
     this.gcGracePeriod =
@@ -164,6 +173,7 @@ export class RealtimeObject {
 
   /**
    * @internal
+   * @spec RTO8
    */
   handleObjectMessages(objectMessages: ObjectMessage[]): void {
     if (this._state !== ObjectsState.synced) {
@@ -175,7 +185,7 @@ export class RealtimeObject {
       return;
     }
 
-    this._applyObjectMessages(objectMessages);
+    this._applyObjectMessages(objectMessages, ObjectsOperationSource.channel); // RTO8b
   }
 
   /**
@@ -223,8 +233,9 @@ export class RealtimeObject {
 
   /**
    * @internal
+   * @spec RTO15
    */
-  async publish(objectMessages: ObjectMessage[]): Promise<void> {
+  async publish(objectMessages: ObjectMessage[]): Promise<API.PublishResult> {
     this._channel.throwIfUnpublishableState();
 
     const encodedMsgs = objectMessages.map((x) => x.encode(this._client));
@@ -238,7 +249,123 @@ export class RealtimeObject {
       );
     }
 
+    // RTO15h
     return this._channel.sendState(encodedMsgs);
+  }
+
+  /**
+   * Publishes ObjectMessages and applies them locally upon receiving the ACK from the server.
+   *
+   * @internal
+   * @spec RTO20
+   */
+  async publishAndApply(objectMessages: ObjectMessage[]): Promise<void> {
+    // RTO20b
+    const publishResult = await this.publish(objectMessages);
+
+    this._client.Logger.logAction(
+      this._client.logger,
+      this._client.Logger.LOG_MICRO,
+      'RealtimeObject.publishAndApply()',
+      `received ACK for ${objectMessages.length} message(s), applying locally; channel=${this._channel.name}`,
+    );
+
+    // RTO20c - check required information is available
+    const siteCode = this._channel.connectionManager.connectionDetails?.siteCode;
+    // RTO20c1
+    if (!siteCode) {
+      this._client.Logger.logAction(
+        this._client.logger,
+        this._client.Logger.LOG_ERROR,
+        'RealtimeObject.publishAndApply()',
+        `operations will not be applied locally: siteCode not available from connectionDetails; channel=${this._channel.name}`,
+      );
+      return;
+    }
+    // RTO20c2
+    if (!publishResult.serials || publishResult.serials.length !== objectMessages.length) {
+      this._client.Logger.logAction(
+        this._client.logger,
+        this._client.Logger.LOG_ERROR,
+        'RealtimeObject.publishAndApply()',
+        `operations will not be applied locally: PublishResult.serials has unexpected length (expected ${objectMessages.length}, got ${publishResult.serials?.length}); channel=${this._channel.name}`,
+      );
+      return;
+    }
+
+    // RTO20d
+    const syntheticMessages: ObjectMessage[] = [];
+    for (let i = 0; i < objectMessages.length; i++) {
+      const serial = publishResult.serials[i];
+
+      // RTO20d1
+      if (serial === null) {
+        this._client.Logger.logAction(
+          this._client.logger,
+          this._client.Logger.LOG_MICRO,
+          'RealtimeObject.publishAndApply()',
+          `operation will not be applied locally: serial is null in PublishResult (index ${i}); channel=${this._channel.name}`,
+        );
+        continue;
+      }
+
+      // RTO20d2, RTO20d3
+      syntheticMessages.push(
+        ObjectMessage.fromValues(
+          {
+            ...objectMessages[i],
+            serial, // RTO20d2a
+            siteCode, // RTO20d2b
+          },
+          this._client.Utils,
+          this._client.MessageEncoding,
+        ),
+      );
+    }
+
+    // RTO20e - Wait for sync to complete if not synced
+    if (this._state !== ObjectsState.synced) {
+      this._client.Logger.logAction(
+        this._client.logger,
+        this._client.Logger.LOG_MICRO,
+        'RealtimeObject.publishAndApply()',
+        `waiting for sync to complete before applying ${syntheticMessages.length} message(s); channel=${this._channel.name}`,
+      );
+
+      await new Promise<void>((resolve, reject) => {
+        const cleanup = () => {
+          this._eventEmitterInternal.off(onSynced);
+          this._channel.internalStateChanges.off(onChannelState);
+        };
+        const onSynced = () => {
+          cleanup();
+          resolve();
+        };
+        // RTO20e1
+        const onChannelState = () => {
+          cleanup();
+          reject(
+            new this._client.ErrorInfo(
+              `the operation could not be applied locally due to the channel entering the ${this._channel.state} state whilst waiting for objects sync to complete`,
+              92008,
+              400,
+              this._channel.errorReason || undefined,
+            ),
+          );
+        };
+        this._eventEmitterInternal.once(ObjectsEvent.synced, onSynced);
+        this._channel.internalStateChanges.once(['detached', 'suspended', 'failed'], onChannelState);
+      });
+    }
+
+    // RTO20f - Apply synthetic messages
+    this._client.Logger.logAction(
+      this._client.logger,
+      this._client.Logger.LOG_MICRO,
+      'RealtimeObject.publishAndApply()',
+      `applying ${syntheticMessages.length} message(s); channel=${this._channel.name}`,
+    );
+    this._applyObjectMessages(syntheticMessages, ObjectsOperationSource.local);
   }
 
   /**
@@ -272,12 +399,16 @@ export class RealtimeObject {
     this._applySync();
     // should apply buffered object operations after we applied the sync.
     // can use regular object messages application logic
-    this._applyObjectMessages(this._bufferedObjectOperations);
+    this._applyObjectMessages(this._bufferedObjectOperations, ObjectsOperationSource.channel); // RTO5c6
 
     this._bufferedObjectOperations = [];
     this._syncObjectsDataPool.clear(); // RTO5c4
     this._currentSyncId = undefined; // RTO5c3
     this._currentSyncCursor = undefined; // RTO5c3
+
+    // RTO5c9 - Clear appliedOnAckSerials
+    this._appliedOnAckSerials.clear();
+
     this._stateChange(ObjectsState.synced);
   }
 
@@ -356,7 +487,8 @@ export class RealtimeObject {
     existingObjectUpdates.forEach(({ object, update }) => object.notifyUpdated(update));
   }
 
-  private _applyObjectMessages(objectMessages: ObjectMessage[]): void {
+  /** @spec RTO9 */
+  private _applyObjectMessages(objectMessages: ObjectMessage[], source: ObjectsOperationSource): void {
     for (const objectMessage of objectMessages) {
       if (!objectMessage.operation) {
         this._client.Logger.logAction(
@@ -368,6 +500,20 @@ export class RealtimeObject {
         continue;
       }
 
+      const serial = objectMessage.serial;
+
+      // RTO9a3 - Skip if already applied on ACK
+      if (serial && this._appliedOnAckSerials.has(serial)) {
+        this._client.Logger.logAction(
+          this._client.logger,
+          this._client.Logger.LOG_MICRO,
+          'RealtimeObject._applyObjectMessages()',
+          `skipping message: already applied on ACK; serial=${serial}, channel=${this._channel.name}`,
+        );
+        this._appliedOnAckSerials.delete(serial);
+        continue;
+      }
+
       const objectOperation = objectMessage.operation;
 
       switch (objectOperation.action) {
@@ -376,7 +522,7 @@ export class RealtimeObject {
         case ObjectOperationAction.MAP_SET:
         case ObjectOperationAction.MAP_REMOVE:
         case ObjectOperationAction.COUNTER_INC:
-        case ObjectOperationAction.OBJECT_DELETE:
+        case ObjectOperationAction.OBJECT_DELETE: {
           // we can receive an op for an object id we don't have yet in the pool. instead of buffering such operations,
           // we can create a zero-value object for the provided object id and apply the operation to that zero-value object.
           // this also means that all objects are capable of applying the corresponding *_CREATE ops on themselves,
@@ -384,8 +530,16 @@ export class RealtimeObject {
           // so to simplify operations handling, we always try to create a zero-value object in the pool first,
           // and then we can always apply the operation on the existing object in the pool.
           this._objectsPool.createZeroValueObjectIfNotExists(objectOperation.objectId);
-          this._objectsPool.get(objectOperation.objectId)!.applyOperation(objectOperation, objectMessage);
+          const applied = this._objectsPool
+            .get(objectOperation.objectId)!
+            .applyOperation(objectOperation, objectMessage, source); // RTO9a2a3
+
+          // RTO9a2a4
+          if (source === ObjectsOperationSource.local && applied && serial) {
+            this._appliedOnAckSerials.add(serial);
+          }
           break;
+        }
 
         default:
           this._client.Logger.logAction(
--- a/test/realtime/liveobjects.test.js
+++ b/test/realtime/liveobjects.test.js
@@ -180,6 +180,139 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
   }
 
   /**
+   * Creates an interceptor that holds OBJECT messages and allows controlled release, to test ACK-before-echo scenarios.
+   * Call when CONNECTING or CONNECTED to intercept messages on the active transport.
+   *
+   * Note: An echo is a subset of OBJECT messages — specifically, OBJECT messages that originated from
+   * this client and are being echoed back. The name "echo" reflects how this interceptor is intended
+   * to be used in tests, but it doesn't actually filter for echoes — it intercepts ALL OBJECT messages.
+   *
+   * Returns an object with:
+   *   - heldEchoes: array of { message, release } objects, where `message` is the held OBJECT message
+   *     and `release()` returns a Promise that resolves when processing completes
+   *   - waitForEcho(): resolves immediately if an OBJECT message has already been intercepted, otherwise
+   *     waits until the next OBJECT message is intercepted. Only one call can be pending at a time;
+   *     subsequent calls overwrite the previous callback.
+   *   - releaseAll(): releases all held OBJECT messages and returns a Promise that resolves when
+   *     all processing completes
+   *   - restore(): restores normal message handling
+   */
+  function createEchoInterceptor(helper, client) {
+    helper.recordPrivateApi('call.connectionManager.activeProtocol.getTransport');
+    const transport = client.connection.connectionManager.activeProtocol.getTransport();
+    const originalOnProtocolMessage = transport.onProtocolMessage;
+    const heldEchoes = [];
+    let onEchoIntercepted = null;
+
+    helper.recordPrivateApi('replace.transport.onProtocolMessage');
+    transport.onProtocolMessage = function (message) {
+      if (message.action === 19) {
+        // OBJECT message
+        heldEchoes.push({
+          message,
+          // release() calls channel.processMessage directly and returns the Promise,
+          // allowing callers to await processing completion
+          release: () => {
+            helper.recordPrivateApi('call.channel.processMessage');
+            const channel = client.channels.get(message.channel);
+            return channel.processMessage(message);
+          },
+        });
+        if (onEchoIntercepted) {
+          onEchoIntercepted();
+          onEchoIntercepted = null;
+        }
+        return;
+      }
+      helper.recordPrivateApi('call.transport.onProtocolMessage');
+      originalOnProtocolMessage.call(transport, message);
+    };
+
+    return {
+      heldEchoes,
+      waitForEcho: () =>
+        new Promise((resolve) => {
+          if (heldEchoes.length > 0) {
+            resolve();
+          } else {
+            onEchoIntercepted = resolve;
+          }
+        }),
+      releaseAll: async () => {
+        while (heldEchoes.length > 0) {
+          await heldEchoes.shift().release();
+        }
+      },
+      restore: () => {
+        helper.recordPrivateApi('replace.transport.onProtocolMessage');
+        transport.onProtocolMessage = originalOnProtocolMessage;
+      },
+    };
+  }
+
+  /**
+   * Creates an interceptor that holds ACK messages and allows controlled release, to test echo-before-ACK scenarios.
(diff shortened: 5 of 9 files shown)


src/plugins/liveobjects/realtimeobject.ts
```diff
@@ -18,6 +19,18 @@ export enum ObjectsEvent {
...
+/** @spec RTO22 */
+export enum ObjectsOperationSource {
+  local = 'local', // RTO22a - applied upon ACK
+  channel = 'channel', // RTO22b - received over channel
+}
+
+/** @spec RTO7c */
+interface BufferedAckEntry {
+  objectMessage: ObjectMessage; // RTO7c1
+  signal: () => void; // RTO7c2
```
lawrence-forooghian: Document what this does (i.e. causes the promise returned by `publishAndApply()` to resolve

src/plugins/liveobjects/realtimeobject.ts
```diff
@@ -18,6 +19,18 @@ export enum ObjectsEvent {
   synced = 'synced',
 }
 
+/** @spec RTO22 */
+export enum ObjectsOperationSource {
+  local = 'local', // RTO22a - applied upon ACK
```
lawrence-forooghian: these comments on the cases are unnecessary

src/plugins/liveobjects/realtimeobject.ts
```diff
@@ -356,9 +426,11 @@ export class RealtimeObject {
     existingObjectUpdates.forEach(({ object, update }) => object.notifyUpdated(update));
   }
 
-  private _applyObjectMessages(objectMessages: ObjectMessage[]): void {
+  /** @spec RTO9 */
+  private _applyObjectMessages(objectMessages: ObjectMessage[], source: ObjectsOperationSource): void {
     for (const objectMessage of objectMessages) {
       if (!objectMessage.operation) {
+        // RTO9a1 - warn and skip
```
lawrence-forooghian: this comment is unrelated to the PR, remove

src/plugins/liveobjects/realtimeobject.ts
```diff
@@ -368,6 +440,14 @@ export class RealtimeObject {
         continue;
       }
 
+      const serial = objectMessage.serial;
+
+      // RTO9a3 - Skip if already applied on ACK
+      if (serial && this._appliedOnAckSerials.has(serial)) {
+        this._appliedOnAckSerials.delete(serial);
```
lawrence-forooghian, line 513: find the other spec points that say to log a debug or trace message, and do that here (c.f. timeserials check), and update the spec too

src/common/lib/client/realtimechannel.ts
```diff
@@ -520,13 +520,18 @@ class RealtimeChannel extends EventEmitter {
     await this.sendMessage(msg);
   }
 
-  async sendState(objectMessages: WireObjectMessage[]): Promise<void> {
+  /** @spec RTO15h */
```
lawrence-forooghian: don't add spec annotation here, and continue the returning of undefined from the sendMessage method, move the handling into `publish()`

src/common/lib/client/realtimechannel.ts
```diff
@@ -520,13 +520,18 @@ class RealtimeChannel extends EventEmitter {
...
+  async sendState(objectMessages: WireObjectMessage[]): Promise<API.PublishResult> {
     const msg = protocolMessageFromValues({
       action: actions.OBJECT,
       channel: this.name,
       state: objectMessages,
     });
-    await this.sendMessage(msg);
+    const result = await this.sendMessage(msg);
+    if (!result) {
+      throw new this.client.ErrorInfo('No PublishResult returned for OBJECT message', 50000, 500);
```
lawrence-forooghian: this isn't how we throw ErrorInfo; look elsewhere in this file and copy

src/plugins/liveobjects/realtimeobject.ts
```diff
@@ -241,6 +260,44 @@ export class RealtimeObject {
     return this._channel.sendState(encodedMsgs);
   }
 
+  /**
+   * @internal
```
lawrence-forooghian, line 259: Add a brief comment here explaining what this method does

src/plugins/liveobjects/realtimeobject.ts
```diff
@@ -241,6 +260,44 @@ export class RealtimeObject {
...
+          serial, // RTO20c1a
+          siteCode: this._channel.connectionManager.connectionDetails?.siteCode, // RTO20c1b
+        },
+        this._client.Utils,
+        this._client.MessageEncoding,
+      );
+
+      if (this._state !== ObjectsState.synced) {
+        // RTO20c2 - Buffer and wait for sync
+        await new Promise<void>((resolve) => {
```
lawrence-forooghian: let's have some debug-level logging here

src/plugins/liveobjects/realtimeobject.ts
```diff
@@ -241,6 +260,44 @@ export class RealtimeObject {
...
+  async publishAndApply(objectMessages: ObjectMessage[]): Promise<void> {
+    // RTO20b
+    const publishResult = await this.publish(objectMessages);
+
+    // RTO20c
+    for (let i = 0; i < objectMessages.length; i++) {
+      const originalMsg = objectMessages[i];
+      const serial = publishResult.serials[i];
+
+      if (serial === null) continue; // Discarded by conflation
```
lawrence-forooghian: - `siteCode` can be extracted outside this loop — let's log a warning if it's absent, and update the spec to match
- let's log a warning and skip the message if `serial` is absent — update the spec too

src/plugins/liveobjects/livecounter.ts
```diff
@@ -117,15 +118,17 @@ export class LiveCounter extends LiveObject<LiveCounterData, LiveCounterUpdate>
...
       );
-      return;
+      return false; // RTLC7b
+    }
+
+    // RTLC7c - Only update siteTimeserials if source is CHANNEL
+    if (source === ObjectsOperationSource.channel) {
+      this._siteTimeserials[opSiteCode] = opSerial;
     }
-    // should update stored site serial immediately. doesn't matter if we successfully apply the op,
```
lawrence-forooghian, line 122: maintain this comment, inside the `if`

src/plugins/liveobjects/livecounter.ts
```diff
@@ -117,15 +118,17 @@ export class LiveCounter extends LiveObject<LiveCounterData, LiveCounterUpdate>
         'LiveCounter.applyOperation()',
         `skipping ${op.action} op: op serial ${opSerial.toString()} <= site serial ${this._siteTimeserials[opSiteCode]?.toString()}; objectId=${this.getObjectId()}`,
       );
-      return;
+      return false; // RTLC7b
+    }
+
+    // RTLC7c - Only update siteTimeserials if source is CHANNEL
```
lawrence-forooghian: just the spec point, no commentary, please. ditto for LiveMap

src/plugins/liveobjects/livecounter.ts
lawrence-forooghian: ditto for LiveMap if applicable

test/realtime/liveobjects.test.js
```diff
@@ -163,6 +163,126 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
     });
   }
 
+  /**
+   * Creates an interceptor that holds OBJECT messages (echoes) and allows controlled release, to test ACK-before-echo scenarios.
+   * Call when CONNECTING or CONNECTED to intercept messages on the active transport.
+   *
+   * Returns an object with:
+   *   - heldEchoes: array of held messages
+   *   - waitForEcho(): resolves when at least one echo has been intercepted
```
lawrence-forooghian: isn't this actually "when the next echo is intercepted"? ditto below? and double-check that this contract is sufficient for our tests

test/realtime/liveobjects.test.js
```diff
@@ -587,7 +707,7 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
               'arrayKey',
               'mapKey',
             ];
-            expect(valuesMap.size()).to.equal(valueMapKeys.length, 'Check nested map has correct number of keys');
```
lawrence-forooghian, line 590: undo these changes to the "nested" language; they're unrelated to this PR

test/realtime/liveobjects.test.js
```diff
@@ -3330,17 +3429,13 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
             helper.recordPrivateApi('read.DefaultInstance._value');
             expectInstanceOf(map._value, 'LiveMap', 'Check map instance is of an expected class');
 
-            expect(nestedMap, 'Check nested map exists').to.exist;
```
lawrence-forooghian, line 3448: undo these changes where you've removed the word "nested" for no reason — check this throughout the PR

test/realtime/liveobjects.test.js
```diff
@@ -8194,6 +8219,477 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
       }, client);
     });
 
+    /**
```
lawrence-forooghian: we can remove this comment

test/realtime/liveobjects.test.js
```diff
@@ -8194,6 +8219,477 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
...
+     */
+    describe('Apply on ACK', () => {
+      /**
+       * Operations applied locally on ACK
+       *
+       * Verify that after a write operation promise resolves, the value is immediately
+       * visible without waiting for the echo.
+       */
+      describe('Operations applied locally on ACK', () => {
+        it('creating a LiveCounter applies immediately on ACK', async function () {
```
lawrence-forooghian: the scenarios within this `describe` all seem very similar; can we DRY them up with scenario arrays like we do in the existing tests?

test/realtime/liveobjects.test.js
```diff
@@ -8194,6 +8219,477 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
...
+            await channel.attach();
+            const root = await channel.object.get();
+
+            // create a counter
+            await root.set('counter', LiveCounter.create(10));
+            const counter = root.get('counter');
+            expect(counter.value()).to.equal(10, 'Check counter has initial value of 10');
+
+            // hold echoes so we can verify value comes from ACK, not echo
+            createEchoInterceptor(helper, client);
```
lawrence-forooghian: if we were to DRY up, we'd probably need to do the intercepting before creating the counter (i.e. creating the counter would not be part of the test setup)

test/realtime/liveobjects.test.js
```diff
@@ -8194,6 +8219,477 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
...
+
+            // create a counter
+            await root.set('counter', LiveCounter.create(10));
+            const counter = root.get('counter');
+
+            // set up ACK interceptor (hold ACKs, let OBJECT messages through)
+            const interceptor = createAckInterceptor(helper, client);
+
+            // set up subscription to wait for the echo
+            const echoApplied = new Promise((resolve) => {
```
lawrence-forooghian: make it clear from name it's a promise

test/realtime/liveobjects.test.js
```diff
@@ -8194,6 +8219,477 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
...
+            // inject ATTACHED with HAS_OBJECTS to start sync
+            const attachedPm = createPM({
+              action: 11, // ATTACHED
+              channel: channel.name,
+              flags: 1 << 7, // HAS_OBJECTS flag
+            });
+            await transport.onProtocolMessage(attachedPm);
+
+            // complete sync with state that has OLD timeserials (before our operation)
+            // using timeserial 'aaa@0-0' which is older than any real operation timeserial
```
lawrence-forooghian: this test doesn't seem right:

1. how do you know 'aaa@0-0' is older than any real operation timeserial? 
2. how do you know that the echo has been rejected due to the siteTimeserials check? you don't know that the siteCode that will be used when applying on ack is necessarily `aaa` i.e. this comparison might not even happen

How could we improve on this? Perhaps we need to focus the test on, instead of thinking about "older timeserial", just making sure that the fake OBJECT_SYNC contains data to which our operation hasn't been applied — i.e. by making sure that our siteCode doesn't even appear in siteTimeserials, e.g. by using a clearly fake site code such as "fake"?

test/realtime/liveobjects.test.js
```diff
@@ -8194,6 +8219,477 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
...
+              receivedEvents.push({
+                action: event.message?.operation?.action,
+                amount: event.message?.operation?.counterOp?.amount,
+              });
+            });
+
+            // 1. trigger operation via SDK (applied locally on ACK)
+            await counter.increment(5);
+
+            expect(receivedEvents.length).to.equal(1, 'Check event fires for locally-applied operation');
```
lawrence-forooghian, line 8882: maybe just add a comment stating that we don't need to wait for the event — we expect it to have been emitted before the `increment()` call returns?

test/realtime/liveobjects.test.js
```diff
@@ -163,6 +163,126 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
     });
   }
 
+  /**
+   * Creates an interceptor that holds OBJECT messages (echoes) and allows controlled release, to test ACK-before-echo scenarios.
```
lawrence-forooghian: an echo is a subset of OBJECT message (i.e. one that we sent).

we can keep the name but there should be a comment here explaining that the name just reflects how it's intended to be used, and that it doesn't actually perform any special filtering for echoes

src/plugins/liveobjects/realtimeobject.ts
```diff
@@ -18,6 +19,22 @@ export enum ObjectsEvent {
...
+  channel = 'channel',
+}
+
+/** @spec RTO7c */
+interface BufferedAckEntry {
+  objectMessages: ObjectMessage[]; // RTO7c1
+  /** When called, causes the promise returned by `publishAndApply()` to resolve.
+   *
+   * @spec RTO7c2
+   */
```
VeskeR: nitpick:
move the comment to the new line:
```
  /**
   * When called, causes the promise returned by `publishAndApply()` to resolve.
   * @spec RTO7c2
   */
```

src/plugins/liveobjects/realtimeobject.ts
```diff
@@ -238,7 +261,83 @@ export class RealtimeObject {
       );
     }
 
-    return this._channel.sendState(encodedMsgs);
+    // RTO15h
+    const result = await this._channel.sendState(encodedMsgs);
+    if (!result) {
+      throw new this._client.ErrorInfo('No PublishResult returned for OBJECT message', 40000, 400);
```
VeskeR: `RTO15h` says to return `PublishResult` in the same way as `RealtimeChannel#publish` [RTL6j](https://sdk.ably.com/builds/ably/specification/main/features/#RTL6j).
Now I appreciate that spec doesn't specify how to handle situations when  `PublishResult` is missing (neither it explains when/why it could happen), but looking at ably-js implementation for channel.publish at https://github.com/ably/ably-js/blob/441cec20d8cb6575b3052b6a0604e8cb9744fad4/src/common/lib/client/realtimechannel.ts#L291, the current behavior is to default to the empty `serials` list, instead of throwing an error.

Do you think it would be more appropriate to do the same here (return empty `serials` list), and let other parts that rely on serials to handle this case if the expected serial is missing?
For example, I can see this publishResult in only used in the new `publishAndApply` and it includes a null check for a serial: https://github.com/ably/ably-js/pull/2155/changes#diff-211bf3115a3d050a1cfad9daf12519fb8056c16aea5be1e589f8b16b84fea4e4R303-R306

VeskeR: Outside of the scope of this PR, but if defaulting to the empty `serials` is an expected behavior, we should add it to the spec for `channel.publish` and for the new `RTO15h` from https://github.com/ably/specification/pull/419/changes#diff-c598156b2e1e7159ee7addc9b73d820e220a41135e64bbde772aa5b838340f25R223

src/plugins/liveobjects/realtimeobject.ts
```diff
@@ -238,7 +261,83 @@ export class RealtimeObject {
...
+    if (this._state !== ObjectsState.synced) {
+      // RTO20e - Buffer and wait for sync
+      this._client.Logger.logAction(
+        this._client.logger,
+        this._client.Logger.LOG_MICRO,
+        'RealtimeObject.publishAndApply()',
+        `buffering ${syntheticMessages.length} message(s) until sync completes; channel=${this._channel.name}`,
+      );
+      await new Promise<void>((resolve) => {
+        this._bufferedAcks.push({ objectMessages: syntheticMessages, signal: resolve }); // RTO20e1
```
VeskeR: Discussed this on a call, posting here a summary and required actions:

This call essentially waits for the channel's LiveObjects state to become SYNCED. (Note: as discussed in https://github.com/ably/specification/pull/419#discussion_r2747222232, this will be rewritten to simply await a SYNCED event instead of dealing with `_bufferedAcks` directly, but the underlying issue remains the same.)

The problem: if during `publishAndApply` we experience a connection loss in such a way that publish still succeeds (it was able to send or retry the operation and receive an ACK), but as a result we enter the SYNCING state, we will wait indefinitely for the LiveObjects state to become SYNCED again. This may never happen if we are stuck in a disconnection loop or experiencing network disturbances and are never able to receive the full sync sequence. As a result, a mutation operation (map.set, map.remove, counter.increment) will hang indefinitely until the connection fully restores and we are able to complete the sync.

This was not the case prior to this PR. RealtimeObject.publish relied only on `ConnectionManager.send`, which handled message queuing and had clear cut-off points for invalid states (see https://github.com/ably/ably-js/blob/441cec20d8cb6575b3052b6a0604e8cb9744fad4/src/common/lib/transport/connectionmanager.ts#L1750-L1782 and https://github.com/ably/ably-js/blob/441cec20d8cb6575b3052b6a0604e8cb9744fad4/src/common/lib/transport/connectionmanager.ts#L1261-L1268). The newly introduced await for the SYNCED event no longer provides the same guarantees.

Context: conceptually, this is not an entirely new pattern. The same behavior exists for presence.get(), which waits for the presence sync event (https://github.com/ably/ably-js/blob/441cec20d8cb6575b3052b6a0604e8cb9744fad4/src/common/lib/client/presencemap.ts#L179-L192), and for object.get(), which waits for the SYNCED state (https://github.com/ably/ably-js/blob/441cec20d8cb6575b3052b6a0604e8cb9744fad4/src/plugins/liveobjects/realtimeobject.ts#L87).

However, this problem is exacerbated by the nature of mutation operations compared to presence.get() / object.get(). You would expect .get() calls to happen once at the beginning of the app lifecycle to retrieve the initial state, after which it is automatically synced over time. If there are connection issues at startup, it is much more intuitive for the end user to refresh the page and try again when something doesn't load.

Mutation operations, on the other hand, are smaller, more intentional, and usually user-initiated. Imagine a button that increments a counter and displays a loader while waiting for `publishAndApply` to complete. If it hangs due to the issue described above, the user will see an indefinite loader on a button, which degrades the user experience drastically - it would force the user to refresh the entire page.

As such, the infinite-await problem is much more pronounced for mutation operations, and we should think about the general approach we want to take in these cases.

We may proposed a general rule: a publicly available async API should never hang indefinitely and must have a clear cut-off point. For `presence.get()`, `object.get()`, and `publishAndApply`, this could be achieved by using Promise.race with a terminal connection or channel state that we consider "bad enough" to justify throwing an error - for example, the same states that cause a queued message to fail in `ConnectionManager`.

___

### Actions:
decide whether this issue is critical enough to fix for mutation operations in this PR, or open a separate issue to address it for mutation operations and `presence.get()` / `object.get()` together. Might be worth opening a DR for this.

src/plugins/liveobjects/realtimeobject.ts
```diff
@@ -238,7 +261,83 @@ export class RealtimeObject {
...
+      this._client.logger,
+      this._client.Logger.LOG_MICRO,
+      'RealtimeObject.publishAndApply()',
+      `received ACK for ${objectMessages.length} message(s), applying locally; channel=${this._channel.name}`,
+    );
+
+    // RTO20c
+    const siteCode = this._channel.connectionManager.connectionDetails?.siteCode;
+    if (!siteCode) {
+      throw new this._client.ErrorInfo(
```
VeskeR: Discussed this on a call, posting here a summary and required actions:

`publishAndApply` may throw after a successful publish - that is, after the server has accepted the operation - but before applying the change locally. In this case, the caller receives an error suggesting the operation failed. However, since the server already accepted the operation, the echoed message will eventually arrive over the channel and the operation will be applied locally through the normal flow (the same behavior we had before apply-on-ACK, where all changes were applied upon receiving the echo). This creates an ambiguous situation: we signal failure, but the operation did succeed server-side, and the local state will eventually reflect it.

There are two places where this can happen:
1. `siteCode` is not present in `connectionDetails` - this would be a server error; the server is expected to always provide `siteCode`.
2. The `PublishResult` contains a null serial for one of the published operations - _most likely_ a server error as well (**we should confirm with the Realtime team that there are no legitimate situations where a serial is not returned**).

In practice, both cases are very unlikely since they require the server to behave incorrectly by omitting required fields.

The question is how we expect developers to handle these errors. There are two scenarios:

- End-user-initiated action (e.g. updating a counter by clicking a button): the developer catches the error, surfaces it to the user, and the user decides whether to retry. They may also observe the echoed operation arriving and see that the action actually succeeded, choosing not to retry.
- Automated action (e.g. incrementing a counter on each page visit): if the developer retries on any error, it could result in a double count when the error was thrown after a successful publish. The correct solution here is to support message metadata for LiveObjects operations, enabling the caller to pass a message id for idempotent publishing (see https://ably.com/docs/pub-sub/advanced#idempotency). This was outlined in the https://ably.atlassian.net/wiki/spaces/LOB/pages/4235722804/LODR-042+LiveObjects+Realtime+Client+API+Improvements#Message-metadata but has not been implemented yet. The current `publishAndApply` implementation escalates the need for message ID support in LiveObjects operations.

___

### Summary:
no changes required in this PR. The scenarios where we throw after a successful publish correspond to genuine server-side errors where something has gone seriously wrong. If customers raise the need for reliable retries of LiveObjects operations in the future, we can address it by  implementing message IDs for idempotent publishing.

### Actions:
We should confirm with the Realtime team that PublishResult always includes serials for published operations.

src/plugins/liveobjects/realtimeobject.ts
lawrence-forooghian: This error no longer exists now that this PR is based on top of #2166

src/plugins/liveobjects/realtimeobject.ts
lawrence-forooghian: I think the gist here is the following:

- it has always been possible for the LiveObjects mutation methods to fail even though the operation has been accepted by Realtime (e.g. by becoming `SUSPENDED` when we haven't yet received the ACK); this does not change that
- in such a situation, it would be good to have an idempotency mechanism that allows the user to retry without accidental repeated operations; this work is already being considered and is unrelated to this PR
- if the server does not send the data that the client needs in order to apply-on-ACK then the client (obviously) cannot apply-on-ACK; we need to convince ourselves this can not happen, and we can continue discussing this in https://github.com/ably/specification/pull/419#discussion_r2798512505

src/plugins/liveobjects/realtimeobject.ts
lawrence-forooghian: I've added new spec point RTO20e1 (https://github.com/ably/specification/commit/57f4449d703fcd0544ac98488f8d555617254613) to handle this specifically for apply-on-ACK (implementation to come shortly). I'll make a note for us to think about what we should do for `presence.get()` and `object.get()`.

src/plugins/liveobjects/realtimeobject.ts
lawrence-forooghian: done

src/plugins/liveobjects/realtimeobject.ts
paddybyers: > * The current `publishAndApply` implementation escalates the need for message ID support in LiveObjects operations.

I don't really agree with this. For operations that are non-idempotent, there are other much more likely scenarios that will result in the client re-attempting an operation that had in fact succeeded (eg the client fails to see the ack). We already need to support ids.

paddybyers: On the point being made: I think the publish promise should resolve if the publish succeeded, and not otherwise. There shouldn't be any valid circumstances that, following receipt of the ack, the apply fails. If something happens, out of spec, that causes the apply to fail, then we should be treating this as a protocol error that invalidates the transport, leading to reconnection and resync.

lawrence-forooghian: > There shouldn't be any valid circumstances that, following receipt of the ack, the apply fails.

That's interesting — what do you think about [this discussion](https://github.com/ably/ably-js/pull/2155#discussion_r2813084019) in that case?

paddybyers: I think the conclusion in that discussion is correct, that all operations will eventually complete, even if that's ultimately triggered by a channel or connection state change. The fact that `presence.get()` doesn't do that right now is a spec bug I think.

lawrence-forooghian: > I think the conclusion in that discussion is correct, that all operations will eventually complete

What do you mean by "eventually"? The point that Andrii was making was that, if the LiveObjects mutation methods don't apply the operation — and thus don't complete — until the objects sync state becomes `SYNCED` (and if there are no other circumstances in which they complete) then they may never complete.

Thus, in response, I today specified a new behaviour in which if, whilst waiting for the state to become `SYNCED`, the channel becomes `DETACHED` , `SUSPENDED`, or `FAILED`, then the mutation methods' promises will reject, and the operation will not be applied locally. But this new behaviour contradicts your principle of "There shouldn't be any valid circumstances that, following receipt of the ack, the apply fails."

paddybyers: Sorry, I meant that we should ensure that all operations eventually complete, following the same approach that you proposed.

> But this new behaviour contradicts your principle of "There shouldn't be any valid circumstances that, following receipt of the ack, the apply fails."

Yes, sorry I should have been clearer. I'm saying that if you get an ack, and attempt to apply the operation locally as a result, then that should always succeed. In the sync case you're not attempting to apply it. There's also the conflation case (see https://github.com/ably/specification/pull/419#discussion_r2813913295) where you might not attempt to apply the operation on ack.

src/plugins/liveobjects/realtimeobject.ts
lawrence-forooghian: implementation in https://github.com/ably/ably-js/pull/2155/commits/7de874893a9573f6931f5670e43e44985a9421d7 — will squash pre-merge

lawrence-forooghian:
this is now up to date with spec commit 57f4449; the only outstanding conversation there is https://github.com/ably/specification/pull/419#discussion_r2747215603 which I'll discuss with Paddy shortly

src/plugins/liveobjects/realtimeobject.ts
lawrence-forooghian: I've implemented the changes from https://github.com/ably/specification/pull/419/commits/24baaa2fcf93b1cd3858d00945aad33ef37b4c97, which handle:

- server misbehaviour (no `siteCode`, malformed `ACK`)
- `serial === null`

src/plugins/liveobjects/realtimeobject.ts
```diff
@@ -238,9 +269,119 @@ export class RealtimeObject {
...
+        const rejectHandle = {
+          reject: (err: ErrorInfo) => {
+            cleanup();
+            reject(err);
+          },
+        };
+        this._eventEmitterInternal.once(ObjectsEvent.synced, onSynced);
+        this._pendingOperations.push(rejectHandle);
+      });
+    }
```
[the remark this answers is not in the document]
lawrence-forooghian: What is the scenario in which there'd be a transition to suspended between the return of `publish` and the `synced` check?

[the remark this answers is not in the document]
lawrence-forooghian: > At this point, control returns to publishAndApply but before the next line executes, the JavaScript event loop processes other queued tasks

Is that possible? once control returns to publishAndApply doesn't execution proceed synchronously until we next yield to the event loop?

src/plugins/liveobjects/realtimeobject.ts
```diff
@@ -238,9 +269,119 @@ export class RealtimeObject {
...
+        };
+        const rejectHandle = {
+          reject: (err: ErrorInfo) => {
+            cleanup();
+            reject(err);
+          },
+        };
+        this._eventEmitterInternal.once(ObjectsEvent.synced, onSynced);
+        this._pendingOperations.push(rejectHandle);
+      });
```
VeskeR, line 358: It seems to me that this whole sequence with `_pendingOperations` and coupling between `actOnChannelState` and `publishAndApply` is rather unnecessary and arguably hard to read.
Would it be possible to just rely on `Promise.race` call and `channel.once` to become detached/suspended/failed?

Something like this:

```typescript
  if (this._state !== ObjectsState.synced) {
    await Promise.race([
      this._eventEmitterInternal.once(ObjectsEvent.synced),
      this._channel.once(['detached', 'suspended', 'failed']),
    ]);

    // RTO20e1
    if (['detached', 'suspended', 'failed'].includes(this._channel.state)) {
      throw new this._client.ErrorInfo(
        `objects sync could not be completed due to the channel entering the ${this._channel.state} state`,
        92008,
        400,
        this._channel.errorReason || undefined,
      );
    }
  }
```

src/plugins/liveobjects/realtimeobject.ts
VeskeR: 👍 will keep this conversation unresolved for visibility.

src/plugins/liveobjects/realtimeobject.ts
VeskeR: We updated the implementation to log instead of throw according to the spec.
Will keep this conversation unresolved for visibility.

src/plugins/liveobjects/livecounter.ts
```diff
@@ -72,15 +72,15 @@ export class LiveCounter extends LiveObject<LiveCounterData, LiveCounterUpdate>
...
    *
-   * This does not modify the underlying data of this LiveCounter object. Instead, the change will be applied when
-   * the published COUNTER_INC operation is echoed back to the client and applied to the object following the regular
-   * operation application procedure.
+   * The change will be applied locally when the ACK is received from Realtime.
    *
-   * @returns A promise which resolves upon receiving the ACK message for the published operation message.
+   * @returns A promise which resolves upon receiving the ACK message for the published operation message
+   * and applying the operation locally.
+   * @spec RTLC12g
```
VeskeR:
```suggestion
   * @spec RTLC12
```

src/plugins/liveobjects/livemap.ts
```diff
@@ -265,38 +265,39 @@ export class LiveMap<T extends Record<string, Value> = Record<string, Value>>
...
    *
-   * This does not modify the underlying data of this LiveMap object. Instead, the change will be applied when
-   * the published MAP_SET operation is echoed back to the client and applied to the object following the regular
-   * operation application procedure.
+   * The change will be applied locally when the ACK is received from Realtime.
    *
-   * @returns A promise which resolves upon receiving the ACK message for the published operation message.
+   * @returns A promise which resolves upon receiving the ACK message for the published operation message
+   * and applying the operation locally.
+   * @spec RTLM20g
```
VeskeR:
```suggestion
   * @spec RTLM20
```

src/plugins/liveobjects/livemap.ts
```diff
@@ -265,38 +265,39 @@ export class LiveMap<T extends Record<string, Value> = Record<string, Value>>
...
    *
-   * This does not modify the underlying data of this LiveMap object. Instead, the change will be applied when
-   * the published MAP_REMOVE operation is echoed back to the client and applied to the object following the regular
-   * operation application procedure.
+   * The change will be applied locally when the ACK is received from Realtime.
    *
-   * @returns A promise which resolves upon receiving the ACK message for the published operation message.
+   * @returns A promise which resolves upon receiving the ACK message for the published operation message
+   * and applying the operation locally.
+   * @spec RTLM21g
```
VeskeR:
```suggestion
   * @spec RTLM21
```

src/plugins/liveobjects/realtimeobject.ts
```diff
@@ -164,6 +175,7 @@ export class RealtimeObject {
 
   /**
    * @internal
+   * @spec RTO8b
```
VeskeR:
```suggestion
   * @spec RTO8
```

test/realtime/liveobjects.test.js
```diff
@@ -8309,6 +8352,570 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
...
+                ctx.get('counter').increment(5);
+                ctx.get('counter').increment(3);
+              });
+              expect(counter.value()).to.equal(18, 'Check batch increments applied on ACK');
+            },
+          },
+        ];
+
+        for (const scenario of applyOnAckScenarios) {
+          it(scenario.description, async function () {
```
VeskeR: I suggest using existing `forScenarios`  helper as it gives some niceties like allowing to specify only/skip for individual scenarios, and a `allTransportsAndProtocols` flag  to test on all transports/protocols when relevant for a test

src/plugins/liveobjects/realtimeobject.ts
lawrence-forooghian: Yeah, the reason I didn't was because I didn't want to use `once()` (because that would be an RTC10 violation) and we didn't have another obvious channel state event emitter to use. Will rebase this on https://github.com/ably/ably-js/pull/2167 and use the `_internalStateChanges` that adds

lawrence-forooghian: Done

src/plugins/liveobjects/livecounter.ts
lawrence-forooghian: Done

src/plugins/liveobjects/livemap.ts
lawrence-forooghian: Done

src/plugins/liveobjects/livemap.ts
lawrence-forooghian: Done

src/plugins/liveobjects/realtimeobject.ts
lawrence-forooghian: Done

test/realtime/liveobjects.test.js
lawrence-forooghian: Thanks, done

src/common/lib/util/eventemitter.ts
```diff
@@ -226,6 +226,7 @@ class EventEmitter {
    * @param event the name of the event to listen to
    */
   once(event: string): Promise<void>;
+  once(event: string[]): Promise<void>;
```
VeskeR: Mind adding a jsdoc for the overload?
something like:
```
  /**
   * Listen for a single occurrence of any of the specified events.
   * Resolves when the first matching event is emitted.
   * @param event the names of the events to listen to
   */
```

src/plugins/liveobjects/realtimeobject.ts
```diff
@@ -238,9 +249,113 @@ export class RealtimeObject {
...
+        this._client.Logger.LOG_MICRO,
+        'RealtimeObject.publishAndApply()',
+        `waiting for sync to complete before applying ${syntheticMessages.length} message(s); channel=${this._channel.name}`,
+      );
+
+      await Promise.race([
+        this._eventEmitterInternal.once(ObjectsEvent.synced),
+        // RTO20e1
+        this._channel._internalStateChanges.once(['detached', 'suspended', 'failed']).then(() => {
+          throw new this._client.ErrorInfo(
```
VeskeR: If synced wins the race, the `_internalStateChanges.once` listener is still registered. When the channel eventually enters detached/suspended/failed, the chained .then() will throw with nobody awaiting it,
and I belive will cause an unhandled promise rejection.

I appreciate the intent to avoid a separate channel state check as in my https://github.com/ably/ably-js/pull/2155#discussion_r2823337052 (even though I don't see a scenario where we become synced but are
in the detached/suspended/failed channel state - I agree we shouldn't create such a dependency here).

What do you think about the following approach?
```typescript
await new Promise<void>((resolve, reject) => {
  const onSynced = () => {
    this._channel._internalStateChanges.off(onChannelState);
    resolve();
  };
  const onChannelState = () => {
    this._eventEmitterInternal.off(ObjectsEvent.synced, onSynced);
    reject(new this._client.ErrorInfo(
      `the operation could not be applied locally due to the channel entering the ${this._channel.state} state whilst waiting for objects sync to complete`,
      92008,
      400,
      this._channel.errorReason || undefined,
    ));
  };
  this._eventEmitterInternal.once(ObjectsEvent.synced, onSynced);
  this._channel._internalStateChanges.once(['detached', 'suspended', 'failed'], onChannelState);
});
```

This accomplishes the same intent without dangling listeners or unhandled rejections.

src/plugins/liveobjects/realtimeobject.ts
VeskeR: see https://github.com/ably/ably-js/pull/2155#discussion_r2828223041

test/realtime/liveobjects.test.js
```diff
@@ -8309,6 +8352,567 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
...
+                // start increment - this will publish and wait for ACK
+                const incrementPromise = counter.increment(5);
+
+                // wait for ACK to be intercepted, then release it
+                // this lets publishAndApply proceed past publish into the sync-wait
+                await interceptor.waitForAck();
+                interceptor.releaseAll();
+
+                // yield to the event loop so publishAndApply reaches the sync wait
+                await new Promise((resolve) => setTimeout(resolve, 0));
```
VeskeR, line 8820: suggest replacing setTimeout with:
```
          helper.recordPrivateApi('call.Platform.nextTick');
          await new Promise((res) => nextTick(res));
```

test/realtime/liveobjects.test.js
```diff
@@ -8309,6 +8352,567 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
...
+
+            // create a counter
+            await root.set('counter', LiveCounter.create(10));
+            const counter = root.get('counter');
+
+            // set up ACK interceptor (hold ACKs, let OBJECT messages through)
+            const interceptor = createAckInterceptor(helper, client);
+
+            // set up subscription to wait for the echo
+            const echoAppliedPromise = new Promise((resolve) => {
```
VeskeR: can use `waitForCounterUpdate` helper here instead

test/realtime/liveobjects.test.js
```diff
@@ -8309,6 +8352,567 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
...
+            const counter = root.get('counter');
+            const counterId = counter.instance().id;
+
+            // wait for create echo and extract the serial
+            await interceptor.waitForEcho();
+            helper.recordPrivateApi('read.ProtocolMessage.state');
+            const createEcho = interceptor.heldEchoes[0].message;
+            // the create has two operations (COUNTER_CREATE and MAP_SET);
+            // get the last one's serial, which is what siteTimeserials will be set to
+            const createState = createEcho.state[createEcho.state.length - 1];
```
VeskeR: I suggest to do `createEcho.state.find(x => x.operation.action === LiveObjectsHelper.ACTIONS.COUNTER_CREATE)` instead to determinitstically extract the message for `COUNTER_CREATE `, as the order of operations may change in the future. More specifically, see https://ably.atlassian.net/browse/AIT-270

test/realtime/liveobjects.test.js
```diff
@@ -8309,6 +8352,567 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
...
+            const counter = root.get('counter');
+            const counterId = counter.instance().id;
+
+            // now simulate a new OBJECT_SYNC sequence starting
+            // inject an ATTACHED with HAS_OBJECTS to trigger SYNCING state
+            helper.recordPrivateApi('call.connectionManager.activeProtocol.getTransport');
+            helper.recordPrivateApi('call.transport.onProtocolMessage');
+            helper.recordPrivateApi('call.makeProtocolMessageFromDeserialized');
+            const transport = client.connection.connectionManager.activeProtocol.getTransport();
+            const attachedPm = createPM({
```
VeskeR: We already have `injectAttachedMessage` helper method in 'Sync events' tests. Move it to the top of the file and reuse it here

test/realtime/liveobjects.test.js
```diff
@@ -8309,6 +8352,567 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
...
+            // now inject an OBJECT_SYNC that doesn't include our operation
+            // this simulates a re-sync where the server state doesn't include our operation yet
+            // the sync will reset the counter to 10 and clear appliedOnAckSerials
+            helper.recordPrivateApi('call.connectionManager.activeProtocol.getTransport');
+            helper.recordPrivateApi('call.transport.onProtocolMessage');
+            helper.recordPrivateApi('call.makeProtocolMessageFromDeserialized');
+            const transport = client.connection.connectionManager.activeProtocol.getTransport();
+
+            // inject ATTACHED with HAS_OBJECTS to start sync
+            const attachedPm = createPM({
```
VeskeR: ditto `injectAttachedMessage`

test/realtime/liveobjects.test.js
```diff
@@ -8309,6 +8352,567 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
...
+
+                await root.set('counter', LiveCounter.create(10));
+                const counter = root.get('counter');
+
+                // inject ATTACHED with HAS_OBJECTS to trigger SYNCING state
+                helper.recordPrivateApi('call.connectionManager.activeProtocol.getTransport');
+                helper.recordPrivateApi('call.transport.onProtocolMessage');
+                helper.recordPrivateApi('call.makeProtocolMessageFromDeserialized');
+                const transport = client.connection.connectionManager.activeProtocol.getTransport();
+                const attachedPm = createPM({
```
VeskeR: ditto `injectAttachedMessage`

test/realtime/liveobjects.test.js
```diff
@@ -8309,6 +8352,567 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
...
+            // no need to wait for event - with apply-on-ACK, the subscription callback
+            // is invoked synchronously before increment() returns
+            expect(receivedEvents.length).to.equal(1, 'Check event fires for locally-applied operation');
+            expect(receivedEvents[0]).to.deep.equal(
+              { action: 'counter.inc', amount: 5 },
+              'Check event from local apply has correct structure',
+            );
+
+            // 2. trigger operation via REST (received over Realtime)
+            const realtimeEventPromise = new Promise((resolve) => {
```
VeskeR: ditto `waitForCounterUpdate`

test/realtime/liveobjects.test.js
```diff
@@ -8309,6 +8352,567 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
...
+
+            counter.subscribe((event) => {
+              receivedEvents.push({
+                action: event.message?.operation?.action,
+                amount: event.message?.operation?.counterOp?.amount,
+              });
+            });
+
+            // 1. trigger operation via SDK (applied locally on ACK)
+            await counter.increment(5);
```
VeskeR, line 8878: We may receive an echo before the ACK for the published operation, so this doesn't guarantee that we test the "subscription triggered for applied on ACK" route.
Need to intercept echoes for this case, and then `restore()` before proceeding to "2. received over Realtime" scenario

src/plugins/liveobjects/realtimeobject.ts
lawrence-forooghian: Good point — will do (will do same approach as in 3d1535a — with a single cleanup function — because I find it marginally more readable)

src/common/lib/util/eventemitter.ts
lawrence-forooghian: I'm going to drop the commit that added this since it's not needed any more (we're no longer using the promise-returning variant of `once()`)

test/realtime/liveobjects.test.js
lawrence-forooghian: We don't actually care here about finding the `COUNTER_CREATE` — we just want to find the serial that establishes the initial `siteTimeserials` value i.e. the last thing in the array. I think the variable naming here makes this a bit confusing, will rename to `initialEcho` etc

test/realtime/liveobjects.test.js
lawrence-forooghian: Ah yes, missed that, will add

src/common/lib/util/eventemitter.ts
lawrence-forooghian: Done

src/plugins/liveobjects/realtimeobject.ts
lawrence-forooghian: Done

test/realtime/liveobjects.test.js
lawrence-forooghian: Done

test/realtime/liveobjects.test.js
lawrence-forooghian: Done

test/realtime/liveobjects.test.js
lawrence-forooghian: Done

test/realtime/liveobjects.test.js
lawrence-forooghian: Done

test/realtime/liveobjects.test.js
lawrence-forooghian: Done

test/realtime/liveobjects.test.js
lawrence-forooghian: Done

test/realtime/liveobjects.test.js
VeskeR: >  we just want to find the serial that establishes the initial siteTimeserials value

_on a counter_.

`siteTimeserials` are stored on a per-object basis, and the echoed message holds two operations: `COUNTER_CREATE` and `MAP_SET`. The timeserial from `COUNTER_CREATE` will be applied to a counter, and `MAP_SET` will be applied to the map that holds the counter (root in this case).
Since we intend to construct a serial between the counter create operation and a future increment operation (we even have [a comment](https://github.com/ably/ably-js/pull/2155/changes#diff-1ac1502266f3161e8a56e86a03079a4d38070c39339cdbd45ea90b5c968e1993R8598) explicitly stating that), I think we should ensure we actually get a `COUNTER_CREATE` timeserial.

Since this all happens in a controlled environment on a sandbox, and we make `COUNTER_CREATE` and `MAP_SET` operations at the same time, their timeserials will likely be close enough that taking the latest one still sort of works - but for the wrong reasons (and we now depend on how timeserials are constructed server-side - we expect COUNTER_CREATE and MAP_SET serials to be similar enough). So I still think we should find the serial for `COUNTER_CREATE` so it all checks out.

lawrence-forooghian: Ah yes, sorry, fixed

VeskeR: 👍

PR #2156, opened 2026-01-26
LiveObjects: Test receiving two create ops for the same object ID

lawrence-forooghian:
These tests demonstrate the necessity of the `createOperationIsMerged` flag, which I had previously incorrectly proposed removing — one reason for which was that no tests failed when I tried removing it. See [internal conversation](https://ably-real-time.slack.com/archives/C09SY1AQGK0/p1769459426372989?thread_ts=1769195388.456439&cid=C09SY1AQGK0).

Tests written by Claude.

## Summary by CodeRabbit

* **Tests**
  * Added test coverage for live object creation: verifies duplicate create operations for map and counter objects are ignored so the original create governs final state; follow-up operations confirm a later duplicate create does not change bindings or values.

### DIFF
--- a/test/realtime/liveobjects.test.js
+++ b/test/realtime/liveobjects.test.js
@@ -1358,6 +1358,75 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
         },
 
         {
+          description: 'only one MAP_CREATE operation is applied for the same object ID',
+          action: async (ctx) => {
+            const { entryInstance, objectsHelper, channel } = ctx;
+
+            // It's possible for multiple MAP_CREATE operations, with different serials, to be received
+            // for the same object ID. The object ID is derived from the operation's content, so they will
+            // have identical content. The client should only merge one of these operations into the object's data.
+
+            // create new map and set on root
+            const mapId = objectsHelper.fakeMapObjectId();
+            await objectsHelper.processObjectOperationMessageOnChannel({
+              channel,
+              serial: lexicoTimeserial('aaa', 0, 0),
+              siteCode: 'aaa',
+              state: [
+                objectsHelper.mapCreateOp({
+                  objectId: mapId,
+                  entries: {
+                    foo: { timeserial: lexicoTimeserial('aaa', 0, 0), data: { string: 'bar' } },
+                  },
+                }),
+              ],
+            });
+            await objectsHelper.processObjectOperationMessageOnChannel({
+              channel,
+              serial: lexicoTimeserial('aaa', 1, 0),
+              siteCode: 'aaa',
+              state: [objectsHelper.mapSetOp({ objectId: 'root', key: mapId, data: { objectId: mapId } })],
+            });
+
+            expect(entryInstance.get(mapId).size()).to.equal(1, 'Check map has 1 key after first MAP_CREATE');
+            expect(entryInstance.get(mapId).get('foo').value()).to.equal(
+              'bar',
+              'Check map has correct "foo" value after first MAP_CREATE',
+            );
+
+            // send another MAP_CREATE op for the same object ID, from a different site, with a later entry timeserial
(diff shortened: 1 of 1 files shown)


test/realtime/liveobjects.test.js
```diff
@@ -1849,6 +1918,52 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
...
+              siteCode: 'aaa',
+              state: [objectsHelper.mapSetOp({ objectId: 'root', key: counterId, data: { objectId: counterId } })],
+            });
+
+            expect(entryInstance.get(counterId).value()).to.equal(
+              10,
+              'Check counter has value 10 after first COUNTER_CREATE',
+            );
+
+            // send another COUNTER_CREATE op for the same object ID, with a later serial ('bbb' > 'aaa')
```
VeskeR: nitpick: in this case it's not that `'bbb' > 'aaa'` (is greater) means that operation can be applied. Operations to be applied on an object have their siteTimeserials compared, and in this case the site is different, so the op can be applied based on that check.
For example, see `'COUNTER_CREATE object operation messages are applied based on the site timeserials vector of the object'` test, where ops with different site code are always applied.

test/realtime/liveobjects.test.js
```diff
@@ -1357,6 +1357,75 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
...
+              state: [objectsHelper.mapSetOp({ objectId: 'root', key: mapId, data: { objectId: mapId } })],
+            });
+
+            expect(entryInstance.get(mapId).size()).to.equal(1, 'Check map has 1 key after first MAP_CREATE');
+            expect(entryInstance.get(mapId).get('foo').value()).to.equal(
+              'bar',
+              'Check map has correct "foo" value after first MAP_CREATE',
+            );
+
+            // send another MAP_CREATE op for the same object ID, with a later serial and entry timeserial ('ccc' > 'aaa')
```
VeskeR: nitpick: ditto `'ccc' > 'aaa'` is not really the reason the op can be applied

test/realtime/liveobjects.test.js
```diff
@@ -1357,6 +1357,75 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
...
+                  entries: {
+                    foo: { timeserial: lexicoTimeserial('ccc', 0, 0), data: { string: 'bar' } },
+                  },
+                }),
+              ],
+            });
+
+            // verify the second CREATE was not applied by checking that a MAP_SET with an intermediate
+            // timeserial ('bbb') can still be applied. if the second CREATE had been wrongly applied,
+            // the entry's timeserial would be 'ccc' and this MAP_SET would be rejected.
```
VeskeR, line 1414: On the contrary, this is correct. Operations on map entries additionally compare the timeserial of the latest operation applied to them.

First, the siteTimeserials check: `bbb` is different from `aaa` (or potentially `ccc` if it were incorrectly applied), so it passes.

Then comes the entry timeserial check. This performs a straightforward lexicographical comparison of the latest timeserial applied to the key. Since `aaa` < `bbb` < `ccc`, the test will pass only if `ccc` was not applied earlier.

VeskeR (approved):
LGTM, have a nitpick about the timeserial comment for second CREATE message. Will leave up to you to update the comment

test/realtime/liveobjects.test.js
lawrence-forooghian: Thanks — yep, the key mistake throughout these tests was stating that the `siteTimeserials` check passes due to lexicographically-greater serial, whereas it's actually because of different `siteCode`; will fix.

lawrence-forooghian: Fixed

lawrence-forooghian: Will resolve the other two comments here

PR #2188, opened 2026-04-01
Improve idle timer and connection log messages

mattheworiordan:
## Summary

When running Ably in a server-side environment (e.g. Temporal durable execution where connections are long-lived and transported), I started seeing repeated idle timer error log messages like:

```
19:53:09.769 Ably: Transport.onIdleTimerExpire(): No activity seen from realtime in 158448ms; assuming connection has dropped
20:08:14.963 Ably: Transport.onIdleTimerExpire(): No activity seen from realtime in 904542ms; assuming connection has dropped
20:15:46.999 Ably: Transport.onIdleTimerExpire(): No activity seen from realtime in 436798ms; assuming connection has dropped
21:17:06.799 Ably: Transport.onIdleTimerExpire(): No activity seen from realtime in 379413ms; assuming connection has dropped
```

These are confusing for developers because:
- They don't explain what the SDK will do next (is it disconnecting? reconnecting? is this fatal?)
- They don't include a connection ID, so with multiple connections you can't tell which one is affected
- There's no corresponding log when the connection recovers, so you see alarming messages with no resolution
- The error code (80003) is created internally but never shown in the log
- They refer to "realtime" instead of "Ably"

_Note: I believe these disconnects are expected, they are probably triggered by me opening & closing my laptop_

As we think about how Ably is increasingly used server-side, these messages need to be more useful. This is a quick win to make the logs actionable.

### Changes

- **Idle timer expiry**: now includes connection ID, error code 80003/408, and states the SDK will disconnect and reconnect automatically
- **Reconnection log**: new LOG_MINOR message when connection is resumed or reestablished after disconnection — a bookend so you can see the resolution
- **Connection state freshness**: now explains that the connection state TTL was exceeded and a new connection will be established with channels reattaching; includes connection ID
- **Naming**: replace "realtime" with "Ably" in user-facing log messages

## Test plan

- [x] Existing `idle_transport_timeout` test passes
- [x] Linter clean
- [ ] Review log output in a real server-side environment

## Summary by CodeRabbit

* **Chores**
  * Improved logging clarity for connection transitions by indicating whether a connection was resumed or reestablished, and including the prior connection identifier when different.
  * Made idle-timeout disconnect messages more informative and Ably-specific, reporting freshness thresholds and (when available) the connection identifier; the message now includes the fixed error code 80003 and status 408.

### DIFF
--- a/src/common/lib/transport/connectionmanager.ts
+++ b/src/common/lib/transport/connectionmanager.ts
@@ -644,6 +644,7 @@ class ConnectionManager extends EventEmitter {
     this.activeProtocol = new Protocol(transport);
     this.host = transport.params.host;
 
+    const prevConnId = this.connectionId;
     const connectionKey = connectionDetails.connectionKey;
     if (connectionKey && this.connectionKey != connectionKey) {
       this.setConnection(connectionId, connectionDetails, !!error);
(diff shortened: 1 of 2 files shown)


ttypic (commented):
LGTM

PR #2190, opened 2026-04-17
Improve docstring for channels#release and add log

SimonWoolf:
https://github.com/ably/specification/pull/448

The current docstring almost makes it sound like we expect people to call release(). 99% of people should not. There's a reason the original ably-js comment on this method was   `/* Included to support certain niche use-cases; most users should ignore this. Please do not use this unless you know what you're doing */`

## Summary by CodeRabbit

* **Documentation**
  * Clarified channel "release" behavior, including warnings about using channel references after release and updated state preconditions.
* **Bug Fixes**
  * Releasing a channel now ensures channels in non-terminal states are detached before being released, and release operations are logged and completed asynchronously to avoid leaving dangling references.

### DIFF
--- a/src/common/lib/client/baserealtime.ts
+++ b/src/common/lib/client/baserealtime.ts
@@ -223,15 +223,29 @@ class Channels extends EventEmitter {
    * Please do not use this unless you know what you're doing */
   release(name: string) {
     name = String(name);
+    Logger.logAction(this.logger, Logger.LOG_MAJOR, 'Channels.release()', 'Releasing references to channel ' + name);
     const channel = this.all[name];
     if (!channel) {
       return;
     }
-    const releaseErr = channel.getReleaseErr();
-    if (releaseErr) {
-      throw releaseErr;
+    const s = channel.state;
+    if (s === 'initialized' || s === 'detached' || s === 'failed') {
+      delete this.all[name];
+      return;
     }
-    delete this.all[name];
+    channel
+      .detach()
+      .catch((err) => {
+        Logger.logAction(
+          this.logger,
+          Logger.LOG_ERROR,
+          'Channels.release()',
+          'Error detaching channel ' + name + ' prior to release: ' + Utils.inspectError(err),
+        );
+      })
+      .then(() => {
+        delete this.all[name];
+      });
   }
 }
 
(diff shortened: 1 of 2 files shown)


SimonWoolf:
....I've just looked through and found https://github.com/ably/docs/pull/1057#discussion_r598570764 where apparently we decided to change the spec to make release() implicitly detach first, but then didn't do that in ably-js? bleh. So put this on hold till we figure out what to do with that

SimonWoolf:
I've switched the implementation to an RTS4a-compliant one, ie implicit detach rather than error on unexpected channel state.

(honestly I would prefer an error, this is a dangerous method and I think making it automagic is a step in the wrong direction, but all other SDKs implement the other behaviour and can't be changed to the ably-js semantics without a breaking change, so, shrug, this is the simplest way to uniformity)

src/common/lib/client/baserealtime.ts
```diff
@@ -223,15 +223,29 @@ class Channels extends EventEmitter {
...
-    const releaseErr = channel.getReleaseErr();
-    if (releaseErr) {
-      throw releaseErr;
+    const s = channel.state;
+    if (s === 'initialized' || s === 'detached' || s === 'failed') {
+      delete this.all[name];
+      return;
     }
-    delete this.all[name];
+    channel
```
AndyTWF, line 236: Does this necessitate a major version bump as we're changing the behaviour? Is that the plan?

SimonWoolf: It's a semantics change, but in the direction of being strictly more permissive, which makes it a backwards-compatible change. If someone who was using the function before without it throwing would continue to work. (and note how it would be naturally async now but I did the slightly-awkward .catch().then() so that I'm not changing the signature of the function to return a promise).

This is why I went this route rather than changing ably-java etc to be stricter, which I would have preferred, but that _would_ require a major version bump.

(in some sense, any observable behaviour change could be relied on by someone: it's not literally impossible that someone is relying on the behaviour that release() throws for an attached channel. but it's pretty damn unlikely, and an interpretation of semver where such a change would require a major version bump would imply there can be no such thing as a minor or patch bump. every code change other than pure refactor causes some observable behaviour change or there would be no point doing it, anyone can rely on anything. so ultimately it has to be a judgement over what people are likely to be relying on. cf https://xkcd.com/1172/ )

AndyTWF: Okie dokie!

AndyTWF (approved):

PR #2221, opened 2026-05-19
Propose an alternative to `PathEvent.bubbles`

lawrence-forooghian:
**Note: This PR is based on top of #2222.**

After noticing that the path-based API spec PR (https://github.com/ably/specification/pull/427) didn't include map-entry events nor the bubbling-exclusion mechanism that's implemented in `ably-js`, I wanted to get a better understanding of this exclusion mechanism and why it exists.

From what I can tell, it exists to make sure that if there's a map at path "myMap", and this map emits a `LiveMapUpdate` having key "myKey", then a subscription that covers the path "myMap" will only receive one event (for "myMap"), as opposed to two (for "myMap" and "myMap.myKey").

If that _is_ the only reason that this mechanism exists, then I don't think it's very obvious currently; the intended behaviour isn't documented. I think the implementation could be clearer if it makes the rules explicit:

- for a given `LiveObjectUpdate`, a given subscription receives at most one event per path-to-root
- in the case where a subscription covers both the "myMap" and "myMap.myKey" paths, "myMap" wins

That's what I've tried to do here.

Perhaps I've misunderstood the intent of `bubbles` (which, given that the tests still pass, would suggest a test gap), or perhaps others don't find what I'm proposing clearer (in which case, `bubbles` needs better documentation that explains its motivation). Thoughts, please.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed path-based subscriptions in LiveObjects to correctly notify once per unique path when the same object is referenced from multiple keys, ensuring accurate event handling in complex object hierarchies.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ably/ably-js/pull/2221?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/plugins/liveobjects/pathobjectsubscriptionregister.ts
+++ b/src/plugins/liveobjects/pathobjectsubscriptionregister.ts
@@ -22,12 +22,16 @@ export interface SubscriptionEntry {
  * Event data that LiveObjects provide when notifying of changes
  */
 export interface PathEvent {
-  /** The path where the event occurred */
-  path: Path;
+  /**
+   * Candidate paths for surfacing this event to subscriptions. For a given
+   * subscription, the first candidate path it covers is used as the path of
+   * `event.object` passed to its listener. The caller is responsible for
+   * ordering this array so that the preferred path comes first when more
+   * than one is covered.
+   */
+  candidatePaths: Path[];
   /** Object message that caused this event */
   message?: ObjectMessage;
-  /** Whether this event should bubble up to parent paths. Defaults to true if not specified. */
-  bubbles?: boolean;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/plugins/liveobjects/pathobjectsubscriptionregister.ts
```diff
@@ -18,15 +18,15 @@ export interface SubscriptionEntry {
...
+ * itself which of these paths (if any) to fire on.
  */
-export interface PathEvent {
-  /** The path where the event occurred */
-  path: string[];
-  /** Object message that caused this event */
+export interface LiveObjectUpdateNotification {
+  /** All paths in the LiveObjects tree affected by this update */
+  touchedPaths: string[][];
+  /** Object message that caused this update, if any */
```
lawrence-forooghian: avoid churn, no need for the "if any"

lawrence-forooghian:
@coderabbitai review

lawrence-forooghian:
@coderabbitai review

lawrence-forooghian:
Replaced by #2223

PR #2223, opened 2026-05-19
Propose an alternative to `PathEvent.bubbles` to make intent clearer

lawrence-forooghian:
**Note: This PR is based on top of #2222.**

After noticing that the path-based API spec PR (https://github.com/ably/specification/pull/427) didn't include map-entry events nor the bubbling-exclusion mechanism that's implemented in `ably-js`, I wanted to get a better understanding of this exclusion mechanism and why it exists.

From what I can tell, it exists to make sure that if there's a map at path "myMap", and this map emits a `LiveMapUpdate` having key "myKey", then a subscription that covers the path "myMap" will only receive one event (for "myMap"), as opposed to two (for "myMap" and "myMap.myKey").

If that _is_ the only reason that this mechanism exists, then I don't think it's very obvious currently; the intended behaviour isn't documented. I think the implementation could be clearer if it makes the rules explicit:

- for a given `LiveObjectUpdate`, a given subscription receives at most one event per path-to-root
- in the case where a subscription covers both the "myMap" and "myMap.myKey" paths, "myMap" wins

That's what I've tried to do here.

Perhaps I've misunderstood the intent of `bubbles` (which, given that the tests still pass, would suggest a test gap), or perhaps others don't find what I'm proposing clearer (in which case, `bubbles` needs better documentation that explains its motivation). Thoughts, please.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved path-based subscriptions so listeners are reliably notified once per reference path when the same object appears under multiple paths; notification selection and ordering have been refined.

* **Tests**
  * Added tests verifying subscription callbacks fire for each referenced path when an object is accessed via multiple root references.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ably/ably-js/pull/2223?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/plugins/liveobjects/liveobject.ts
+++ b/src/plugins/liveobjects/liveobject.ts
@@ -299,40 +299,44 @@ export abstract class LiveObject<
 
   /**
    * Notifies path-based subscriptions about changes to this object.
-   * For LiveMapUpdate events, also creates non-bubbling events for each updated key.
+   * For LiveMapUpdate events, each updated key also contributes a candidate
+   * path one segment deeper than this object's own path.
    */
   private _notifyPathSubscriptions(update: TUpdate): void {
-    const paths = this.getFullPaths();
+    const pathsToThis = this.getFullPaths();
 
-    if (paths.length === 0) {
+    if (pathsToThis.length === 0) {
       // No paths to this object, skip notification
       return;
     }
 
     // Do not expose object sync messages as they do not represent a single operation on an object
     const operationObjectMessage = update.objectMessage?.isOperationMessage() ? update.objectMessage : undefined;
-    const pathEvents: PathEvent[] = paths.map((path) => ({
-      path,
-      message: operationObjectMessage,
-      bubbles: true,
-    }));
 
-    // For LiveMapUpdate, also create non-bubbling events for each updated key
-    if (update._type === 'LiveMapUpdate') {
-      const updatedKeys = Object.keys(update.update);
+    // Call notifyPathEvent() once for each path-to-this. Since
+    // notifyPathEvent() emits at most one event on each subscription, this
+    // means that we emit at most one event per path-to-this.
+    for (const pathToThis of pathsToThis) {
+      const priorityOrderedCandidatePaths: Path[] = [pathToThis];
 
-      for (const key of updatedKeys) {
-        for (const basePath of paths) {
-          pathEvents.push({
-            path: [...basePath, key],
-            message: operationObjectMessage,
-            bubbles: false,
-          });
+      // For LiveMapUpdate, also add a candidate path per updated key. We insert these after
+      // pathToThis so that notifyPathEvent() picks pathToThis in the case where a given subscription
+      // covers multiple candidate paths (that is, we favour the shorter path).
+      if (update._type === 'LiveMapUpdate') {
+        const updatedKeys = Object.keys(update.update);
+
+        for (const key of updatedKeys) {
+          priorityOrderedCandidatePaths.push([...pathToThis, key]);
         }
       }
-    }
 
-    this._realtimeObject.getPathObjectSubscriptionRegister().notifyPathEvents(pathEvents);
+      const pathEvent: PathEvent = {
+        priorityOrderedCandidatePaths,
+        message: operationObjectMessage,
+      };
+
+      this._realtimeObject.getPathObjectSubscriptionRegister().notifyPathEvent(pathEvent);
+    }
   }
 
   private _isNoopUpdate(update: TUpdate | LiveObjectUpdateNoop): update is LiveObjectUpdateNoop {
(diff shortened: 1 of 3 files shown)


src/plugins/liveobjects/pathobjectsubscriptionregister.ts
```diff
@@ -22,12 +22,16 @@ export interface SubscriptionEntry {
  * Event data that LiveObjects provide when notifying of changes
  */
 export interface PathEvent {
-  /** The path where the event occurred */
-  path: Path;
+  /**
+   * Candidate paths for surfacing this event to subscriptions. For a given
+   * subscription, the first candidate path it covers is used as the path of
+   * `event.object` passed to its listener. The caller is responsible for
+   * ordering this array so that the preferred path comes first when more
```
lawrence-forooghian: I also considered just making the rule be shortest-path-wins, which would achieve the same thing but perhaps not be as clear

src/plugins/liveobjects/liveobject.ts
```diff
@@ -299,40 +299,41 @@ export abstract class LiveObject<
...
-
-      for (const key of updatedKeys) {
-        for (const basePath of paths) {
-          pathEvents.push({
-            path: [...basePath, key],
-            message: operationObjectMessage,
-            bubbles: false,
-          });
+
+    // For each subscription, emit at most one event per path-to-this
```
VeskeR: don't think this comment belongs in _notifyPathSubscriptions, it's reads to me more like something PathObjectSubscriptionRegister manages  - ensuring that for each subscritpion at most one event per path is emitted

src/plugins/liveobjects/pathobjectsubscriptionregister.ts
```diff
@@ -22,12 +22,16 @@ export interface SubscriptionEntry {
...
+   * than one is covered.
+   */
+  candidatePaths: Path[];
```
VeskeR: make it `orderedCandidatePaths` so it's a bit clearer to the caller

test/realtime/liveobjects.test.js
```diff
@@ -5590,6 +5590,64 @@ define(['ably', 'shared_helper', 'chai', 'liveobjects', 'liveobjects_helper'], f
           },
         },
 
+        {
+          description:
+            'PathObject.subscribe() fires once per path when the same object is referenced from multiple paths',
```
VeskeR, line 5595: > I don't know whether the current behaviour is intentional or not, but it
seemed worth writing a test for since Claude told me I'd introduced a
regression of this behaviour when I was trying to do a refactor.

Intentional, if an object is observable from multiple paths then each path should be notified of a change.
It is correct that current Realtime API doesn't allow to end with such a state, but it is possible via REST API atm, and we also discussed before that we may have such a functionality for Realtime - you would pass an `Instance` to collection set methods (`map.set`, `list.add` etc) instead of a ValueType.
Thanks for closing the gap

src/plugins/liveobjects/pathobjectsubscriptionregister.ts
VeskeR: I think letting the caller decide is fine, and we can just write it out explicitly in the caller (_notifyPathSubscriptions) that the intention is to have shortest-path-win semantics

VeskeR (approved):
> From what I can tell, it exists to make sure that if there's a map at path "myMap", and this map emits a LiveMapUpdate having key "myKey", then a subscription that covers the path "myMap" will only receive one event (for "myMap"), as opposed to two (for "myMap" and "myMap.myKey").
If that is the only reason that this mechanism exists

Correct. I like the new approach, no need to have different path notification resolutions based on bubbling, and no concept of bubbling at all. Update just now says "here are all affected paths that subscribers might be interested in, in priority order", and subscribers react to that.

LGTM, with some minor nitpicks below. Also linter has failed

test/realtime/liveobjects.test.js
lawrence-forooghian: Thanks, I've updated the commit message

src/plugins/liveobjects/liveobject.ts
lawrence-forooghian: Yes but `PathObjectSubscriptionRegister` isn't aware that it's being invoked once for each path-to-this; all that it does is ensure that each subscription emits at most one event per invocation. It's `_notifyPathSubscriptions` which is responsible for doing one invocation per path-to-this

lawrence-forooghian: Will try making it clearer though

lawrence-forooghian:
> Also linter has failed

fixed in #2225

src/plugins/liveobjects/liveobject.ts
lawrence-forooghian: Tweaked the message

src/plugins/liveobjects/pathobjectsubscriptionregister.ts
lawrence-forooghian: Called it `priorityOrderedCandidatePaths`

src/plugins/liveobjects/pathobjectsubscriptionregister.ts
lawrence-forooghian: Have added a comment alluding to this

PR #2228, opened 2026-05-22
ci: harden workflows

emptyhammond:
Four independent commits addressing findings on `.github/workflows/`:

  - **template-injection in `publish-cdn.yml`** — the `workflow_dispatch` `version` input was interpolated directly into a `run:` shell command. Now passed via a step `env:` var so the shell handles it as a
  variable instead of splicing it into the command string.
  - **secrets-inherit in `features.yml`** — replaced `secrets: inherit` with an explicit passthrough of `ABLY_AWS_ACCOUNT_ID_SDK`, the only secret the called reusable workflow
  (`ably/features/.github/workflows/sdk-features.yml`) declares.
  - **cache-poisoning in `docs.yml`** — split the single job into a `build` job (uses `setup-node`, uploads typedoc as an artifact) and a `publish` job (downloads the artifact, configures AWS, runs
  `sdk-upload-action`). The publishing job no longer uses any cache-restoring setup action.
  - **SHA-pin every action** — every external action across all 10 workflows, plus the `ably/features` reusable-workflow `@main` ref, pinned to its current 40-char commit SHA, with the version preserved in a
  trailing `# vN` comment.

## Summary by CodeRabbit

* **Chores**
  * Updated build and deployment pipelines to use pinned versions of build tools for improved stability and security.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ably/ably-js/pull/2228?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/bundle-report.yml
+++ b/.github/workflows/bundle-report.yml
@@ -12,13 +12,13 @@ jobs:
       deployments: write
       id-token: write
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@ee0669bd1cc54295c223e0bb666b733df41de1c5 # v2
       - name: Reconfigure git to use HTTP authentication
         run: >
           git config --global url."https://github.com/".insteadOf
           ssh://git@github.com/
       - name: Use Node.js 20.x
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@f1f314fca9dfce2769ece7d933488f076716723e # v1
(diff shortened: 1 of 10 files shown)


owenpearson (approved):
Looks good! I think the coderabbit comments about `persist-credentials` seem valid, so it's worth doing that separately and incorporating into the other PRs

PR #2229, opened 2026-05-22
ci: don't persist GITHUB_TOKEN in checkouts

emptyhammond:
**`persist-credentials: false`** on every `actions/checkout` block. By default the action writes the workflow `GITHUB_TOKEN` into `.git/config`, where any later step can read it.
  Nothing in this repo's workflows needs that — no step does `git push` / `tag` / `commit`, the submodule URL is HTTPS-public, and the two `git+ssh://git@github.com/ably-forks/…` entries
  in `package-lock.json` resolve to public repos via the existing URL rewrite, so `npm ci` clones them anonymously. Opt out so the token is wiped from disk once checkout completes.

 **Explicit `permissions:` blocks** on every workflow / reusable-workflow caller that didn't already have one. Most are `contents: read` set at the workflow level (so jobs inherit a
  restrictive default). The docs publish job keeps `deployments: write` + `id-token: write` for the AWS-OIDC role + GitHub Deployment that powers docs previews, and gains an explicit
  `contents: read` so the override doesn't drop checkout access. The features caller needs `id-token: write` so the reusable workflow's `configure-aws-credentials` step can assume its
  role.

## Summary by CodeRabbit

* **Chores**
  * Disabled credential persistence in checkout steps across CI workflows to reduce credential exposure.
  * Added explicit top-level workflow permissions (e.g., contents: read) for several pipelines.
  * Declared additional permission (id-token: write) for the features pipeline to support token-based operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ably/ably-js/pull/2229?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/bundle-report.yml
+++ b/.github/workflows/bundle-report.yml
@@ -13,6 +13,8 @@ jobs:
       id-token: write
     steps:
       - uses: actions/checkout@ee0669bd1cc54295c223e0bb666b733df41de1c5 # v2
+        with:
+          persist-credentials: false
       - name: Reconfigure git to use HTTP authentication
         run: >
           git config --global url."https://github.com/".insteadOf
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -5,6 +5,9 @@ on:
     branches:
       - main
 
+permissions:
+  contents: read
+
 jobs:
   lint:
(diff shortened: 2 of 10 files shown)


owenpearson (approved):
looks good, ty 👍

PR #2233, opened 2026-05-26
DX-1209: inline remediation guidance on SDK ErrorInfo throw sites

umair-ably:
## Intent

> Note: the field was called `hint` in earlier rounds of this PR and in the review comments below; it is now named `remediation`.

This PR extends the message-vs-remediation split DX-1204 introduced (`err.message` says *what* failed; `err.remediation` says *how* to fix it) to ~80 client-side `ErrorInfo` throw sites in the SDK. Where applicable, remediations also forecast the **next** likely failure — typically a server-side rejection driven by channel-namespace / capability config the agent can't see from inside the SDK alone.

## Why

LLMs (and humans) recover faster from SDK errors when the actionable next step is named in the error itself, rather than left to be inferred or searched. We saw two recurring failure modes when agents hit SDK errors:

1. **Spiralling** — agent keeps mutating code to chase an error whose root cause is outside the code (dashboard config, capability gap, wrong execution context). Without a remediation that surfaces the *next* wall, agents fix the client-side check, then hit the server-side rejection, then loop, often burning upwards of 40 tool calls before quitting. Quite often, an agent would also choose a non-optimal solution just so the code "works". 
2. **Quitting** — agent recognises the wall but has no path forward, gives up, or papers over the failure with empty catches.

DX-1209's remediations are designed to convert both behaviours into "investigate, then commit to the idiomatic fix" — naming the second wall up front (so the agent doesn't spiral into it), and pointing at concrete diagnostics or replacement APIs (so the agent doesn't quit).

## Outcome and Evidence

Same emulated-constraint LLM testing as DX-1204: subagents fix a failing Node test, no `WebFetch` / `WebSearch`, SDK as a black box visible only through runtime errors. Four experiments below summarise ~120 piloted runs across Opus 4.7, Sonnet 4.6, Haiku 4.5 × pre/post × n=5.

### How we scored fixes

The interesting question wasn't just "did the test go green?" but "*how* did the agent make it green?" Every passing run was bucketed by reading the final `test.js`:

- **IDIOMATIC** — the agent fixed the problem the way the SDK was designed to be used. For the annotations test that means `channels.get(name, { modes: ['subscribe', 'annotation_subscribe'] })` plus a channel name in a namespace where Mutable Messages is enabled. The code looks like documentation.
- **HALF** — the agent got the API call right (e.g. correct `modes`) but wrapped the result in a `try/catch` they don't appear to understand, or paired the fix with an unrelated workaround. Test passes; intent is partial.
- **HACKY** — the agent papered over the symptom rather than fixing the cause: empty `.catch(() => {})` on a real failure, `channel.modes.push('annotation_subscribe')` after attach to satisfy a snapshot assertion, swapping `push.activate()` for a private monkey-patch of `Platform.Config.push`. Test passes; the code that lands in the repo doesn't reflect how the SDK is meant to be used.

### Experiment 1: annotation_subscribe remediation with second-wall forecast

`channel.annotations.subscribe()` without `annotation_subscribe` in `ChannelOptions.modes` throws 93001. Pre-PR the message named the symptom only; post-PR the remediation names the modes API call shape **and** forecasts that the channel namespace must have Mutable Messages enabled in the Ably dashboard.

The idiomatic fix is two steps: add `annotation_subscribe` to `ChannelOptions.modes`, and use a channel name in a namespace that has Mutable Messages enabled. Pre-PR agents tended to get the first step and miss the second, then either swallow the server-side rejection (HACKY) or fence-sit between the right modes call and a defensive try/catch (HALF). Post-PR, the remediation primes them to expect the second wall, so they go looking for the right namespace immediately.

- With a snapshot-only assertion (test checks that `channel.modes` after attach contains `annotation_subscribe`), **IDIOMATIC rate among passing runs: pre 67% → post 83%**. The HALF bucket collapses (3 → 0): no more "I added the modes and threw a catch around it just in case." Sonnet specifically shifts from HACKY (no-await + empty `.catch`) to IDIOMATIC (probes namespaces, finds `test:`, commits to the fix).
- With end-to-end verification (test publishes a real annotation via REST and asserts the realtime listener receives it), every passing run is IDIOMATIC on both sides — the harder test rejects every hacky path, so by definition only idiomatic fixes pass. The metric that moves instead is **quit rate**: Haiku abandons 4/5 pre vs 1/5 post.

The annotation remediation also includes a short anti-hack line — *"Note: appending to channel.modes after attach() does not enable the mode server-side."* That sentence exists because one Haiku run literally did `channel.modes.push('annotation_subscribe')` to satisfy a snapshot assertion. The remediation now names that gaming pattern explicitly so future agents (and humans) don't repeat it. Same line was added to the LiveObjects remediation for the same reason. This provides some evidence going forwards that in a similar vain to prompting LLMs, our remediations can and should also include advice on what _not_ to do. 

### Experiment 2: pointing at the Ably CLI from inside the remediation

The dashboard line in the annotations remediation surfaces a real wall ("this is a config issue, not a code issue") but doesn't tell an agent how to *see* the config. From inside a Node process the agent has no view of which channel namespaces actually have Mutable Messages enabled; without that visibility, weaker models quit, capable models burn tool calls probing namespace prefixes by trial and error.

Adding a final sentence — *"If you have the Ably CLI installed, `ably apps rules list` will show which channel namespaces have Mutable Messages enabled"* — bridges that gap. The CLI is the natural diagnostic surface for this class of question (it speaks to the same backend the SDK does, with the same credentials), and agents are willing to invoke it when the remediation tells them it exists. The conditional phrasing matters:

- **CLI installed, conditional remediation**: Haiku abandons **7/10 → 0/10**; mean tool_uses on passes **27.5 → 11.6** (-58%). Every Haiku ran `ably apps rules list`, found a working namespace, and moved on. No more namespace-probing spirals.
- **CLI absent, conditional remediation**: 5/5 Opus pass. Agents see *"if you have the CLI installed"*, register that they don't, and fall back to namespace probing without burning time trying to invoke a missing binary. 1 Sonnet run actually installed the CLI using npx and then inspected the namespaces. 
- **CLI absent, imperative remediation** (an earlier experiment with *"Use `ably apps rules list`"*): **3/5 Opus runs timed out** trying to invoke the missing CLI. The imperative phrasing turned "the CLI helps if you have it" into "the CLI is required" — and capable agents committed hardest to that wrong path.

The conditional form is therefore shipped: it's strictly better when the CLI is absent, equivalent when present, and tells production users that the CLI is the right diagnostic for any error rooted in dashboard or capability config.

NOTE: Just this experiment was a strong signal that the CLI paired with these remediations was a powerful combination in reducing the number of tool calls whilst reaching an idiomatic solution quicker. I will investigate how we can surface installing the CLI at a sooner, and more appropriate stage e.g. perhaps we log CLI install instructions when the SDK is first initialised. 

### Experiment 3: LiveObjects portability

I originally didn't apply a remediation to `object_subscribe` mode missing (40024) because the error message alone seemed descriptive enough, however, there's evidence that shows we can still help weaker models here.

Idiomatic fix mirrors annotations: `channels.get(name, { modes: [..., 'object_subscribe'] })`. The test channel name happens to work in the default namespace, so there's no second-wall to navigate here; the remediation's job is just to surface the modes API.

- Opus / Sonnet: no measurable delta (5/5 pass either side, ~5 tool calls — the existing error message names the missing mode plainly enough that capable agents fix it from training data alone).
- Haiku: pre **3/5 pass + 2 abandons → post 5/5 pass**. Same idiomatic fix every time; the remediation just makes it discoverable at the weakest tier. The second-wall forecast pattern transfers cleanly across SDK surfaces and eliminates weakest-tier quit rate without regressing capable models.

### Experiment 4: telling agents to stop fighting the SDK when the API is structurally wrong for the context

Some errors signal *"you're using the wrong API entirely, not a fixable mistake"* — e.g. `push.activate()` on a Node server, where there is no device to register. The idiomatic fix is to recognise that `push.activate()` is the wrong API in this context and pivot to `client.push.admin.publish(...)` / `deviceRegistrations.save(...)`, which are designed for server-side push management. The HACKY alternative the eval kept seeing was monkey-patching `Platform.Config.push` with a hand-rolled storage object so the SDK's platform check would pass — an entirely different code path that simulates being a browser inside Node.

The original remediation named the alternative (*"Use Push admin"*) but left the door open to retry; Opus runs spent **22–30 tool calls** building those monkey-patches. The remediation was true but too soft, so agents read it as advice instead of an instruction to stop.

Rephrasing to lead with the impossibility — *"push.activate() registers this process as a push target — it cannot succeed in Node.js/server contexts. Use client.push.admin.publish(...) / deviceRegistrations.save(...) instead."* — dropped Opus tool counts to **mean 7.3 (range 6–9)** across 3 confirmation runs, with **zero monkey-patches**. The transcripts shift from "puzzle to solve" to "cite the remediation, accept the SDK is right, wrap the call narrowly and move on." Same pattern applied to all 6 push platform-not-supported throw sites (2 in `push.ts`, 4 in `pushactivation.ts`).

### Where the remediations do *not* help

Two scenarios where the eval showed no measurable delta — included for honest scoping:

- **Server-relayed `40160 / "Channel access denied"`** (constrained-key presence test): the server's own error already names the failure clearly enough that capable agents decode it without needing the client-side remediation. Adding the capability forecast didn't move tool counts.
- **Token revocation under token auth (40162)**: the SDK message itself names the fix verbatim (*"Cannot revoke tokens when using token auth"* + the remediation already in place names "construct a separate Ably.Rest client with ClientOptions.key"). All 12 piloted runs converged in 4–6 tool calls regardless of pre/post.

These are useful as the negative space around DX-1209's value: remediations help most where the message is symptomatic-only, the fix is non-obvious, or the next failure is hidden behind a wall the agent can't see from inside the process. Where the message already encodes the fix, the remediation is a no-op.

---

The cross-cutting story across all four experiments: DX-1209's remediations **shorten the loop between hitting an SDK error and committing to the idiomatic fix**. Where agents previously spiralled (mutating code against a wall that needs a dashboard change) or quit (no path forward visible from inside the process), the remediations now name the next wall up front, point at the diagnostic tool that can see it, and — when the call is structurally wrong for the context — tell the agent to pivot rather than keep fighting. Tool-count averages move modestly; abandon rates at the weakest tier collapse; and the code that lands in your repo is materially more likely to follow the SDK's intended design.

## Summary

- Adopts the **`message` (what failed) / `remediation` (how to fix)** split that DX-1204 introduced for `detectV1Callback` and applies it to every `ErrorInfo` throw in the SDK where a remediation adds value. The split lets agents (and humans) read the *what* in `err.message` and act on the *how* in `err.remediation` without parsing prose.
- Covers ~80 inline throw sites across 19 files — the high-traffic codes LLM agents and migrating users will trip on first: channel-state / publish-shape / presence-clientId / auth-options / push-state / message-serial / capability / pagination / plugin-missing / derived-channel / vcdiff / endpoint-vs-host.
- **Server-relayed codes are deliberately out of scope here** (40140 token expired, 40160 capability, 40300 forbidden, 42910 rate-limit, etc.). Propagating remediations from the server is being tracked separately.

## Themes / where this sits

This is the foundation slice of `LLMEvalUpdatedPlan.md` A1 (TKT-6) and aligns with the team's "Theme 3 — errors guide self-healing" target for the quarter. It's defence-in-depth alongside DX-1204 (v1-callback runtime throws with a remediation) and DX-1205 (legacy-import shims with a remediation).

## What's in scope vs deliberately not

**Remediation added** (~80 sites):
- `realtimechannel.ts` — channel options validation, publish shape (40013), max size (40009), detach-while-failed, `invalidStateError` (90001), release-state, attach/detach timeout (90007), untilAttach, `sendUpdate` missing serial (40003), `sync()` not-attached.
- `realtimepresence.ts` — clientId missing for enter/update/leave (40012), leave-in-incompatible-state (90001), `get()` on suspended (91005), untilAttach, auto-re-enter failure (91004).
- `auth.ts` — no auth options (40106), incompatible key (40102 ×2), authUrl/authCallback errors (40170 ×9), no-renewable-token (40171), no/invalid key (40101 ×2), empty/wildcard/non-string clientId (40012 ×3), token clientId mismatch (RSA15c), token-shape rejection.
- `baseclient.ts`, `defaults.ts`, `rest.ts`, `baserealtime.ts`, `paginatedresource.ts`, `utils.ts`, `connectionmanager.ts`, `basemessage.ts` — see commit message for the per-file rundown.
- `push.ts`, `pushactivation.ts`, `pushchannel.ts`, `getW3CDeviceDetails.ts` — push platform/state, registration callback, device-identity preconditions.
- `restchannel.ts`, `restchannelmixin.ts`, `restannotations.ts`, `realtimeannotations.ts` — message-serial requirements, annotation shape / mode.

**Skipped** (per A1 scope):
- Transport `80xxx` connection-lifecycle codes — auto-recovered by the SDK; a remediation would encourage user retry code that fights the SDK.
- `50000` internal errors — only actionable advice is *file an issue*, not worth a per-site remediation.
- **Server-relayed codes** (`40140` family, `40160`, `40300`, `42910`, etc.) — the SDK does not construct these, the server does. Propagating server-supplied remediations is separate ongoing work.

## What's next

As mentioned in the above section; server-supplied remediations is ongoing work. Once this is ready, we can revisit these current remediations to see if we still need to forecast the next error, or if the combination of client error + server error works just as well. 

I'd also like to surface the CLI install steps in the SDK initialisation journey.

## Release Notes

* **New Features**
  * Enhanced error messages across the SDK with actionable hints and remediation guidance for authentication, channels, presence, push notifications, and other operations.

* **Chores**
  * Increased minimal bundle size thresholds and introduced automated validation to ensure error hints remain consistent and helpful across updates.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ably/ably-js/pull/2233?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


test/unit/error-hints.test.js
```diff
@@ -0,0 +1,77 @@
+'use strict';
+
+/*
+ * DX-1209 — hint pinning tests for inline SDK throw sites.
```
umair-ably: I'm not bedded to the hint-coverage script and/or this test file but we need something to test the presence of hints... I'm leaning more towards the script as opposed to exhaustively testing all the hints here (although these few as a smokescreen don't hurt). Happy to hear your thoughts here @AndyTWF

src/plugins/push/getW3CDeviceDetails.ts
```diff
@@ -28,17 +28,19 @@ export async function getW3CPushDeviceDetails(machine: ActivationStateMachine) {
...
 
   const swUrl = machine.client.options.pushServiceWorkerUrl;
   if (!swUrl) {
-    machine.handleEvent(
-      new GettingPushDeviceDetailsFailed(new ErrorInfo('Missing ClientOptions.pushServiceWorkerUrl', 400, 40000)),
-    );
+    const err = new ErrorInfo('Missing ClientOptions.pushServiceWorkerUrl', 400, 40000);
+    err.hint =
+      'Set ClientOptions.pushServiceWorkerUrl to the path of your service worker (e.g. "/ably-push-sw.js") so the SDK can register it for web push.';
+    machine.handleEvent(new GettingPushDeviceDetailsFailed(err));
```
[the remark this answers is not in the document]
umair-ably: good spot... this was preexisting but will fix

src/common/lib/client/auth.ts
```diff
@@ -269,7 +272,12 @@ class Auth {
     /* RSA10a: authorize() call implies token auth. If a key is passed it, we
      * just check if it doesn't clash and assume we're generating a token from it */
     if (authOptions && authOptions.key && this.authOptions.key !== authOptions.key) {
-      throw new ErrorInfo('Unable to update auth options with incompatible key', 40102, 401);
+      {
+        const err = new ErrorInfo('Unable to update auth options with incompatible key', 40102, 401);
+        err.hint =
+          'auth.authorize() cannot change the API key — the new authOptions.key differs from the one the client was constructed with. To use a different key, construct a new Ably client.';
```
AndyTWF: Regular hyphens over em-dashes

src/common/lib/client/auth.ts
```diff
@@ -486,27 +494,33 @@ class Auth {
           }
           if (Platform.BufferUtils.isBuffer(body)) body = body.toString();
           if (!contentType) {
-            cb(new ErrorInfo('authUrl response is missing a content-type header', 40170, 401), null);
+            const err = new ErrorInfo('authUrl response is missing a content-type header', 40170, 401);
+            err.hint =
+              'Have your auth endpoint return a Content-Type of application/json (TokenDetails/TokenRequest), text/plain (token string) or application/jwt.';
```
AndyTWF: Imo this sounds too imperative and more akin with docs than actual error, perhaps a better form would be (and ditto styling elsewhere):

`Auth endpoints may return a Content-Type of application/json (TokenDetails/TokenRequest), text/plain (token string) or application/jwt.`

src/common/lib/client/auth.ts
```diff
@@ -486,27 +494,33 @@ class Auth {
...
-                  contentType +
-                  ', should be either text/plain, application/jwt or application/json',
-                40170,
-                401,
-              ),
-              null,
+            const err = new ErrorInfo(
+              'authUrl responded with unacceptable content-type ' +
+                contentType +
+                ', should be either text/plain, application/jwt or application/json',
```
AndyTWF: should? or must?

"either" usually implies a binary choice, so the more natural message would be `authUrl responded with unacceptable Content-Type. Expected one of: text/plain, application/jwt or application/json`

src/common/lib/client/auth.ts
```diff
@@ -486,27 +494,33 @@ class Auth {
...
+              40170,
+              401,
             );
+            err.hint =
+              'Update your auth endpoint to return Content-Type application/json, text/plain or application/jwt — the SDK cannot parse other content types.';
```
AndyTWF: Ditto my previous comment on imperative and em-dashes

src/common/lib/client/auth.ts
```diff
@@ -486,27 +494,33 @@ class Auth {
...
+            cb(err, null);
             return;
           }
           if (json) {
             if ((body as string).length > MAX_TOKEN_LENGTH) {
-              cb(new ErrorInfo('authUrl response exceeded max permitted length', 40170, 401), null);
+              const err = new ErrorInfo('authUrl response exceeded max permitted length', 40170, 401);
+              err.hint =
+                'authUrl payloads must be under 128 KB. Your endpoint is returning more than a TokenDetails/TokenRequest object — return only the token shape, not wrapping JSON or extra fields.';
```
AndyTWF: Is it true that they _are_ returning something other than TokenDetails? Could the TokenDetails just be large (e.g. a big capabilities block)?

src/common/lib/client/auth.ts
```diff
@@ -628,7 +645,10 @@ class Auth {
           tokenRequestCallbackTimeoutExpired = true;
           const msg = 'Token request callback timed out after ' + timeoutLength / 1000 + ' seconds';
           Logger.logAction(this.logger, Logger.LOG_ERROR, 'Auth.requestToken()', msg);
-          reject(new ErrorInfo(msg, 40170, 401));
+          const err = new ErrorInfo(msg, 40170, 401);
+          err.hint =
+            'Your authCallback/authUrl did not respond in time. Make sure the callback invokes its callback parameter (or resolves its promise) on every code path.';
```
AndyTWF: "Make sure the callback" is case-specific though, doesn't apply to authUrl?

AFAIK you can't have a promise in authCallback, so this part doesn't apply?

src/common/lib/client/auth.ts
```diff
@@ -648,29 +668,37 @@ class Auth {
...
-                40170,
-                401,
-              ),
+            const err = new ErrorInfo(
+              'Token string exceeded max permitted length (was ' + tokenRequestOrDetails.length + ' bytes)',
+              40170,
+              401,
             );
+            err.hint =
+              'Tokens must be under 128 KB. Your endpoint is returning more than the token itself — return only the token string or a TokenDetails object.';
```
AndyTWF: Per previous - could be large capabilities.

src/common/lib/client/auth.ts
```diff
@@ -648,29 +668,37 @@ class Auth {
...
             !(contentType && contentType.indexOf('application/jwt') > -1)
           ) {
-            reject(
-              new ErrorInfo(
-                "Token was double-encoded; make sure you're not JSON-encoding an already encoded token request or details",
-                40170,
-                401,
-              ),
+            const err = new ErrorInfo(
+              "Token was double-encoded; make sure you're not JSON-encoding an already encoded token request or details",
```
AndyTWF: Is the hint here beneficial, as it's basically repeating the error message?

src/common/lib/client/auth.ts
```diff
@@ -775,18 +810,26 @@ class Auth {
...
+      const err = new ErrorInfo('Invalid key specified', 40101, 403);
+      err.hint =
+        'API keys are "appId.keyId:secret". Copy the full key including the colon from the Ably dashboard. If you have the Ably CLI installed, `ably auth keys list` shows the keys configured on the current app.';
+      throw err;
     }
 
     if (tokenParams.clientId === '') {
-      throw new ErrorInfo('clientId can’t be an empty string', 40012, 400);
+      const err = new ErrorInfo('clientId can’t be an empty string', 40012, 400);
+      err.hint = 'Pass a non-empty clientId, or omit the field entirely to mint an anonymous token.';
```
AndyTWF: I don't think we'd use the word "mint" in our docs

src/common/lib/client/baserealtime.ts
```diff
@@ -199,11 +199,14 @@ class Channels extends EventEmitter {
...
     } else if (channelOptions) {
       if (channel._shouldReattachToSetOptions(channelOptions, channel.channelOptions)) {
-        throw new ErrorInfo(
+        const err = new ErrorInfo(
           'Channels.get() cannot be used to set channel options that would cause the channel to reattach. Please, use RealtimeChannel.setOptions() instead.',
           40000,
           400,
         );
+        err.hint =
+          'channels.get(name) returns the existing channel — call channel.setOptions(opts) to change params/modes (this triggers a re-attach if modes or params changed). Only the first channels.get() for a name applies options without a re-attach.';
```
AndyTWF: Changing options via `channels.get()` is deprecated anyway, so I don't think we need to point people to using `get` for non-attaching options. See RTS3c

src/common/lib/client/push.ts
```diff
@@ -37,11 +37,19 @@ class Push {
         return;
       }
       if (!this.stateMachine) {
-        reject(new ErrorInfo('This platform is not supported as a target of push notifications', 40000, 400));
+        {
+          const err = new ErrorInfo('This platform is not supported as a target of push notifications', 40000, 400);
+          err.hint =
+            'push.activate() registers this process as a push target — it cannot succeed in Node.js/server contexts (there is no device to register). Use client.push.admin to manage other devices from a server: client.push.admin.publish(recipient, payload) to send to a device or clientId, client.push.admin.deviceRegistrations.save(device) to register a device record.';
```
AndyTWF: Would a better hint just to be to say what platforms DO support push?

src/common/lib/client/realtimeannotations.ts
```diff
@@ -71,11 +71,14 @@ class RealtimeAnnotations {
...
     if ((this.channel.state === 'attached' && this.channel._mode & flags.ANNOTATION_SUBSCRIBE) === 0) {
-      throw new ErrorInfo(
+      const err = new ErrorInfo(
         "You are trying to add an annotation listener, but you haven't requested the annotation_subscribe channel mode in ChannelOptions, so this won't do anything (we only deliver annotations to clients who have explicitly requested them)",
         93001,
         400,
       );
+      err.hint =
+        'Re-create the channel with annotation_subscribe in modes: realtime.channels.get(name, { modes: ["subscribe", "annotation_subscribe", ...] }). If the subsequent attach is rejected by the server, the channel namespace must have "Message annotations, updates, and deletes" enabled in the Ably dashboard, and your API key must have annotation-subscribe capability on this channel. If you have the Ably CLI installed, `ably apps rules list` shows which channel namespaces have Mutable Messages enabled, and `ably auth keys list` shows your key\'s capabilities. Note: appending to channel.modes after attach() does not enable the mode server-side — the array reflects what the server granted, not what you requested.';
```
AndyTWF: `If the subsequent attach is rejected by the server, the channel namespace must have` - this is perhaps a bit ambiguous. Is this saying "you need to go and do this" or is it saying "it must have" implying it's the cause

src/common/lib/client/realtimechannel.ts
```diff
@@ -173,12 +179,15 @@ class RealtimeChannel extends EventEmitter {
...
   invalidStateError(): ErrorInfo {
-    return new ErrorInfo(
+    const err = new ErrorInfo(
       'Channel operation failed as channel state is ' + this.state,
       90001,
       400,
       this.errorReason || undefined,
     );
+    err.hint =
+      'Inspect channel.errorReason for the underlying cause, then call channel.attach() to recover. From "failed" or "suspended", a fresh attach() is required before further channel operations.';
```
AndyTWF: Not true on `suspended`, the connection and thus attachment will eventually recover

src/common/lib/client/realtimechannel.ts
```diff
@@ -271,23 +280,29 @@ class RealtimeChannel extends EventEmitter {
...
       params = args[1];
     } else {
-      throw new ErrorInfo(
+      const err = new ErrorInfo(
         'The single-argument form of publish() expects a message object or an array of message objects',
         40013,
         400,
       );
+      err.hint =
+        'Call publish(name, data) for a single event, or publish(message | message[]) with a Message-shaped object. If the resulting publish is rejected by the server, your token/API-key capability must include "publish" on this channel. If you have the Ably CLI installed, `ably auth keys list` shows your key\'s capabilities.';
```
AndyTWF: There are all sorts of reasons why a publish may be rejected, not just capabilities.

src/common/lib/client/realtimechannel.ts
```diff
@@ -504,8 +523,12 @@ class RealtimeChannel extends EventEmitter {
     switch (this.state) {
       case 'initialized':
       case 'detaching':
-      case 'detached':
-        throw new PartialErrorInfo('Unable to sync to channel; not attached', 40000);
+      case 'detached': {
+        const err = new PartialErrorInfo('Unable to sync to channel; not attached', 40000);
+        err.hint =
```
AndyTWF: Sync is an internal method, there's nothing that a user or LLM can do to fix it.

src/common/lib/client/realtimepresence.ts
```diff
@@ -73,7 +76,10 @@ class RealtimePresence extends EventEmitter {
 
   private async _updateImpl(data: unknown): Promise<void> {
     if (isAnonymousOrWildcard(this)) {
-      throw new ErrorInfo('clientId must be specified to update presence data', 40012, 400);
+      const err = new ErrorInfo('clientId must be specified to update presence data', 40012, 400);
+      err.hint =
+        'Set ClientOptions.clientId (or include clientId in the token) before calling presence.update(). To update on behalf of another identity, use presence.updateClient(clientId, data). If the resulting presence message is rejected by the server, your token/API-key capability must include "presence" on this channel. If you have the Ably CLI installed, `ably auth keys list` shows your key\'s capabilities.';
```
AndyTWF: Or, you don't have a wildcard clientId from key or via token. You'll get rejected by the server if your token has a clientId

src/common/lib/client/realtimepresence.ts
```diff
@@ -206,11 +218,14 @@ class RealtimePresence extends EventEmitter {
     /* Special-case the suspended state: can still get (stale) presence set if waitForSync is false */
     if (this.channel.state === 'suspended') {
       if (waitForSync) {
-        throw ErrorInfo.fromValues({
+        const err = ErrorInfo.fromValues({
```
AndyTWF: General observation... rather than having `const err = ...; err.hint =`, should we be updating ErrorInfo.fromValues to handle the `hint` field?

src/common/lib/client/restchannel.ts
```diff
@@ -106,11 +106,14 @@ class RestChannel {
...
       params = args[1];
     } else {
-      throw new ErrorInfo(
+      const err = new ErrorInfo(
         'The single-argument form of publish() expects a message object or an array of message objects',
         40013,
         400,
       );
+      err.hint =
+        'Call publish(name, data) for a single event, or publish(message | message[]) with a Message-shaped object. If the resulting publish is rejected by the server, your token/API-key capability must include "publish" on this channel. If you have the Ably CLI installed, `ably auth keys list` shows your key\'s capabilities.';
```
AndyTWF: Ditto previous about there being more reasons than capabilities for server rejection. If the server is rejecting, should we just be letting that error carry itself rather than over prescribing / predicting the future here?

src/common/lib/transport/connectionmanager.ts
```diff
@@ -1945,11 +1948,17 @@ class ConnectionManager extends EventEmitter {
     } else if (err.statusCode === HttpStatusCodes.Forbidden) {
       const msg = 'Client configured authentication provider returned 403; failing the connection';
       Logger.logAction(this.logger, Logger.LOG_ERROR, 'ConnectionManager.actOnErrorFromAuthorize()', msg);
-      this.notifyState({ state: 'failed', error: new ErrorInfo(msg, 80019, 403, err) });
+      const wrapped = new ErrorInfo(msg, 80019, 403, err);
+      wrapped.hint =
+        'Your authUrl/authCallback returned 403. Fix the auth endpoint: the request reached Ably, but your server refused to mint a token. Inspect cause for the underlying error.';
```
AndyTWF: This hint is too narrow. Per RSA4d, this eventuality could happen because the Ably server returns 403 when trying to turn a TokenRequest into a TokenDetails.

src/common/lib/util/defaults.ts
```diff
@@ -175,10 +175,14 @@ export function getHosts(options: NormalisedClientOptions): string[] {
 
 function checkHost(host: string): void {
   if (typeof host !== 'string') {
-    throw new ErrorInfo('host must be a string; was a ' + typeof host, 40000, 400);
+    const err = new ErrorInfo('host must be a string; was a ' + typeof host, 40000, 400);
+    err.hint = 'Pass `endpoint` as a single hostname string (e.g. "main.realtime.ably.net"), not an array or object.';
```
AndyTWF: The endpoint for `main` would just be `main`

src/common/lib/util/utils.ts
```diff
@@ -470,11 +470,16 @@ export function matchDerivedChannel(name: string) {
   const regex = /^(\[([^?]*)(?:(.*))\])?(.+)$/; // eslint-disable-line
   const match = name.match(regex);
   if (!match || !match.length || match.length < 5) {
-    throw new ErrorInfo('regex match failed', 400, 40010);
+    const err = new ErrorInfo('regex match failed', 400, 40010);
```
AndyTWF: Is this an opportunity to also improve the message, not just the hint?

src/plugins/push/getW3CDeviceDetails.ts
```diff
@@ -28,17 +28,19 @@ export async function getW3CPushDeviceDetails(machine: ActivationStateMachine) {
   const permission = await Notification.requestPermission();
 
   if (permission !== 'granted') {
-    machine.handleEvent(
-      new GettingPushDeviceDetailsFailed(new ErrorInfo('User denied permission to send notifications', 400, 40000)),
-    );
+    const err = new ErrorInfo('User denied permission to send notifications', 400, 40000);
```
AndyTWF: Actual error to fix here from coderabbit

scripts/hint-coverage.ts
```diff
@@ -0,0 +1,278 @@
...
+ *
+ * Rules of thumb when adding entries:
+ *   - Pin SDK API identifiers (`enterClient`, `defaultTokenParams`, ...)
+ *     and Ably product/feature words (`annotation_subscribe`, `Mutable
+ *     Messages`, ...). They're the public-ish surface.
+ *   - Avoid pinning prose ("must", "the resulting"). Wording is allowed
+ *     to evolve.
+ *   - Use `matches` (regex) only when the token genuinely varies.
+ */
+const RUBRIC: Record<number, RubricEntry> = {
```
AndyTWF: Is this likely to cause us development friction, feels brittle?

scripts/hint-coverage.ts
```diff
@@ -0,0 +1,278 @@
...
+  //   message-keyed rules in a follow-up if/when needed.
+};
+
+const SRC_ROOT = path.resolve(__dirname, '..', 'src');
+
+function stripBlockComments(src: string): string {
+  return src.replace(/\/\*[\s\S]*?\*\//g, (m) => m.replace(/[^\n]/g, ' '));
+}
+
+function extractHints(filePath: string): HintEntry[] {
```
AndyTWF: Subsequent to my comment about expanding ErrorInfo.fromValues to take a hint directly and not having an extra `err.hint = ...` line each time, would this script being an AST rather than a regex be more sturdy here?

AndyTWF (commented):
Thanks for this. Generally - we have a fair number of hints that are pretty much re-wordings of the actual error messages. Are hints necessary in this case if they don't add additional information/substance, or are they just noise? Otherwise we could trim down the message and instead lean more heavily on the hint.

src/common/lib/types/errorinfo.ts
```diff
@@ -46,15 +52,43 @@ export default class ErrorInfo extends Error implements IPartialErrorInfo, API.E
...
+    } else {
+      super(messageOrValues);
+      if (typeof Object.setPrototypeOf !== 'undefined') {
+        Object.setPrototypeOf(this, ErrorInfo.prototype);
+      }
+      this.code = code as number;
+      this.statusCode = statusCode as number;
+      this.cause = cause;
+      this.detail = detail;
     }
```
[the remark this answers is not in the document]
umair-ably: This is intentional in that SDK originating errors don't have synthesised href's added... the aim here is to use a constructor without the href for SDK originating errors, and .fromValues for server errors (as is today). 

Does this help clarify the issue and resolve your comment?

src/common/lib/types/basemessage.ts
```diff
@@ -212,14 +217,20 @@ export async function decodeData(
             }
           case 'vcdiff':
             if (!context.plugins || !context.plugins.vcdiff) {
-              throw new ErrorInfo('Missing Vcdiff decoder (https://github.com/ably-forks/vcdiff-decoder)', 40019, 400);
+              throw new ErrorInfo({
+                message: 'Missing Vcdiff decoder (https://github.com/ably-forks/vcdiff-decoder)',
+                code: 40019,
+                statusCode: 400,
+                hint: 'You enabled the delta channel option but did not provide the Vcdiff plugin. Install @ably/vcdiff-decoder and pass it in ClientOptions.plugins.vcdiff.',
+              });
```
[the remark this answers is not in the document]
umair-ably: fixed

src/plugins/push/pushchannel.ts
```diff
@@ -105,7 +115,12 @@ class PushChannel {
     if (deviceIdentityToken) {
       return deviceIdentityToken;
     } else {
-      throw new this.client.ErrorInfo('Cannot subscribe from client without deviceIdentityToken', 50000, 500);
+      throw new this.client.ErrorInfo({
+        message: 'Cannot subscribe from client without deviceIdentityToken',
+        code: 50000,
+        statusCode: 500,
+        hint: 'Activate this device first by awaiting client.push.activate(registerCallback) - the device must hold an identity token before subscribing to push.',
+      });
```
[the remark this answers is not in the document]
umair-ably: sorted

lmars:
@umair-ably I've not reviewed the diff yet, but here's a review just based on the PR description:

> This PR extends the message-vs-hint split DX-1204 introduced (err.message says what failed; err.hint says how to fix it)

It would be useful to explain why we're doing a message-vs-hint split rather than expanding the error messages themselves, since [DX-1204](https://ably.atlassian.net/browse/DX-1204) says nothing about it (it seems to be about re-introducing deprecated callbacks, is it definitely the correct ticket?). I'm not necessarily saying we shouldn't, but we should be deliberate in why this is a new field.

For example, I haven't looked at the diff yet, but let's say this introduces `ErrorInfo{message: "invalid key", hint: "The provided key must have format x.y:z"}`, then why not just update the message to `invalid key: the provided key must have format x.y:z`?

> Experiment 2: pointing at the Ably CLI from inside the hint ... "If you have the Ably CLI installed, `ably apps rules list` will show which channel namespaces have Mutable Messages enabled"

If they have the Ably CLI installed, I would suggest we instead consider a command like `ably diagnose 93001` which internally knows to lookup the namespace and to return an appropriate response, even suggesting the command to run to create the relevant rule? This way the error message in the SDKs stays lightweight (i.e. "if you have the Ably CLI, check `ably diagnose <error code>`") and we centralise the diagnostic checks in the CLI. WDYT?

> Experiment 3: LiveObjects portability

OOI, what exactly was this experiment? As in, what was the prompt?

[DX-1204]: https://ably.atlassian.net/browse/DX-1204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

scripts/hint-coverage.ts
```diff
@@ -0,0 +1,374 @@
...
+ *     and Ably product/feature words (`annotation_subscribe`, `Mutable
+ *     Messages`, ...). They're the public-ish surface.
+ *   - Avoid pinning prose ("must", "the resulting"). Wording is allowed
+ *     to evolve.
+ *   - Use `matches` (regex) only when the token genuinely varies.
+ */
+const RUBRIC: Record<number, RubricEntry> = {
+  40009: {
+    description: 'publish payload exceeds maxMessageSize',
+    require: [{ kind: 'contains', value: 'maxMessageSize' }],
```
lmars: I don't really understand the purpose of this. The code has errors with hints, but we're concerned that a developer updates a hint to make it not as useful as we think it should be, so we're adding separate code to make sure those hints contain specific words? That feels awfully brittle.

scripts/hint-coverage.ts
```diff
@@ -0,0 +1,374 @@
...
+  },
+  40019: {
+    description: 'missing plugin',
+    // 40019 is shared between modular-plugin missing (ably/modular) and the
+    // vcdiff-decoder missing site. Either install path is acceptable.
+    require: [{ kind: 'matches', value: /ably\/modular|@ably\/vcdiff-decoder/ }],
```
lmars: Case in point from above, 40019 has the title "Required client library plugin not present", why does the hint for that _have_ to include "ably/modular" or "@ably/vcdiff-decoder"?

scripts/moduleReport.ts
```diff
@@ -5,8 +5,12 @@ import { promisify } from 'util';
 import { gzip } from 'zlib';
 import Table from 'cli-table';
 
-// The maximum size we allow for a minimal useful Realtime bundle (i.e. one that can subscribe to a channel)
-const minimalUsefulRealtimeBundleSizeThresholdsKiB = { raw: 107, gzip: 33 };
+// The maximum size we allow for a minimal useful Realtime bundle (i.e. one that can subscribe to a channel).
+// Raised from { raw: 107, gzip: 33 } in DX-1209 to absorb the ~80 inline err.hint strings now embedded at SDK throw sites,
+// plus the ErrorInfo options-object constructor overload that lets throw sites pass message/code/statusCode/hint in one
+// call (~10 KiB raw, ~3 KiB gzipped in total). To recalibrate after further hint changes, temporarily delete the hints to
+// measure the hint-free baseline, then add a margin and restore.
```
lmars: This is all noise, if someone wants to know the history (i.e. "raised from 107 to 118 due to $REASONS"), look at the git history?

src/common/lib/client/auth.ts
```diff
@@ -158,7 +158,12 @@ class Auth {
         const msg =
           'No authentication options provided; need one of: key, authUrl, or authCallback (or for testing only, token or tokenDetails)';
         Logger.logAction(this.logger, Logger.LOG_ERROR, 'Auth()', msg);
-        throw new ErrorInfo(msg, 40160, 401);
+        throw new ErrorInfo({
+          message: msg,
+          code: 40160,
+          statusCode: 401,
+          hint: 'Pass one of ClientOptions.{ key, authUrl, authCallback, token, tokenDetails }. For production, prefer authUrl or authCallback so the API key stays on your server.',
```
lmars: The error message says:

"No authentication options provided; need one of: key, authUrl, or authCallback (or for testing only, token or tokenDetails)"

and then this hint says:

"Pass one of ClientOptions.{ key, authUrl, authCallback, token, tokenDetails }. For production, prefer authUrl or authCallback so the API key stays on your server."

which just seems like a re-wording of the existing message, so perhaps just update the message?

src/common/lib/client/auth.ts
```diff
@@ -269,7 +274,12 @@ class Auth {
     /* RSA10a: authorize() call implies token auth. If a key is passed it, we
      * just check if it doesn't clash and assume we're generating a token from it */
     if (authOptions && authOptions.key && this.authOptions.key !== authOptions.key) {
-      throw new ErrorInfo('Unable to update auth options with incompatible key', 40102, 401);
+      throw new ErrorInfo({
+        message: 'Unable to update auth options with incompatible key',
+        code: 40102,
+        statusCode: 401,
+        hint: 'auth.authorize() cannot change the API key - the new authOptions.key differs from the one the client was constructed with. To use a different key, construct a new Ably client.',
```
lmars: Again, this just seems like an alternative error message? I won't keep saying this, I think we should consider whether this initiative should be "improve the error message" rather than adding a new field.

scripts/hint-coverage.ts
umair-ably: yeah i wasn't sold on this from the get go which i flagged in another comment. I was trying to avoid writing a unit test to assert we actually have hints, but also wanted a way to track we're adding hints. Paddy raised a fair point in that we don't test particular error messages so pointless to try test hints too. I've removed this script

umair-ably:
> @umair-ably I've not reviewed the diff yet, but here's a review just based on the PR description:
> 
> > This PR extends the message-vs-hint split DX-1204 introduced (err.message says what failed; err.hint says how to fix it)
> 
> It would be useful to explain why we're doing a message-vs-hint split rather than expanding the error messages themselves, since [DX-1204](https://ably.atlassian.net/browse/DX-1204) says nothing about it (it seems to be about re-introducing deprecated callbacks, is it definitely the correct ticket?). I'm not necessarily saying we shouldn't, but we should be deliberate in why this is a new field.
> 
> For example, I haven't looked at the diff yet, but let's say this introduces `ErrorInfo{message: "invalid key", hint: "The provided key must have format x.y:z"}`, then why not just update the message to `invalid key: the provided key must have format x.y:z`?
> 

That PR is correct in that it first added the hint field because we needed it for that work too. 

As for why we have a hint vs longer error messages... my stance is the error message is for what went wrong, the hint is for how to fix it. They serve very different purposes. In the example you picked, granted, it does seem like we can just extend the error message, but for more complex examples e.g. telling the user they need to enable mutable messages in the dashboard and/or need a specific mode enabled, I think the separation is a clear win. 

We're also not the only ones going down this route. There's a similar (if not, more detailed) experiment [here](https://github.com/NousResearch/hermes-agent/issues/722#issue-4042367016)... fairly certain supabase also have hints in their error envelope.  

> > Experiment 2: pointing at the Ably CLI from inside the hint ... "If you have the Ably CLI installed, `ably apps rules list` will show which channel namespaces have Mutable Messages enabled"
> 
> If they have the Ably CLI installed, I would suggest we instead consider a command like `ably diagnose 93001` which internally knows to lookup the namespace and to return an appropriate response, even suggesting the command to run to create the relevant rule? This way the error message in the SDKs stays lightweight (i.e. "if you have the Ably CLI, check `ably diagnose <error code>`") and we centralise the diagnostic checks in the CLI. WDYT?
> 

Yeah I like this! Not going to commit to this now as it's beyond the scope of this work but we should iterate towards this. 

> > Experiment 3: LiveObjects portability
> 
> OOI, what exactly was this experiment? As in, what was the prompt?

I created a failing test which tries to subscribe to a liveobjects backed channel without the objects_subsribe mode. All the agent has to do is extend channels.get with the correct modes and the test would pass. 

The prompt was to make the test pass without doing any web searches, and when it's done to write a few sentences on what it did to fix it and what it thinks helped it to fix it. 

I ran this before and after adding a hint. I also thought the error message was enough but there was still an improvement for haiku with the hint instead of relying on the error message alone. 


[DX-1204]: https://ably.atlassian.net/browse/DX-1204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

src/common/lib/client/auth.ts
umair-ably: I've gone through and reduced the messages themselves where appropriate so we keep the structure I've highlighted in another comment i.e. the message is what went wrong, the hint is how to fix it.

lmars:
@umair-ably apologies for taking so long on this one.

> As for why we have a hint vs longer error messages... my stance is the error message is for what went wrong, the hint is for how to fix it. They serve very different purposes. In the example you picked, granted, it does seem like we can just extend the error message, but for more complex examples e.g. telling the user they need to enable mutable messages in the dashboard and/or need a specific mode enabled, I think the separation is a clear win.

When you say "they serve different purposes", I still don't see how? You explain them semantically that `message` is what went wrong, and `hint` is for how to fix it, but when do you expect an LLM to want to view one and not the other? Surely it's always useful to tell the LLM both what went wrong _and_ how to fix it? In which case a single message explaining both will suffice?

> telling the user they need to enable mutable messages in the dashboard and/or need a specific mode enabled, I think the separation is a clear win

I don't see how putting that in a separate field rather than in the existing message is a clear win?

> I ran this before and after adding a hint. I also thought the error message was enough but there was still an improvement for haiku with the hint instead of relying on the error message alone.

Ok thanks, but it seems adding the extra information to the error message would have the same effect?

lmars:
@umair-ably also to speed up the review cycle, I'll assume we do want a separate field, here's where I'm at from the server perspective.

Internally, I'm writing a Markdown doc per error code which has `title` and `summary` in frontmatter to be displayed in list views (i.e. a view of all the errors a user has encountered this month, grouped by error code), and the body to be displayed in detailed views about a specific error code.

The `title` is a short description of what occurred like "Channel backlog too large", and the `summary` is a slightly longer description which includes why it occurred like "The client requested to resume or rewind a channel whose backlog held more messages than a single replay can deliver. The channel attaches, but the messages beyond that limit are not replayed."

Within the Markdown body, I've proposed three sections:
1. What you should do
2. Why it happens
3. What you'll see

For the "Channel backlog too large" error, here's the proposed "What you should do" section:

> ## What you should do
>
> The channel stays attached and keeps delivering new messages, so nothing is broken. Whether to act on the messages that weren't replayed depends on your application:
>
> - If it can tolerate missing them, no action is needed and the error can be ignored.
> - If it cannot, enable [persistence](https://ably.com/docs/storage-history/storage) on the channel and have your application fetch the messages it did not receive from [history](https://ably.com/docs/storage-history/history) whenever it encounters this error.

ISTM this "What you should do" I'm proposing is the same as what you describe "hint" as in this PR.

So, long story short, should we align the two, call the field `what_you_should_do` in the `ErrorInfo`, and maintain it in these Markdown documents?

I should add that this matches [what Cloudflare do](https://blog.cloudflare.com/rfc-9457-agent-error-pages/).

umair-ably:
Sorry @lmars, I'm slightly more confused now... 

We had previously discussed and agreed adding a top level `hint` field (because the first proposition of `ErrorDetail` clashed), but your first message from today recommends dropping this and adding the info to the error message itself. 

Your second message  is in favour of the separation (I think?) but renaming the field from `hint` to `what_you_should_do`? 

If it's a choice between the two, then naturally I'm in favour of the second option but I'd propose some other names to consider:

- `remediation`
- `guidance`

`what_you_should_do` is too prescriptive and wordy imo

lmars:
@umair-ably yes sorry, I wanted to give you two perspectives at once so that we can discuss them in parallel 🙂 

On the first point about adding a new field, I know we discussed it, but then seeing the actual implementation in this PR, and reading the hints versus messages (which in many cases seem to be saying the same thing just worded differently), I am questioning whether there is a need to add a new field versus just making the message more descriptive, since it does not seem to me that it matters to an LLM whether the information it needs is spread across two fields or all in one field. So _really_, I'm asking for a more concrete justification.

On the second point, I'm assuming you give a more concrete justification, and then providing a review given that I then agree with adding a new field, and trying to align it with what I've been doing server side since we spoke.

> If it's a choice between the two

I'm going to concede the first point and assume we're adding a new field 👍 

> I'd propose some other names to consider ... `what_you_should_do` is too prescriptive and wordy imo

Happy to discuss alternatives, I don't like `hint` as I think it indicates a suggestion that is somehow secondary to understanding the error and something you might ignore, I think being prescriptive is _exactly_ what we want, especially for LLMs: be direct, this is what you should do regarding this error.

That being said, I'll now review the actual changes independent of what the field is called.

umair-ably:
@lmars 

> I don't like hint as I think it indicates a suggestion that is somehow secondary to understanding the error and something you might ignore

That's fair but it's exactly this reason why I want a separate field. There's something about an LLM seeing an explicit field which prods it into knowing there's something it can do to fix it. The liveobjects hint showed exactly this. The error was already clear enough on it's own, but introducing a hint just made it explicit that the agent can self-heal here. 

I'm not wedded to `hint` at all, so more than happy to change this :)

lmars:
Ok thanks, so my suggestion is `what_you_should_do` because:
- it's very clear what this field is: it's what you should do regarding this error (I think this isn't a name we would have picked in the past, but for LLMs, I think clarity and explicitness is important)
- it matches the wording we'll have in the docs (i.e. a "What you should do" header)
- it matches what Cloudflare do (although I don't think they made a particularly scientific decision)

umair-ably:
@lmars I can get behind `what_you_should_do` but I imagine @AndyTWF and @m-hulbert would have some thoughts - we've explicitly tried not to solely optimise for LLMs whilst ignoring human readability/expectations in the past

src/common/lib/client/auth.ts
```diff
@@ -486,27 +495,39 @@ class Auth {
           }
           if (Platform.BufferUtils.isBuffer(body)) body = body.toString();
           if (!contentType) {
-            cb(new ErrorInfo('authUrl response is missing a content-type header', 40170, 401), null);
+            const err = new ErrorInfo({
+              message: 'authUrl response is missing a content-type header',
+              code: 40170,
+              statusCode: 401,
+              hint: 'Auth endpoints may return a Content-Type of application/json (TokenDetails/TokenRequest), text/plain (token string) or application/jwt.',
```
lmars: We describe the hint as "how to fix it", so I think this should be worded like "Update the server that authUrl points at to return a compatible Content-Type HTTP header, (explain what are compatible)".

Should we consider committing some guidelines to the repo explaining how the hint should be worded and formatted?

src/common/lib/client/auth.ts
```diff
@@ -486,27 +495,39 @@ class Auth {
...
             return;
           }
           if (json) {
             if ((body as string).length > MAX_TOKEN_LENGTH) {
-              cb(new ErrorInfo('authUrl response exceeded max permitted length', 40170, 401), null);
+              const err = new ErrorInfo({
+                message: 'authUrl response exceeded max permitted length',
+                code: 40170,
+                statusCode: 401,
+                hint: 'authUrl payloads must be under 128 KB. If your TokenDetails legitimately contains a large capability, trim unused fields or set authOptions.suppressMaxLengthCheck. Otherwise check the endpoint is returning only the token shape, not wrapped in extra JSON.',
```
lmars: > trim unused fields

This only applies to JWTs, if using `requestToken` then Ably will return a persisted token if a literal token would be too large. It does concern me a little that we're embedding advice in code that might not align with our own documentation.

> suppressMaxLengthCheck

This error is not wrapped in such a check?

> Otherwise check the endpoint is returning only the token shape, not wrapped in extra JSON.

This is a bit abstract, if this is useful advice we want to keep, then it should probably say "check the endpoint is returning a JSON object like X" so there is a concrete check that the JSON is in the right "shape".

src/common/lib/client/auth.ts
```diff
@@ -585,7 +606,12 @@ class Auth {
         'Auth()',
         'library initialized with a token literal without any way to renew the token when it expires (no authUrl, authCallback, or key). See https://help.ably.io/error/40171 for help',
       );
-      throw new ErrorInfo(msg, 40171, 403);
+      throw new ErrorInfo({
+        message: msg,
+        code: 40171,
+        statusCode: 403,
+        hint: 'Initialise the client with one of ClientOptions.{ key, authUrl, authCallback } so the SDK can refresh tokens. A bare token/tokenDetails alone cannot be renewed once expired.',
```
lmars: The error message here is:

> Need a new token, but authOptions does not include any way to request one (no authUrl, authCallback, or key)

and I think this hint is just a rewording of that message.

I think in this case, if the wording of the message isn't optimal that the hint is fixing, we should update the message too.

src/common/lib/client/auth.ts
```diff
@@ -628,7 +654,13 @@ class Auth {
           tokenRequestCallbackTimeoutExpired = true;
           const msg = 'Token request callback timed out after ' + timeoutLength / 1000 + ' seconds';
           Logger.logAction(this.logger, Logger.LOG_ERROR, 'Auth.requestToken()', msg);
-          reject(new ErrorInfo(msg, 40170, 401));
+          const err = new ErrorInfo({
+            message: msg,
+            code: 40170,
+            statusCode: 401,
+            hint: 'authCallback did not invoke its callback within the timeout, or authUrl did not respond. Check that the callback runs to completion on every code path and that authUrl is reachable.',
```
lmars: > authCallback did not invoke its callback within the timeout, or authUrl did not respond

I suggest "a request to retrieve a new token using authCallback or authUrl did not complete within the configured timeout" (which I think is just an improved error message, not a hint).

> Check that the callback runs to completion on every code path

This is a little abstract, how do you expect this to be checked? As a developer, what I'd probably do is either look at my logs to see why it took so long, or if I don't have any logs, add some.

src/common/lib/client/auth.ts
```diff
@@ -648,29 +680,41 @@ class Auth {
         /* the response from the callback might be a token string, a signed request or a token details */
         if (typeof tokenRequestOrDetails === 'string') {
           if (tokenRequestOrDetails.length === 0) {
-            reject(new ErrorInfo('Token string is empty', 40170, 401));
+            const err = new ErrorInfo({
+              message: 'Token string is empty',
+              code: 40170,
+              statusCode: 401,
+              hint: 'Return a non-empty token string, or a TokenDetails/TokenRequest object.',
```
lmars: Return a non-empty token string where? I think this could be a little more forthcoming.

lmars (commented):
I've started to review the actual hints, and I'm finding each to be a little vague in terms of what it's telling the user to actually do about the error, so before I continue:
- what guidelines did we use to produce these hints?
- is it worth only including a hint when we think there is a genuine reason for it?

umair-ably:
@lmars I've made some more tweaks in the latest commit. 

Re: guidelines - these were created using a dynamic workflow following some basic rules e.g. error message is what went wrong, hint is guidance on what can or should be done next - followed by some generic "follow the code to verify what's happening". 

The recent commit has tried to codify some standards e.g. imperative voice. I'll commit this to the repo in a separate repo once this lands. 

Re: when to add a hint. I think we've touched on this before - it's easier to mandate a hint everywhere than trying to draw some arbitrary line and miss useful cases. Likewise, I've shown some evidence that even a hint we don't think is useful has proved some use to LLMs. I think the risk of adding a hint when it is not needed is so small that it shouldn't be blocking this PR - we can certainly revisit it if it is something that is happening more often than not, but I don't think that is the case.

src/common/lib/client/auth.ts
```diff
@@ -155,10 +155,14 @@ class Auth {
...
-        const msg =
-          'No authentication options provided; need one of: key, authUrl, or authCallback (or for testing only, token or tokenDetails)';
+        const msg = 'No authentication options provided';
         Logger.logAction(this.logger, Logger.LOG_ERROR, 'Auth()', msg);
-        throw new ErrorInfo(msg, 40160, 401);
+        throw new ErrorInfo({
+          message: msg,
+          code: 40160,
+          statusCode: 401,
+          hint: 'Pass one of ClientOptions.{ key, authUrl, authCallback, token, tokenDetails }. For production, prefer authUrl or authCallback so the API key stays on your server.',
```
lmars: I think "For production, prefer authUrl or authCallback so the API key stays on your server." is confusing, what if this code is running on a server? How about the following which matches [Recommended authentication](https://ably.com/docs/auth#recommended) from docs:


```suggestion
          hint: 'Set one of the following in ClientOptions: key, authUrl, authCallback, token, or tokenDetails. For production use, prefer authUrl or authCallback for client-side (browser, mobile apps), or key for server-side.',
```

src/common/lib/client/auth.ts
```diff
@@ -269,7 +273,12 @@ class Auth {
     /* RSA10a: authorize() call implies token auth. If a key is passed it, we
      * just check if it doesn't clash and assume we're generating a token from it */
     if (authOptions && authOptions.key && this.authOptions.key !== authOptions.key) {
-      throw new ErrorInfo('Unable to update auth options with incompatible key', 40102, 401);
+      throw new ErrorInfo({
+        message: 'Unable to update auth options with incompatible key',
+        code: 40102,
+        statusCode: 401,
+        hint: 'auth.authorize() cannot change the API key - the new authOptions.key differs from the one the client was constructed with. To use a different key, construct a new Ably client.',
```
lmars: The general guidance is that "`err.message` says what failed; `err.hint` says how to fix it", and here "the new authOptions.key differs from the one the client was constructed with" is more "what failed" than it is "how to fix it".

Saying this here because I'm wondering if the prompts we're using to generate these hints needs to be a bit more explicit that `hint` should be to-the-point about specifically what needs to be done, rather than expanding on what the error means (which is what `err.message` is for).

If I were writing this based on the guidance:


```suggestion
        message: 'authorize called with a key that does not match the existing key being used by the client',
        code: 40102,
        statusCode: 401,
        hint: 'To use a different key, construct a new Ably client with the key as a client option.',
```

This is more about updating the guidelines than it is about this specific case. I won't keep suggesting this for other cases, I'll just focus on anything that is inaccurate, and leave you to decide if any futher message/hint splitting is useful.

src/common/lib/client/auth.ts
```diff
@@ -486,27 +495,39 @@ class Auth {
...
             return;
           }
           if (json) {
             if ((body as string).length > MAX_TOKEN_LENGTH) {
-              cb(new ErrorInfo('authUrl response exceeded max permitted length', 40170, 401), null);
+              const err = new ErrorInfo({
+                message: 'authUrl JSON response exceeded the maximum permitted length of 128 KB',
+                code: 40170,
+                statusCode: 401,
+                hint: 'Make your authUrl endpoint return only the token payload (a TokenDetails or TokenRequest object, or a token string), not an envelope wrapping it in extra fields. If a TokenDetails genuinely needs a large capability, narrow the capability to the channels and operations the client actually needs.',
```
lmars: > Make your authUrl endpoint return only the token payload

How do we know this is the reason? Can an encoded `TokenDetails` object not itself exceed this limit?

> If a TokenDetails genuinely needs a large capability

Why does this jump into capabilities?

I think here, behind the scenes we've thought of two reasons the token might be too large: either it's wrapped which makes it bigger, or its got a very large capability, and then we're hinting at a fix for both without stating those might be the reasons?

This is what I'd probably write:

> If the server behind authUrl is returning Ably tokens wrapped with extra fields, consider removing the fields that are not required. If the client is being granted a long list of capabilities that make the token itself very large, consider using [wildcards](https://ably.com/docs/auth/capabilities#wildcards) rather than long lists of channels.

src/common/lib/client/auth.ts
```diff
@@ -648,29 +683,41 @@ class Auth {
         /* the response from the callback might be a token string, a signed request or a token details */
         if (typeof tokenRequestOrDetails === 'string') {
           if (tokenRequestOrDetails.length === 0) {
-            reject(new ErrorInfo('Token string is empty', 40170, 401));
+            const err = new ErrorInfo({
+              message: 'Token string is empty',
+              code: 40170,
+              statusCode: 401,
+              hint: 'Return a non-empty value from your authCallback/authUrl: a token string, a TokenRequest, or a TokenDetails object. The callback returned an empty string.',
```
lmars:
```suggestion
              hint: 'Return a non-empty value from your authCallback/authUrl: a token string, a TokenRequest, or a TokenDetails object.',
```

src/common/lib/client/auth.ts
```diff
@@ -681,18 +728,25 @@ class Auth {
             'Expected token request callback to call back with a token string or token request/details object, but got a ' +
             typeof tokenRequestOrDetails;
           Logger.logAction(this.logger, Logger.LOG_ERROR, 'Auth.requestToken()', msg);
-          reject(new ErrorInfo(msg, 40170, 401));
+          const err = new ErrorInfo({
+            message: msg,
+            code: 40170,
+            statusCode: 401,
+            hint: 'authCallback must invoke its callback with (err, tokenStringOrTokenDetailsOrTokenRequest). authUrl must respond with a token string or TokenDetails/TokenRequest JSON.',
```
lmars:
```suggestion
            hint: 'If authenticating with an authCallback, update it to callback with (err, tokenStringOrTokenDetailsOrTokenRequest). If authenticating with an authUrl, update the server to respond with a token string or TokenDetails/TokenRequest JSON.',
```

src/common/lib/client/auth.ts
```diff
@@ -681,18 +728,25 @@ class Auth {
...
+          const err = new ErrorInfo({
+            message:
               'Token request/details object exceeded max permitted stringified size (was ' + objectSize + ' bytes)',
-              40170,
-              401,
-            ),
-          );
+            code: 40170,
+            statusCode: 401,
+            hint: 'Token objects must serialise to under 128 KB. Trim unused fields from your TokenDetails/TokenRequest, and if a large capability is required, narrow it to the channels and operations the client actually needs.',
```
lmars: This seems like the same hint as above but worded completely differently? Should we be a bit more consistent?

src/common/lib/client/auth.ts
```diff
@@ -704,7 +758,13 @@ class Auth {
           const msg =
             'Expected token request callback to call back with a token string, token request object, or token details object';
           Logger.logAction(this.logger, Logger.LOG_ERROR, 'Auth.requestToken()', msg);
-          reject(new ErrorInfo(msg, 40170, 401));
+          const err = new ErrorInfo({
+            message: msg,
+            code: 40170,
+            statusCode: 401,
+            hint: 'Return a token string, a TokenRequest (an object with a `keyName` field), or a TokenDetails (an object with an `issued` field) from your authCallback/authUrl. The returned object had neither field, so it matched neither shape; note that a bare `token` field is not enough to identify a TokenDetails here, the `issued` field is what the SDK keys on.',
```
lmars: This is being more explicit in what `TokenRequest` is (i.e. "an object with a `keyName` field"), but none of the other hints have done that, should they?

src/common/lib/client/auth.ts
```diff
@@ -775,18 +835,33 @@ class Auth {
 
     const key = authOptions.key;
     if (!key) {
-      throw new ErrorInfo('No key specified', 40101, 403);
+      throw new ErrorInfo({
+        message: 'No key specified',
+        code: 40101,
+        statusCode: 403,
+        hint: 'createTokenRequest needs an API key. Pass ClientOptions.key on the client or { key } in the authOptions argument. Token-auth clients cannot construct token requests themselves.',
```
lmars:
```suggestion
        hint: 'Pass ClientOptions.key on the client or { key } in the authOptions argument to createTokenRequest.',
```

src/common/lib/client/auth.ts
```diff
@@ -972,13 +1049,19 @@ class Auth {
...
-      throw new ErrorInfo(
-        'Can’t use "*" as a clientId as that string is reserved. (To change the default token request behaviour to use a wildcard clientId, instantiate the library with {defaultTokenParams: {clientId: "*"}}), or if calling authorize(), pass it in as a tokenParam: authorize({clientId: "*"}, authOptions)',
-        40012,
-        400,
-      );
+      throw new ErrorInfo({
+        message: 'Can’t use "*" as a clientId as that string is reserved',
+        code: 40012,
+        statusCode: 400,
+        hint: 'Move "*" out of ClientOptions.clientId; for a wildcard identity set defaultTokenParams: { clientId: "*" } on the client, or pass { clientId: "*" } to authorize() as a tokenParam. A client using a raw API key can already assume any clientId, so a wildcard token request succeeds by default; it is rejected only when the token or key issuing it is itself restricted to a single clientId.',
```
lmars: I don't understand this hint, it's saying don't set `clientId = "*"`, but then suggesting to set `clientId = "*"` in a bunch of other places? Isn't it the case that either this client can use a wildcard clientId, or it can't?

src/common/lib/client/baseclient.ts
```diff
@@ -76,21 +76,33 @@ class BaseClient {
       if (!keyMatch) {
         const msg = 'invalid key parameter';
         Logger.logAction(this.logger, Logger.LOG_ERROR, 'BaseClient()', msg);
-        throw new ErrorInfo(msg, 40400, 404);
+        throw new ErrorInfo({
+          message: msg,
+          code: 40400,
+          statusCode: 404,
+          hint: 'ClientOptions.key must be the full "appId.keyId:secret" string copied from the Ably dashboard. If you only have a token, use ClientOptions.token / tokenDetails instead. If you have the Ably CLI installed, `ably auth keys list` shows the keys configured on the current app.',
```
lmars: I don't think we should suggest setting `token` or `tokenDetails` in client options.

src/common/lib/client/baseclient.ts
```diff
@@ -76,21 +76,33 @@ class BaseClient {
...
-          'Can’t use "*" as a clientId as that string is reserved. (To change the default token request behaviour to use a wildcard clientId, use {defaultTokenParams: {clientId: "*"}})',
-          40012,
-          400,
-        );
+      if (!(typeof normalOptions.clientId === 'string' || normalOptions.clientId === null)) {
+        throw new ErrorInfo({
+          message: 'clientId must be either a string or null',
+          code: 40012,
+          statusCode: 400,
+          hint: 'For an anonymous client omit clientId or pass null; to identify the client pass a stable string such as a user id. To authenticate as any clientId, use a wildcard token (defaultTokenParams: { clientId: "*" }) rather than setting clientId here.',
```
lmars: This error means the code passed in something which is not a string or null, so then why explain about wildcard tokens? It's not like if the token has a wildcard client id then the code can use a non string-or-null value here.

src/common/lib/client/paginatedresource.ts
```diff
@@ -147,15 +147,25 @@ export class PaginatedResult<T> {
       return this.get(this._relParams!.first);
     }
 
-    throw new ErrorInfo('No link to the first page of results', 40400, 404);
+    throw new ErrorInfo({
+      message: 'No link to the first page of results',
+      code: 40400,
+      statusCode: 404,
+      hint: 'The public PaginatedResult surface exposes no boolean guard for first(), so wrap the call in try/catch or attach a .catch handler to the promise it returns rather than pre-checking. A first-page link exists only on a page whose paginated REST response carried a Link header with rel="first", so expect first() to reject on results that were not produced by such a query.',
```
lmars: This reads like jargon, what is "the public PaginatedResult surface", and what's a "boolean guard for first()"?

Also, it suggests "wrap the call in try/catch or attach a .catch handler to the promise it returns rather than pre-checking" but how is that helping to fix the error? What if the code already caught the error and now needs to do something to fix it?

I think this is definitely a case where a hint for a hint's sake seems detrimental.

src/common/lib/client/paginatedresource.ts
```diff
@@ -147,15 +147,25 @@ export class PaginatedResult<T> {
...
     if (this.hasCurrent()) {
       return this.get(this._relParams!.current);
     }
 
-    throw new ErrorInfo('No link to the current page of results', 40400, 404);
+    throw new ErrorInfo({
+      message: 'No link to the current page of results',
+      code: 40400,
+      statusCode: 404,
+      hint: 'Call current() only on a PaginatedResult returned by a paginated REST query (such as channel history, presence, or stats) whose response carried pagination Link headers. This page has no rel="current" navigation link, so there is nothing to reload; if you only need to walk pages, use next() together with isLast() instead, since those are the public navigation methods this result type exposes.',
```
lmars: > Call current() only on a PaginatedResult returned by a paginated REST query (such as channel history, presence, or stats) whose response carried pagination Link headers

How exactly should the developer do that though?

src/common/lib/client/realtimeannotations.ts
```diff
@@ -69,13 +69,15 @@ class RealtimeAnnotations {
...
-        93001,
-        400,
-      );
+      throw new ErrorInfo({
+        message:
+          "You are trying to add an annotation listener, but you haven't requested the annotation_subscribe channel mode in ChannelOptions, so this won't do anything (we only deliver annotations to clients who have explicitly requested them)",
+        code: 93001,
+        statusCode: 400,
+        hint: 'Re-create the channel with annotation_subscribe in modes, e.g. realtime.channels.get(name, { modes: ["subscribe", "annotation_subscribe"] }), since appending to channel.modes after attach() does not enable the mode server-side. If the subsequent attach is rejected by the server, confirm the channel namespace has "Message annotations, updates, appends, and deletes" enabled in the Ably dashboard and that your API key has annotation-subscribe capability on this channel. If you have the Ably CLI installed, `ably apps rules list` shows which namespaces have it enabled and `ably auth keys list` shows the capabilities of your key.',
```
lmars: Should this be suggesting `setOptions` to update the modes, because won't `channels.get(name)` with different options throw an error?

src/common/lib/client/realtimechannel.ts
```diff
@@ -419,8 +441,14 @@ class RealtimeChannel extends EventEmitter {
...
         return;
       // RTL5b
-      case 'failed':
-        throw new ErrorInfo('Unable to detach; channel state = failed', 90001, 400);
+      case 'failed': {
+        throw new ErrorInfo({
+          message: 'Unable to detach; channel state = failed',
+          code: 90001,
+          statusCode: 400,
+          hint: 'Release it via channels.release(name) and call channels.get(name) again to start a fresh channel.',
```
lmars: If the code is trying to detach, why call `channels.get(name)` to start a fresh channel, which will attach again?

src/common/lib/client/realtimechannel.ts
```diff
@@ -951,12 +984,22 @@ class RealtimeChannel extends EventEmitter {
   timeoutPendingState(): void {
     switch (this.state) {
       case 'attaching': {
-        const err = new ErrorInfo('Channel attach timed out', 90007, 408);
+        const err = new ErrorInfo({
+          message: 'Channel attach timed out',
+          code: 90007,
+          statusCode: 408,
+          hint: 'The SDK will retry automatically once the connection is healthy; check connection.state and connection.errorReason.',
```
lmars: How do we know the timeout is because the connection is unhealthy?

src/common/lib/client/realtimepresence.ts
```diff
@@ -195,12 +210,11 @@ class RealtimePresence extends EventEmitter {
...
-        const err = new PartialErrorInfo(
-          'Unable to leave presence channel while in ' + channel.state + ' state',
-          90001,
-        );
-        err.code = 90001;
-        throw err;
+        throw new PartialErrorInfo({
+          message: 'Unable to leave presence channel while in ' + channel.state + ' state',
+          code: 90001,
+          hint: 'Inspect channel.errorReason for the cause, then await channel.attach() and retry presence.leave() if the channel state is "failed". If the state is "initialized" no member was ever entered, so there is nothing to leave and no action is needed.',
```
lmars: If the state is "failed", isn't `channel.attach()` going to throw?

lmars (commented):
I'm down to src/common/lib/transport/connectionmanager.ts in the diff but leaving my review so far (I'll come back to it).

umair-ably:
Thanks @lmars - I've applied most of these changes, just checking through them myself once more before pushing up. 

A few I'd like to sanity-check before I finalise the wording:
- realtimechannel.ts:449: does channels.get() actually re-attach on its own? I read it as just returning the channel, so I leaned on attach() to recover and channels.release() to discard instead... does this sound right? 
- realtimepresence.ts:216: I don't think attach() throws from failed (looks like it re-attempts)? The real issue seemed to be that presence members get cleared on failure, so there'd be nothing left to leave?
- auth.ts:1063: on the * contradiction: aren't ClientOptions.clientId (a fixed identity, where * is reserved) and a wildcard token (clientId: "*") two different mechanisms?

lmars:
@umair-ably apologies for the delay, I will re-review the diff, but re your questions above:

> realtimechannel.ts:449: does channels.get() actually re-attach on its own? I read it as just returning the channel, so I leaned on attach() to recover and channels.release() to discard instead... does this sound right?

Sorry I think my comment was misleading, and I'm not that familiar with ably-js internals so I might be incorrect with my understanding of how things work, but this error is returned from `detach()` when the channel state is `failed`, and the hint says:

"Release it via channels.release(name) and call channels.get(name) again to start a fresh channel."

So what I mean is, the developer has called `detach()`, got an error because the channel is in a failed state (does that mean it's already detached?), and we're suggesting they release that channel, get a fresh one, and then I assume call `detach()` on it again, but what's that going to achieve?

I think you're updated hint seems reasonable (although perhaps this is highlighting that I'm perhaps not best placed to scrutinise some of these hints).

> realtimepresence.ts:216: I don't think attach() throws from failed (looks like it re-attempts)? The real issue seemed to be that presence members get cleared on failure, so there'd be nothing left to leave?

Ok thanks, I defer to your judgement on that one (I think I just misunderstand what a "failed" channel state is).

> auth.ts:1063: on the * contradiction: aren't ClientOptions.clientId (a fixed identity, where * is reserved) and a wildcard token (clientId: "*") two different mechanisms?

To try to clarify what I mean, this error is thrown when code does something like `new Ably.Realtime({ clientId: "*" })`, and the suggestion is making a code change to instead do `new Ably.Realtime({ defaultTokenParams: { clientId: "*" } })`, but then the client won't actually get a `clientId` set to `*`, they'll actually not get a client id? So what are we trying to suggest here?

lmars:
@umair-ably I've reviewed the diff, and whilst there are some hints I'd consider re-wording, nothing looks to be incorrect or misleading, so I would be happy to approve, but I notice there is no description of the hint in the diff itself? I think to merge this, we need to add guidance for other developers and coding agents explaining what a hint is, and that every ErrorInfo requires one (if that's the decision we're making here).

lmars (approved):
LGTM

PR #2236, opened 2026-06-02
[DX-1211] Address currently silent failures around missing `modes` 

umair-ably:
Targets https://github.com/ably/ably-js/pull/2233 - review that first please

## Author's Note

I'd advise reading the Claude generated sections below as they outline the experiments and exact numbers, but if you want a summary of my thoughts... 

This PR aims to address 2 silent failures we have today:
- `presence.get()` on a channel without `presence_subscribe` returns an empty list.
- `channel.subscribe()` on a channel without `subscribe` registers a listener that never fires.

The fixes themselves were easy in that we just throw an error and hint pairing when running into these issues. We also introduce a `strictMode` which allows us to keep this as an additive change (log instead of throw) before promoting it to a breaking change in a later release (always throw). We want this strictMode in case people don't wish to upgrade to a breaking change release, but they can still benefit from the improvements we've made here. 

The experiments themselves were the tricky and time-consuming part.

- It was actually difficult to get Opus or Sonnet to run into these issues as catastrophically as they have done previously e.g. as reported on Matt's gist. There could be a mixture of reasons e.g. Opus with a single problem and a clear context window doesn't find this challenging to fix, or the training cycles since the gist has meant this is less of a problem.
- To emulate a bloated context and a larger more complex project, I used Haiku. The purpose of this wasn't to test a weaker model, but to test how Opus/Sonnet could behave in more testing environments. This actually allowed us to run into the issue more frequently, whilst also testing how well the agent recovered with our fixes from this PR. 
- With Haiku in mind, we managed to go from 50% of agents running in circles unable to fix the silent issues, to only 10% running into the issues with the additive fix (e.g. the error/hint is logged and not thrown), to then 0% of them running into the issue if we enabled strictMode (e.g. the error/hint is thrown). 
- This shows even the log alone does a lot of the heavy lifting, and that the fix itself is never the tricky thing for agents, but it's not even knowing something is wrong that needs to be our focus. 
- Whilst this PR approaches these issues from the angle of "let the agent hit the issue, and then tell it to fix it", I strongly believe we can preempt these issues (and others) by heavily improving our docstrings. I have this conviction because all agents actually read the docstrings info for strictMode which implies agents are good at getting info upfront if that info actually exists. Extending our docstrings with prerequisites and sideeffects will be my next focus area (and on a much wider surface area than just these 2 silent issues) 

The full experiment runs and details are at https://github.com/ably-labs/llm-eval/pull/10

## What this PR does

Two SDK calls fail silently today:

- `presence.get()` on a channel without `presence_subscribe` returns an empty list.
- `channel.subscribe()` on a channel without `subscribe` registers a listener that never fires.

Both with no error and no warning. A developer or AI agent gets back "nothing", assumes it works, and
ships broken code.

This PR makes those calls speak up: a warning log by default, and — if you set
`clientOptions.strictMode` — a thrown error with a hint instead of the silent result. This is the v2.x
step of the A5/B5 rollout in DXRFC-022 (warn now; `strictMode` becomes the default in v3).

## How we tested it

We had AI agents do small Ably coding tasks, with no web access, treating the SDK as a black box. Each
task ran against three versions of the SDK:

- **baseline** — today's published SDK (silent).
- **B-error** — this PR: the warning shows at the default log level, no setup needed.
- **C** — `strictMode: true`: the call throws instead of returning silently.

(A fourth, **B-verbose**, is the warning shown only if the developer raised the log level themselves.)

About 370 runs, 0 invalid, on Opus 4.8 and Sonnet 4.6, plus Haiku 4.5 on the last experiment. The
agents can't cheat by swapping in a better API key — the harness blocks it. Every result is compared
against the currently-published `ably@2.21.0`.

We labelled each run by what the agent did:

- **UNDETECTED** — shipped the empty result as "it works." This is the failure we want to remove.
- **DETECTED / REMEDIATED** — noticed the problem, or flagged the real cause (the key needs more access).
- **IDIOMATIC / HACKY** — for fixable cases: a clean fix, vs. one that hides the symptom.

## The main result

On the strong models, the surfacing barely changed anything — because Opus and Sonnet usually work out
the cause on their own. At baseline they shipped broken code only about 10% of the time, so there was
little left to fix.

But that's the easy case: a capable model, full attention, a tiny task. In real use a strong model
often has a full or long context window, or is working in a large unfamiliar codebase, and can't give
each empty result that much attention. We can't easily simulate "Opus with a full context window", so
we used **Haiku as a stand-in for a strong model that can't fully focus** — same task, less reasoning
brought to bear.

That is where the surfacing matters. On the keystone task:

| Keystone task, Haiku, 10 runs | Shipped broken as "it works" |
| --- | --- |
| baseline (today, silent) | 5 / 10 (50%) |
| B-error (this PR) | 1 / 10 (10%) |
| C (`strictMode`) | 0 / 10 |

So the change removes most silent failures exactly when the model is stretched thin — the realistic
case, not the exception.

This is a stand-in, not a direct measurement: we did not run Opus with a deliberately full context.
Haiku is the proxy. Measuring a frontier model under real load is the obvious next test.

## The seven experiments

1. **No-oracle subscribe (keystone).** "Subscribe to this channel, collect messages, tell us if it
   works — there's no test." The key can't subscribe, so nothing arrives. Opus/Sonnet shipped broken
   2/20 at baseline, 1/20 with B-error, 0/10 with C. More telling: at baseline, 0 of 20 runs reached
   the diagnosis from the SDK (they guessed, or tripped a loud server error by probing); with B-error,
   20 of 20 did.
2. **Presence.** "Print who's present" — the key can't read presence, so the list is empty. Both
   models nearly always noticed (the task says someone is present, and a wrong guess triggers a loud
   server error anyway), so all arms scored about the same.
3. **Presence, log-level question.** Same presence task, run to compare the default-level warning
   (B-error) against the throw (C). Result: **B-error ≈ C** — the default warning does the job, and
   `strictMode` is the backstop for the rest.
4. **Debugging a reported bug.** "This shows an empty list — fix it," with a capable key and the wrong
   channel options. Every model and every arm fixed it cleanly in about 6 steps. When the bug is
   reported and the cause is in plain sight, the warning adds nothing.
5. **Discoverability.** "Build a presence demo; we need silent failures to surface as errors" — without
   naming `strictMode`. 100% of runs found `strictMode` in the type definitions and switched it on.
   Agents read the JSDoc.
6. **A realistic silent failure.** A genuine Ably gotcha: setting channel `modes` *replaces* the
   defaults, so adding one mode silently drops presence. We buried this in a multi-file app and in a
   from-scratch build. Two-thirds of agents made the mistake — but on the strong models they all caught
   it themselves, so nothing shipped broken. The "wasted hours debugging" failures we'd heard about did
   not show up on the strong models at this scale.
7. **Haiku.** We re-ran experiments 4 and 1 on Haiku. Debugging (4) stayed flat — even a weak model
   fixes a reported bug. The no-oracle task (1) is where the 50% → 10% → 0% result above came from.

## Where it doesn't help

- Reported bugs in small code — the fix is obvious without the warning.
- Tasks that already signal the problem (e.g. a loud server error one step away).
- Strong models with room to think — they self-diagnose. That is why the value shows up under load.

## Good next step: clearer JSDoc

Experiment 5 showed agents reliably read the SDK's type definitions — 100% found `strictMode` there.
That makes the JSDoc a cheap, high-value place to stop these failures before they happen.

The clearest example is the bug in experiment 6. The JSDoc for `ChannelOptions.modes` just says "An
array of ChannelMode objects." It does not say that setting `modes` **replaces** the defaults instead of
adding to them — which is the exact cause of that silent failure. One added line ("setting `modes`
replaces the default set — list every mode you need") would stop agents making the mistake at all, in
the place they actually look.

This PR makes failures speak up at runtime. Clearer JSDoc on `modes` (and related options) would prevent
some of them up front. The two together are the natural follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added an optional `strictMode` setting (default: `false`) to convert certain previously silent failure cases into thrown errors with clearer, hint-rich messages.

* **Bug Fixes**
  * Improved error handling when required channel modes are missing for `subscribe()` and `presence.get()`, including consistent error codes/hints and reduced repeated warnings (logging is limited unless strict mode is enabled).

* **Tests**
  * Added test coverage for `strictMode` behavior for missing `subscribe` and `presence_subscribe` modes.

### DIFF
--- a/src/common/lib/client/realtimeannotations.ts
+++ b/src/common/lib/client/realtimeannotations.ts
@@ -83,7 +83,7 @@ class RealtimeAnnotations {
         this.logger,
         Logger.LOG_MAJOR,
         'RealtimeAnnotations.subscribe()',
-        err.message + '; hint=' + err.hint,
+        err.message + '; remediation=' + err.remediation,
       );
       // The listener stays registered despite the throw, matching subscribe()'s existing
       // semantics: the listener is always added regardless of attach outcome.
--- a/src/common/lib/client/realtimeannotations.ts
+++ b/src/common/lib/client/realtimeannotations.ts
@@ -77,7 +77,7 @@ class RealtimeAnnotations {
         code: 93001,
         statusCode: 400,
         remediation:
-          'Include "annotation_subscribe" in the channel modes: realtime.channels.get(name, { modes: ["subscribe", "annotation_subscribe", ...] }), or call channel.setOptions({ modes: [...] }) on an existing channel (this triggers a reattach). If the subsequent attach is rejected by the server, check that the channel namespace has "Message annotations, updates, appends, and deletes" enabled in the Ably dashboard and that your API key has annotation-subscribe capability on this channel. If you have the Ably CLI installed, `ably apps rules list` shows which namespaces have it enabled, and `ably auth keys list` shows your key\'s capabilities.',
+          'Include "annotation_subscribe" in the channel modes: realtime.channels.get(name, { modes: ["annotation_subscribe", ...] }), or call channel.setOptions({ modes: [...] }) on an existing channel (this triggers a reattach). If the attach is then rejected, the channel namespace does not have "Message annotations, updates, appends, and deletes" enabled (`ably apps rules list` shows which namespaces do). If the attach succeeds but annotations still are not delivered, your API key lacks the annotation-subscribe capability and the server silently dropped the mode (`ably auth keys list` shows your key\'s capabilities).',
       });
       Logger.logActionNoStrip(
         this.logger,
--- a/src/common/lib/client/realtimeannotations.ts
+++ b/src/common/lib/client/realtimeannotations.ts
@@ -70,8 +70,8 @@ class RealtimeAnnotations {
     }
 
     // explicit check for attach state in case attachOnSubscribe=false
-    if ((this.channel.state === 'attached' && this.channel._mode & flags.ANNOTATION_SUBSCRIBE) === 0) {
-      throw new ErrorInfo({
+    if (this.channel.state === 'attached' && (this.channel._mode & flags.ANNOTATION_SUBSCRIBE) === 0) {
+      const err = new ErrorInfo({
         message:
           "You are trying to add an annotation listener, but you haven't requested the annotation_subscribe channel mode in ChannelOptions, so this won't do anything (we only deliver annotations to clients who have explicitly requested them)",
         code: 93001,
@@ -79,6 +79,15 @@ class RealtimeAnnotations {
         remediation:
           'Enable the mode on the channel with channel.setOptions({ modes: ["subscribe", "annotation_subscribe"] }), which re-attaches with the new mode (calling channels.get(name, { modes }) on an existing channel throws, and appending to channel.modes does not enable it server-side). If the re-attach is rejected by the server, confirm the channel namespace has "Message annotations, updates, appends, and deletes" enabled in the Ably dashboard and that your API key has annotation-subscribe capability on this channel. If you have the Ably CLI installed, `ably apps rules list` shows which namespaces have it enabled and `ably auth keys list` shows the capabilities of your key.',
       });
+      Logger.logActionNoStrip(
+        this.logger,
+        Logger.LOG_MAJOR,
+        'RealtimeAnnotations.subscribe()',
+        err.message + '; hint=' + err.hint,
+      );
+      // The call is about to throw, so undo the listener registration above to avoid leaking a handler.
+      this.subscriptions.off(event, listener);
+      throw err;
     }
   }
 
--- a/src/common/lib/client/realtimeannotations.ts
+++ b/src/common/lib/client/realtimeannotations.ts
@@ -77,7 +77,7 @@ class RealtimeAnnotations {
         code: 93001,
         statusCode: 400,
         remediation:
-          'Enable the mode on the channel with channel.setOptions({ modes: ["subscribe", "annotation_subscribe"] }), which re-attaches with the new mode (calling channels.get(name, { modes }) on an existing channel throws, and appending to channel.modes does not enable it server-side). If the re-attach is rejected by the server, confirm the channel namespace has "Message annotations, updates, appends, and deletes" enabled in the Ably dashboard and that your API key has annotation-subscribe capability on this channel. If you have the Ably CLI installed, `ably apps rules list` shows which namespaces have it enabled and `ably auth keys list` shows the capabilities of your key.',
+          'Include "annotation_subscribe" in the channel modes: realtime.channels.get(name, { modes: ["subscribe", "annotation_subscribe", ...] }), or call channel.setOptions({ modes: [...] }) on an existing channel (this triggers a reattach). If the subsequent attach is rejected by the server, check that the channel namespace has "Message annotations, updates, appends, and deletes" enabled in the Ably dashboard and that your API key has annotation-subscribe capability on this channel. If you have the Ably CLI installed, `ably apps rules list` shows which namespaces have it enabled, and `ably auth keys list` shows your key\'s capabilities.',
       });
       Logger.logActionNoStrip(
         this.logger,
@@ -85,8 +85,8 @@ class RealtimeAnnotations {
         'RealtimeAnnotations.subscribe()',
         err.message + '; hint=' + err.hint,
       );
-      // The call is about to throw, so undo the listener registration above to avoid leaking a handler.
-      this.subscriptions.off(event, listener);
+      // The listener stays registered despite the throw, matching subscribe()'s existing
+      // semantics: the listener is always added regardless of attach outcome.
       throw err;
     }
   }
--- a/src/common/lib/client/realtimeannotations.ts
+++ b/src/common/lib/client/realtimeannotations.ts
@@ -77,7 +77,7 @@ class RealtimeAnnotations {
         code: 93001,
         statusCode: 400,
         remediation:
-          'Include "annotation_subscribe" in the channel modes: realtime.channels.get(name, { modes: ["annotation_subscribe", ...] }), or call channel.setOptions({ modes: [...] }) on an existing channel (this triggers a reattach). If the attach is then rejected, the channel namespace does not have "Message annotations, updates, appends, and deletes" enabled (`ably apps rules list` shows which namespaces do). If the attach succeeds but annotations still are not delivered, your API key lacks the annotation-subscribe capability and the server silently dropped the mode (`ably auth keys list` shows your key\'s capabilities).',
+          'Include "annotation_subscribe" in the channel modes: realtime.channels.get(name, { modes: ["annotation_subscribe", ...] }), or call channel.setOptions({ modes: [...] }) on an existing channel to trigger a reattach. Calling channels.get(name, { modes }) on an existing channel throws. The server only grants the mode if your key or token has the annotation-subscribe capability. Without it the attach succeeds but the server silently drops the mode and annotations are never delivered. `ably auth keys list` shows your key\'s capabilities.',
       });
       Logger.logActionNoStrip(
         this.logger,
--- a/src/common/lib/client/realtimechannel.ts
+++ b/src/common/lib/client/realtimechannel.ts
@@ -526,7 +526,8 @@ class RealtimeChannel extends EventEmitter {
           'The channel was attached without the subscribe mode, so the server will not deliver messages to this listener.',
         code: 90009,
         statusCode: 400,
-        hint: 'Include "subscribe" in the channel modes: realtime.channels.get(name, { modes: ["subscribe", ...] }), or call channel.setOptions({ modes: [...] }) on an existing channel to trigger a reattach. Alternatively, omit modes entirely and ensure your token/API-key capability permits subscribe on this channel. If you have the Ably CLI installed, `ably auth keys list` shows your key\'s capabilities.',
+        remediation:
+          'Include "subscribe" in the channel modes: realtime.channels.get(name, { modes: ["subscribe", ...] }), or call channel.setOptions({ modes: [...] }) on an existing channel to trigger a reattach. Alternatively, omit modes entirely and ensure your token/API-key capability permits subscribe on this channel. If you have the Ably CLI installed, `ably auth keys list` shows your key\'s capabilities.',
       });
       if (this.client.options.strictMode === true) {
         // The listener stays registered despite the throw, matching subscribe()'s existing
@@ -538,7 +539,7 @@ class RealtimeChannel extends EventEmitter {
           this.logger,
           Logger.LOG_ERROR,
           'RealtimeChannel.subscribe()',
-          err.message + '; hint=' + err.hint + Logger.silentFailureLogSuffix(),
+          err.message + '; remediation=' + err.remediation + Logger.silentFailureLogSuffix(),
         );
         this._silentSubscribeWarned = true;
       }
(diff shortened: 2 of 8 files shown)


umair-ably:
@coderabbitai review

umair-ably:
@coderabbitai review

src/common/lib/client/realtimechannel.ts
```diff
@@ -499,18 +500,51 @@ class RealtimeChannel extends EventEmitter {
...
-      return null;
+      stateChange = await this.attach();
+    }
+
+    // Whether or not we attached on subscribe, if the channel ended up attached without the
+    // subscribe mode the server will never deliver messages to this listener.
+    if (this.state === 'attached' && (this._mode & flags.SUBSCRIBE) === 0) {
+      const err = new ErrorInfo({
+        message:
+          'You called channel.subscribe() but the channel was attached without the subscribe mode, so the server will never deliver messages to this listener.',
```
AndyTWF: This message is a bit verbose - sounds a bit like being told off by a teacher, rather than technical 😅

src/common/lib/client/realtimechannel.ts
```diff
@@ -499,18 +500,51 @@ class RealtimeChannel extends EventEmitter {
...
+        code: 93003,
+        statusCode: 400,
+        hint: 'Re-create the channel with subscribe in modes: realtime.channels.get(name, { modes: ["subscribe", ...] }). Your token/API-key capability must permit subscribe on this channel. If you have the Ably CLI installed, `ably auth keys list` shows your key\'s capabilities. Note: appending to channel.modes after attach() does not enable the mode server-side - the array reflects what the server granted, not what you requested.',
```
AndyTWF: > the array reflects what the server granted, not what you requested

What array? The only array I see in the hint here is the requested one.

Do we want to direct people to `setOptions()`, or do we prefer the "know what you want in advance" approach?

src/common/lib/client/realtimechannel.ts
```diff
@@ -499,18 +500,51 @@ class RealtimeChannel extends EventEmitter {
...
+      });
+      if (this.client.options.strictMode === true) {
+        // The call is about to throw, so undo the listener registration above to avoid leaking a handler.
```
AndyTWF: This is a breaking change.

The current behaviour of `subscribe` (with the implicit attach) is that listener is always added regardless of failure.

src/common/lib/client/realtimepresence.ts
```diff
@@ -236,6 +237,24 @@ class RealtimePresence extends EventEmitter {
...
 
     await this.channel.ensureAttached();
+
+    if ((this.channel._mode & flags.PRESENCE_SUBSCRIBE) === 0) {
+      const err = new ErrorInfo({
+        message:
+          'You called presence.get() but the channel was attached without the presence_subscribe mode, so the server has not delivered any members to this client.',
+        code: 93002,
+        statusCode: 400,
+        hint: 'Re-create the channel with presence_subscribe in modes: realtime.channels.get(name, { modes: ["presence_subscribe", ...] }). Your token/API-key capability must permit presence-subscribe on this channel. If you have the Ably CLI installed, `ably auth keys list` shows your key\'s capabilities. Note: appending to channel.modes after attach() does not enable the mode server-side - the array reflects what the server granted, not what you requested.',
```
AndyTWF: Thought on the hints, should it be along the lines of "or ensure your key has the presence subscribe capability, and don't specify the modes explicitly"

test/realtime/channel.test.js
```diff
@@ -2114,5 +2114,42 @@ define(['ably', 'shared_helper', 'async', 'chai'], function (Ably, Helper, async
         await detachPromise;
       }, realtime);
     });
+
+    describe('DX-1211 - subscribe() without subscribe mode', function () {
```
AndyTWF: Test names shouldn't name tickets, what if the test ever changes, it'll make no sense

AndyTWF (changes requested):

src/common/lib/client/realtimechannel.ts
umair-ably: ahh i see - this was a coderabbit suggestion from earlier. Still seems strange the listener is added despite this erroring out? 

I've undone this to align with existing behaviour though

src/common/lib/client/realtimeannotations.ts
```diff
@@ -70,14 +70,23 @@ class RealtimeAnnotations {
...
+        hint: 'Include "annotation_subscribe" in the channel modes: realtime.channels.get(name, { modes: ["subscribe", "annotation_subscribe", ...] }), or call channel.setOptions({ modes: [...] }) on an existing channel (this triggers a reattach). If the subsequent attach is rejected by the server, check that the channel namespace has "Message annotations, updates, and deletes" enabled in the Ably dashboard and that your API key has annotation-subscribe capability on this channel. If you have the Ably CLI installed, `ably apps rules list` shows which channel namespaces have Mutable Messages enabled, and `ably auth keys list` shows your key\'s capabilities.',
```
AndyTWF: Im wondering if this hint is a bit presumptuous? Like, are there other reasons the next attach could be rejected? Or should this hint stop at what you need to do to fix the immediate problem and then let the next error hint at how to fix that? Do we risk confusing LLMs by throwing to many possibilities at them?

src/common/lib/client/realtimepresence.ts
```diff
@@ -236,6 +237,24 @@ class RealtimePresence extends EventEmitter {
...
 
     await this.channel.ensureAttached();
+
+    if ((this.channel._mode & flags.PRESENCE_SUBSCRIBE) === 0) {
+      const err = new ErrorInfo({
+        message:
+          'The channel was attached without the presence_subscribe mode, so the server has not delivered any members to this client.',
+        code: 93002,
+        statusCode: 400,
+        hint: 'Include "presence_subscribe" in the channel modes: realtime.channels.get(name, { modes: ["presence_subscribe", ...] }), or call channel.setOptions({ modes: [...] }) on an existing channel (this triggers a reattach). Alternatively, omit modes entirely and ensure your token/API-key capability permits presence-subscribe on this channel. If you have the Ably CLI installed, `ably auth keys list` shows your key\'s capabilities.',
```
AndyTWF: presence_subscribe isn't a capability we publish in our documentation. It exists on the server, but it's not something we expose as its a subset of `subscribe` which for legacy reasons gives message + presence + objects.

So that's a judgement call to make here as to whether this should become a public capability or not.

In any case, telling an LLM to do this right now would only confuse it further - as `ably.d.ts` doesnt list it as a valid capability, so the LLM would just see a typescript error

AndyTWF (changes requested):

src/common/lib/client/realtimepresence.ts
umair-ably: Good catch — fixed the capability, with one push-back on the mode.

**The real bug (fixed):** the hint named a `presence-subscribe` capability, which doesn't exist. Presence delivery is governed by the `subscribe` capability (the legacy "subscribe ⇒ messages + presence + objects" you describe), so it now reads "…capability permits **subscribe**", matching the sibling subscribe-mode hint.

**On `presence_subscribe` the _mode_ — I'd keep it:** it's a valid `ChannelMode` in `ably.d.ts` (doc-commented "The client will receive presence messages"), so `modes: ["presence_subscribe"]` typechecks — no TS error. And it's required: modes map 1:1 to flag bits (`encodeModesToFlags` OR-sets one bit per mode, no supersetting), and the integration tests prove it — `channel.test.js` attaches `modes: ['subscribe']` and asserts presence subscription is _denied_, and `['publish', 'presence_subscribe']` and asserts message subscribe is denied. So `subscribe` doesn't imply `presence_subscribe`; when a user has restricted `modes` it's the only client-side way to get the flag. The hint leads with "omit modes entirely" for the common case, so we only steer toward `presence_subscribe` when someone has explicitly restricted modes. Whether to document it more prominently is a fair product question, but the hint needs no new public surface.

**Also reassigned the code:** while checking this I found `93002` is the canonical _server_ code for "namespace needs Mutable Messages" (ably-common `errors.json` + `faqs.ably.com/error-code-93002`), so reusing it client-side for presence was a collision. `presence.get()` now throws **`91008`** — the presence block, next to `91005` which `get()` already throws for the suspended case. Reserved in ably/ably-common#345.

— 🤖 Drafted with [Claude Code](https://claude.com/claude-code) (Claude Opus 4.8, 1M context)

src/common/lib/client/realtimeannotations.ts
umair-ably: Agree it was over-reaching — and it was partly wrong. The two things the old tail told you to check "if the attach is rejected" actually fail _differently_ (confirmed by live sandbox traces + the canonical registry):

- **Missing namespace rule** + an explicit `annotation_subscribe` mode → server **rejects the attach** (channel → `failed`, server code `93002` "…namespace with Mutable Messages enabled").
- **Missing `annotation-subscribe` capability** → **no rejection**; the server silently drops the mode, the attach succeeds, and this same `93001` re-fires.

So "if the attach is rejected, check the namespace _and_ the capability" conflated a reject path with a downgrade path. Reworded to split by the symptom you actually see:

> …If the attach is then rejected, the channel namespace does not have "Message annotations, updates, and deletes" enabled (`ably apps rules list` …). If the attach succeeds but annotations still are not delivered, your API key lacks the annotation-subscribe capability and the server silently dropped the mode (`ably auth keys list` …).

On "let the next error hint at the fix" — I'd like to, but the downstream rejection is currently **unhinted**: server errors come through `ErrorInfo.fromValues`, which only adds an `href`, never a `hint`. So the cause has to live here for now. If we add a hint to the server's `93002` rejection (a separate change), we could then trim this branch to just the immediate fix — happy to file that follow-up.

— 🤖 Drafted with [Claude Code](https://claude.com/claude-code) (Claude Opus 4.8, 1M context)

umair-ably:
Pushed a round addressing both review comments, plus a related fix surfaced while verifying them.

**The two threads:**
- **presence `presence_subscribe` capability** — fixed the capability name (`presence-subscribe` → `subscribe`); kept the `presence_subscribe` *mode* (it's a valid `ChannelMode`, typechecks, and is required — `subscribe` doesn't imply it). Details in-thread.
- **annotation hint presumptuousness** — reworded to split the two failure modes by symptom (attach rejected → enable the namespace rule; attach succeeds but still nothing → grant the capability), instead of lumping both under "if the attach is rejected". Details in-thread.

**Error-code collision (surfaced while verifying the above):** `93002` is the canonical *server* code for "namespace needs Mutable Messages" (`faqs.ably.com/error-code-93002`), and `presence.get()` was reusing it client-side. Moved the two new codes to their subsystem blocks:

| Call | was | now |
| --- | --- | --- |
| `presence.get()` without `presence_subscribe` | `93002` (collides) | **`91008`** (presence block) |
| `channel.subscribe()` without `subscribe` | `93003` | **`90009`** (channel block) |

Both are new on this unreleased branch, so the renumber is non-breaking. `93001` (annotation — already released) is unchanged.

⚠️ **Dependency:** these codes are reserved in ably/ably-common#345 — that should land before this is released so they're unique platform-wide.

— 🤖 Drafted with [Claude Code](https://claude.com/claude-code) (Claude Opus 4.8, 1M context)

src/common/lib/client/realtimeannotations.ts
AndyTWF: What's the status of getting `hint` or whatever we decide to call it into Realtime? That's a trivial PR when it comes to it, so would it not be better to solve it properly first time rather than adding technical debt we have to undo later?

umair-ably: added a ticket so we don't lose track of this https://ably.atlassian.net/browse/DX-1460

AndyTWF (approved):

PR #2239, opened 2026-06-08
[DX-1204] Fix v1-callback overloads breaking mocks (never → void)

umair-ably:
## Background

[#2226](https://github.com/ably/ably-js/pull/2226) (DX-1204) re-introduced the v1 callback-style method signatures as `@deprecated` overloads returning `never`, to steer users/LLMs toward the v2 promise API. The **runtime** guard (`detectV1Callback`, which throws a hinted error) is the part the PR's evidence measured, and it is **not changed here**.

@owenpearson reported that the `never` return type makes these methods impossible to mock by assignment — a common testing pattern:

```ts
channel.subscribe = vi.fn(() => Promise.reject()); // ❌ doesn't type-check on main
```

This fails because the assigned value must satisfy every overload, including the one returning `never`, and nothing is assignable to a `never`-returning signature.

## Fix

Change the return type of the 14 deprecated v1-callback overloads from `never` → `void` (auth `authorize`/`createTokenRequest`/`requestToken`, presence `get`/`history`/`subscribe`/`enter`/`update`/`leave`, channel `unsubscribe`/`history`/`subscribe`/`publish`, `connection.ping`).

TypeScript allows a function returning any value to be assigned to a `void`-returning function type (the rule that makes `arr.forEach(x => arr.push(x))` legal), so mock assignment now type-checks. `never` was the only return type that rejected it. (`void` also matches what these methods returned in v1.)

Diff is 14 lines — `): never;` → `): void;` — nothing else.

## Intentionally out of scope: the `no-deprecated` warning on bare references

Owen also noted that `@typescript-eslint/no-deprecated` flags bare references like `expect(channel.subscribe)`. We are **not** addressing that here. It's arguably an upstream issue: for a non-call reference the rule can't resolve a specific overload, so it treats the symbol as deprecated if *any* overload is. Removing the `@deprecated` tag would silence it but would also drop the editor strikethrough + migration hint on v1 call sites, which we want to keep. Consumers hitting the warning can disable the rule on those lines until it's fixed upstream.

## Verification

- `npx tsc --noEmit ably.d.ts modular.d.ts` (the `check` CI gate), `prettier --check`, and `eslint` all pass.
- Mock assignment compiles for all 14 methods (verified against the real `ably.d.ts` with `vitest` and `@types/jest`): `vi.fn(...)`, `vi.fn()`, `jest.fn(...)`, `jest.spyOn`, direct assignment.
- The `@deprecated` strikethrough + migration hint on v1 call sites are unchanged.
- Runtime `detectV1Callback` guard and its tests are unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected TypeScript type definitions for deprecated callback-based method overloads affecting authentication, realtime presence, channels, and connections. Return type annotations have been updated to more accurately reflect method behavior, enhancing type safety, improving IDE autocompletion accuracy, and reducing type-related issues for TypeScript developers using callback-style APIs.

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -2008,7 +2008,7 @@ export declare interface Auth {
     tokenParams: TokenParams | null,
     authOptions: AuthOptions | null,
     callback: StandardCallback<TokenDetails>,
-  ): never;
+  ): void;
   /**
    * Creates and signs an Ably {@link TokenRequest} based on the specified (or if none specified, the client library stored) {@link TokenParams} and {@link AuthOptions}. Note this can only be used when the API `key` value is available locally. Otherwise, the Ably {@link TokenRequest} must be obtained from the key owner. Use this to generate an Ably {@link TokenRequest} in order to implement an Ably Token request callback for use by other clients. Both {@link TokenParams} and {@link AuthOptions} are optional. When omitted or `null`, the default token parameters and authentication options for the client library are used, as specified in the {@link ClientOptions} when the client library was instantiated, or later updated with an explicit `authorize` request. Values passed in are used instead of, rather than being merged with, the default values. To understand why an Ably {@link TokenRequest} may be issued to clients in favor of a token, see [Token Authentication explained](https://ably.com/docs/core-features/authentication/#token-authentication).
    *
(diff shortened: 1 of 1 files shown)


ably.d.ts
```diff
@@ -1991,10 +1991,11 @@ export declare interface Auth {
    */
   authorize(tokenParams?: TokenParams, authOptions?: AuthOptions): Promise<TokenDetails>;
   /**
-   * @deprecated v1 callback signature — no longer supported. Use {@link Auth.authorize | `realtime.auth.authorize(tokenParams)`} and `await` the returned promise. See [the v2 migration guide](https://github.com/ably/ably-js/blob/main/docs/migration-guides/v2/lib.md).
```
owenpearson, line 1994: I really don't like this as a solution, now there's no typescript compiler errors at all for using v1 apis, which is surely a worse position than we were in previously - agents will only discover the error at runtime, and for a web app launched by the user the error will only be visible to the user

owenpearson (changes requested):
the change to `void` is good! i don't like the other one though, sorry:

ably.d.ts
umair-ably: stripped as discussed on slack

owenpearson (approved):

PR #2241, opened 2026-06-09
[ECO-5728] Fix presence auto reenter causes NACKs with "not currently attached" after reconnect from transient disconnect

### LINKED ISSUE #2220: Presence messages NACK'd with "not currently attached" after reconnect from transient disconnect
**Summary**

When a client loses connectivity and a presence operation is invoked while DISCONNECTED, the presence message is queued on the connection-wide queue (RTL6c2). On reconnect, RTL3d flushes that queue immediately, before the channel finishes re-attaching. The server then rejects the presence message with:

`Unable to perform operation on channel: <channel> (not currently attached)`

This is also a likely cause of the intermittent same-error reports originating from the RTP17i auto re-entry path seen in customer Sentry logs.

**Reproduction**

1. Connect and attach to a channel.
2. Force the tab/network offline - connection goes DISCONNECTED (RTN20a); channel remains ATTACHED.
3. While offline, call a presence op (e.g. enterClient). Because the channel is still ATTACHED, RTP16b does not channel-queue it; instead it goes onto the connection-wide queue per RTL6c2.
4. Restore connectivity (RTN20b). Connection resumes (RTN15c6) or re-connects (RTN15c7).
5. Per RTL3d, queued messages are flushed immediately without waiting for channel re-attach.
6. Server NACKs the presence message because the channel is not (yet) attached server-side.

**Root cause**

On the server, presence and sync messages require a real (non-transient) attachment; messages, objects and annotations don't (a transient attach is created for them). So presence is uniquely sensitive to being sent while the channel isn't actually attached server-side.

This was an oversight in the protocol v2 redesign: in v1, channels remained ATTACHED server-side across a connection recovery, so flushing the queue immediately was safe. In v2 that's no longer guaranteed.

**Proposed fixes (one of)**

1. Channel state on disconnect - transition ATTACHED channels to ATTACHING when the connection goes DISCONNECTED. RTP16b would then naturally queue presence at the channel level until re-attach completes. Cleaner, but higher blast radius - needs careful review of lifecycle implications.
2. Presence-aware queueing - RealtimePresence checks the connection state directly and routes to pendingPresence (channel-level queue) whenever the connection isn't CONNECTED, regardless of channel state. Smaller change; slight smell that channel reaches into connection state.

Spec also needs updating: either RTL3d should exclude presence from immediate flush, or the channel-state model needs to change as above.

Related

- Spec items: RTL3d, RTL6c2, RTP16b, RTP17i, RTN15c6/c7, RTN20a/b, RTP8d, RTP8g
- Original Slack threads: https://ably-real-time.slack.com/archives/CURL4U2FP/p1755856188346249, https://ably-real-time.slack.com/archives/C8SPU4589/p1778756591487359

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-5728) by [Unito](https://www.unito.io)

VeskeR:
A presence operation invoked while the connection was transiently DISCONNECTED
(with the channel still ATTACHED) was placed on the connection-wide queue, which
RTL3d flushes on reconnect before the channel re-attaches. The server then
rejected it with "not currently attached" (code 90001), because presence
requires a real, non-transient channel attachment, unlike regular messages.

Now, when the channel is ATTACHED but the connection is not CONNECTED, presence
is queued at the channel level (RTP16d2) and sent once the channel next reaches
ATTACHED (RTP5b), honoring queueMessages.

Two commits:
1. Honor `queueMessages` for channel-level presence queueing (RTP16b), a
   pre-existing gap fixed first.
2. Queue presence at the channel level when the connection is not CONNECTED.

Companion spec PR: https://github.com/ably/specification/pull/490.

Resolves [ECO-5728](https://ably.atlassian.net/browse/ECO-5728)

[ECO-5728]: https://ably.atlassian.net/browse/ECO-5728?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Presence enter/update/leave now fully respect per-channel `queueMessages`: when disabled, presence actions made before attachment/re-attachment are rejected instead of queued.
  * Pending presence operations are re-queued and delivered after a successful re-attach, improving ordering and membership consistency during reconnects.
* **Documentation**
  * Updated `presenceEnterWithoutAttach` docs to clarify queueing behavior when `queueMessages` is enabled by default.
* **Tests**
  * Added coverage for `queueMessages: false` rejection and for ensuring queued presence transmits only after post-reconnect attachment, with membership verification.

### DIFF
--- a/src/common/lib/client/realtimepresence.ts
+++ b/src/common/lib/client/realtimepresence.ts
@@ -396,6 +396,17 @@ class RealtimePresence extends EventEmitter {
     }
   }
 
+  /**
+   * RTL3d: re-queue presence messages that were moved off the connection-wide
+   * queue onto this channel's presence queue, to be sent once the channel next
+   * reaches the ATTACHED state (RTP5b).
+   */
+  requeuePresenceMessages(presenceMessages: WirePresenceMessage[], callback: ErrCallback): void {
+    for (const presence of presenceMessages) {
+      this.pendingPresence.push({ presence, callback });
+    }
+  }
+
   _clearMyMembers(): void {
     this._myMembers.clear();
   }
--- a/src/common/lib/client/realtimepresence.ts
+++ b/src/common/lib/client/realtimepresence.ts
@@ -120,15 +120,21 @@ class RealtimePresence extends EventEmitter {
         return channel.sendPresence([wirePresMsg]);
       case 'initialized':
       case 'detached':
+        // RTP8d: implicitly attach the channel
         channel.attach();
       // eslint-disable-next-line no-fallthrough
       case 'attaching':
-        return new Promise((resolve, reject) => {
-          this.pendingPresence.push({
-            presence: wirePresMsg,
-            callback: (err) => (err ? reject(err) : resolve()),
+        // RTP16b: queue at the channel level only when queueMessages is enabled
+        if (channel.client.options.queueMessages) {
+          return new Promise((resolve, reject) => {
+            this.pendingPresence.push({
+              presence: wirePresMsg,
+              callback: (err) => (err ? reject(err) : resolve()),
+            });
           });
-        });
+        }
+      // RTP16c: not queueable, so reject
+      // eslint-disable-next-line no-fallthrough
       default: {
         const err = new PartialErrorInfo(
           'Unable to ' + action + ' presence channel while in ' + channel.state + ' state',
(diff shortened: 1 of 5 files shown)


src/common/lib/client/realtimepresence.ts
```diff
@@ -114,20 +114,34 @@ class RealtimePresence extends EventEmitter {
...
+    if (channel.state === 'attached' && channel.connectionManager.state.state === 'connected') {
+      return channel.sendPresence([wirePresMsg]);
+    }
+
     switch (channel.state) {
-      case 'attached':
-        return channel.sendPresence([wirePresMsg]);
       case 'initialized':
       case 'detached':
-        channel.attach();
```
ttypic, line 123: it is a bit vague in the spec, I believe right now spec implies that we should always attach on `presence.enter` even if the message is not queued and will result in error. It's a behavioral breaking change otherwise

ttypic (commented):
Looks good, but I think we should add some `ConnectionManager` handling as well. Right now, if the connection drops during a sendPresence call and we don’t receive an ack, those messages are queued in the ConnectionManager-level queue. I believe we should filter them out and move them to the channel-level queue on reconnect. Otherwise, we may still send those nacks.

This is a much less likely scenario, so we could potentially skip it, but I think it’s worth addressing.

src/common/lib/client/realtimepresence.ts
VeskeR: Oh, I see, I went a bit overaggressive here I think. The implicit attach on enter is covered in RTP8, not in RTP16. Will revert this bit.
But I assume the change below to check for `queueMessages` before adding to `pendingPresence` is ok in your mind? It's explicitly covered by RTP16b

ttypic (approved):
LGTM

PR #2242, opened 2026-06-09
DX-1211: Auth interface docstrings (prerequisites, side-effects, failure modes)

umair-ably:
> **Stacked on `DX-1211/silent-failure-hints`.** Review the diff against that base, not `main`. Merge the parent first.

Applies `docstringRules.md` to the public `Auth` interface in `ably.d.ts`, surfacing the call-site prerequisites and failure modes a caller cannot infer from the type signature (the DX-1211 class of silent/architectural pitfall). Same effort as the already-landed RealtimeChannel/RealtimePresence pass; one subsystem = one PR.

### Members (5 audited; 3 deprecated v1-callback overloads skipped)

| Member | What was surfaced |
|---|---|
| `revokeTokens` | basic-auth (API key, **not** token) requirement; non-code "revocable tokens enabled on the key" prerequisite (+ feature-page `@see`) |
| `requestToken` | token-issuing prerequisite (`authCallback`/`authUrl`/`key`); callback-contract detail offloaded to the `@see` |
| `createTokenRequest` | a local API `key` must be available to sign the request |
| `authorize` | re-auth side-effect (resolves once the token takes effect on a `connected` connection); token-issuing prerequisite; **RSA10a key-immutability** (`authorize()` cannot change the API key) |
| `clientId` | adds the canonical `@see` |

### Conventions
- Folded prose, no headings/labels, no em-dashes, **no numeric error codes** (failures framed as "rejects with an `{@link ErrorInfo}`").
- Every member carries an `@see` to the canonical JS API reference.
- Every claim re-derived from `src/` (`file:line`) in an adversarial verify pass before applying.

### Validation
- `npm run docs` (TypeDoc, `treatWarningsAsErrors`) — clean (all `{@link}` resolve)
- `eslint ably.d.ts` — exit 0 · `prettier --check` — clean

### Reviewer notes / follow-ups
- **`@see` anchors:** the canonical reference (ably/docs #3400) is not yet published, so 4 of 5 anchors are convention-derived per §6 and unverified (`#revoke-tokens` is corroborated). `@see` is not validated by TypeDoc, so re-check the slugs when #3400 ships.
- **`clientId`** change is cosmetic (`@see` only) — the drafter's `null`/wildcard value claims were rejected by verification as inaccurate. Fine to drop if you'd rather treat it as a pure accessor.

## Summary by CodeRabbit

* **Documentation**
  * Clarified how client identity (`clientId`) is resolved and how anonymous/unset values behave.
  * Updated `authorize()` guidance on when it can be used and which connection state is required.
  * Tightened token request documentation to match token-auth mechanisms and API-key availability for signing/token creation.
  * Expanded `revokeTokens()` notes to specify API-key/basic authentication and revocation eligibility rules.

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -1999,7 +1999,11 @@ export declare interface RealtimeClient {
  */
 export declare interface Auth {
   /**
-   * A client ID, used for identifying this client when publishing messages or for presence purposes. The `clientId` can be any non-empty string, except it cannot contain a `*`. This option is primarily intended to be used in situations where the library is instantiated with a key. Note that a `clientId` may also be implicit in a token used to instantiate the library. An error is raised if a `clientId` specified here conflicts with the `clientId` implicit in the token.
+   * The client ID currently in effect for this client, used to identify it when publishing messages or entering presence.
+   *
+   * The value is resolved from the `clientId` in {@link ClientOptions} or from the `clientId` implicit in the token in use. A conflict between the two raises an {@link ErrorInfo}.
+   *
+   * The value is unset for an anonymous client, for example a key-authenticated client with no `clientId` configured. Guard against an unset value despite the declared type.
    *
    * @see https://ably.com/docs/pub-sub/api/javascript/realtime/auth#client-id
    */
@@ -2008,13 +2012,13 @@ export declare interface Auth {
   /**
    * Instructs the library to get a new token immediately.
    *
-   * On a realtime client it re-authenticates the live connection, or initiates a connection if not currently connected. The returned promise resolves only once the new token has taken effect on a `connected` connection. It rejects with an {@link ErrorInfo} if re-authentication fails or the connection cannot be (re)established.
+   * On a realtime client it re-authenticates the live connection, or initiates a connection if not currently connected. The returned promise resolves only once the new token has taken effect on a `connected` connection. It rejects with an {@link ErrorInfo} if re-authentication fails or the connection cannot be established.
    *
-   * The client must be able to issue tokens, so a `key`, `authUrl`, or `authCallback` must be configured in {@link ClientOptions}. If one of these is not configured the call rejects with an {@link ErrorInfo}.
+   * The client must have a way to obtain a token, so the resolved {@link AuthOptions} must include one of `authCallback`, `authUrl`, or `key`, or supply a token directly. Without any of these the call rejects with an {@link ErrorInfo}.
    *
    * `authorize()` cannot change the API key, so passing an `authOptions.key` that differs from the one the client was constructed with is rejected with an {@link ErrorInfo}.
    *
-   * Any {@link TokenParams} and {@link AuthOptions} passed in are stored as the new defaults for all subsequent token requests and entirely replace, rather than merge with, the current saved values.
+   * Any {@link TokenParams} and {@link AuthOptions} passed in are stored as the new defaults for subsequent token requests. They replace, rather than merge with, the stored defaults.
    *
    * @param tokenParams - A {@link TokenParams} object.
    * @param authOptions - An {@link AuthOptions} object.
@@ -2046,7 +2050,7 @@ export declare interface Auth {
     callback: StandardCallback<TokenDetails>,
   ): void;
   /**
-   * Creates and signs an Ably {@link TokenRequest} based on the specified {@link TokenParams} and {@link AuthOptions}. If none are specified it uses those previously stored by the library. Use this to implement an Ably Token request callback for use by other clients.
+   * Creates and signs an Ably {@link TokenRequest} based on the specified {@link TokenParams} and {@link AuthOptions}. Use this to implement an Ably Token request callback for use by other clients.
    *
    * An API `key` value must be available locally to sign the request, supplied either in the client's {@link ClientOptions} or as `key` in the `authOptions` argument. Without a `key` the call rejects with an {@link ErrorInfo}, since a token-authenticated client cannot construct token requests itself and must instead obtain the {@link TokenRequest} from the key owner.
    *
@@ -2082,9 +2086,11 @@ export declare interface Auth {
     callback: StandardCallback<TokenRequest>,
   ): void;
   /**
-   * Calls the `requestToken` REST API endpoint to obtain an Ably Token according to the specified {@link TokenParams} and {@link AuthOptions}. Both {@link TokenParams} and {@link AuthOptions} are optional. When omitted or `null`, the client's stored defaults are used, as specified at instantiation or later updated by an `authorize()` request. Any values passed in replace, rather than merge with, those defaults.
+   * Obtains an Ably Token according to the specified {@link TokenParams} and {@link AuthOptions}.
    *
-   * The client must have a way to obtain a token, so the resolved {@link AuthOptions} must include one of `authCallback`, `authUrl`, or `key`. A client given only a literal token or `tokenDetails` with no renewal mechanism cannot request a new token and the call rejects with an {@link ErrorInfo}.
+   * Both {@link TokenParams} and {@link AuthOptions} are optional. When omitted or `null`, the client's stored defaults are used, as specified at instantiation or later updated by an `authorize()` request. Any values passed in replace, rather than merge with, those defaults.
+   *
+   * The client must have a way to obtain a token, so the resolved {@link AuthOptions} must include one of `authCallback`, `authUrl`, or `key`. A client given only a literal token cannot request a new one and the call rejects with an {@link ErrorInfo}.
    *
    * @param TokenParams - A {@link TokenParams} object.
    * @param authOptions - An {@link AuthOptions} object.
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -2126,7 +2126,7 @@ export declare interface Auth {
    *
    * The client making this call must be authenticated with an API key (basic auth), not a token. A token-authenticated client cannot revoke tokens and the call rejects with an {@link ErrorInfo}.
    *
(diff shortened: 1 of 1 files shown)


ably.d.ts
```diff
@@ -1986,15 +1986,22 @@ export declare interface RealtimeClient {
...
+   * Instructs the library to get a new token immediately. On a realtime client it re-authenticates the live connection, or initiates a connection if not connected, and the returned promise resolves only once the new token has taken effect on a `connected` connection, rejecting with an {@link ErrorInfo} if re-authentication fails or the connection cannot be (re)established. The client must be able to issue tokens, so a `key`, `authUrl`, or `authCallback` must be configured in {@link ClientOptions}, otherwise the call rejects with an {@link ErrorInfo}; `authorize()` cannot change the API key, so passing an `authOptions.key` that differs from the one the client was constructed with also rejects. Any {@link TokenParams} and {@link AuthOptions} passed in are stored as the new defaults for all subsequent token requests and entirely replace, rather than merge with, the current saved values.
```
m-hulbert:
```suggestion
   * Instructs the library to get a new token immediately.
   * 
   * On a realtime client it re-authenticates the live connection, or initiates a connection if not currently connected. The returned promise resolves only once the new token has taken effect on a `connected` connection. It rejects with an {@link ErrorInfo} if re-authentication fails or the connection cannot be (re)established. 
   * 
   * The client must be able to issue tokens, so a `key`, `authUrl`, or `authCallback` must be configured in {@link ClientOptions}. If one of these isn't configured the call rejects with an {@link ErrorInfo}.
   *
   * `authorize()` cannot change the API key, so passing an `authOptions.key` that differs from the one the client was constructed with will be rejected. 
   * 
   * Any {@link TokenParams} and {@link AuthOptions} passed in are stored as the new defaults for all subsequent token requests and entirely replace, rather than merge with, the current saved values.
```

AFAIK you can create paragraphs using empty lines, which helps break up this wall of text. I'd also push for shorter sentences where possible as some of those got quite convoluted.

ably.d.ts
```diff
@@ -2017,11 +2024,16 @@ export declare interface Auth {
...
+   * Creates and signs an Ably {@link TokenRequest} based on the specified (or if none specified, the client library stored) {@link TokenParams} and {@link AuthOptions}. Use this to implement an Ably Token request callback for use by other clients. An API `key` value must be available locally to sign the request, supplied either in the client's {@link ClientOptions} or as `key` in the `authOptions` argument; without one the call rejects with an {@link ErrorInfo}, since a token-authenticated client cannot construct token requests itself and must instead obtain the {@link TokenRequest} from the key owner. Both {@link TokenParams} and {@link AuthOptions} are optional; when omitted or `null`, the client library's stored defaults (those set at construction or by a later `authorize` call) are used, and any values passed in replace, rather than merge with, those defaults.
```
m-hulbert:
```suggestion
   * Creates and signs an Ably {@link TokenRequest} based on the specified {@link TokenParams} and {@link AuthOptions}. If no `TokenRequest` or `TokenParams` are specified it uses those previously stored by the library. Use this to implement an Ably Token request callback for use by other clients. 
   * 
   * An API `key` value must be available locally to sign the request, supplied either in the client's {@link ClientOptions} or as `key` in the `authOptions` argument.  Without a `key` the call rejects with an {@link ErrorInfo}, since a token-authenticated client cannot construct token requests itself and must instead obtain the {@link TokenRequest} from the key owner. 
   * 
   * Both {@link TokenParams} and {@link AuthOptions} are optional. When omitted or `null`, the client library's stored defaults (those set at construction or by a later `authorize()` call) are used, and any values passed in replace, rather than merge with, those defaults.
```

ably.d.ts
```diff
@@ -2044,11 +2056,16 @@ export declare interface Auth {
...
+   * Calls the `requestToken` REST API endpoint to obtain an Ably Token according to the specified {@link TokenParams} and {@link AuthOptions}. Both are optional; when omitted or `null`, the client's stored defaults are used, as specified in the {@link ClientOptions} at instantiation or later updated by an `authorize` request, and any values passed in are used instead of, rather than merged with, those defaults. The client must have a usable way to obtain a token, so the resolved {@link AuthOptions} must include one of `authCallback`, `authUrl`, or `key`; a client given only a literal token or `tokenDetails` with no renewal mechanism cannot request a new token and the call rejects with an {@link ErrorInfo}.
```
m-hulbert:
```suggestion
   * Calls the `requestToken` REST API endpoint to obtain an Ably Token according to the specified {@link TokenParams} and {@link AuthOptions}. Both properties are optional. When omitted or `null`, the client's stored defaults are used, as specified in the {@link ClientOptions} at instantiation or later updated by an `authorize()` request. Any values passed in are used instead of, rather than merged with, those defaults.
   *
   * The client must have a way to obtain a token, so the resolved {@link AuthOptions} must include one of `authCallback`, `authUrl`, or `key`. A client given only a literal token or `tokenDetails` with no renewal mechanism cannot request a new token and the call rejects with an {@link ErrorInfo}.
```

On the 2nd paragraph I've made here - unsure if those 2 sentences are supposed to be related. They were separated by a semi-colon but I'm not sure they are.

Small note here on inconsistency between the descriptions too in how we refer to the default values and `authorize()`. 

Could argue whether the user needs to know that it's calling the REST API, but I won't go and rewrite everything as that's not the point 🙂

ably.d.ts
```diff
@@ -2071,11 +2088,21 @@ export declare interface Auth {
     callback: StandardCallback<TokenDetails>,
   ): never;
   /**
-   * Revokes the tokens specified by the provided array of {@link TokenRevocationTargetSpecifier}s. Only tokens issued by an API key that had revocable tokens enabled before the token was issued can be revoked. See the [token revocation docs](https://ably.com/docs/core-features/authentication#token-revocation) for more information.
+   * Revokes the tokens specified by the provided array of {@link TokenRevocationTargetSpecifier}s.
+   * The client must be authenticated with an API key (basic auth), not a token; a
```
m-hulbert: This reads a bit confusing as it makes it sound like the client you're trying to revoke a token _for_ needs to be authenticated by a key.

m-hulbert (commented):
Just a few suggestions on splitting up long sentences and large walls of text into paragraphs, but see what you think.

ably.d.ts
```diff
@@ -1985,16 +1985,35 @@ export declare interface RealtimeClient {
  */
 export declare interface Auth {
   /**
-   * A client ID, used for identifying this client when publishing messages or for presence purposes. The `clientId` can be any non-empty string, except it cannot contain a `*`. This option is primarily intended to be used in situations where the library is instantiated with a key. Note that a `clientId` may also be implicit in a token used to instantiate the library. An error is raised if a `clientId` specified here conflicts with the `clientId` implicit in the token. Find out more about [identified clients](https://ably.com/docs/core-features/authentication#identified-clients).
+   * The client ID currently in effect for this client, used to identify it when publishing messages or entering presence.
```
m-hulbert: I'm not sure about 'in effect' here.

ably.d.ts
```diff
@@ -1985,16 +1985,35 @@ export declare interface RealtimeClient {
...
+   *
```
m-hulbert, line 2003: 'token in use' or more like 'the token that authenticated this client' or the wording from before?

ably.d.ts
```diff
@@ -1985,16 +1985,35 @@ export declare interface RealtimeClient {
...
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/auth#client-id
    */
   clientId: string;
 
   /**
-   * Instructs the library to get a new token immediately. When using the realtime client, it upgrades the current realtime connection to use the new token, or if not connected, initiates a connection to Ably, once the new token has been obtained. Also stores any {@link TokenParams} and {@link AuthOptions} passed in as the new defaults, to be used for all subsequent implicit or explicit token requests. Any {@link TokenParams} and {@link AuthOptions} objects passed in entirely replace, as opposed to being merged with, the current client library saved values.
+   * Instructs the library to get a new token immediately.
+   *
+   * On a realtime client it re-authenticates the live connection, or initiates a connection if not currently connected. The returned promise resolves only once the new token has taken effect on a `connected` connection. It rejects with an {@link ErrorInfo} if re-authentication fails or the connection cannot be established.
```
m-hulbert: Is "live connection" the right terminology here? We're also mixing code-styled and plain text for states.

ably.d.ts
```diff
@@ -2044,11 +2072,20 @@ export declare interface Auth {
     callback: StandardCallback<TokenRequest>,
   ): void;
   /**
-   * Calls the `requestToken` REST API endpoint to obtain an Ably Token according to the specified {@link TokenParams} and {@link AuthOptions}. Both {@link TokenParams} and {@link AuthOptions} are optional. When omitted or `null`, the default token parameters and authentication options for the client library are used, as specified in the {@link ClientOptions} when the client library was instantiated, or later updated with an explicit `authorize` request. Values passed in are used instead of, rather than being merged with, the default values. To understand why an Ably {@link TokenRequest} may be issued to clients in favor of a token, see [Token Authentication explained](https://ably.com/docs/core-features/authentication/#token-authentication).
+   * Obtains an Ably Token according to the specified {@link TokenParams} and {@link AuthOptions}.
```
m-hulbert, line 2089: Thoughts on whether some of these should retain a link to the different methods of authentication?

umair-ably:
Addressed previous outstanding feedback and also cross-checked to ensure the @see links are live and working

m-hulbert (approved):

PR #2243, opened 2026-06-09
DX-1206 RealtimeChannel interface docstrings

umair-ably:
Stacked on #2242 (Auth docstrings, base `DX-1211/auth-docstrings`); review/merge that first. Rewrites the JSDoc on the `RealtimeChannel` public surface in `ably.d.ts` per `docstringRules.md`: terse folded prose that surfaces call-site prerequisites, side-effects, and failure modes (the DX-1211 silent-failure class), with an `@see` companion link to the canonical reference.

## Scope
`ably.d.ts` only (+135/−29). 19 members rewritten, 2 deliberately left as-is.

## Highlights
- **`subscribe()` (DX-1211 core):** documents the silent missing-`subscribe`-mode trap — the listener never fires, the SDK warn-logs rather than rejecting (or rejects with a hinted `ErrorInfo` under `strictMode`). On all four active overloads.
- **`params` / `errorReason`:** declared non-optional but actually `undefined`/`null` until first attach/error — now carry a guard-against-undefined/null caveat (parallel to `modes`).
- **`history()`:** non-code persistence channel-rule prerequisite; fixes a `@param` copy-paste bug ("presence members" → "messages").
- **Message ops** (`getMessage`/`updateMessage`/`deleteMessage`/`appendMessage`/`getMessageVersions`): async **reject** framing (not synchronous throw) + message-interactions channel-rule prerequisite.
- **`push`:** universal missing-plugin throw kept at the call site; `presence`/`annotations` left unchanged (bundled by the default build, so the throw is modular-only).

## Validation
`npm run docs` (TypeDoc `treatWarningsAsErrors`), `prettier --check`, and `eslint` all pass. A clean TypeDoc build confirms every `{@link}` resolves and nothing became undocumented.

## Reviewer notes
- `@see` anchors follow the kebab-case convention (matching the Auth PR) but are unverified — ably/docs #3400 isn't live yet and `@see` is not TypeDoc-validated.
- A secondary feature-page `@see` for the five message-interaction members was deliberately omitted pending a confirmed slug.

## Summary by CodeRabbit

* **Documentation**
  * Clarified message lifecycle behavior, including requirements and semantics for updating, deleting, appending, and retrieving message versions.
  * Expanded realtime channel documentation for states, options, subscriptions, attachment, history, push notifications, and promise behavior.
  * Documented runtime cases where certain channel properties may be unavailable.
  * Clarified that unsubscribing removes local listeners without detaching the channel.

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -2820,9 +2820,7 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
    */
   presence: RealtimePresence;
   /**
-   * A {@link PushChannel} object that manages device push notification subscriptions for this channel.
-   *
-   * Accessing this property requires the `Push` plugin to be registered via {@link ClientOptions.plugins}. If the plugin is absent, the getter throws an {@link ErrorInfo}.
+   * A {@link PushChannel} object that manages device push notification subscriptions for this channel. Accessing this property requires the `Push` plugin to be registered via {@link ClientOptions.plugins}. The default and modular builds do not bundle the `Push` plugin. If the plugin is absent, the getter throws an {@link ErrorInfo} rather than returning a {@link PushChannel}.
    *
    * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#push
    */
@@ -2866,7 +2864,7 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
   /**
    * Retrieves a {@link PaginatedResult} object, containing an array of historical {@link InboundMessage} objects for the channel.
    *
-   * Messages are returned from storage only when message persistence is enabled for the channel by a rule. If message persistence is not enabled, only messages from the last two minutes are returned.
+   * Messages are returned from storage only when message persistence is enabled for the channel by a [rule](https://ably.com/docs/channels#rules). If message persistence is not enabled, only messages from the last two minutes are returned.
    *
    * @param params - A set of parameters which are used to specify which messages should be retrieved.
    * @returns A promise which, upon success, will be fulfilled with a {@link PaginatedResult} object containing an array of {@link InboundMessage} objects. Upon failure, the promise will be rejected with an {@link ErrorInfo} object which explains the error.
@@ -2875,7 +2873,6 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
    * const result = await channel.history({ limit: 25 });
    * ```
    * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#history
-   * @see https://ably.com/docs/storage-history/storage
    */
   history(params?: RealtimeHistoryParams): Promise<PaginatedResult<InboundMessage>>;
   /**
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -2725,7 +2725,7 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
   /**
    * An {@link ErrorInfo} object describing the last error which occurred on the channel, if any. This is `null` until an error occurs, for example a transition in channel state to `failed` or `suspended`. Guard against `null` despite the declared type.
    *
-   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#error-reason
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#properties
    */
   errorReason: ErrorInfo;
   /**
@@ -2737,7 +2737,7 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
    *
    * It is `undefined` before the channel attaches for the first time, so guard against `undefined` despite the declared type.
    *
-   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#params
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#properties
    */
   params: ChannelParams;
   /**
@@ -2745,11 +2745,11 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
    *
    * It is `undefined` before the channel attaches for the first time. When the server grants no modes, it is also `undefined`, not an empty array. Guard against `undefined` despite the declared type.
    *
-   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#modes
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#properties
    */
   modes: ResolvedChannelMode[];
   /**
-   * Deregisters the given listener for the specified event name. This only removes the local listeners and does not detach the channel. The channel stays attached, so the server keeps streaming its messages to the client while the client's capability permits.
+   * Deregisters the given listener for the specified event name. This removes local listeners only and does not detach the channel. The server keeps streaming messages to an attached channel, subject to the client's capabilities.
    *
    * @param event - The event name.
    * @param listener - An event listener function.
@@ -2757,7 +2757,7 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
    */
   unsubscribe(event: string, listener: messageCallback<InboundMessage>): void;
   /**
-   * Deregisters the given listener from all event names in the array. This only removes the local listeners and does not detach the channel. The channel stays attached, so the server keeps streaming its messages to the client while the client's capability permits.
+   * Deregisters the given listener from all event names in the array. This removes local listeners only and does not detach the channel. The server keeps streaming messages to an attached channel, subject to the client's capabilities.
    *
    * @param events - An array of event names.
    * @param listener - An event listener function.
@@ -2765,21 +2765,21 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
    */
   unsubscribe(events: Array<string>, listener: messageCallback<InboundMessage>): void;
   /**
-   * Deregisters all listeners for the given event name. This only removes the local listeners and does not detach the channel. The channel stays attached, so the server keeps streaming its messages to the client while the client's capability permits.
+   * Deregisters all listeners for the given event name. This removes local listeners only and does not detach the channel. The server keeps streaming messages to an attached channel, subject to the client's capabilities.
    *
    * @param event - The event name.
    * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#unsubscribe
    */
   unsubscribe(event: string): void;
   /**
-   * Deregisters all listeners for all event names in the array. This only removes the local listeners and does not detach the channel. The channel stays attached, so the server keeps streaming its messages to the client while the client's capability permits.
+   * Deregisters all listeners for all event names in the array. This removes local listeners only and does not detach the channel. The server keeps streaming messages to an attached channel, subject to the client's capabilities.
    *
    * @param events - An array of event names.
    * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#unsubscribe
    */
   unsubscribe(events: Array<string>): void;
   /**
-   * Deregisters listeners registered with the supplied {@link MessageFilter}. The filter is compared by object identity, so it must be the same object previously passed to {@link RealtimeChannel.subscribe | `subscribe()`}. A filter object not previously passed to `subscribe()` removes nothing and no error is raised. This only removes the local listeners and does not detach the channel. The channel stays attached, so the server keeps streaming its messages to the client while the client's capability permits.
+   * Deregisters listeners registered with the supplied {@link MessageFilter}. The filter is matched by object identity, so pass the same object given to {@link RealtimeChannel.subscribe | `subscribe()`}. Any other filter object removes nothing and no error is raised. This removes local listeners only and does not detach the channel. The server keeps streaming messages to an attached channel, subject to the client's capabilities.
    *
    * @param filter - A {@link MessageFilter}.
    * @param listener - An event listener function.
@@ -2787,14 +2787,14 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
    */
   unsubscribe(filter: MessageFilter, listener?: messageCallback<InboundMessage>): void;
   /**
-   * Deregisters the given listener from all event names. This only removes the local listeners and does not detach the channel. The channel stays attached, so the server keeps streaming its messages to the client while the client's capability permits.
+   * Deregisters the given listener from all event names. This removes local listeners only and does not detach the channel. The server keeps streaming messages to an attached channel, subject to the client's capabilities.
    *
    * @param listener - An event listener function.
    * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#unsubscribe
    */
   unsubscribe(listener: messageCallback<InboundMessage>): void;
   /**
-   * Deregisters all listeners to messages on this channel. This only removes the local listeners and does not detach the channel. The channel stays attached, so the server keeps streaming its messages to the client while the client's capability permits.
+   * Deregisters all listeners to messages on this channel. This removes local listeners only and does not detach the channel. The server keeps streaming messages to an attached channel, subject to the client's capabilities.
    *
    * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#unsubscribe
    */
@@ -2822,7 +2822,7 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
   /**
    * A {@link PushChannel} object that manages device push notification subscriptions for this channel. Accessing this property requires the `Push` plugin to be registered via {@link ClientOptions.plugins}. The default and modular builds do not bundle the `Push` plugin. If the plugin is absent, the getter throws an {@link ErrorInfo} rather than returning a {@link PushChannel}.
    *
-   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#push
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#properties
    */
   push: PushChannel;
   /**
@@ -2906,7 +2906,7 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
    */
   setOptions(options: ChannelOptions): Promise<void>;
   /**
-   * Registers a listener for messages with a given event name on this channel. Implicitly attaches the channel unless {@link ChannelOptions.attachOnSubscribe} is `false`. Without the `subscribe` mode the listener never fires and no error is raised.
+   * Registers a listener for messages with a given event name on this channel. Implicitly attaches the channel unless {@link ChannelOptions.attachOnSubscribe} is `false`. Without the `subscribe` mode the server delivers no messages and the listener never fires.
    *
    * @param event - The event name.
    * @param listener - An event listener function.
@@ -2919,7 +2919,7 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
    */
   subscribe(event: string, listener?: messageCallback<InboundMessage>): Promise<ChannelStateChange | null>;
   /**
-   * Registers a listener for messages on this channel for multiple event name values. Implicitly attaches the channel unless {@link ChannelOptions.attachOnSubscribe} is `false`. Without the `subscribe` mode the listener never fires and no error is raised.
+   * Registers a listener for messages on this channel for multiple event name values. Implicitly attaches the channel unless {@link ChannelOptions.attachOnSubscribe} is `false`. Without the `subscribe` mode the server delivers no messages and the listener never fires.
    *
    * @param events - An array of event names.
    * @param listener - An event listener function.
@@ -2934,7 +2934,7 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
   /**
    * {@label WITH_MESSAGE_FILTER}
    *
-   * Registers a listener for messages on this channel that match the supplied filter. Implicitly attaches the channel unless {@link ChannelOptions.attachOnSubscribe} is `false`. Without the `subscribe` mode the listener never fires and no error is raised.
+   * Registers a listener for messages on this channel that match the supplied filter. Implicitly attaches the channel unless {@link ChannelOptions.attachOnSubscribe} is `false`. Without the `subscribe` mode the server delivers no messages and the listener never fires.
    *
    * @param filter - A {@link MessageFilter}.
    * @param listener - An event listener function.
@@ -2947,7 +2947,7 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
    */
   subscribe(filter: MessageFilter, listener?: messageCallback<InboundMessage>): Promise<ChannelStateChange | null>;
   /**
-   * Registers a listener for messages on this channel. Implicitly attaches the channel unless {@link ChannelOptions.attachOnSubscribe} is `false`. Without the `subscribe` mode the listener never fires and no error is raised.
+   * Registers a listener for messages on this channel. Implicitly attaches the channel unless {@link ChannelOptions.attachOnSubscribe} is `false`. Without the `subscribe` mode the server delivers no messages and the listener never fires.
    *
    * @param callback - An event listener function.
    * @returns A promise which, upon successful attachment to the channel, will be fulfilled with a {@link ChannelStateChange} object. If the channel was already attached the promise will be resolved with `null`. The promise also resolves with `null` when {@link ChannelOptions.attachOnSubscribe} is `false`, in which case no attach is attempted. Upon failure, the promise will be rejected with an {@link ErrorInfo} object.
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -2681,7 +2681,9 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
    */
   readonly name: string;
   /**
-   * An {@link ErrorInfo} object describing the last error which occurred on the channel, if any.
+   * An {@link ErrorInfo} object describing the last error which occurred on the channel, if any. This is `null` until an error occurs, for example a transition to `failed` or `suspended`, so guard against `null` despite the declared type; inspect it to diagnose a failed or suspended channel.
+   *
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#error-reason
    */
   errorReason: ErrorInfo;
   /**
@@ -2689,54 +2691,70 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
    */
   readonly state: ChannelState;
   /**
-   * Optional [channel parameters](https://ably.com/docs/realtime/channels/channel-parameters/overview) that configure the behavior of the channel.
+   * Optional [channel parameters](https://ably.com/docs/realtime/channels/channel-parameters/overview) that configure the behavior of the channel. After the channel attaches this reflects the parameters the server confirmed on the most recent attach, which may differ from those requested via {@link ChannelOptions.params}, and is `undefined` before the channel attaches for the first time, so guard against `undefined` despite the declared type.
+   *
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#params
    */
   params: ChannelParams;
   /**
-   * An array of {@link ResolvedChannelMode} objects.
+   * An array of {@link ResolvedChannelMode} objects reflecting the modes the server granted on the most recent attach, which may differ from the modes requested via {@link ChannelOptions.modes}. The value is `undefined` before the channel attaches for the first time, and is also `undefined` (rather than an empty array) when the server grants no modes, so callers should guard against `undefined` despite the declared type.
+   *
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#modes
    */
   modes: ResolvedChannelMode[];
   /**
-   * Deregisters the given listener for the specified event name. This removes an earlier event-specific subscription.
+   * Deregisters the given listener for the specified event name, removing an earlier event-specific subscription. This only removes the local listener and does not detach the channel, so messages may continue to arrive for any other registered listeners.
    *
    * @param event - The event name.
    * @param listener - An event listener function.
+   * @example
+   * ```ts
+   * channel.unsubscribe('event', listener);
+   * ```
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#unsubscribe
    */
   unsubscribe(event: string, listener: messageCallback<InboundMessage>): void;
   /**
-   * Deregisters the given listener from all event names in the array.
+   * Deregisters the given listener from all event names in the array. This only removes the local listeners and does not detach the channel.
    *
    * @param events - An array of event names.
    * @param listener - An event listener function.
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#unsubscribe
    */
   unsubscribe(events: Array<string>, listener: messageCallback<InboundMessage>): void;
   /**
-   * Deregisters all listeners for the given event name.
+   * Deregisters all listeners for the given event name. This only removes the local listeners and does not detach the channel.
    *
    * @param event - The event name.
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#unsubscribe
    */
   unsubscribe(event: string): void;
   /**
-   * Deregisters all listeners for all event names in the array.
+   * Deregisters all listeners for all event names in the array. This only removes the local listeners and does not detach the channel.
    *
    * @param events - An array of event names.
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#unsubscribe
    */
   unsubscribe(events: Array<string>): void;
   /**
-   * Deregisters all listeners to messages on this channel that match the supplied filter.
+   * Deregisters listeners to messages on this channel that match the supplied filter, removing an earlier filtered subscription. This only removes the local listeners and does not detach the channel.
    *
    * @param filter - A {@link MessageFilter}.
    * @param listener - An event listener function.
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#unsubscribe
    */
   unsubscribe(filter: MessageFilter, listener?: messageCallback<InboundMessage>): void;
   /**
-   * Deregisters the given listener (for any/all event names). This removes an earlier subscription.
+   * Deregisters the given listener (for any/all event names), removing an earlier subscription. This only removes the local listener and does not detach the channel.
    *
    * @param listener - An event listener function.
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#unsubscribe
    */
   unsubscribe(listener: messageCallback<InboundMessage>): void;
   /**
-   * Deregisters all listeners to messages on this channel. This removes all earlier subscriptions.
+   * Deregisters all listeners to messages on this channel, removing all earlier subscriptions. This only removes the local listeners and does not detach the channel, so the channel stays attached.
+   *
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#unsubscribe
    */
   unsubscribe(): void;
   /**
@@ -2760,7 +2778,9 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
    */
   presence: RealtimePresence;
   /**
-   * A {@link PushChannel} object.
+   * A {@link PushChannel} object that manages device push-notification subscriptions for this channel. Accessing this property requires the Push plugin to be registered via {@link ClientOptions.plugins}, which neither the default nor the modular build bundles; if the plugin is absent, the getter throws an {@link ErrorInfo} rather than returning a {@link PushChannel}.
+   *
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#push
    */
   push: PushChannel;
   /**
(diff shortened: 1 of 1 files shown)


ably.d.ts
```diff
@@ -2627,62 +2627,80 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
    */
   readonly name: string;
   /**
-   * An {@link ErrorInfo} object describing the last error which occurred on the channel, if any.
+   * An {@link ErrorInfo} object describing the last error which occurred on the channel, if any. This is `null` until an error occurs, for example a transition to `failed` or `suspended`, so guard against `null` despite the declared type; inspect it to diagnose a failed or suspended channel.
```
m-hulbert:
```suggestion
   * An {@link ErrorInfo} object describing the last error which occurred on the channel, if any. This is `null` until an error occurs, for example a transition in channel state to `failed` or `suspended`. Guard against `null` despite the declared type and inspect it to diagnose a channel in the failed or suspended state.
```

ably.d.ts
```diff
@@ -2627,62 +2627,80 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#error-reason
    */
   errorReason: ErrorInfo;
   /**
    * The current {@link ChannelState} of the channel.
    */
   readonly state: ChannelState;
   /**
-   * Optional [channel parameters](https://ably.com/docs/realtime/channels/channel-parameters/overview) that configure the behavior of the channel.
+   * Optional [channel parameters](https://ably.com/docs/realtime/channels/channel-parameters/overview) that configure the behavior of the channel. After the channel attaches this reflects the parameters the server confirmed on the most recent attach, which may differ from those requested via {@link ChannelOptions.params}, and is `undefined` before the channel attaches for the first time, so guard against `undefined` despite the declared type.
```
m-hulbert:
```suggestion
   * Optional channel parameters that configure the behavior of the channel. After the channel attaches this reflects the parameters the server confirmed on the most recent attach, which may differ from those requested via {@link ChannelOptions.params}. It is `undefined` before the channel attaches for the first time, so guard against `undefined` despite the declared type.
```

Appreciate this existed before, but it's odd that we link out to the docs on this and the TypeDoc reference and then the `@see` link is to some different docs again. I think drop the initial link to docs.

ably.d.ts
```diff
@@ -2627,62 +2627,80 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
+   *
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#params
    */
   params: ChannelParams;
   /**
-   * An array of {@link ResolvedChannelMode} objects.
+   * An array of {@link ResolvedChannelMode} objects reflecting the modes the server granted on the most recent attach, which may differ from the modes requested via {@link ChannelOptions.modes}. The value is `undefined` before the channel attaches for the first time, and is also `undefined` (rather than an empty array) when the server grants no modes, so callers should guard against `undefined` despite the declared type.
```
m-hulbert: `which may differ from the modes requested via {@link ChannelOptions.modes}` how and why? 

I'd also remove the brackets and long sentences as per other suggestions.

ably.d.ts
```diff
@@ -2627,62 +2627,80 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
+   *
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#modes
    */
   modes: ResolvedChannelMode[];
   /**
-   * Deregisters the given listener for the specified event name. This removes an earlier event-specific subscription.
+   * Deregisters the given listener for the specified event name, removing an earlier event-specific subscription. This only removes the local listener and does not detach the channel, so messages may continue to arrive for any other registered listeners.
```
m-hulbert: I think if we're saying other listeners may still be receiving messages then we should also state that messages will still be _streamed_ to the client if they have permissions, unless they detach maybe?

ably.d.ts
```diff
@@ -2627,62 +2627,80 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
+   * @example
+   * ```ts
+   * channel.unsubscribe('event', listener);
+   * ```
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#unsubscribe
    */
   unsubscribe(event: string, listener: messageCallback<InboundMessage>): void;
   /**
-   * Deregisters the given listener from all event names in the array.
+   * Deregisters the given listener from all event names in the array. This only removes the local listeners and does not detach the channel.
```
m-hulbert: Is there a reason to only include the additional info on 1 overload?

ably.d.ts
```diff
@@ -2627,62 +2627,80 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
+   * Deregisters listeners to messages on this channel that match the supplied filter, removing an earlier filtered subscription. This only removes the local listeners and does not detach the channel.
    *
    * @param filter - A {@link MessageFilter}.
    * @param listener - An event listener function.
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#unsubscribe
    */
   unsubscribe(filter: MessageFilter, listener?: messageCallback<InboundMessage>): void;
   /**
-   * Deregisters the given listener (for any/all event names). This removes an earlier subscription.
+   * Deregisters the given listener (for any/all event names), removing an earlier subscription. This only removes the local listener and does not detach the channel.
```
m-hulbert: I think this needs the brackets removed.

ably.d.ts
```diff
@@ -2627,62 +2627,80 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
    *
    * @param listener - An event listener function.
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#unsubscribe
    */
   unsubscribe(listener: messageCallback<InboundMessage>): void;
   /**
-   * Deregisters all listeners to messages on this channel. This removes all earlier subscriptions.
+   * Deregisters all listeners to messages on this channel, removing all earlier subscriptions. This only removes the local listeners and does not detach the channel, so the channel stays attached.
```
m-hulbert: `and does not detach the channel, so the channel stays attached.` a bit redundant at the end.

ably.d.ts
```diff
@@ -2706,30 +2724,48 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
    */
   presence: RealtimePresence;
   /**
-   * A {@link PushChannel} object.
+   * A {@link PushChannel} object that manages device push-notification subscriptions for this channel. Accessing this property requires the Push plugin to be registered via {@link ClientOptions.plugins}, which neither the default nor the modular build bundles; if the plugin is absent, the getter throws an {@link ErrorInfo} rather than returning a {@link PushChannel}.
```
m-hulbert:
```suggestion
   * A {@link PushChannel} object that manages device push notification subscriptions for this channel. Accessing this property requires the `Push` plugin to be registered via {@link ClientOptions.plugins}.  The default and modular builds do not bundle the `Push` plugin. If the plugin is absent, the getter throws an {@link ErrorInfo} rather than returning a {@link PushChannel}.
```

ably.d.ts
```diff
@@ -2706,30 +2724,48 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
+   * Attach to this channel, ensuring it is created in the Ably system so that messages published on the channel are received by any listeners registered with {@link RealtimeChannel.subscribe | `subscribe()`}, and emitting the resulting state change to any listeners registered via {@link EventEmitter.on | `on()`} or {@link EventEmitter.once | `once()`}. As a convenience you need not call this directly: it is invoked implicitly when {@link RealtimeChannel.subscribe | `subscribe()`} is called, when {@link RealtimePresence.enter | `enter()`} or {@link RealtimePresence.subscribe | `subscribe()`} are called on this channel's {@link RealtimePresence} object, or when `get()` is called on its `RealtimeObject`. The returned promise rejects with an {@link ErrorInfo} if the connection is unusable or the channel transitions to `detached`, `suspended`, or `failed` instead of attaching.
```
m-hulbert:
```suggestion
   * Attach to this channel, ensuring it is created in the Ably system so that messages published on the channel are received by any listeners registered with {@link RealtimeChannel.subscribe | `subscribe()`}. Also emits the resulting state change to any listeners registered via {@link EventEmitter.on | `on()`} or {@link EventEmitter.once | `once()`}. 
   *
   * As a convenience you need not call this directly: it is invoked implicitly when {@link RealtimeChannel.subscribe | `subscribe()`} is called, when {@link RealtimePresence.enter | `enter()`} or {@link RealtimePresence.subscribe | `subscribe()`} are called on this channel's {@link RealtimePresence} object, or when `get()` is called on its `RealtimeObject`. 
   *
   * The returned promise rejects with an {@link ErrorInfo} if the connection is unusable or the channel transitions to the `detached`, `suspended`, or `failed` state instead of attaching.
```

ably.d.ts
```diff
@@ -2706,30 +2724,48 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
+   * @example
+   * ```ts
+   * await channel.attach();
+   * ```
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#attach
    */
   attach(): Promise<ChannelStateChange | null>;
   /**
-   * Detach from this channel. Any resulting channel state change is emitted to any listeners registered using the {@link EventEmitter.on | `on()`} or {@link EventEmitter.once | `once()`} methods. Once all clients globally have detached from the channel, the channel will be released in the Ably service within two minutes.
+   * Detach from this channel. Any resulting channel state change is emitted to any listeners registered using the {@link EventEmitter.on | `on()`} or {@link EventEmitter.once | `once()`} methods. Once all clients globally have detached from the channel, the channel is released in the Ably service within two minutes. Detaching from a `suspended` or already `detached` channel resolves without further action, but detaching from a `failed` channel rejects with an {@link ErrorInfo}, in which case release the channel and get it again to start afresh.
```
m-hulbert:
```suggestion
   * Detach from this channel. Any resulting channel state change is emitted to any listeners registered using the {@link EventEmitter.on | `on()`} or {@link EventEmitter.once | `once()`} methods. Once all clients globally have detached from the channel, the channel is released in the Ably service within two minutes.
   * 
   * Detaching from a channel in the `suspended` or `detached` state resolves without further action, but detaching from a channel in the `failed` state rejects with an {@link ErrorInfo}, in which case release the channel and get it again to start afresh.
```

Should release and get be links here too?

ably.d.ts
```diff
@@ -2706,30 +2724,48 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
+   * @example
+   * ```ts
+   * await channel.detach();
+   * ```
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#detach
    */
   detach(): Promise<void>;
   /**
-   * Retrieves a {@link PaginatedResult} object, containing an array of historical {@link InboundMessage} objects for the channel. If the channel is configured to persist messages, then messages can be retrieved from history for up to 72 hours in the past. If not, messages can only be retrieved from history for up to two minutes in the past.
+   * Retrieves a {@link PaginatedResult} object, containing an array of historical {@link InboundMessage} objects for the channel. Messages are returned from durable storage only when message persistence is enabled for the channel by a channel rule; without it, only messages from the last two minutes, the service's default retention, are returned.
```
m-hulbert:
```suggestion
   * Retrieves a {@link PaginatedResult} object, containing an array of historical {@link InboundMessage} objects for the channel. Messages are returned from storage only when message persistence is enabled for the channel by a rule. If message persistence is not enabled, only messages from the last two minutes are returned.
```

ably.d.ts
```diff
@@ -2747,43 +2783,68 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
    */
   history(params: RealtimeHistoryParams | null, callback: StandardCallback<PaginatedResult<InboundMessage>>): never;
   /**
-   * Sets the {@link ChannelOptions} for the channel.
+   * Sets the {@link ChannelOptions} for the channel. If the channel is already attached or attaching and the new {@link ChannelOptions.modes} or {@link ChannelOptions.params} differ from the current ones, this triggers a live re-attach to apply them, and the returned promise resolves only once the server confirms the new options, rejecting with an {@link ErrorInfo} if the channel detaches or fails in the meantime. Otherwise the new options are stored for the next attach. The promise also rejects with an {@link ErrorInfo} when the supplied options are invalid.
```
m-hulbert: This is a very long sentence and it isn't clear what a "live re-attach" is.

ably.d.ts
```diff
@@ -2747,43 +2783,68 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
+   * @example
+   * ```ts
+   * await channel.setOptions({ params: { rewind: '1' } });
+   * ```
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#set-options
    */
   setOptions(options: ChannelOptions): Promise<void>;
   /**
-   * Registers a listener for messages with a given event name on this channel. The caller supplies a listener function, which is called each time one or more matching messages arrives on the channel.
+   * Registers a listener for messages with a given event name on this channel. The caller supplies a listener function, which is called each time one or more matching messages arrives on the channel. Implicitly attaches the channel unless {@link ChannelOptions.attachOnSubscribe} is `false`. Requires the `subscribe` mode (granted by default unless {@link ChannelOptions.modes} excludes it); if the channel attaches without it the server never delivers messages, so the listener silently never fires and a warning is logged rather than the call rejecting (or it rejects with a hinted {@link ErrorInfo} when {@link ClientOptions.strictMode} is enabled).
```
m-hulbert: Mmm, this is a lot of extra info getting into modes here - do you think that's really required?

ably.d.ts
```diff
@@ -2802,28 +2863,43 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
    */
   subscribe(event: string, listener: messageCallback<InboundMessage>, callback: ErrorCallback): never;
   /**
-   * Publishes a single message to the channel with the given event name and payload. When publish is called with this client library, it won't attempt to implicitly attach to the channel, so long as [transient publishing](https://ably.com/docs/realtime/channels#transient-publish) is available in the library. Otherwise, the client will implicitly attach.
+   * Publishes a single message to the channel with the given event name and payload. Publishing does not attach the channel, so unlike {@link RealtimeChannel.subscribe | `subscribe()`}, {@link RealtimePresence.enter | `enter()`}, and {@link RealtimePresence.get | `get()`} a publish-only client need not attach or subscribe first.
```
m-hulbert: I don't think we should be saying unlike X, Y or Z as someone is just looking at this method. If you think we _really_ need to state something, I'd go more along the lines of "the client does not need to attach to the channel before publishing to it", but even that seems redundant to a certain extent.

ably.d.ts
```diff
@@ -2842,51 +2918,81 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
+   * @example
+   * ```ts
+   * const stateChange = await channel.whenState('attached');
+   * ```
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#when-state
    */
   whenState(targetState: ChannelState): Promise<ChannelStateChange | null>;
   /**
-   * Retrieves the latest version of a specific message by its serial identifier.
+   * Retrieves the latest version of a specific message by its serial identifier. Retrievable messages require message interactions (mutable messages) to be enabled for the channel by a channel rule. The supplied serial or {@link Message} must carry a populated `serial`, which is present on a {@link Message} received from a subscribe callback but not on a freshly constructed one; if the serial is missing the promise rejects with an {@link ErrorInfo}.
```
m-hulbert: We don't call things "mutable messages", "message interactions" nor "channel rules".

'Retrievable messages' also reads strangely - as does what the subsequent sentence is trying to say. I think you're saying `serial` only exists if you've received a message via subscribe(), not if you've constructed one which is arguable more confusing to read than what it's trying to solve IMO.

ably.d.ts
```diff
@@ -2842,51 +2918,81 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
+   * @example
+   * ```ts
+   * const message = await channel.getMessage(serial);
+   * ```
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#get-message
    */
   getMessage(serialOrMessage: string | Message): Promise<Message>;
   /**
-   * Publishes an update to an existing message with patch semantics. Non-null `name`, `data`, and `extras` fields in the provided message will replace the corresponding fields in the existing message, while null fields will be left unchanged.
+   * Publishes an update to an existing message with patch semantics. Non-null `name`, `data`, and `extras` fields in the provided message will replace the corresponding fields in the existing message, while null fields will be left unchanged. The namespace must have message interactions (updates) enabled by a channel rule, and the supplied `message` must carry the `serial` it was given when published (pass the {@link Message} received in a subscribe callback, not a freshly constructed object); otherwise the promise rejects with an {@link ErrorInfo}.
```
m-hulbert: As above (haven't been commenting when a previous comment applies to every overload) but also should `Message` and `Serial` be capitalised?

m-hulbert (commented):
A few more comments on this one outside of language. There are some things I think make the descriptions more complicated rather than simpler and a few things to tidy up in terms of naming.

ably.d.ts
AndyTWF: Yes I think we need to be very clear here. It does not detach the channel and any registered listeners may receive messages. But the key part is not detaching the channel - messages will still be sent by the server and will be billed accordingly

ably.d.ts
AndyTWF: I think we need to be careful about the term "created", as that gives the impression that they are persistent entities that require explicit operations to delete, which isn't true.

I would advise against mentioning `EventEmitter` here as that's an internal implementation detail, perhaps instead saying `RealtimeChannel.on()`.

It is also invoked implicitly for RealtimePresence.update() and RealtimePresence.get().

I think that simply referencing `RealtimeObject` is perhaps a bit generic and might confuse things - do we want to reference liveobjects directly here?

The promise also rejects if the channel transitions to `detaching`.

ably.d.ts
AndyTWF: Ditto my previous on mentioning EventEmitter.

Worth discussing the behaviour of calling detach() on a detaching channel?

I think there's an opportunity here to explicitly highlight that this means messages on this channel will not be sent from the server to the client, but that any registered listeners locally will remain

ably.d.ts
```diff
@@ -2706,30 +2724,48 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
+   * Retrieves a {@link PaginatedResult} object, containing an array of historical {@link InboundMessage} objects for the channel. Messages are returned from durable storage only when message persistence is enabled for the channel by a channel rule; without it, only messages from the last two minutes, the service's default retention, are returned.
    *
-   * @param params - A set of parameters which are used to specify which presence members should be retrieved.
+   * @param params - A set of parameters which are used to specify which messages should be retrieved.
    * @returns A promise which, upon success, will be fulfilled with a {@link PaginatedResult} object containing an array of {@link InboundMessage} objects. Upon failure, the promise will be rejected with an {@link ErrorInfo} object which explains the error.
+   * @example
+   * ```ts
+   * const result = await channel.history({ limit: 25 });
+   * ```
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#history
```
AndyTWF, line 2875: A general question - do we have a plan for keeping this in sync with the docs to avoid 404s?

ably.d.ts
```diff
@@ -2747,43 +2783,68 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
+   * await channel.subscribe(['event1', 'event2'], (message) => console.log(message.data));
+   * ```
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#subscribe
    */
   subscribe(events: Array<string>, listener?: messageCallback<InboundMessage>): Promise<ChannelStateChange | null>;
   /**
    * {@label WITH_MESSAGE_FILTER}
    *
-   * Registers a listener for messages on this channel that match the supplied filter.
+   * Registers a listener for messages on this channel that match the supplied filter. Implicitly attaches the channel unless {@link ChannelOptions.attachOnSubscribe} is `false`. Requires the `subscribe` mode (granted by default unless {@link ChannelOptions.modes} excludes it); if the channel attaches without it the server never delivers messages, so the listener silently never fires and a warning is logged rather than the call rejecting (or it rejects with a hinted {@link ErrorInfo} when {@link ClientOptions.strictMode} is enabled).
```
AndyTWF: Granted by default IFF the key/token has the necessary capability

ably.d.ts
```diff
@@ -2842,51 +2918,81 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
    */
   publish(name: string, data: any, callback: ErrorCallback): never;
   /**
-   * If the channel is already in the given state, returns a promise which immediately resolves to `null`. Else, calls {@link EventEmitter.once | `once()`} to return a promise which resolves the next time the channel transitions to the given state.
+   * If the channel is already in the given state, returns a promise which immediately resolves to `null`. Else, calls {@link EventEmitter.once | `once()`} to return a promise which resolves with the {@link ChannelStateChange} the next time the channel transitions to the given state.
```
AndyTWF: Calling EventEmitter.once is an internal detail, does the consumer need to know how it works?

ably.d.ts
AndyTWF: Mutable messages != message interactions - the latter is an unrelated, deprecated feature

ably.d.ts
```diff
@@ -2842,51 +2918,81 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
+   * @example
+   * ```ts
+   * const result = await channel.deleteMessage(message);
+   * ```
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#delete-message
    */
   deleteMessage(message: Message, operation?: MessageOperation, options?: PublishOptions): Promise<UpdateDeleteResult>;
   /**
-   * Appends data to an existing message. The supplied `data` field is appended to the previous message's data, while all other fields (`name`, `extras`) replace the previous values if provided.
+   * Appends data to an existing message. The supplied `data` field is appended to the previous message's data, while all other fields (`name`, `extras`) replace the previous values if provided. Requires message interactions (mutable messages) to be enabled for the channel by a channel rule, and the supplied {@link Message} must carry the `serial` it received from a subscribe callback, otherwise the call rejects with an {@link ErrorInfo}.
```
AndyTWF: Ditto here and other places mentioning "message interactions" - its unrelated to this

ably.d.ts
umair-ably: fair! I've moved towards keeping things as simple as possible and just mentioned that subscribe is required

ably.d.ts
umair-ably: not directly, the pr description highlights that some of these point at ably/docs PR #3400 which is yet to be merged.

I'll create a ticket to track how we can avoid 404s going forwards, likely some CI step

ably.d.ts
umair-ably: not serial, no. `serial` is the field on `Message` (lowercase in the API)

umair-ably:
Thanks both - @AndyTWF @m-hulbert 

I've heavily refactored the prompt/workflow this was generated from, and introduced a new review step which covers some language and content style guidelines with your feedback from this PR in mind. 

Latest commit shows the output of this which I've also checked through. Might still be some stragglers but leaning into this approach so it's easier to apply to all the other docstrings too

ably.d.ts
m-hulbert: A few things on this one:

* Where are the documented silent-failure paths?
* We shouldn't be including a version in the string, as these will only be visible to anyone on version 2.x+ 
* I don't think we should have an example in this instance - we just state it returns the legacy values. 
* I'd then move the info about the default to just before the final sentence.

ably.d.ts
```diff
@@ -1985,22 +1985,14 @@ export declare interface RealtimeClient {
...
-   *
-   * Any {@link TokenParams} and {@link AuthOptions} passed in are stored as the new defaults for all subsequent token requests and entirely replace, rather than merge with, the current saved values.
+   * Instructs the library to get a new token immediately. On a realtime client it re-authenticates the live connection, or initiates a connection if not connected, and the returned promise resolves only once the new token has taken effect on a `connected` connection, rejecting with an {@link ErrorInfo} if re-authentication fails or the connection cannot be (re)established. The client must be able to issue tokens, so a `key`, `authUrl`, or `authCallback` must be configured in {@link ClientOptions}, otherwise the call rejects with an {@link ErrorInfo}; `authorize()` cannot change the API key, so passing an `authOptions.key` that differs from the one the client was constructed with also rejects. Any {@link TokenParams} and {@link AuthOptions} passed in are stored as the new defaults for all subsequent token requests and entirely replace, rather than merge with, the current saved values.
```
m-hulbert: These are some really long sentences - can we split them up?

ably.d.ts
```diff
@@ -2032,11 +2024,7 @@ export declare interface Auth {
...
-   *
-   * Both {@link TokenParams} and {@link AuthOptions} are optional. When omitted or `null`, the client's stored defaults are used, as specified at instantiation or later updated by an `authorize()` request. Any values passed in replace, rather than merge with, those defaults.
+   * Creates and signs an Ably {@link TokenRequest} based on the specified (or if none specified, the client library stored) {@link TokenParams} and {@link AuthOptions}. Use this to implement an Ably Token request callback for use by other clients. An API `key` value must be available locally to sign the request, supplied either in the client's {@link ClientOptions} or as `key` in the `authOptions` argument; without one the call rejects with an {@link ErrorInfo}, since a token-authenticated client cannot construct token requests itself and must instead obtain the {@link TokenRequest} from the key owner. Both {@link TokenParams} and {@link AuthOptions} are optional; when omitted or `null`, the client library's stored defaults (those set at construction or by a later `authorize` call) are used, and any values passed in replace, rather than merge with, those defaults.
```
m-hulbert: As above - I'd make the semi-colon a fullstop to split up the long sentences and make it easier to parse.

ably.d.ts
```diff
@@ -2068,9 +2056,7 @@ export declare interface Auth {
...
-   *
-   * The client must have a way to obtain a token, so the resolved {@link AuthOptions} must include one of `authCallback`, `authUrl`, or `key`. A client given only a literal token or `tokenDetails` with no renewal mechanism cannot request a new token and the call rejects with an {@link ErrorInfo}.
+   * Calls the `requestToken` REST API endpoint to obtain an Ably Token according to the specified {@link TokenParams} and {@link AuthOptions}. Both are optional; when omitted or `null`, the client's stored defaults are used, as specified in the {@link ClientOptions} at instantiation or later updated by an `authorize` request, and any values passed in are used instead of, rather than merged with, those defaults. The client must have a usable way to obtain a token, so the resolved {@link AuthOptions} must include one of `authCallback`, `authUrl`, or `key`; a client given only a literal token or `tokenDetails` with no renewal mechanism cannot request a new token and the call rejects with an {@link ErrorInfo}.
```
m-hulbert: Two big sentences - sensing a theme 🙂

ably.d.ts
```diff
@@ -2103,10 +2089,10 @@ export declare interface Auth {
...
   /**
    * Revokes the tokens specified by the provided array of {@link TokenRevocationTargetSpecifier}s.
-   *
-   * The client making this call must be authenticated with an API key (basic auth), not a token. A token-authenticated client cannot revoke tokens and the call rejects with an {@link ErrorInfo}.
-   *
-   * Only tokens issued by an API key that had revocable tokens enabled before the token was issued can be revoked.
+   * The client must be authenticated with an API key (basic auth), not a token; a
+   * token-authenticated client cannot revoke tokens and the call rejects with an {@link ErrorInfo}.
+   * Only tokens issued by an API key that had revocable tokens enabled before the token was issued
+   * can be revoked.
```
m-hulbert: Some odd formatting here and similar comment to before in that we should be explicit in that the client revoking the tokens needs to be auth'd with an API key. Also semi-colon should be a fullstop again.

ably.d.ts
m-hulbert: Any reason this one has an API ref link and a conceptual one?

ably.d.ts
```diff
@@ -2662,62 +2648,82 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
+   * An array of {@link ResolvedChannelMode} objects reflecting the modes the server granted on the most recent attach. The server grants only the modes the key or token capability permits, so this may be a subset of the modes requested via {@link ChannelOptions.modes}.
+   *
+   * It is `undefined` before the channel attaches for the first time, and also when the server grants no modes rather than an empty array, so guard against `undefined` despite the declared type.
+   *
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#modes
    */
   modes: ResolvedChannelMode[];
   /**
-   * Deregisters the given listener for the specified event name. This removes an earlier event-specific subscription.
+   * Deregisters the given listener for the specified event name, removing an earlier event-specific subscription. This only removes the local listeners and does not detach the channel. The server keeps sending messages on the channel and they are billed accordingly.
```
m-hulbert: I think we need a bit more exposition here around the behaviour of messages being streamed to clients if they aren't detached and have the relevant capabilities (and we should omit the billing part as that may change in the future).

ably.d.ts
```diff
@@ -2741,30 +2747,58 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
   /**
-   * Attach to this channel ensuring the channel is created in the Ably system and all messages published on the channel are received by any channel listeners registered using {@link RealtimeChannel.subscribe | `subscribe()`}. Any resulting channel state change will be emitted to any listeners registered using the {@link EventEmitter.on | `on()`} or {@link EventEmitter.once | `once()`} methods. As a convenience, `attach()` is called implicitly if {@link RealtimeChannel.subscribe | `subscribe()`} for the channel is called, or {@link RealtimePresence.enter | `enter()`} or {@link RealtimePresence.subscribe | `subscribe()`} are called on the {@link RealtimePresence} object for this channel, or `get()` is called on the `RealtimeObject` object for this channel.
+   * Attach to this channel so that messages published on it are received by any listeners registered with {@link RealtimeChannel.subscribe | `subscribe()`}. It also emits the resulting state change to any registered listeners.
+   *
+   * As a convenience you need not call this directly. It is invoked implicitly by `subscribe()`, by presence `enter()`, `subscribe()`, `update()`, or `get()`, or by LiveObjects `get()`.
```
m-hulbert: I think we should be more precise on the methods; `channel.subscribe()`, `presence.enter()` etc and link to them too.

ably.d.ts
```diff
@@ -2741,30 +2747,58 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
+   * await channel.attach();
+   * ```
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#attach
    */
   attach(): Promise<ChannelStateChange | null>;
   /**
-   * Detach from this channel. Any resulting channel state change is emitted to any listeners registered using the {@link EventEmitter.on | `on()`} or {@link EventEmitter.once | `once()`} methods. Once all clients globally have detached from the channel, the channel will be released in the Ably service within two minutes.
+   * Detach from this channel. Any resulting channel state change is emitted to any registered listeners.
+   *
+   * Detaching stops the server sending messages on this channel to the client. Locally registered listeners remain registered.
```
m-hulbert: ... and behaviour will resume if they reattach? Maybe unnecessary.

ably.d.ts
```diff
@@ -2741,30 +2747,58 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
+   * @example
+   * ```ts
+   * await channel.detach();
+   * ```
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#detach
    */
   detach(): Promise<void>;
   /**
-   * Retrieves a {@link PaginatedResult} object, containing an array of historical {@link InboundMessage} objects for the channel. If the channel is configured to persist messages, then messages can be retrieved from history for up to 72 hours in the past. If not, messages can only be retrieved from history for up to two minutes in the past.
+   * Retrieves a {@link PaginatedResult} object, containing an array of historical {@link InboundMessage} objects for the channel. Messages are returned from storage only when message persistence is enabled for the channel by a rule. If message persistence is not enabled, only messages from the last two minutes are returned.
```
m-hulbert: Objects or messages?

ably.d.ts
```diff
@@ -2784,41 +2818,70 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
+   * @example
+   * ```ts
+   * await channel.setOptions({ params: { rewind: '1' } });
+   * ```
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#set-options
    */
   setOptions(options: ChannelOptions): Promise<void>;
   /**
-   * Registers a listener for messages with a given event name on this channel. The caller supplies a listener function, which is called each time one or more matching messages arrives on the channel.
+   * Registers a listener for messages with a given event name on this channel. The caller supplies a listener function, which is called each time one or more matching messages arrives on the channel. Implicitly attaches the channel unless {@link ChannelOptions.attachOnSubscribe} is `false`. Without the `subscribe` mode the listener never fires.
```
m-hulbert: `Without the `subscribe` mode the listener never fires.` I assume that's a silent failure?

ably.d.ts
```diff
@@ -2877,51 +2955,103 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
+   * ```ts
+   * const stateChange = await channel.whenState('attached');
+   * ```
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#when-state
    */
   whenState(targetState: ChannelState): Promise<ChannelStateChange | null>;
   /**
    * Retrieves the latest version of a specific message by its serial identifier.
    *
+   * The channel must be configured to allow message updates, enabled by a rule.
```
m-hulbert: Maybe a `@see` or link on 'rules' for info?

src/common/lib/client/auth.ts
m-hulbert: `The callback returned an empty string` I don't think that's part of a hint.

m-hulbert: I think this is mixing part of what went wrong with how to fix it.

m-hulbert: `since the key you passed has no colon-separated secret` - I'd argue this should be in the message itself rather than the hint.

src/common/lib/client/realtimeannotations.ts
```diff
@@ -76,7 +76,7 @@ class RealtimeAnnotations {
...
+        hint: 'Include "annotation_subscribe" in the channel modes: realtime.channels.get(name, { modes: ["annotation_subscribe", ...] }), or call channel.setOptions({ modes: [...] }) on an existing channel (this triggers a reattach). If the attach is then rejected, the channel namespace does not have "Message annotations, updates, and deletes" enabled (`ably apps rules list` shows which namespaces do). If the attach succeeds but annotations still are not delivered, your API key lacks the annotation-subscribe capability and the server silently dropped the mode (`ably auth keys list` shows your key\'s capabilities).',
```
m-hulbert: Would the part from "If the attach is then rejected..." not result in different errors?

src/common/lib/client/restchannel.ts
m-hulbert: Semi-colons to fullstops.

src/common/lib/types/basemessage.ts
m-hulbert: This one reads a bit oddly.

m-hulbert (commented):
These are looking a lot better - added a few comments but they're mostly about long sentences and replacing some semi-colons with fullstops to make things easier to read.

ably.d.ts
```diff
@@ -2667,62 +2709,80 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
+   * An array of {@link ResolvedChannelMode} objects reflecting the modes the server granted on the most recent attach. The server grants only the modes the key or token capability permits, so this may be a subset of the modes requested via {@link ChannelOptions.modes}.
+   *
+   * It is `undefined` before the channel attaches for the first time. When the server grants no modes, it is also `undefined`, not an empty array. Guard against `undefined` despite the declared type.
+   *
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#modes
    */
   modes: ResolvedChannelMode[];
   /**
-   * Deregisters the given listener for the specified event name. This removes an earlier event-specific subscription.
+   * Deregisters the given listener for the specified event name. This only removes the local listeners and does not detach the channel. The channel stays attached, so the server keeps streaming its messages to the client while the client's capability permits.
```
m-hulbert:
```suggestion
   * Deregisters the given listener for the specified event name. This only removes local listeners and does not detach the channel. If the channel is not detached, the server continues to stream messages to the client, if the client's capabilities permit it to receive them.
```

ably.d.ts
```diff
@@ -2667,62 +2709,80 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
-   * Deregisters all listeners for all event names in the array.
+   * Deregisters all listeners for all event names in the array. This only removes the local listeners and does not detach the channel. The channel stays attached, so the server keeps streaming its messages to the client while the client's capability permits.
    *
    * @param events - An array of event names.
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#unsubscribe
    */
   unsubscribe(events: Array<string>): void;
   /**
-   * Deregisters all listeners to messages on this channel that match the supplied filter.
+   * Deregisters listeners registered with the supplied {@link MessageFilter}. The filter is compared by object identity, so it must be the same object previously passed to {@link RealtimeChannel.subscribe | `subscribe()`}. A filter object not previously passed to `subscribe()` removes nothing and no error is raised. This only removes the local listeners and does not detach the channel. The channel stays attached, so the server keeps streaming its messages to the client while the client's capability permits.
```
m-hulbert: Could simplify this somewhat.

ably.d.ts
```diff
@@ -2789,41 +2878,70 @@ export declare interface RealtimeChannel extends EventEmitter<channelEventCallba
...
+   * @example
+   * ```ts
+   * await channel.subscribe({ name: 'event' }, (message) => console.log(message.data));
+   * ```
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-channel#subscribe
    */
   subscribe(filter: MessageFilter, listener?: messageCallback<InboundMessage>): Promise<ChannelStateChange | null>;
   /**
-   * Registers a listener for messages on this channel. The caller supplies a listener function, which is called each time one or more messages arrives on the channel.
+   * Registers a listener for messages on this channel. Implicitly attaches the channel unless {@link ChannelOptions.attachOnSubscribe} is `false`. Without the `subscribe` mode the listener never fires and no error is raised.
```
m-hulbert: Is `Without the `subscribe` mode the listener never fires and no error is raised` truly necessary or does this actually make reading the description more complex than it needs to be?

m-hulbert (approved):
A few minor comments left, but mostly LGTM now.

m-hulbert (approved):

PR #2244, opened 2026-06-10
DX-1206: RealtimePresence interface docstrings

umair-ably:
Rewrites the JSDoc on the `RealtimePresence` public surface in `ably.d.ts` per `docstringRules.md`: terse folded prose that surfaces call-site prerequisites, side-effects, and failure modes (the DX-1211 silent-failure class), with an `@see` companion link to the canonical reference. Previously stacked on #2243, now rebased onto `main` since that merged.

## Scope
`ably.d.ts` only (+108/−19). 11 members rewritten (counting overload groups once).

## Highlights
- **`subscribe()` (DX-1211 core):** documents the silent missing-`presence_subscribe`-mode trap — the channel attaches but the server never delivers presence events, so the listener never fires; also notes `attachOnSubscribe: false` skips the implicit attach. On both active overloads.
- **`get()`:** same mode prerequisite, but the failure shape differs — resolves with an **empty array** rather than rejecting. Also documents the one state that does reject: `suspended`, unless `waitForSync: false` is passed to read the last known members.
- **`syncComplete`:** the trap that a `true` value means only "no sync in progress" — it is also `true` once the local set is cleared, so it does not mean the set is populated — and points to `get()` as the way to wait for a sync.
- **`enter()`/`update()`/`leave()`:** identified-client prerequisite (rejects on missing/wildcard `clientId`), with pointers to the `*Client()` variants; `enter()` documents automatic re-enter on re-attach and that a re-enter failure surfaces as a channel `update` event, not a rejection.
- **`leave()`/`leaveClient()`:** neither implicitly attaches. Requires an `attached` channel; on an `attaching` channel the call is queued until it attaches unless `ClientOptions.queueMessages` is disabled, and it rejects in any other state or when the connection is unusable.
- **`enterClient()`/`updateClient()`/`leaveClient()`:** wildcard-`clientId` key/token prerequisite is server-enforced, so violations reject with a server-returned `ErrorInfo`.
- **`unsubscribe()`:** all six overloads clarify the removal is local-only — no detach, no presence leave.
- **`history()`:** non-code persistence channel-rule prerequisite, and the 2-minute default retention without it.

## Review passes since the first push
- Consistency and concision pass against the merged #2242 and #2243 prose: blank-line paragraphs, no semicolons, one shared wording per repeated prerequisite, and the merged `subscribe()` mode clause and `unsubscribe()` tail adopted verbatim. The `strictMode` mechanism moved out of `get()` to the `@see`, matching how RealtimeChannel shipped.
- Two claims corrected against `realtimepresence.ts`: `leave()` no longer asserts a client must have entered before leaving (the SDK enforces no such thing), and the `attaching` case is now described as queued rather than as proceeding.

## Validation
`npx typedoc --skipErrorChecking` and `prettier --check` pass; a clean TypeDoc build confirms every `{@link}` resolves and nothing became undocumented. (`npm run docs` is red on this branch and on `main` for an unrelated reason: `test/uts` sits in the TypeDoc tsconfig.)

## Reviewer notes
- `@see` anchors are now verified against the live reference pages, which went live with ably/docs #3400. All eleven resolve. `syncComplete` points at `#properties` rather than `#sync-complete`: the pages carry member-name alias anchors inside method headings, but properties are table rows under one `#properties` heading, so a per-property anchor does not exist. #2280 adds a CI check for exactly this.
- During this pass a **confirmed library bug** was found (presence ops queued while the channel is `attaching` can resolve silently without ever reaching the server in some failure paths). The queueing itself is now documented, but the bug is deliberately *not*, pending a decision on whether to fix the code instead.

## Summary by CodeRabbit

* **Documentation**
  * Clarified TypeScript documentation for realtime presence operations, including `syncComplete`, `get`, `subscribe`, `unsubscribe`, and `history`.
  * Expanded guidance on attach/queuing behavior, server delivery constraints, client identification rules (identified vs wildcard), and implicit-attach differences.
  * Documented when historical presence messages are returned and how presence membership is re-established after re-attaches.
  * Confirmed no public type signatures or APIs were changed.

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -2230,6 +2230,8 @@ export declare interface RealtimePresence {
    *
    * Requires the `presence_subscribe` mode. Without it the call resolves with an empty array.
    *
+   * On a channel in the `suspended` state the call rejects with an {@link ErrorInfo}, unless `waitForSync` is `false` in {@link RealtimePresenceParams}, which resolves with the last known members.
+   *
    * @param params - A set of parameters which are used to specify which presence members should be retrieved.
    * @returns A promise which, upon success, will be fulfilled with an array of {@link PresenceMessage} objects. Upon failure, the promise will be rejected with an {@link ErrorInfo} object which explains the error.
    * @example
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -2168,51 +2168,69 @@ export declare interface Presence {
  */
 export declare interface RealtimePresence {
   /**
-   * Indicates whether the presence set synchronization between Ably and the clients on the channel has been completed. Set to `true` when the sync is complete.
+   * Indicates whether the presence set synchronization between Ably and the clients on the channel has been completed. Set to `true` when the sync is complete, and back to `false` whenever a new sync starts (typically after a (re)attach). The value is also `true` after the local presence set is cleared (when the channel becomes detached or failed, or attaches without the server reporting any presence members), so it indicates only that no sync is in progress (though it is initially `false`, before any sync has started), not that the channel is attached or that a sync ever ran. To wait for an in-progress sync, call {@link RealtimePresence.get | `get()`}, which by default resolves only once the sync completes.
+   *
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-presence#sync-complete
    */
   syncComplete: boolean;
   /**
-   * Deregisters a specific listener that is registered to receive {@link PresenceMessage} on the channel for a given {@link PresenceAction}.
+   * Deregisters a specific listener that is registered to receive {@link PresenceMessage} on the channel for a given {@link PresenceAction}. This only removes the local listener; it does not detach the channel or remove this client from the presence set (use {@link RealtimePresence.leave | `leave()`} for that), and presence events may continue to arrive for any other registered listeners.
    *
    * @param presence - A specific {@link PresenceAction} to deregister the listener for.
    * @param listener - An event listener function.
+   * @example
+   * ```ts
+   * channel.presence.unsubscribe('enter', listener);
+   * ```
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-presence#unsubscribe
    */
   unsubscribe(presence: PresenceAction, listener: messageCallback<PresenceMessage>): void;
   /**
-   * Deregisters a specific listener that is registered to receive {@link PresenceMessage} on the channel for a given array of {@link PresenceAction} objects.
+   * Deregisters a specific listener that is registered to receive {@link PresenceMessage} on the channel for a given array of {@link PresenceAction} objects. This only removes the local listener and does not detach the channel.
    *
    * @param presence - An array of {@link PresenceAction} objects to deregister the listener for.
    * @param listener - An event listener function.
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-presence#unsubscribe
    */
   unsubscribe(presence: Array<PresenceAction>, listener: messageCallback<PresenceMessage>): void;
   /**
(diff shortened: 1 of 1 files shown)


ably.d.ts
```diff
@@ -2168,51 +2168,75 @@ export declare interface Presence {
...
+   * A `true` value means only that no sync is in progress. It does not mean the presence set is populated, since the flag is also `true` once the local set is cleared, and it returns to `false` whenever a new sync starts.
+   *
+   * To wait for an in-progress sync, call {@link RealtimePresence.get | `get()`}, which by default resolves only once the sync completes.
+   *
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-presence#properties
    */
   syncComplete: boolean;
   /**
-   * Deregisters a specific listener that is registered to receive {@link PresenceMessage} on the channel for a given {@link PresenceAction}.
+   * Deregisters a specific listener that is registered to receive {@link PresenceMessage} on the channel for a given {@link PresenceAction}. This removes local listeners only and does not detach the channel or remove this client from the presence set.
```
m-hulbert: Do we need the note we had about still being streamed messages if you have the capabilities, unless the channel is detached? Can take the wording from regular unsubscribe()

ably.d.ts
```diff
@@ -2306,10 +2354,17 @@ export declare interface RealtimePresence {
    */
   enter(data: any, callback: ErrorCallback): void;
   /**
-   * Updates the `data` payload for a presence member. If called before entering the presence set, this is treated as an {@link PresenceActions.ENTER} event.
+   * Updates the `data` payload for a presence member. If called before entering the presence set, this is treated as an {@link PresenceActions.ENTER} event. Implicitly attaches the channel if it is not already attached.
+   *
+   * Requires an identified client. If the `clientId` is unset, or is the wildcard `*`, the call rejects with an {@link ErrorInfo}. Use {@link RealtimePresence.updateClient | `updateClient()`} to update on behalf of another identity.
```
m-hulbert, line 2359: Would you ever get to the position of being able to call `update()` or `leave()` without having a `clientId` set?

umair-ably: pretty easily i think... someone could be using an api key or a token with no client id set? this method also isn't gated on a prior enter, but instead can act like an enter too

ably.d.ts
```diff
@@ -2384,7 +2384,7 @@ export declare interface RealtimePresence {
   /**
    * Leaves the presence set for the channel.
    *
-   * Requires an identified client. If the `clientId` is unset, or is the wildcard `*`, the call rejects with an {@link ErrorInfo}. Use {@link RealtimePresence.leaveClient | `leaveClient()`} to leave on behalf of another identity, which requires a wildcard `clientId` on the API key or token.
+   * Requires an identified client. If the `clientId` is unset, or is the wildcard `*`, the call rejects with an {@link ErrorInfo}. Use {@link RealtimePresence.leaveClient | `leaveClient()`} to leave on behalf of another identity.
```
umair-ably: leaveClient()'s own docstring already states the wildcard prerequisite, so leave() restating it is duplication... not removing this would mean needing to highlight the whole  "...requires an API key or token bound to a wildcard clientId, which lets the client claim any clientId per operation but does not give it an identity of its own." again

ably.d.ts
m-hulbert: Ah, yeah fair point on `update()` being treated as an `enter()`.

The same doesn't apply for `leave()` though, right?

umair-ably: fair! [updated](https://github.com/ably/ably-js/pull/2244/commits/2d3e75056199aa31aa10ad0b368cb16b0b07300e)

m-hulbert (approved):

PR #2245, opened 2026-06-10
DX-1618: RealtimeAnnotations & RestAnnotations interface docstrings

umair-ably:
Stacked on #2244. One commit per change:

1. **Docstrings** — rewrite all `RealtimeAnnotations`/`RestAnnotations` docstrings so prerequisites, side-effects and failure modes are discoverable at the call site (the DX-1211 effort): the default-off `annotation_subscribe` mode, the mutable-messages channel rule, the `annotation-publish`/`annotation-subscribe` capabilities, the unencrypted-annotation-data caveat, and each member's reject behaviour. Server-side failure shapes were verified by live traces against sandbox.
2. **`RestAnnotations.delete` declaration** — the implementation has existed since 04fed9c4 but was never declared on the interface, so the working method is uncallable from TypeScript. Public API addition (changelog-worthy).
3. **src fixes** — annotation validation errors now name the calling method instead of always saying `annotations.publish()`, and the 93001 hint no longer tells users to expect a rejected attach for a missing capability (the server actually silently grants fewer modes; live-traced).

## Summary by CodeRabbit

* **Documentation**
  * Expanded annotation API documentation covering attachment, permissions, namespaces, validation, channel state, encryption, default actions, and serial requirements.
* **Bug Fixes**
  * Improved annotation validation errors by identifying the operation that caused the issue.
  * Added clearer validation handling for publishing, deleting, and retrieving annotations.


src/common/lib/client/realtimeannotations.ts
```diff
@@ -23,7 +23,7 @@ class RealtimeAnnotations {
 
   async publish(msgOrSerial: string | Message, annotationValues: Partial<Properties<Annotation>>): Promise<void> {
     const channelName = this.channel.name;
-    const annotation = constructValidateAnnotation(msgOrSerial, annotationValues);
+    const annotation = constructValidateAnnotation(msgOrSerial, annotationValues, 'publish');
```
m-hulbert, line 26: Looks like a dodgy rebase here and in restannotations.ts?

ably.d.ts
```diff
@@ -2466,96 +2466,195 @@ export declare interface RealtimePresence {
...
-   * Note that if you want to receive individual realtime annotations (instead of just the rolled-up summaries), you will need to request the annotation_subscribe ChannelMode in ChannelOptions, since they are not delivered by default. In general, most clients will not bother with subscribing to individual annotations, and will instead just look at the summary updates.
+   * Registers a listener that is called each time an {@link Annotation} matching a given type is received on the channel. Implicitly attaches the channel unless {@link ChannelOptions.attachOnSubscribe} is `false`.
+   *
+   * Individual annotations are delivered only to clients that request the `annotation_subscribe` mode via {@link ChannelOptions.modes}, since most clients can rely on the rolled-up summary updates instead.
+   *
+   * The mode requires message annotations to be enabled for the channel's namespace by a [rule](https://ably.com/docs/messages/annotations), without which the attach rejects with an {@link ErrorInfo}. It also requires the `annotation-subscribe` capability on the key or token, without which the server silently drops the mode.
```
m-hulbert: I don't think it's the mode that requires the rule necessarily. 

The part about requiring the capability should probably be next to the mode itself, too I'd have thought.

ably.d.ts
```diff
@@ -2466,96 +2466,195 @@ export declare interface RealtimePresence {
...
+   *
+   * The call rejects with an {@link ErrorInfo} on a channel in the `failed` state, before the listener is registered. It also rejects when the channel attaches without the mode, in which case the listener stays registered. When {@link ChannelOptions.attachOnSubscribe} is `false` and the channel is not yet attached, the call resolves and a listener that lacks the mode never fires.
```
m-hulbert: >It also rejects when the channel attaches without the mode, in which case the listener stays registered.

Is this correct?

ably.d.ts
```diff
@@ -2466,96 +2466,195 @@ export declare interface RealtimePresence {
...
+   * @example
+   * ```ts
+   * await channel.annotations.subscribe((annotation) => console.log(annotation.name));
+   * ```
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-annotations#annotations-subscribe
    */
   subscribe(listener?: messageCallback<Annotation>): Promise<void>;
   /**
-   * Deregisters a specific listener that is registered to receive {@link Annotation} on the channel for a given type.
+   * Deregisters a specific listener that is registered to receive {@link Annotation} on the channel for a given type. This removes local listeners only and does not detach the channel. The server keeps streaming annotations to an attached channel while the `annotation_subscribe` mode is granted.
```
m-hulbert: This makes it sound like the mode is what's controlling the streaming and you should use the mode to stop it.

ably.d.ts
```diff
@@ -2466,96 +2466,195 @@ export declare interface RealtimePresence {
...
+   * ```ts
+   * await channel.annotations.publish(message, { type: 'emoji:distinct.v1', name: '👍' });
+   * ```
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-annotations#annotations-publish
    */
   publish(message: Message, annotation: OutboundAnnotation): Promise<void>;
   /**
-   * Publish a new annotation for a message (alternative form where you only have the
-   * serial of the message, not a complete Message object)
+   * Publish a new annotation for a message (alternative form where you only have the serial of the message, not a complete {@link Message} object). If the annotation specifies no `action`, it defaults to `annotation.create`.
```
m-hulbert: The part in brackets suggests you've read the other method. We should remove the part about it being an alternative.

ably.d.ts
```diff
@@ -2466,96 +2466,195 @@ export declare interface RealtimePresence {
...
+   *
+   * The supplied serial or {@link Message} must carry a populated `serial`. A newly constructed {@link Message} has none, so the call rejects with an {@link ErrorInfo}.
```
m-hulbert: This then states you can pass a message, but the above paragraph suggests you only have the serial.

ably.d.ts
```diff
@@ -2466,96 +2466,195 @@ export declare interface RealtimePresence {
...
+   * await channel.annotations.delete(message, { type: 'emoji:distinct.v1', name: '👍' });
+   * ```
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-annotations#annotations-delete
    */
   delete(message: Message, annotation: OutboundAnnotation): Promise<void>;
   /**
-   * Publish an annotation removal request for a message, to remove it from the summary
-   * summaries. The semantics of the delete (and what fields are required) are different
-   * for each annotation type; see annotation types documentation for more details.
+   * Publish an annotation removal request for a message, to remove it from the message's annotation summary (alternative form where you only have the serial of the message, not a complete {@link Message} object). The semantics of a delete, and the fields it requires, differ by annotation type.
```
m-hulbert: Similar comment to above, but we also took the paraentheses out above.

ably.d.ts
```diff
@@ -2466,96 +2466,195 @@ export declare interface RealtimePresence {
...
+   * await channel.annotations.delete(message.serial, { type: 'emoji:distinct.v1', name: '👍' });
+   * ```
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-annotations#annotations-delete
    */
   delete(messageSerial: string, annotation: OutboundAnnotation): Promise<void>;
   /**
-   * Get all annotations for a given message (as a paginated result)
+   * Get all annotations for a given message, as a paginated result.
+   *
+   * This is a REST request even on a realtime client. It does not implicitly attach the channel and can be called in any channel or connection state. It does not require the `annotation_subscribe` mode.
```
m-hulbert: Is the fact it's a REST request relevant?

ably.d.ts
```diff
@@ -2747,56 +2846,114 @@ export declare interface Channel {
...
+   * @example
+   * ```ts
+   * await channel.annotations.delete(messageSerial, { type: 'emoji:distinct.v1', name: '👍' });
+   * ```
+   * @see https://ably.com/docs/pub-sub/api/javascript/rest/rest-annotations#annotations-delete
    */
   delete(messageSerial: string, annotation: OutboundAnnotation): Promise<void>;
   /**
-   * Get all annotations for a given message (as a paginated result)
+   * Get all annotations for a given message, as a paginated result.
```
m-hulbert: "as a paginated result" isn't something we've included in the other descriptions I don't think. That's just info for the return. (or if we did then we linked to `PaginatedResult`.

src/common/lib/client/realtimeannotations.ts
umair-ably: Not a rebase artefact: the extracted `_publish()` was sitting below `delete()`, so the diff rendered `delete()` as a moved method. `2eeea742` puts it directly after `publish()`, leaving `delete()` in place with only its call line changed. The `methodName` argument is what makes the validation error say `annotations.delete()` rather than always `annotations.publish()`.

ably.d.ts
umair-ably: Agreed on both counts, fixed in `2eeea742`. The capability now sits with the mode it gates, and the namespace rule is its own paragraph with its own consequence (without it the attach rejects).

ably.d.ts
umair-ably: Yes: the listener is registered before the attach and the 93001 path never removes it, with a comment in `realtimeannotations.ts` saying so deliberately. Since `_processIncoming` still emits to it, `2eeea742` now adds what to do about it, rather than leaving it as a curiosity.

ably.d.ts
umair-ably: Fair, that did read as a switch. `2eeea742` uses the channel and presence wording instead, "subject to the client's capabilities".

ably.d.ts
umair-ably: Removed in `2eeea742`. Each overload now documents its own declared parameter, so the cross-reference isn't needed.

ably.d.ts
umair-ably: Same fix: the `Message` overload keeps "The supplied `Message` must carry a populated `serial`", the serial overload now just requires a non-empty string. This matches how merged `getMessage` and `updateMessage` already vary that sentence by signature.

ably.d.ts
umair-ably: Done, dropped from all six serial-form overloads in `2eeea742`.

ably.d.ts
umair-ably: No, it's mechanism. `2eeea742` drops it and keeps the three consequences: no implicit attach, any channel or connection state, no mode required. Merged `RealtimeChannel.history` doesn't mention it either.

ably.d.ts
umair-ably: Dropped from all four `get` descriptions in `2eeea742`; `@returns` already links `PaginatedResult`. Worth noting the merged `history` members do open by naming it, so happy to link it instead if you'd prefer parity.

ably.d.ts
```diff
@@ -2466,96 +2466,195 @@ export declare interface RealtimePresence {
...
+   * await channel.annotations.delete(message.serial, { type: 'emoji:distinct.v1', name: '👍' });
+   * ```
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-annotations#annotations-delete
    */
   delete(messageSerial: string, annotation: OutboundAnnotation): Promise<void>;
   /**
-   * Get all annotations for a given message (as a paginated result)
+   * Get all annotations for a given message.
+   *
+   * Does not implicitly attach the channel, and can be called in any channel or connection state. It does not require the `annotation_subscribe` mode.
```
m-hulbert, line 2625: Minor: but I don't think the attach point is relevant here as we're documenting a behaviour doesn't happen.

ably.d.ts
```diff
@@ -2466,96 +2466,195 @@ export declare interface RealtimePresence {
...
+   * ```
+   * @see https://ably.com/docs/pub-sub/api/javascript/realtime/realtime-annotations#annotations-get
    */
   get(message: Message, params: GetAnnotationsParams | null): Promise<PaginatedResult<Annotation>>;
   /**
-   * Get all annotations for a given message (as a paginated result) (alternative form
-   * where you only have the serial of the message, not a complete Message object)
+   * Get all annotations for a given message.
+   *
+   * Does not implicitly attach the channel, and can be called in any channel or connection state. It does not require the `annotation_subscribe` mode.
```
m-hulbert, line 2644: As above.

m-hulbert (approved):
Approved with 2 minor nitpicks from the docstring side.

ably.d.ts
umair-ably: Discussed offline but not addressing as the Realtime versions of these methods do implicitly attach, so keeping these in case humans/LLMs assume the same for the REST methods

AndyTWF (approved):
Mark asked me to check the actual SDK changes, which LGTM

PR #2248, opened 2026-06-15
feat(react): allow omitting channelName to use the nearest ChannelProvider

owenpearson:
[AIT-985](https://ably.atlassian.net/browse/AIT-985)

Implements https://ably.atlassian.net/wiki/spaces/CHN/pages/5150801924/PSDR18+Optional+channelName+in+ably-js+React+channel+hooks

[AIT-985]: https://ably.atlassian.net/browse/AIT-985?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * React hooks now support omitting the channel name to infer it from the nearest `ChannelProvider`, with correct nested precedence and `ablyId` scoping.
  * Added listener-only/callback-only overloads for `useChannel`, `usePresenceListener`, and `useChannelStateListener`, plus optional channelName support in `useChannelInstance` and `usePresence`.
* **Documentation**
  * Expanded guidance on channel-name inference, including the required `usePresence(undefined, presenceData)` usage to avoid ambiguity.
* **Bug Fixes**
  * Improved error handling when neither an explicit channel name nor a nearby provider channel is available.
* **Tests**
  * Added coverage for inference, nested selection, presence behavior, and missing-provider failure cases.

### DIFF
--- a/docs/react.md
+++ b/docs/react.md
@@ -71,6 +71,36 @@ const { channel } = useChannel('your-channel-name', (message) => {
 
 Every time a message is sent to `your-channel-name` it'll be logged to the console. You can do whatever you need to with those messages.
 
+#### Omitting the channel name
+
+When a channel-based hook (`useChannel`, `usePresence`, `usePresenceListener`, `useChannelStateListener`) is called without a channel name, it resolves to the channel of the closest enclosing `ChannelProvider`. This avoids repeating the name when there is a surrounding provider:
+
+```jsx
+<ChannelProvider channelName="your-channel-name">
+  <Component />
+</ChannelProvider>;
+
+// Inside Component — no channel name needed:
+const { channel } = useChannel((message) => {
+  console.log(message);
+});
+```
+
+When providers are nested, the nearest one wins. You can still pass an explicit channel name to target a specific enclosing provider by name:
+
+```jsx
+<ChannelProvider channelName="outer">
+  <ChannelProvider channelName="inner">
+    {/* useChannel(...) resolves to "inner"; useChannel("outer", ...) resolves to "outer" */}
+  </ChannelProvider>
+</ChannelProvider>
+```
+
+> [!NOTE]
+> Channel resolution is scoped by `ablyId`. A hook called without an `ablyId` uses the `default` one, so it only resolves to a `ChannelProvider` that also uses the `default` `ablyId`. If your `ChannelProvider` uses a non-default `ablyId`, you must pass it to the hook for the channel to be inferred, for example `useChannel({ ablyId: 'your-ably-id' })`.
+
+<!-- -->
+
 > [!NOTE]
 > Our react hooks are designed to run on the client-side, so if you are using server-side rendering, make sure that your components which use Ably react hooks are only rendered on the client side.
 
@@ -196,6 +226,13 @@ usePresence('your-channel-name', { foo: 'bar' });
 usePresence('another-channel-name', 123);
 usePresence('third-channel-name', 'initial status');
 ```
+
+> [!NOTE]
+> Unlike the other channel-based hooks, `usePresence` cannot infer the channel from the nearest `ChannelProvider` when you call it with presence data only. Because the presence data can itself be a string or object, it is indistinguishable from a channel name or options object, so `usePresence(presenceData)` would be treated as a channel name. To infer the channel while still passing presence data, pass `undefined` as the first argument:
+>
+> ```javascript
+> usePresence(undefined, { foo: 'bar' });
+> ```
 
 If you're using `TypeScript` there are type hints to make sure that value passed to `updateStatus` is of the same `type` as your initial constraint, or a provided generic type parameter:
 
--- a/src/platform/react-hooks/src/AblyReactHooks.ts
+++ b/src/platform/react-hooks/src/AblyReactHooks.ts
@@ -1,7 +1,13 @@
 import * as Ably from 'ably';
 
 export type ChannelNameAndOptions = {
-  channelName: string;
+  /**
+   * The name of the channel to use. When omitted, channel hooks resolve to the
+   * channel named by the closest enclosing `ChannelProvider`. Resolution is
+   * scoped by `ablyId`, so a channel is only inferred from a `ChannelProvider`
+   * that uses the same `ablyId` (the `default` one when none is provided).
+   */
+  channelName?: string;
   ablyId?: string;
   skip?: boolean;
 
--- a/src/platform/react-hooks/src/hooks/useChannel.ts
+++ b/src/platform/react-hooks/src/hooks/useChannel.ts
@@ -19,6 +19,8 @@ export interface ChannelResult {
 
 type SubscribeArgs = [string, AblyMessageCallback] | [AblyMessageCallback];
 
+// The channel name can be omitted to use the channel from the nearest ChannelProvider.
+export function useChannel(callbackOnMessage?: AblyMessageCallback): ChannelResult;
 export function useChannel(
   channelNameOrNameAndOptions: ChannelParameters,
   callbackOnMessage?: AblyMessageCallback,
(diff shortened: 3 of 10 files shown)


src/platform/react-hooks/src/hooks/useChannel.ts
```diff
@@ -30,19 +32,34 @@ export function useChannel(
...
 
 export function useChannel(
-  channelNameOrNameAndOptions: ChannelParameters,
+  channelNameOrNameAndOptionsOrCallback?: ChannelParameters | AblyMessageCallback,
   eventOrCallback?: string | AblyMessageCallback,
   callback?: AblyMessageCallback,
 ): ChannelResult {
+  // The first argument is the message callback when the channel name is
+  // inferred from a surrounding ChannelProvider, e.g. useChannel(listener).
+  const inferChannelFromCallback = typeof channelNameOrNameAndOptionsOrCallback === 'function';
```
VeskeR: nit: it's not that we _infer the channel name from the callback_ (we can't infer it from the _callback_), it's rather if first argument _is callback_, then we should infer from the nearest provider.
So I think this should be `inferChannelFromCallback` -> `inferChannelFromNearestProvider`

src/platform/react-hooks/src/hooks/useChannelStateListener.ts
```diff
@@ -16,22 +19,34 @@ export function useChannelStateListener(
...
 
 export function useChannelStateListener(
-  channelNameOrNameAndAblyId: ChannelNameAndAblyId | string,
+  channelNameOrNameAndAblyIdOrListener?: ChannelNameAndAblyId | string | ChannelStateListener,
   stateOrListener?: Ably.ChannelState | Ably.ChannelState[] | ChannelStateListener,
   listener?: (stateChange: Ably.ChannelStateChange) => any,
 ) {
+  // The first argument is the listener when the channel name is inferred from a
+  // surrounding ChannelProvider, e.g. useChannelStateListener(listener).
+  const inferChannelFromListener = typeof channelNameOrNameAndAblyIdOrListener === 'function';
```
VeskeR: nit: ditto variable name

src/platform/react-hooks/src/hooks/usePresence.ts
```diff
@@ -14,14 +14,17 @@ export interface PresenceResult<T> {
   channelError: Ably.ErrorInfo | null;
 }
 
+// The channel name can be omitted to use the channel from the nearest ChannelProvider.
 export function usePresence<T = any>(
-  channelNameOrNameAndOptions: ChannelParameters,
+  channelNameOrNameAndOptions?: ChannelParameters,
```
VeskeR, line 27: this should also allow `| T` and the code below should branch based on the first argument could be a message data `T` - similar to what `useChannel` does now.
As it stands right now, calling `usePresence(messageData)` will have an unexpected result - `messageData` is treated as channel options, underlying `channel.presence.enter` is called with `undefined`.

VeskeR: UPDATE: having a second thought - `| T` wouldn't work, would it, as the presence data itself can be of any type, usually a `string` or `object`, indistinguishable from the `ChannelParameters` which themselves are `string` or `object`.
This does go against the nice "just omit the name" mental mode we have for AIT SDK/chat and for  the rest of the hooks. I can see this wouldn't been a problem if `usePresence` accepted only one parameter to begin with, so you would have `usePresence({ channelName, data })`, so you would just drop channel name, but this goes beyond what we're trying to achieve in this PR. Should we just bite the bullet and document explicitly that for `usePresence` you have to write `usePresence(undefined, data)` if you want to infer the channel name but still pass the message data? WDYT?

src/platform/react-hooks/src/hooks/usePresenceListener.ts
```diff
@@ -17,24 +17,45 @@ export interface PresenceListenerResult<T> {
...
   onPresenceMessageReceived?: OnPresenceMessageReceived<T>,
+): PresenceListenerResult<T>;
+
+export function usePresenceListener<T = any>(
+  channelNameOrNameAndOptionsOrListener?: ChannelParameters | OnPresenceMessageReceived<T>,
+  onPresenceMessageReceived?: OnPresenceMessageReceived<T>,
 ): PresenceListenerResult<T> {
+  // The first argument is the listener when the channel name is inferred from a
+  // surrounding ChannelProvider, e.g. usePresenceListener(listener).
+  const inferChannelFromListener = typeof channelNameOrNameAndOptionsOrListener === 'function';
```
VeskeR: nit: ditto variable name

src/platform/react-hooks/src/hooks/useInferredChannelName.test.tsx
```diff
@@ -0,0 +1,219 @@
...
+      ).toThrowError(/no parent ChannelProvider to infer one from/);
+    } finally {
+      consoleError.mockRestore();
+    }
+  });
+
+  /** @nospec */
+  it('usePresenceListener(listener) listens on the surrounding channel', async () => {
+    const Component = () => {
+      const { presenceData } = usePresenceListener<{ foo: string }>();
```
VeskeR: this doesn't actually pass in `listener` like the test name states

src/platform/react-hooks/src/hooks/useInferredChannelName.test.tsx
```diff
@@ -0,0 +1,219 @@
...
+      await otherClient.channels.get('inner').publish({ text: 'hello' });
+    });
+
+    const list = screen.getAllByRole('messages')[0];
+    expect(list.childElementCount).toBe(1);
+    expect(list.children[0].innerHTML).toBe('hello');
+  });
+
+  /** @nospec */
+  it('useChannel() returns a publish bound to the surrounding channel', async () => {
```
VeskeR, line 60: would also like a test to check the publish also works for a derived surrounding channel

docs/react.md
```diff
@@ -71,6 +71,31 @@ const { channel } = useChannel('your-channel-name', (message) => {
 
 Every time a message is sent to `your-channel-name` it'll be logged to the console. You can do whatever you need to with those messages.
 
+#### Omitting the channel name
+
+When a channel based hook (`useChannel`, `usePresence`, `usePresenceListener`, `useChannelStateListener`) is called without a channel name, it resolves to the channel of the closest enclosing `ChannelProvider`. This avoids repeating the name when there is a surrounding provider:
```
VeskeR: One caveat though: `ChannelProvider`'s context is keyed per-`ablyId`, by default  we use `default` value.
So strictly speaking, `useChannel()` (with `ablyId` equal to `default` implicitly) will not pick up a `<ChannelProvider ablyId="x">`; you still need to use `useChannel({ ablyId: 'x' })`.
I'd say this is intended, but we should be more clear in the docs and in the jsdocs for `ChannelNameAndOptions`

VeskeR (changes requested):
Couple of nitpicks, and a question about `usePresence(undefined, data)` and inferring with non-default `ablyId`.

src/platform/react-hooks/src/hooks/usePresence.ts
owenpearson: Interesting.. I think usePresence({ channelName, data }) is the right API long term, but agree that it's not in the scope of this PR and it's a bit messy having two ways to supply presence data (and what if someone provides both?). I think just documenting it explicitly with `usePresence({}, data)` is probably fine for now and it doesn't make anything worse, so I think I'll do that

owenpearson: done that now 👍

owenpearson:
The failing tests here are all from the pre-existing error code issue, relevant tests are passing

src/platform/react-hooks/src/AblyReactHooks.ts
```diff
@@ -1,7 +1,11 @@
 import * as Ably from 'ably';
 
 export type ChannelNameAndOptions = {
-  channelName: string;
+  // Optional: when omitted, channel hooks resolve to the channel named by the
+  // closest enclosing `ChannelProvider`. Resolution is scoped by `ablyId`, so a
+  // channel is only inferred from a `ChannelProvider` that uses the same `ablyId`
+  // (the `default` one when none is provided).
```
VeskeR: Should this be a jsdoc? I believe the field is exposed to the developers, so they might benefit from the tooltip.

VeskeR (approved):
LGTM, one question about jsdoc vs code coment, up to you.

src/platform/react-hooks/src/AblyReactHooks.ts
owenpearson: Good point, changed it to jsdoc

PR #2253, opened 2026-06-26
feat(push): iOS Live Activity support via APNs broadcast channels

ttypic:
Spec: https://github.com/ably/specification/pull/500

Summary

Adds Push Admin support for managing iOS Live Activities through APNs broadcast channels. This introduces a two-part API: creating a broadcast channel, then driving the Live Activity lifecycle (start → update → end) over it.

1. push.admin.createApnsBroadcast(options)

Creates an APNs broadcast channel for an iOS Live Activity. Call once before starting the activity and persist the returned ids for the session.

```js
const { id, apnsChannelId } = await rest.push.admin.createApnsBroadcast({
  messageStoragePolicy: 1, // 1 = cache last update so late-joining devices get current state; 0 = disabled
});
```

- New `Admin.createApnsBroadcast` method (`POST /push/apnsBroadcastChannels`).
- New types: `PushApnsBroadcastOptions`, `PushApnsBroadcast`.

2. `push.admin.liveActivity` lifecycle API

A new LiveActivity class exposed as push.admin.liveActivity, managing the lifecycle of a Live Activity over a broadcast channel:

```js
// Push-to-start to all devices subscribed to the given channels
await rest.push.admin.liveActivity.start({
  recipient: { channels: ['channel-1'], deviceId: 'optional-device' },
  apnsBroadcast: id,
  apns: { /* APNs Live Activity start payload */ },
});

// Update content-state (single push; APNs fans out to subscribers)
await rest.push.admin.liveActivity.update({
  apnsBroadcast: id,
  apns: { /* APNs update payload */ },
  headers: { 'apns-priority': 5 },
});

// End the activity on all devices and clean up the channel
await rest.push.admin.liveActivity.end({
  apnsBroadcast: id,
  apns: { /* APNs end payload */ },
});
```

- start → `POST /push/apnsBroadcastChannels/{id}/start` (supports targeting all subscribed channels, optionally restricted to a single deviceId).
- update → `POST /push/apnsBroadcastChannels/{id}/broadcast` (optional apns-priority / apns-expiration headers).
- end → `POST /push/apnsBroadcastChannels/{id}/end`; the broadcast id is no longer valid afterwards.
- APNs payloads are passed through as-is.
- New types: `PushLiveActivity`, `PushLiveActivityStartParams`, `PushLiveActivityUpdateParams`, `PushLiveActivityEndParams`.

## Summary by CodeRabbit

* **New Features**
  * Added iOS Live Activity support via APNs broadcast channels.
  * Introduced APIs to create APNs broadcast channels and manage Live Activities (start, update, end), including support for APNs payload and optional request headers.
* **Tests**
  * Added REST unit tests for APNs broadcast channel creation and Live Activity start/update/end requests, validating request structure and HTTP error handling.

### DIFF
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -3767,13 +3767,13 @@ export declare interface PushLiveActivityStartParams {
         /**
          * One or more Ably channel names. Devices subscribed to any of these channels will receive the push-to-start notification. Provide either `channels` or `deviceId`.
          */
-        channels?: string[];
+        channels: string[];
       }
     | {
         /**
          * Restrict the push-to-start notification to a single device. Provide either `channels` or `deviceId`.
          */
-        deviceId?: string;
+        deviceId: string;
       };
   /**
    * The `id` returned from {@link PushAdmin.createApnsBroadcast}.
@@ -3831,6 +3831,8 @@ export declare interface PushLiveActivityEndParams {
 export declare interface PushApnsBroadcastOptions {
   /**
    * Set to `1` to cache the last update payload so late-joining devices receive the current content state immediately on subscription. Set to `0` to disable caching.
+   *
+   * @see https://developer.apple.com/documentation/usernotifications/sending-channel-management-requests-to-apns
    */
   messageStoragePolicy: 0 | 1;
 }
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -3703,6 +3703,39 @@ export declare interface PushAdmin {
    * @returns A promise which resolves upon success of the operation and rejects with an {@link ErrorInfo} object upon its failure.
    */
   publish(recipient: any, payload: any): Promise<void>;
+  /**
+   * Creates an APNs broadcast channel for use with an iOS Live Activity. Call once before starting the Live Activity and persist the returned ids for the session.
+   *
+   * @experimental This is a preview feature and may change in a future non-major release.
+   *
+   * @param options - Options for the broadcast, including the `messageStoragePolicy`.
+   * @returns A promise resolving to the broadcast `{ id, apnsChannelId }`.
+   */
+  createApnsBroadcast(options: PushApnsBroadcastOptions): Promise<PushApnsBroadcast>;
+}
+
+/**
+ * Options for creating an APNs broadcast channel via {@link PushAdmin.createApnsBroadcast}.
+ */
+export declare interface PushApnsBroadcastOptions {
+  /**
+   * Set to `1` to cache the last update payload so late-joining devices receive the current content state immediately on subscription. Set to `0` to disable caching.
+   */
+  messageStoragePolicy: 0 | 1;
+}
+
+/**
+ * The result of creating an APNs broadcast channel via {@link PushAdmin.createApnsBroadcast}.
+ */
+export declare interface PushApnsBroadcast {
+  /**
+   * The opaque Ably broadcast id.
+   */
+  id: string;
+  /**
+   * The Apple-assigned channel id.
+   */
+  apnsChannelId: string;
 }
 
 /**
--- a/ably.d.ts
+++ b/ably.d.ts
@@ -3712,6 +3712,117 @@ export declare interface PushAdmin {
    * @returns A promise resolving to the broadcast `{ id, apnsChannelId }`.
    */
   createApnsBroadcast(options: PushApnsBroadcastOptions): Promise<PushApnsBroadcast>;
+  /**
+   * Controls the lifecycle of iOS Live Activities over an APNs broadcast channel created with {@link PushAdmin.createApnsBroadcast}.
+   *
+   * @experimental This is a preview feature and may change in a future non-major release.
+   */
+  liveActivity: PushLiveActivity;
+}
+
+/**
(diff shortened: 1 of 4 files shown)


src/common/lib/client/push.ts
```diff
@@ -218,8 +315,6 @@ class DeviceRegistrations {
 
     Utils.mixin(headers, client.options.headers);
 
-    if (client.options.pushFullWait) Utils.mixin(params, { fullWait: 'true' });
```
VeskeR, line 221: what happened here? why is this removed in this PR?

VeskeR: Also would like to know what this `fullWait` is and if we need/don't need/decided not to use for live activities

ably.d.ts
```diff
@@ -3703,6 +3703,147 @@ export declare interface PushAdmin {
...
+}
+
+/**
+ * Parameters for {@link PushLiveActivity.start}.
+ */
+export declare interface PushLiveActivityStartParams {
+  /**
+   * The targeted recipients of the push-to-start notification.
+   */
+  recipient: {
```
VeskeR: [RSH1e1](https://github.com/ably/specification/pull/500/changes#diff-50fdfe2f469d90cc9711ac4f5fd1810ff73ca731ec38ce20ee5297ad2648fa57R1053) describes this as 
> is either a list of `channels` or a single `deviceId`

Can we improve the type by having a union between `{ channels } | { deviceId }` so at least one field is required?

VeskeR: and possible have an explicit check and throw an error in the `.start()` method?
and then a corresponding test for possible combinations

ably.d.ts
```diff
@@ -3703,6 +3703,147 @@ export declare interface PushAdmin {
...
+   */
+  apnsBroadcast: string;
+  /**
+   * A valid APNs Live Activity start payload. The payload is passed through to APNs as-is.
+   */
+  apns: any;
+  /**
+   * Optional APNs delivery headers, such as `apns-priority` and `apns-expiration`. When supplied, these are included in the request body under a `headers` key.
+   */
+  headers?: any;
```
VeskeR: can we at least use `Record<string, any>` for headers in public types, like in the implementation?
same for all headers below

ably.d.ts
```diff
@@ -3703,6 +3703,147 @@ export declare interface PushAdmin {
...
+}
+
+/**
+ * Options for creating an APNs broadcast channel via {@link PushAdmin.createApnsBroadcast}.
+ */
+export declare interface PushApnsBroadcastOptions {
+  /**
+   * Set to `1` to cache the last update payload so late-joining devices receive the current content state immediately on subscription. Set to `0` to disable caching.
+   */
+  messageStoragePolicy: 0 | 1;
```
VeskeR, line 3837: can we have an actual string enum for this in public API types? and then decode to the integer when sending to the server

src/common/lib/client/push.ts
ttypic: deleted by accident, it's debug option, if true it's wait until all expensive denormalization happens for device, we don't have this on liveactivity

ably.d.ts
ttypic: added error, narrowed down type, but it's not for UTS, it won't matter for statically-typed languages

ably.d.ts
ttypic: used Record<string, string>, because backend expects string anyway, it will throw an error if something else will come

ably.d.ts
ttypic: discussed earlier, when doing DR, don't remember why, but agreed to use numbers, I'll ask why, but let's not block on this.

ably.d.ts
VeskeR: > but it's not for UTS, it won't matter for statically-typed languages

Huh, interesting. I see why we wouldn't want it to be in the UTS, but not including the tests in languages where it actually helps seems wrong

ably.d.ts
```diff
@@ -3703,6 +3703,150 @@ export declare interface PushAdmin {
...
+        /**
+         * One or more Ably channel names. Devices subscribed to any of these channels will receive the push-to-start notification. Provide either `channels` or `deviceId`.
+         */
+        channels?: string[];
+      }
+    | {
+        /**
+         * Restrict the push-to-start notification to a single device. Provide either `channels` or `deviceId`.
+         */
+        deviceId?: string;
```
VeskeR: can we make `channels` and `deviceId` not optional here?

ably.d.ts
VeskeR: uhh, sure. A bit uncomfortable for me to approve as we don't have raw integers anywhere else in the codebase, it's always a string enum in public types.

VeskeR (approved):
Approving, two small oustanding points:
- make non optional `channels` and `deviceId`
- As discussed on standup, `messageStoragePolicy` is Apple's related field, so keeping it as an integer would be fine as this is what Apple uses. We agreed to at least include the link to the Apple's docs for this. Potentially we may decide to change to the enum if the team agrees it will be of benefit.

PR #2257, opened 2026-07-07
[AIT-1131] fix(uts): align objects unit tests with reconciled specs and close audit gaps

sacOO7:
- Fixes https://ably.atlassian.net/browse/AIT-1131
- Based on updated spec at -> https://github.com/ably/specification/pull/499

## Intent

The objects unit-test specs were reconciled with the canonical LiveObjects spec in [ably/specification#499](https://github.com/ably/specification/pull/499), and an exhaustive spec↔test audit of the derived ably-js suite surfaced a handful of verification gaps (dropped assertions, one vacuous assertion, tests exercising the wrong scenario). This PR brings the derived unit tests in line with the reconciled specs and closes every audit finding, so the suite verifies exactly what the specs prescribe.

Two structurally new tests are added; everything else is assertion-level convergence. **Suite: 309 passing, 0 failing, 4 pending (all four documented in `deviations.md`).**

### 1. Spec reconciliation (ably/specification#499 logic changes)

- **RTO23e (ensure-active-channel)** — new test `RTO23e/get-reattaches-detached-0`: `get()` on a DETACHED channel re-attaches per RTL33b and resolves with the root PathObject. The existing get()-on-FAILED test is relabeled to `RTO23e/get-rejects-failed-0` (rejects 90001 per RTL33c) and the now-duplicate old RTO23b body is removed.
- **RTO25b** — access preconditions are exercised through a real access method, `[...root.keys()]` (the precondition throws on first iteration of the generator), on both DETACHED and FAILED — the FAILED variant is the second new test.
- **RTO20e/RTO20e1** — the sync-wait test now asserts mid-sync state (operation not settled, value unchanged while SYNCING) before completing the sync; the RTO20e1 Test ID is corrected to `fails-on-channel-detached-0` (the ERROR→FAILED adaptation remains, documented).
- **RTO24c1 / RTO24b2a1 (depth semantics)** — the spec adopted ably-js's self-counts-as-depth-1 model, so the RTO24c1 test is rewritten to `depth: 1` with the spec's new stimuli, and the two `path_object_subscribe` depth tests go spec-verbatim (single-candidate COUNTER_INC grandchild stimuli, expected counts drop 3→2). Three inline deviation adaptations retire as a result.
- **Quiescence controls** — negative "listener did not fire" assertions now await a positive control delivery through the same dispatch before asserting silence (RTINS16f, both depth tests, the RUN_DEVIATIONS-gated RTO24b2b); RTINS16g asserts identity on the delivered event's object rather than the pre-existing handle (the exact vacuous form the spec fixed).
- **Canonical harness constants** — `standard_test_pool.ts` exports `SITE_CODE` and `ackSerial(msgSerial, i)`; the echo-dedup, ack-after-echo and RTO5c9 replay tests reference them instead of hardcoded literals.

### 2. Audit fixes (verification gaps found by the spec↔test audit)

- **RTO3** now tests what the spec asks: a fresh, unsynced pool initializes with a zero-value root LiveMap (previously ran against a synced pool).
- **RTO4b** captures the root LiveMap's internal `notifyUpdated` so the spec's update-diff (`name: 'removed'`) and `objectMessage`-unpopulated assertions are genuinely verified — the previous `updates[0].objectMessage` assertion was vacuous (events expose `.message`, so it passed unconditionally).
- **RTO5c9** regains the spec's serial-replay steps, proving `appliedOnAckSerials` is actually cleared on re-sync rather than dedup-rejected.
- **`update.noop == true`** assertions restored at all 7 rejection sites (6 in `live_map`, 1 in `live_counter`) via the existing `captureNotifyUpdated` helper; **`statusCode == 400`** restored at 3 sites; `object_id` and `RTPO13c5` UTS tags aligned with the spec's Test IDs (traceability).
- **RTLO4b4c1** now exercises the genuine noop suppression path (already-merged COUNTER_CREATE) with the spec's follow-up control increment and `length == 2` assertion.

### 3. `deviations.md`

Retired the entries superseded by the reconciled specs (get()-re-attaches-on-DETACHED, RTO24c1 depth semantics, RTLM24 strict-`>` comparison — all now spec-mandated behavior); reframed RTLM9b to the throw-instead-of-false family; added the new **RTLC9h** entry (ably-js applies missing-number COUNTER_INC as NaN instead of nooping) and registry entries for the four skipped tests, plus a null/undefined language-mapping umbrella entry.

## Spec defect found during implementation

Implementing the spec's canonical `ack_serial = "ack-{m}:{i}"` verbatim broke 7 tests: the value sorts **before** the standard pool's `"t:0"` entry timeserials under the string LWW comparison (RTLM9), so locally applied MAP_SETs on pool entries are rejected as stale (the new RTO24c1 serials `"s:1"/"s:2"` have the same problem). Any strictly-deriving SDK will hit this. This PR uses the sort-safe `t:{msgSerial+1}:{i}` form inside the spec-named `ackSerial()` helper; the corresponding spec-side fix is prepared on the spec branch to accompany ably/specification#499.

## Validation

- `npm run test:uts:unit` scope: **309 passing, 0 failing, 4 pending** — the pending four are deliberate, documented skips.
- Mechanical `// UTS:` tag ↔ spec `Test ID` set-diff against the reconciled specs is clean: every spec Test ID has a derived test and every tag resolves, with only the documented exceptions (`RTLMV4b`, untranslatable to JS).
- Prettier clean.

## Related

- [ably/specification#499](https://github.com/ably/specification/pull/499) — the spec reconciliation these tests derive from (plus the pending serial-ordering fix noted above).
- #2256 — the sibling PR covering the objects *integration* tests.

## Summary by CodeRabbit

* **Documentation**
  * Expanded UTS deviation notes to clarify coverage gaps, refined subscription argument order, and updated channel `clear()`/lifecycle edge-case expectations.

* **Bug Fixes**
  * Prevented cleanup routines from deleting the root object when it isn’t present in synced object IDs.

* **Tests**
  * Updated object/map/path/channel tests with stronger, observable assertions (noop suppression, dispatch counts, detach/failure scenarios, depth filtering with quiescence waits).
  * Introduced shared test constants/helpers to standardize serial and site code usage.
  * Adjusted timeouts and error assertions for more reliable validation.

### DIFF
--- a/src/plugins/liveobjects/objectspool.ts
+++ b/src/plugins/liveobjects/objectspool.ts
@@ -44,10 +44,13 @@ export class ObjectsPool {
 
   /**
    * Deletes objects from the pool for which object ids are not found in the provided array of ids.
+   *
+   * @spec RTO5c2 - remove objects whose ids were not received during the sync sequence
+   * @spec RTO5c2a - the root object must never be removed (RTO3b), even if absent from the sync
    */
   deleteExtraObjectIds(objectIds: string[]): void {
     const poolObjectIds = [...this._pool.keys()];
-    const extraObjectIds = poolObjectIds.filter((x) => !objectIds.includes(x));
+    const extraObjectIds = poolObjectIds.filter((x) => !objectIds.includes(x) && x !== ROOT_OBJECT_ID);
 
     extraObjectIds.forEach((x) => this._pool.delete(x));
   }
--- a/test/uts/deviations.md
+++ b/test/uts/deviations.md
@@ -128,7 +128,7 @@ These tests assert spec behavior but are skipped by default because they are kno
 
 **ably-js behavior**: `shouldFallback` only receives the error object, not response headers. The `Server` header is not inspected anywhere in the fallback decision path.
 
-**Test**: `RSC15l4 - CloudFront Server header triggers fallback`.
+**Tests**: `RSC15l4 - CloudFront Server header triggers fallback` (unit, `rest/unit/fallback.test.ts`; also skipped at integration tier in `rest/integration/proxy/rest_fallback.test.ts`).
 
 **Issue**: [#2197](https://github.com/ably/ably-js/issues/2197)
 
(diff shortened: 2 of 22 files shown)


ttypic (approved):
LGTM

PR #2259, opened 2026-07-09
feat(react): add useObject hook for LiveObjects

owenpearson:
[AIT-1133](https://ably.atlassian.net/browse/AIT-1133)

Implements [PSDR20](https://ably.atlassian.net/wiki/spaces/CHN/pages/5171445785/PSDR20+LiveObjects+react+hooks+for+ably-js)

Adds a new `ably/liveobjects/react` entry point exporting `useObject`, which subscribes to a channel's LiveObjects state (optionally a nested node selected by navigating from the root) and re-renders on change, built on useSyncExternalStore with a cached compact() snapshot.

[AIT-1133]: https://ably.atlassian.net/browse/AIT-1133?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added `useObject` for React LiveObjects, supporting root and nested object nodes, selector-based subscriptions, cached reactive `value`, and layered loading/error states (`error`, `connectionError`, `channelError`).
  * Added a new package entry point for React LiveObjects (`ably/liveobjects/react`) with CommonJS and ESM support.
* **Documentation**
  * Documented `useObject`, including React 18+ requirements, supported subscription patterns, and loading/error/ready behavior.
* **Tests**
  * Added React hook tests covering lifecycle, selector changes, identity stability, timing edge cases, and cleanup.
* **Chores**
  * Added TypeScript type support and exports for the new React LiveObjects entry point.

### DIFF
--- a/docs/react.md
+++ b/docs/react.md
@@ -24,6 +24,7 @@ The hooks provide a simplified syntax for interacting with Ably, and manage the 
   - [useChannel](#usechannel)
   - [usePresence](#usepresence)
   - [usePresenceListener](#usepresencelistener)
+  - [useObject](#useobject)
 
 ## <!-- /code_chunk_output -->
 
@@ -318,6 +319,83 @@ interface MyPresenceType {
 ```
 
 `presenceData` is a good way to store synchronised, per-client metadata, so types here are especially valuable.
+
+### useObject
+
+The `useObject` hook subscribes to the [LiveObjects](https://ably.com/docs/liveobjects) state on a channel and re-renders your component when it changes. It is exported from its own entry point, `ably/liveobjects/react`, so apps that don't use LiveObjects don't need to import any of its code:
+
+```javascript
+import { useObject } from 'ably/liveobjects/react';
+```
+
+Using the hook requires:
+
+- the `LiveObjects` plugin passed to the Ably client (see the [LiveObjects documentation](https://ably.com/docs/liveobjects/quickstart/javascript)),
+- the `OBJECT_SUBSCRIBE` channel mode (plus `OBJECT_PUBLISH` if you write to objects), set via `ChannelProvider`'s channel options,
+- React 18 or later (the hook is built on `useSyncExternalStore`).
+
+```jsx
+<ChannelProvider channelName="your-channel-name" options={{ modes: ['OBJECT_SUBSCRIBE', 'OBJECT_PUBLISH'] }}>
+  <Component />
+</ChannelProvider>
+```
+
+Called with no arguments, `useObject` subscribes to the channel's root object. `value` is a plain-data snapshot of the object (the result of [`compact()`](https://ably.com/docs/liveobjects/concepts/objects#compact)), and `object` is the live `PathObject` for the subscribed node, which you can use to navigate and to write:
+
+```jsx
+const Scoreboard = () => {
+  const { value, object } = useObject();
+
+  if (!object) return <p>Loading…</p>;
+
+  return <button onClick={() => object.get('scores').get('alice').increment(1)}>Alice: {value.scores.alice}</button>;
+};
+```
+
+Pass a selector to subscribe to a nested node instead. The selector receives the root `PathObject` and must navigate to a descendant using `get`/`at`; the hook subscribes to the node it returns, so the component only re-renders for changes within that node. Navigating to the narrowest node you actually render is the main performance lever:
+
+```jsx
+const AliceScore = () => {
+  const { value, object } = useObject((root) => root.get('scores').get('alice'));
+
+  return <button onClick={() => object?.increment(1)}>Alice: {value ?? 0}</button>;
+};
+```
+
+If you're using TypeScript, annotate the selector's parameter with the shape of your channel's objects and the whole navigation chain is checked at compile time — `value` and `object` are then typed from the chain with no further annotation:
+
+```tsx
+import { LiveCounter, LiveMap, PathObject } from 'ably/liveobjects';
+
+type GameRoot = LiveMap<{
+  scores: LiveMap<{ alice: LiveCounter; bob: LiveCounter }>;
+  title: string;
+}>;
+
+const AliceScore = () => {
+  // value: number | undefined, object: PathObject<LiveCounter> | undefined
+  const { value, object } = useObject((root: PathObject<GameRoot>) => root.get('scores').get('alice'));
+
+  return <button onClick={() => object?.increment(1)}>Alice: {value ?? 0}</button>;
+};
+```
+
+For the no-selector form, provide the root type as a type parameter instead: `useObject<GameRoot>()`.
+
+The root object resolves asynchronously (after the channel attaches and the initial sync completes), so the hook's readiness is derived from the result: `object === undefined && error === null` means loading, `error !== null` means the object could not be resolved (for example the `LiveObjects` plugin is missing, or the channel lacks the object modes — `error` carries ably-js's own `ErrorInfo`), and a defined `object` means ready, with `value` carrying the data once the node exists.
+
+Two things to note about `value`:
+
+- It is a cached snapshot: its identity only changes when the subscribed node changes, so it is safe to use in dependency arrays.
+- It is a by-value copy: the same logical node reached through two different hooks is not reference-equal. Compare fields rather than snapshots, and use `object` when you need a stable identity.
+
+Like the other channel hooks, `useObject` accepts the standard options (`channelName`, `ablyId`, `skip`, `onConnectionError`, `onChannelError`) and returns `connectionError` / `channelError`:
+
+```javascript
+const { value, error, connectionError, channelError } = useObject((root) => root.get('scores'), {
+  channelName: 'your-channel-name',
+});
+```
 
 ### useConnectionStateListener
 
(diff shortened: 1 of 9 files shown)


docs/react.md
```diff
@@ -319,6 +320,83 @@ interface MyPresenceType {
...
+- the `OBJECT_SUBSCRIBE` channel mode (plus `OBJECT_PUBLISH` if you write to objects), set via `ChannelProvider`'s channel options,
+- React 18 or later (the hook is built on `useSyncExternalStore`).
+
+```jsx
+<ChannelProvider channelName="your-channel-name" options={{ modes: ['OBJECT_SUBSCRIBE', 'OBJECT_PUBLISH'] }}>
+  <Component />
+</ChannelProvider>
+```
+
+Called with no arguments, `useObject` subscribes to the channel's root object. `value` is a plain-data snapshot of the object (the result of [`compact()`](https://ably.com/docs/liveobjects/concepts/objects#compact)), and `object` is the live `PathObject` for the subscribed node, which you can use to navigate and to write:
```
VeskeR: I know this is annoying, but we don't use "root" terminology anymore, see https://ably.atlassian.net/wiki/spaces/LOB/pages/4235722804/LODR-042+LiveObjects+Realtime+Client+API+Improvements#Entrypoint:
> We also remove any reference to the “root” object altogether; this concept was previously used to distinguish between the root entrypoint object and nested objects. In a single-object-per-channel mental model, this concept is no longer needed.

Essentially, in public docs we just say "channel's object". Should update throughout the public docs/jsdocs/types.
It's okay to use "root" internally though, if it would make things clearer in the implementation.

docs/react.md
```diff
@@ -319,6 +320,83 @@ interface MyPresenceType {
...
+  // value: number | undefined, object: PathObject<LiveCounter> | undefined
+  const { value, object } = useObject((root: PathObject<GameRoot>) => root.get('scores').get('alice'));
+
+  return <button onClick={() => object?.increment(1)}>Alice: {value ?? 0}</button>;
+};
+```
+
+For the no-selector form, provide the root type as a type parameter instead: `useObject<GameRoot>()`.
+
+The root object resolves asynchronously (after the channel attaches and the initial sync completes), so the hook's readiness is derived from the result: `object === undefined && error === null` means loading, `error !== null` means the object could not be resolved (for example the `LiveObjects` plugin is missing, or the channel lacks the object modes — `error` carries ably-js's own `ErrorInfo`), and a defined `object` means ready, with `value` carrying the data once the node exists.
```
VeskeR: This is the first mention of `error` and it's not clear that this is one of the return values from `useObject` and what it contains. Probably one more paragraph with code sample above to showcase it?

docs/react.md
```diff
@@ -319,6 +320,83 @@ interface MyPresenceType {
...
+import { LiveCounter, LiveMap, PathObject } from 'ably/liveobjects';
+
+type GameRoot = LiveMap<{
+  scores: LiveMap<{ alice: LiveCounter; bob: LiveCounter }>;
+  title: string;
+}>;
+
+const AliceScore = () => {
+  // value: number | undefined, object: PathObject<LiveCounter> | undefined
+  const { value, object } = useObject((root: PathObject<GameRoot>) => root.get('scores').get('alice'));
```
VeskeR: nit: can we pass a type parameter to the hook in all code samples for consistency with "no-selector" form? Could be cleaner

src/platform/react-hooks/src/liveobjects/useObject.ts
```diff
@@ -0,0 +1,262 @@
...
+          value = node.compact();
+          onStoreChange();
+        });
+        return () => subscription.unsubscribe();
+      },
+      getSnapshot: () => value,
+    };
+  }, [node]);
+
+  const value = useSyncExternalStore(store.subscribe, store.getSnapshot, getServerSnapshot);
```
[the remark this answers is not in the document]
VeskeR: @owenpearson  my Claude run also flagged this as an issue. Could you please take a look? suggested change is to have:
```tsx
subscribe: (onStoreChange) => {
  const subscription = node.subscribe(() => { value = node.compact(); onStoreChange(); });
  const latest = node.compact();          // catch a change between render and subscribe
  if (!dequal(latest, value)) { value = latest; onStoreChange(); }
  return () => subscription.unsubscribe();
}
```

docs/react.md
VeskeR: I think this statement also should be updated to reflect that `useObject` requires 18+

src/platform/react-hooks/src/liveobjects/useObject.ts
```diff
@@ -0,0 +1,262 @@
...
+export type UseObjectOptions = ChannelNameAndOptions;
+
+/**
+ * The surface common to every {@link PathObject} variant that {@link useObject}
+ * relies on. Every node a selector can return satisfies this type; it exists so
+ * the hook's signatures can constrain and infer concrete `PathObject` subtypes
+ * (which is what makes the selector's navigation chain determine the result
+ * type) without forcing them into a single `PathObject<V>` shape.
+ */
+export interface ObjectNode {
```
VeskeR, line 35: It's not immediately clear to me why this is required and we can't use the types from `liveobjects.d.ts` where we have `<T extends Value = Value> -> PathObject<T>`. Could you please clarify it, in a response here or in an inline code comment?

VeskeR (commented):
LGTM overall, couple of minor requests below and a question about types.

Also, one of the core benefits of the typed navigation selector we chose is, well, that types flow through. We'd want to ensure this actually works. I think it's worthwhile to add couple of simple checks to our package tests in https://github.com/ably/ably-js/blob/main/test/package/browser/template/src/ReactApp.tsx, and check that `useObject` resolves to correct types as expected

docs/react.md
```diff
@@ -319,6 +320,83 @@ interface MyPresenceType {
...
+  return <button onClick={() => object?.increment(1)}>Alice: {value ?? 0}</button>;
+};
+```
+
+If you're using TypeScript, annotate the selector's parameter with the shape of your channel's objects and the whole navigation chain is checked at compile time — `value` and `object` are then typed from the chain with no further annotation:
+
+```tsx
+import { LiveCounter, LiveMap, PathObject } from 'ably/liveobjects';
+
+type GameRoot = LiveMap<{
```
VeskeR: Also noticed that `useObject` hook expects a `LiveMap` type for the channel object, but non-hooks version `RealtimeObject.get` is typed `get<T extends Record<string, Value>>(): Promise<PathObject<LiveMap<T>>>;` https://github.com/ably/ably-js/blob/main/liveobjects.d.ts#L589, so it expects an _inner_ object shape.
Not sure if we can make `useObject` work the same due to `ObjectNode`

src/platform/react-hooks/src/liveobjects/useObject.ts
```diff
@@ -0,0 +1,262 @@
...
+ * The channel is taken from the nearest `ChannelProvider` unless `channelName`
+ * is given. `Root` describes the channel's root LiveMap and defaults to an
+ * untyped map.
+ *
+ * Requires the `LiveObjects` plugin on the Realtime client and the
+ * `OBJECT_SUBSCRIBE` channel mode (plus `OBJECT_PUBLISH` to write); otherwise
+ * `error` carries ably-js's reason. Readiness is derived from the result (see
+ * {@link UseObjectResult.object}), not a status enum.
+ */
+export function useObject<Root extends Value = LiveMap<Record<string, Value>>>(
```
VeskeR: I think the unintended effect of `Root extends Value` is that you can pass `LiveCounter:`
```tsx
useObject<LiveCounter>()
```
which is nonsensical

src/platform/react-hooks/src/liveobjects/useObject.ts
```diff
@@ -0,0 +1,262 @@
...
+/**
+ * Subscribe to a nested LiveObjects node, selected by navigating from a typed
+ * root, and re-render on change. Annotate the selector's parameter with the
+ * channel's root shape (`(root: PathObject<MyRoot>) => ...`) and the whole
+ * navigation chain — wrong keys included — is checked at compile time.
+ *
+ * Channel resolution, readiness, and plugin/modes requirements are as for the
+ * root overload.
+ */
+export function useObject<Root extends ObjectNode, N extends ObjectNode>(
```
VeskeR: wdyt about these types for `useObject`'s selector? the root is always a map, so it doesn't need to be `Root extends ObjectNode` - we can type it with the real thing (same style as `channel.object.get<T>()`):

```ts
export function useObject<T extends Record<string, Value> = Record<string, Value>, N extends ObjectNode = ObjectNode>(
  selector: (root: PathObject<LiveMap<T>>) => N,
  options?: UseObjectOptions,
): UseObjectResult<N>;
```

As a result you can't do `useObject<LiveCounter>()`

owenpearson: That looks good to me. I've changed to this but had to make a slight tweak to prevent the weird case where someone declares the first type param but not the second (ie `useObject<State>` would be broken because the node type wouldn't be inferred)

VeskeR (approved):
LGTM.
Looks like there is a lint issue to fix before merging

PR #2261, opened 2026-07-12
fix: replace in-flight presence sync sequences

### LINKED ISSUE #2215: New presence sync does not discard previous in-flight sync (RTP18a)
## Spec requirement

**RTP18a**: _"If a new SYNC is initiated while a previous SYNC sequence is in progress, the previous sync should be discarded and the new one should replace it."_ The residual members set should be reset to the current map contents when a new sync starts mid-flight.

## Current behavior

`startSync()` in `presencemap.ts` (line ~131-145) has an explicit guard:

```js
if (!this.syncInProgress) {
  this.residualMembers = Utils.copy(map);
  this.setInProgress(true);
}
```

When called during an in-progress sync, `syncInProgress` is already true, so `residualMembers` is NOT reset. This means members from the first sync's `residualMembers` set persist, and when the second sync completes, `endSync()` may incorrectly evict members that were present in the second sync sequence but not explicitly seen (because they were already removed from `residualMembers` during the first sync's processing).

## Root cause

The `if (!this.syncInProgress)` guard in `startSync()` was likely intended to be idempotent for duplicate START messages, but it also prevents reset when a genuinely new sync sequence begins. The fix would be to always reset `residualMembers = Utils.copy(map)` regardless of `syncInProgress`.

## UTS test

- `RTP18a - new sync discards previous in-flight sync` in `test/uts/realtime/unit/presence/presence_sync.test.ts`

Found via UTS (Universal Test Specification) compliance testing.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-5727) by [Unito](https://www.unito.io)

w3lld1:
## Summary

- reset residual presence members when a new sync sequence replaces one already in progress
- track the sync sequence ID so pages from the same multi-message sync continue accumulating normally
- enable the RTP18a UTS regression test as a conforming test

## Validation

- `npm run test:uts:unit` (1,052 passing, 55 pending)
- `npx mocha --no-config --require tsx/cjs test/uts/realtime/unit/presence/presence_sync.test.ts` (14 passing)
- `npx eslint src/common/lib/client/presencemap.ts src/common/lib/client/realtimepresence.ts`
- `npx prettier --check src/common/lib/client/presencemap.ts src/common/lib/client/realtimepresence.ts test/uts/realtime/unit/presence/presence_sync.test.ts`
- `git diff --check`

Fixes #2215

## Summary by CodeRabbit

* **Bug Fixes**
  * Reworked realtime presence synchronization so repeated sync starts rebuild the residual membership snapshot from the latest state.
  * Improved sync-sequence handling by tracking a sync identifier from the sync channel serial, preventing duplicate/incorrect member-sync initialization.
  * Ensures absent entries are excluded when rebuilding residual state and clears the identifier when the sync sequence completes.
* **Tests**
  * Updated realtime presence sync tests and comments to reflect the “new sync discards previous in-flight sync” behavior.

### DIFF
--- a/src/common/lib/client/presencemap.ts
+++ b/src/common/lib/client/presencemap.ts
@@ -137,11 +137,9 @@ export class PresenceMap extends EventEmitter {
       'PresenceMap.startSync()',
       'channel = ' + this.presence.channel.name + '; syncInProgress = ' + syncInProgress,
     );
-    /* we might be called multiple times while a sync is in progress */
-    if (!this.syncInProgress) {
-      this.residualMembers = Utils.copy(map);
-      this.setInProgress(true);
-    }
+    /* RTP18a: a new sync replaces any sync already in progress */
+    this.residualMembers = Utils.copy(map);
+    this.setInProgress(true);
   }
 
   endSync() {
--- a/src/common/lib/client/presencemap.ts
+++ b/src/common/lib/client/presencemap.ts
@@ -139,7 +139,7 @@ export class PresenceMap extends EventEmitter {
     );
     /* RTP18a: a new sync replaces any sync already in progress. Exclude
      * tombstones that have already emitted their leave event. */
-    this.residualMembers = {};
+    this.residualMembers = Object.create(null) as Record<string, PresenceMessage>;
     for (const memberKey in map) {
       if (map[memberKey].action !== 'absent') {
         this.residualMembers[memberKey] = map[memberKey];
--- a/src/common/lib/client/presencemap.ts
+++ b/src/common/lib/client/presencemap.ts
@@ -137,8 +137,14 @@ export class PresenceMap extends EventEmitter {
       'PresenceMap.startSync()',
       'channel = ' + this.presence.channel.name + '; syncInProgress = ' + syncInProgress,
     );
-    /* RTP18a: a new sync replaces any sync already in progress */
-    this.residualMembers = Utils.copy(map);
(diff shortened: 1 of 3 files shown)


w3lld1:
Addressed the tombstone edge case in `77864256`: replacement syncs now exclude `absent` entries from `residualMembers`, preventing a second synthesized leave for members already removed during the previous sync.

Validation:

- `pnpm exec prettier --check src/common/lib/client/presencemap.ts`
- `pnpm exec eslint src/common/lib/client/presencemap.ts`
- focused presence sync suite: 14 passing

w3lld1:
Thanks — addressed in `2359d365`: replacement sync residual maps now use `Object.create(null)`, preserving the null-prototype behavior while still filtering tombstones. The PR checks are green.

ttypic (approved):
LGTM

w3lld1:
The latest CI run exposed a bootstrap-sync regression in `leave_published_for_members_on_presenceless_attached`: the first server sync ID was treated as a replacement for the sync started by ATTACHED, which discarded a member received in between.

Fixed in `bc59c3c8` by registering the first sync ID without restarting the active bootstrap sync; later ID changes still replace an in-flight sync.

Fresh checks:
- focused Node regression: 1 passing
- focused RTP18a/RTP18c/RTP19a UTS selectors: 5 passing
- ESLint and Prettier on the changed file
- `git diff --check`

ttypic:
Thank you @w3lld1! rerunning CI

PR #2262, opened 2026-07-14
[AIT-850] feat: react native push plugin

ttypic:
Demo app: https://github.com/ably-labs/rn-push-notifications
RFC: https://ably.atlassian.net/wiki/x/CADCKQE

Adds a push plugin for React Native that mirrors the web Push plugin. Users supply an async storage implementation (matching `@react-native-async-storage/async-storage`) and a requestToken callback returning `{ transportType: 'fcm' | 'apns', token }`, keeping the SDK free of native module dependencies. `create()` reuses the existing activation state machine, local device factory, and push channel via re-export, and carries a `pushConfig` which the core installs into `Platform.Config.push` at client construction. The plugin registers under `ClientOptions.plugins.Push` unchanged.

Push activation is no longer browser-only, so the availability error messages and remediations now point React Native users at the `ably/react-native-push` plugin. The activation hint constant copies in `pushactivation.ts` and `push.ts` are kept byte-identical as required.

## Summary by CodeRabbit

* **New Features**
  * Added a React Native push plugin (FCM/APNs) with a new `react-native-push` package export and typings.
  * Introduced an async `getDevice()` API across clients; synchronous `device()` is deprecated.
* **Bug Fixes**
  * Improved push activation/deactivation and persistence for async storage, including safer hydration and clearer error handling.
  * Expanded device push recipient support for FCM and APNs with stricter request-token validation.
* **Build & Release**
  * Added React Native push build output, npm scripts, and bundle validation.
* **Tests**
  * Added/expanded React Native push activation tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,10 @@
       "types": "./push.d.ts",
       "import": "./build/push.js"
     },
+    "./react-native-push": {
+      "types": "./react-native-push.d.ts",
+      "default": "./build/react-native-push.js"
+    },
     "./liveobjects": {
       "import": {
         "types": "./liveobjects.d.mts",
@@ -60,6 +64,7 @@
     "promises.d.ts",
     "promises.js",
     "push.d.ts",
+    "react-native-push.d.ts",
     "resources/**",
     "src/**",
     "react/**"
@@ -183,6 +188,7 @@
     "build:react:mjs": "tsc --project src/platform/react-hooks/tsconfig.mjs.json && cp src/platform/react-hooks/res/package.mjs.json react/mjs/package.json",
     "build:react:cjs": "tsc --project src/platform/react-hooks/tsconfig.cjs.json && cp src/platform/react-hooks/res/package.cjs.json react/cjs/package.json",
     "build:push": "grunt build:push",
+    "build:react-native-push": "grunt build:react-native-push",
     "build:liveobjects": "grunt build:liveobjects",
     "requirejs": "grunt requirejs",
     "lint": "eslint .",
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "promises.js",
     "push.d.ts",
     "react-native-push.d.ts",
+    "react-native-push.js",
     "resources/**",
     "src/**",
     "react/**"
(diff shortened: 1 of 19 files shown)


src/common/lib/client/baseclient.ts
```diff
@@ -145,17 +147,63 @@ class BaseClient {
...
+            'Use await client.getDevice() instead. device() is deprecated and will become asynchronous in the next major release.',
+        });
+      }
       this._device = this.push.LocalDevice.load(this);
     }
     return this._device;
   }
 
+  /** RSH8 */
+  async getDevice(): Promise<LocalDevice & API.LocalDevice> {
```
ttypic, line 188: Another option is to keep `device()` and add `initDevice()`/`loadDevice()` to initialize device (we read it from the cache anyway)

src/plugins/push/pushactivation.ts
owenpearson, line 860: I think we need to call `storage.remove` on these two fields here because the persist method never removes existing keys. This is a bug for web push too but I think it's more likely to be an issue for react native because of how native apps work.

package.json
```diff
@@ -30,6 +30,10 @@
       "types": "./push.d.ts",
       "import": "./build/push.js"
     },
+    "./react-native-push": {
```
owenpearson, line 33: worth adding a shim (like callbacks.js) for older metro versions? i think there will be apps using older versions of metro which don't support package exports

test/uts/rest/unit/push/push_activation_react_native.test.ts
```diff
@@ -0,0 +1,340 @@
...
+    };
+  });
+
+  after(function () {
+    (Module as any)._load = originalModuleLoad;
+  });
+
+  afterEach(function () {
+    restoreAll();
+    // the Push constructor installs the plugin-supplied push config into the Platform singleton
```
owenpearson: I think this comment is stale

owenpearson (changes requested):
Looks great overall, nice one! A few things I want to flag before approval

package.json
ttypic: added

src/plugins/push/pushactivation.ts
ttypic: added

test/uts/rest/unit/push/push_activation_react_native.test.ts
ttypic: fixed

owenpearson (approved):
:shipit:

PR #2266, opened 2026-07-21
Support running the node test suite against ABLY_LOCAL_SANDBOX_URL

lmars:
Part of ongoing work to let the SDK test suite run against Ably-compatible open source servers, not just the cloud sandbox.

When `ABLY_LOCAL_SANDBOX_URL` is set, each test provisions its app through a local sandbox's `POST /apps` (the same appspec as the cloud sandbox) instead of the cloud, and routes every client at the isolated server that response describes (its own endpoint/port/tls). Teardown and stats fixtures target the local sandbox too. `ABLY_TEST_TRANSPORTS` can restrict the suite to a transport subset for servers that implement only some.

All of this is inert unless `ABLY_LOCAL_SANDBOX_URL` is set — the normal cloud-sandbox path is unchanged. The wiring lives in `test/common/globals/environment.js` and `test/common/modules/{testapp_manager,client_module,shared_helper}.js`.

## Summary by CodeRabbit

* **Tests**
  * Improved test reliability by preventing race conditions in realtime presence synchronization checks.
  * Added/updated REST authentication coverage for embedded JWTs, including encrypted variants.
  * Expanded REST request testing for put/patch/delete using a dedicated echo-service setup.
* **Test Configuration**
  * Added configuration to control which transports are used during tests.
  * Added local sandbox routing support (including overrides via URL query), ensuring app setup, stats, and cleanup target the sandbox consistently.

### DIFF
--- a/test/common/globals/environment.js
+++ b/test/common/globals/environment.js
@@ -7,7 +7,18 @@ define(function (require) {
     port = environment.ABLY_PORT || 80,
     tlsPort = environment.ABLY_TLS_PORT || 443,
     tls = 'ABLY_USE_TLS' in environment ? environment.ABLY_USE_TLS.toLowerCase() !== 'false' : true,
-    logLevel = environment.ABLY_LOG_LEVEL || defaultLogLevel;
+    logLevel = environment.ABLY_LOG_LEVEL || defaultLogLevel,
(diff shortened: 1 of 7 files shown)


owenpearson (approved):
looks good 👍

PR #2267, opened 2026-07-22
[AIT-1183] Add push token updates support for React Native

ttypic:
This pull request adds support for push token updates and synchronization in React Native.

## Summary by CodeRabbit

* **New Features**
  * Added `push.updateToken()` to sync refreshed FCM/APNs device tokens after activation, using `{ transportType, token }`.
  * Persists updated push recipient details for consistent future sessions.
* **Documentation**
  * Updated React Native TypeScript declarations and examples to reflect token rotation behavior and the `{ transportType, token }` payload.
* **Bug Fixes**
  * Improved update synchronization by patching the correct device-registration endpoint and handling sync-failure states, retries, and queued concurrent updates safely.
* **Tests**
  * Added unit tests covering FCM/APNs mapping, cold-start updates, validation, retries, callback routing, concurrency, and deactivation race handling.

### DIFF
--- a/src/common/lib/client/push.ts
+++ b/src/common/lib/client/push.ts
@@ -22,6 +22,14 @@ const PUSH_ACTIVATION_NOT_AVAILABLE_HINT =
 const PUSH_DEACTIVATION_NOT_AVAILABLE_HINT =
   'Run push.deactivate() in a browser environment with service worker support, or in React Native using the ably/react-native-push plugin. From a server, call client.push.admin.deviceRegistrations.remove(deviceId) to remove a device registration.';
 
+const PUSH_TOKEN_UPDATE_NOT_AVAILABLE_HINT =
+  'Call push.updateToken() in React Native with the ably/react-native-push plugin configured, passing plugins: { Push: ReactNativePush.create({ storage, requestToken }) } in the client options. From a server, use client.push.admin.deviceRegistrations.save(device) to modify a device registration instead.';
+
+interface PushDeviceToken {
+  transportType: 'apns' | 'fcm';
+  token: string;
+}
+
 class Push {
   client: BaseClient;
   admin: Admin;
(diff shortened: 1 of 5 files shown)


src/common/lib/client/push.ts
```diff
@@ -136,6 +141,78 @@ class Push {
...
+      client,
+      '/push/deviceRegistrations',
+      requestBody,
+      authDetails.headers,
+      authDetails.params,
+      format,
+      true,
+    );
+
+    await device.persist();
```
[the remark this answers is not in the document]
ttypic: no, it's not returning new identity token

src/common/lib/client/push.ts
```diff
@@ -136,6 +141,78 @@ class Push {
     machine.handleEvent(new pushPlugin.CalledDeactivate(machine, deregisterCallback));
     await deactivated;
   }
+
+  async updateToken(token: PushDeviceToken): Promise<void> {
```
owenpearson, line 148: I think `activate`/`deactivate` have some serialisation (ie they don't interfere with eachother when called concurrently), but we should probably do this here too so that this works concurrently with an in-flight activate/deactivate

src/common/lib/client/push.ts
owenpearson, line 198: should we invoke the customer's registerCallback after this is all done?

src/common/lib/client/push.ts
```diff
@@ -136,6 +141,78 @@ class Push {
...
+      throw new ErrorInfo({
+        message: 'Push token cannot be updated because the device is not activated for push notifications',
+        code: 40000,
+        statusCode: 400,
+        remediation:
+          'Call client.push.activate() and await its completion before calling push.updateToken(). Wire updateToken() to your platform token refresh listener only after activation has completed.',
+      });
+    }
+
+    device.push.recipient =
```
owenpearson, line 194: this mutates the local device even if the PATCH subsequently fails. i don't think it's a big issue because it will reset next time the device activates, but it's probably best to avoid this state mismatch by creating a copy of the device for the request

src/common/lib/client/push.ts
```diff
@@ -136,6 +141,78 @@ class Push {
...
+  async updateToken(token: PushDeviceToken): Promise<void> {
+    const pushPlugin = this.client.options.plugins?.Push;
+    if (!pushPlugin) {
+      throw Utils.createMissingPluginError('Push');
+    }
+    const machine = this.stateMachine;
+    if (!machine) {
+      throw new ErrorInfo({
+        message:
+          'This platform is not supported as a target of push notifications: push activation requires a React Native environment with the ably/react-native-push plugin',
```
owenpearson: This error is weird, the remediation (PUSH_ACTIVATION_NOT_AVAILABLE_HINT) talks about service workers/web but the message you added here says that push needs a react native environment.

owenpearson (changes requested):
looks pretty good but a few things i think we need to fix:

src/common/lib/client/push.ts
ttypic: changed implementation to use state machine, so shouldn't be an issue any more

src/common/lib/client/push.ts
ttypic: now I persist device first (it's established pattern inside state machine)

src/common/lib/client/push.ts
ttypic: now it should invokes registerCallback

owenpearson (approved):
all good now, thanks!

PR #2268, opened 2026-07-22
packages: add types-only skeleton for the PDR-091 device/server split

ttypic:
Introduce the public type surface and package.json declarations for the
new per-side Pub/Sub packages, so the shape can be reviewed before any
implementation or build wiring lands:

- `@ably/pubsub-common` (internal, never published): self-contained copies
  of the cross-side types, with client options decomposed into
  CommonClientOptions/HttpOptions/RealtimeOptions and no value
  declarations
- `@ably/pubsub-device`: `createClient()` returning a realtime client with
  required clientId and device push (activate/deactivate), without push
  admin or stats
- `@ably/pubsub-server`: `createHttpClient()` and `createRealtimeClient()`,
  with push admin, stats and token revocation, without device push
- `@ably/pubsub-react`: the React hooks surface typed against `DeviceClient`
- `@ably/pubsub-react-native`: device client with app-supplied push
  storage/token acquisition; no exports map so
  older Metro resolvers can resolve it


packages/pubsub-server/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@ably/pubsub-server",
+  "version": "3.0.0",
```
AndyTWF, line 3: As this would be an entirely new package from ably-js, I think there's an argument that we start it at 1.0.0

packages/pubsub-react-native/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "@ably/pubsub-react-native",
```
AndyTWF, line 2: I know that React/Native implies device, but would we want to male this `@ably/pubsub-device-react-native` for completeness?

packages/pubsub-common/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "@ably/pubsub-common",
```
AndyTWF, line 2: Would this need to actually be published, or just bundled into the actual packages - maintaining a separate npm package could lead to friction potentially?

packages/pubsub-device/index.d.ts
```diff
@@ -0,0 +1,369 @@
...
+   *
+   * @returns a {@link PaginatedResult} object containing an array of {@link PushChannelSubscription} objects for each push channel subscription active for the local device.
+   */
+  listSubscriptions(): Promise<PaginatedResult<PushChannelSubscription>>;
+}
+
+/**
+ * The realtime channel object exposed by the device client. Extends the shared realtime channel surface with device push functionality.
+ */
+export interface RealtimeChannel extends RealtimeChannelBase {
```
lmars, line 231: I'm not overly familiar with TypeScript, but from a developer's point of view, is it the intention that all types they encounter will come from the device package? So for example, the type of the `channel.presence` field would be `device.RealtimePresence`, not `common.RealtimePresence`?

lmars (commented):
This is a larger change than I was expecting, what do you think about the ongoing maintenance of this approach with so many things duplicated?

PR #2276, opened 2026-07-24
Only update attachSerial for non-resumed attaches (RTL15c)

SimonWoolf:
Implements [ably/specification#503](https://github.com/ably/specification/pull/503) (spec 6.1.1).

`attachSerial` is used as the `fromSerial` bound for `untilAttach` history, which returns history contiguous with the realtime message stream. A resumed `ATTACHED` (`resumed=true`) indicates a brief disconnection that did **not** break message continuity, so it must not move the point up to which `untilAttach` history is contiguous.

Previously the SDK updated `attachSerial` on every `ATTACHED` regardless of the `RESUMED` flag; now it only does so when `resumed` is `false` (RTL15a → RTL15c).

## Changes

- **`realtimechannel.ts`**: guard the `attachSerial` update on `!resumed` in the `ATTACHED` handler. `channelSerial` still updates unconditionally per RTL15b.
- **UTS ports** (`test/uts/.../channel_properties.test.ts`): renamed the two `RTL15a` tests to `RTL15c`, and added a new test asserting `attachSerial` is left unchanged by a resumed `ATTACHED` (while `channelSerial` still updates, confirming the message was processed).

`attachSerial` is not part of the public `ably.d.ts` (experimental/unstable per RTL15), so there's no public docstring to update.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed channel attachment handling so resumed attachments no longer overwrite the existing attachment serial.
  * Ensured channel serial behavior remains correct after resumed attachment flows.

* **Tests**
  * Updated unit tests to cover RTL15c scenarios for attach serial progression.
  * Added a test verifying that `RESUMED` ATTACHED updates channel serial but preserves attach serial.

### DIFF
--- a/src/common/lib/client/realtimechannel.ts
+++ b/src/common/lib/client/realtimechannel.ts
@@ -655,13 +655,18 @@ class RealtimeChannel extends EventEmitter {
       isSync = false;
     switch (message.action) {
       case actions.ATTACHED: {
-        this.properties.attachSerial = message.channelSerial;
+        const resumed = message.hasFlag('RESUMED');
+        // RTL15c: only update attachSerial for non-resumed attaches. A resumed
+        // attach hasn't reset the period of message continuity, so it must not
+        // move the point up to which untilAttach history is contiguous.
+        if (!resumed) {
+          this.properties.attachSerial = message.channelSerial;
+        }
         this._mode = message.getMode();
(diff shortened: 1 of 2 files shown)


ttypic (approved):
LGTM

PR #2277, opened 2026-07-29
Encrypt annotation data payloads on encrypted channels

SimonWoolf:
Annotation `data` payloads were published in plaintext even on channels with a cipher configured.

This was, originally, deliberate and documented that annotations would be exempt from e2e encryption, since the multiple.v1 annotation type used data payloads like `{"count":1}`, so the server had to read them.

Before the public release I argued for changing that API so aggregation didn't use `data`, specifically to stop the server having to parse payloads and to preserve the "Ably never reads your payload" property, and pointed to the ability to do end-to-end encryption as a benefit of the change. That was agreed and I implemented it in ably-js in April 2025 (adding the `count` field), but I forgot to actually enable e2e encryption at the same time, which would then have been possible. So the rationale for disabling e2e encryption for annotations no longer applies.

Before releasing, we need to either confirm nobody is using annotation `data` payloads on encrypted channels or contact anyone who is, cf PUB-3816 and https://github.com/ably/realtime/pull/8575.

Also fixed the CipherOptions type which has been broken since the typescriptification.

## Summary by CodeRabbit

* **Bug Fixes**
  * Annotation data now respects channel encryption settings when published through Realtime and REST.
  * Improved encryption handling for messages, presence updates, and annotations.
  * Prevented annotation data from being exposed as plaintext in transmitted payloads.
  * Improved annotation validation and channel state handling during resumed connections.

* **Tests**
  * Added coverage confirming encrypted annotation data round-trips correctly.
  * Added checks that annotation payloads do not contain plaintext data.

### DIFF
--- a/src/common/lib/client/realtimeannotations.ts
+++ b/src/common/lib/client/realtimeannotations.ts
@@ -9,7 +9,6 @@ import RestAnnotations, { RestGetAnnotationsParams, constructValidateAnnotation 
 import type { PaginatedResult } from './paginatedresource';
 import type Message from '../types/message';
 import type { Properties } from '../util/utils';
-import type { CipherOptions } from '../types/basemessage';
 
 class RealtimeAnnotations {
   private channel: RealtimeChannel;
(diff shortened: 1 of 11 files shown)


ttypic (approved):
LGTM

PR #2281, opened 2026-07-30
Check error codes against the ably-common registry at compile time

lmars:
Generates the set of registered Ably error codes from the registry in `ably-common` and narrows `ErrorInfo.code` to it, so that an unregistered code — or a transposed `code`/`statusCode` pair — fails to compile.

**Depends on ably/ably-common#357**, which adds the generator. Don't merge this before that one; see [Submodule pin](#submodule-pin) below.

## Why

`new ErrorInfo(message, code, statusCode)` takes two adjacent numbers that the compiler accepted in either order, and codes were written as bare numeric literals with nothing tying them to the registry. Registry codes are 5–6 digits and an HTTP status is 3, so a status can never be a member of the generated union — a transposition is caught exactly, not heuristically.

`src/common/lib/types/errorcodes.ts` is generated by `npm run generate:errorcodes-ts` and committed. The type erases at compile time, so this costs nothing in the bundle: `errorcodes` appears **zero** times in `build/ably-node.js` and `build/ably.min.js`.

## Checked vs unchecked, split by name

Errors this SDK raises are checked. Errors decoded from the server are not, because the server chooses those codes and may use ones a given client version doesn't know about.

| | |
| --- | --- |
| `fromValues` | checked — for errors this repository raises |
| `fromWireValues` | unchecked — for a response body, a response header, or the `error` field of a `ProtocolMessage` |

`fromValues` was previously reached by both kinds of caller, so a single signature couldn't be both checked and honest. Splitting it makes the safe path the default and leaves every unchecked site greppable. There are ten: nine genuine server decodes, plus the application's `authCallback` in `auth.ts`, whose code originates in user code and so can't be validated here.

Several of those wire sites previously compiled only because they were *cast* to `ErrorInfo` — asserting that untrusted JSON carries a registry-valid code. `fromWireValues` takes `IConvertibleToErrorInfo` (`code: number`), so the typing now matches the data.

`ErrorInfoValues` / `PartialErrorInfoValues` are the checked shapes for the options-object constructor, which is the form new code tends to prefer since it's the one that carries `remediation`. Closing it cost exactly one cast, because every existing site already passed a valid code.

## Three defects this exposed

⚠️ **The first two change observable error codes.**

- **Five sites had `code` and `statusCode` the wrong way round** — `src/common/types/http.ts` (×2), the Node and web `crypto.ts`, and web `http.ts`. Their `err.code` / `err.statusCode` were transposed and are now correct. The web crypto site had no valid pairing in *either* order (400/50000); it now matches its corrected Node sibling.
- **`connectionerrors.ts` `unknownChannelErr` used `UNKNOWN_CONNECTION_ERR` (50002)**, which is why `UNKNOWN_CHANNEL_ERR` (50001) was unused. The registry gives 50001 as *Internal channel error* and `protocol.ts` already uses it that way. Channel errors now report 50001.
- **`normaliseAuthcallbackError` took its code off an `any`**, so neither the callback's value nor the SDK's own fallbacks were checked. The fallbacks are now annotated constants; the callback's value is an explicit cast.

`ConnectionErrorCodes` uses `as const satisfies Record<string, ErrorCode>` so keys stay literal *and* values get checked. A plain `Record<string, ErrorCode>` annotation would have kept accepting a misspelled key — which is the shape of the bug fixed above.

## CI

Two steps in `check.yml`:

1. **`tsc --noEmit -p tsconfig.json`** — nothing typechecked `src/` before this. The build is esbuild and the UTS suite runs under `tsx`; both strip types without checking them, and `tsc --noEmit ably.d.ts modular.d.ts` only covered the two standalone declaration files. Without this step none of the above is enforced anywhere but an editor. It passes with no other change.
2. **Regenerate and `git diff --exit-code`** — at the *pinned* submodule commit, so a reference to a code that isn't registered in `ably-common` can't merge here first.

## Submodule pin

The pin is **provisional**. It points at the unmerged `ably-common` branch `error-code-typescript-gen`, which the two steps above need in order to run at all — the current pin `496da5e` predates `errors/codes/` existing. Once ably/ably-common#357 merges, repoint to `main`.

Only `protocol/` differs between the old pin and this one. `test-resources/` — all ably-js actually reads from the submodule — is byte-identical, so the bump is safe for the fixtures.

Note that because the drift check runs against the pinned commit, CI here will fail if that branch is force-pushed while this PR is open. That's expected, not a real failure.

## Verification

- `tsc --noEmit -p tsconfig.json`: 0 errors
- 1390 UTS unit tests + 18 unit tests passing
- Node and browser builds succeed; `errorcodes` absent from both bundles
- eslint and Prettier clean (2 pre-existing warnings in a react-hooks fake, untouched)
- Generator output is deterministic across runs and already Prettier-clean as emitted; 261 union members for 261 registry files
- Drift check verified to *fail* on a synthetic registry change and pass once reverted
- Negative-tested every construction path: status-in-code-slot and unregistered codes rejected across both classes, both constructor forms, and both `fromValues`; valid codes and the explicit `fromWireValues` escape hatch accepted

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected error-code handling across HTTP, authentication, transport, and channel workflows.
  * Improved decoding of service errors, preserving their intended messages, codes, and status details.
  * Corrected classifications for unexpected request failures and unknown-channel errors.

* **Type Safety**
  * Added a comprehensive registry of recognized error codes and strengthened validation.
  * Added tooling and checks to keep error-code definitions synchronized.

* **Documentation**
  * Added guidance for maintaining error codes and selecting appropriate error details.


.github/workflows/check.yml
```diff
@@ -28,6 +28,13 @@ jobs:
       - run: npm run lint
       - run: npm run format:check
       - run: npx tsc --noEmit ably.d.ts modular.d.ts
+      # src/ is compiled by esbuild, which strips types without checking them, so this is
```
VeskeR: This is not true.
`npm ci` above runs our `prepare` scripts which in turn runs `webpack:all` and typechecks - somewhat incidentally IMO, but still.
Not against having the `npx tsc --noEmit -p tsconfig.json` explicitly, but this is redundant at the moment ,and will never be reached in case there is an error - it will fail on `npm ci` step first

src/common/lib/types/errorinfo.ts
```diff
@@ -44,19 +53,45 @@ export interface IConvertibleToPartialErrorInfo {
   href?: string;
 }
 
+/**
+ * The values an error raised by this SDK is constructed from. Identical to
+ * {@link IConvertibleToErrorInfo} except that `code` must be a registered Ably error code,
+ * so that an unregistered code, or a `code`/`statusCode` transposition, fails to compile.
+ * This is the shape to use for any error written in this repository.
+ */
+export interface ErrorInfoValues extends Omit<IConvertibleToErrorInfo, 'code'> {
```
VeskeR: don't think we need Omit here as we're narrowing the type from number to ErrorCode

src/common/lib/types/errorinfo.ts
```diff
@@ -44,19 +53,45 @@ export interface IConvertibleToPartialErrorInfo {
...
+  code: ErrorCode;
+}
+
+/** As {@link ErrorInfoValues}, for the partial case. */
+export interface PartialErrorInfoValues extends Omit<IConvertibleToPartialErrorInfo, 'code'> {
```
VeskeR: ditto no need for Omit

src/common/lib/types/errorinfo.ts
```diff
@@ -165,13 +208,13 @@ export class PartialErrorInfo extends Error implements IPartialErrorInfo {
...
-      result.href = 'https://help.ably.io/error/' + result.code;
-    }
-    return result;
+  /** As {@link ErrorInfo.fromValues}, for the partial case. */
+  static fromValues(values: PartialErrorInfoValues): PartialErrorInfo {
+    return withHelpHref(new PartialErrorInfo(values));
+  }
+
+  /** As {@link ErrorInfo.fromWireValues}, for the partial case. */
+  static fromWireValues(values: IConvertibleToPartialErrorInfo): PartialErrorInfo {
```
VeskeR: this is not used anywhere atm, so don't include it

src/common/lib/types/errorcodes.ts
VeskeR, line 1: Worth including in `.prettierignore` so a generate script is the only way the file changes

VeskeR (commented):
Couple of nits below and one more ask:
I think we should mention somewhere what a contributor should do if they introduce an error code not assignable to currently generated `ErrorCode`. The expected fflow is add to ably-common, bump ably-common version here, regenerate error codes. Maybe to CONTRIBUTING.md or CLAUDE.md?

src/common/lib/types/errorinfo.ts
VeskeR: removed in https://github.com/ably/ably-js/pull/2281/changes/195d3e4549c535c381a741ce13ac71abdef60905

src/common/lib/types/errorinfo.ts
VeskeR: changed in https://github.com/ably/ably-js/pull/2281/changes/195d3e4549c535c381a741ce13ac71abdef60905

src/common/lib/types/errorinfo.ts
VeskeR: changed in https://github.com/ably/ably-js/pull/2281/changes/195d3e4549c535c381a741ce13ac71abdef60905

.github/workflows/check.yml
VeskeR: updated comment in https://github.com/ably/ably-js/pull/2281/changes/195d3e4549c535c381a741ce13ac71abdef60905

src/common/lib/types/errorcodes.ts
VeskeR: added in https://github.com/ably/ably-js/pull/2281/changes/195d3e4549c535c381a741ce13ac71abdef60905

VeskeR:
Added expected steps to add new errors codes in CLAUDE.md in https://github.com/ably/ably-js/pull/2281/changes/195d3e4549c535c381a741ce13ac71abdef60905

VeskeR (approved):
Approving as I've addressed my feedback and think PR is good to merge

PR #2284, opened 2026-08-09
[AIT-1274] LiveObjects: fail `get()`'s sync wait when the channel enters DETACHED/SUSPENDED/FAILED (RTO23c1)

sacOO7:
- Fixes https://ably.atlassian.net/browse/AIT-1274

## Problem

`RealtimeObject.get()` waits for the objects sync state to reach `SYNCED` before returning (RTO23c), but that wait listened only for the internal `synced` event. If the channel left a usable state while a `get()` was parked, the returned promise **never settled**:

- a **solicited `channel.detach()`** rests the channel in `DETACHED` with the objects data cleared (RTO27a) — the parked `get()` hangs until the user happens to re-attach;
- a channel entering **`FAILED`** — including a `SUSPENDED` channel whose connection then fails terminally (`propogateConnectionInterruption` maps connection `failed` → channel `failed`) — is terminal: the data is cleared, no automatic recovery exists, and the promise can never resolve.

This was also internally inconsistent: `publishAndApply` already rejects deterministically in the identical wait (RTO20e1, 92008), and `get()` itself rejects when the channel is *already* `FAILED` at entry (`ensureAttached`, 90001) — but hung when the channel *became* unusable mid-wait.

The spec now covers this as **RTO23c1** (companion PR: https://github.com/ably/specification/pull/514): the parked `get()` must fail with `ErrorInfo` `code` 92008, `statusCode` 400, and `cause` set to the channel's `errorReason`, regardless of the state the channel transitioned from.

## Fix

One mechanism, shared by both waiters:

- `actOnChannelState` — the plugin's single channel-state entry point, invoked on every transition — now emits an internal-only `ObjectsInternalEvent.syncWaitFailed` on `detached` / `suspended` / `failed`, *before* the RTO27a data clearing (drain-then-clear, matching ably-cocoa and ably-java). The event is emitted on `_eventEmitterInternal` only, so it is not observable through the public `RealtimeObject#on()` API. Because it hooks the state handler rather than a specific event source, every route into the three states is covered — including `SUSPENDED` → `FAILED`.
- A new `_waitForSyncedOrChannelFailure(failureDescription)` helper races `synced` against `syncWaitFailed` and builds the 92008/400/cause `ErrorInfo`; `get()` and `publishAndApply` differ only in the message prefix mandated by their respective spec points (`'the object could not be retrieved'` vs `'the operation could not be applied locally'`). `publishAndApply` is refactored onto the helper, replacing its previous `internalStateChanges` subscription. Both listeners are removed on either outcome.
- `realtimechannel.ts` (4 lines): `notifyState` invokes `actOnChannelState` *before* assigning `this.errorReason`, so the state-change `reason` is now forwarded as an argument — exactly as the adjacent presence handler (`_presence.actOnChannelState(state, hasPresence, reason)`) already does. Without this the rejection's `cause` would be lost.

RTO27 data semantics are unchanged: `SUSPENDED` still retains the objects data (RTO27b); only `DETACHED`/`FAILED` clear it (RTO27a).

## Tests

- The three `RTO23c1` UTS unit tests (`fails-on-channel-{detached,suspended,failed}-0`) previously existed as **skipped deviations**; the skip guards and the `test/uts/deviations.md` entry are removed now that the behaviour is implemented. The `failed` case asserts the `cause` (`cause.code === 90000`), which exercises the reason-forwarding.
- New `RTO5a5`/`RTO5a6` UTS unit tests (`objects_pool.test.ts`) covering `OBJECT_SYNC` `channelSerial` handling: absent serial (single-message sync) and malformed serial treated as absent. The existing implementation already conforms — these add the spec-derived coverage.

All tests derive from the UTS unit spec cases added in https://github.com/ably/specification/pull/514.

## Verification

- `test/uts/objects/unit`: **325 passing, 0 failing, 0 pending** (previously 3 pending deviations).
- `test/uts/realtime/unit` channels + presence sweep (for the `realtimechannel.ts` change): **326 passing, 0 new failures**.
- `tsc --noEmit` and `eslint` clean on the touched sources.

## Summary by CodeRabbit

* **Bug Fixes**
  * Object retrieval and updates now fail promptly with a clear synchronization error when the channel is detached, suspended, or failed.
  * Error details and channel failure causes are preserved for improved diagnosis.
  * Retry guidance now reflects the channel state and operation status.
  * Object data is retained during temporary suspension and cleared appropriately after permanent channel failures.
  * Synchronization completes correctly when incoming messages omit or contain invalid channel sequence information.


ttypic (approved):
LGTM

PR #2285, opened 2026-08-10
fix: fall back to the base transport when websocket handshakes are rejected outright

VeskeR:
## Problem

A browser behind a proxy that **rejects** the WebSocket upgrade — a 403, say — never connects. It rotates through its fallback hosts on `web_socket`, reports `80003 Unable to connect (and no more fallback hosts to try)`, and repeats indefinitely. `xhr_polling` is never attempted.

If the proxy instead silently *drops* the handshake, the same client connects over `xhr_polling` in ~10s. Only the fast-failure case is broken.

The failover to the base transport hangs off `webSocketGiveUpTimer` (`webSocketConnectTimeout`, 10s), which only helps when a websocket attempt is *slow*. On fast rejection, `tryFallbackHosts()` exhausts the host list in milliseconds and calls `notifyState('disconnected')` — which cancels that timer. Since 2.0 replaced connect-on-base-then-upgrade with connect-on-websocket-then-fall-back, the timers are the only route to the base transport.

## Fix

Exhausting every host on `web_socket` now runs the websocket connectivity check and decides on its result, instead of writing the attempt off:

| Connectivity check | Behaviour |
| --- | --- |
| fails | take up the base transport immediately |
| hangs | unchanged — the give-up timer takes it up at `webSocketConnectTimeout` |
| succeeds | unchanged — `disconnected`, since websockets work here and the base transport talks to the same unhealthy hosts |

That last row is what keeps `try_fallback_hosts_on_placement_constraint` passing: a client told to move datacenter must not hop back to the primary host on a different transport. It is also the condition the give-up timer already applies, so the two paths now agree.

This only affects attempts that fail on *every* host over websocket without the transport becoming viable. Clients with no base transport available skip the check entirely.

## Testing

Three new tests in `test/realtime/transports.test.js`, using a `rejectingWebSocket()` drop-in that fires `onerror` then `close(1006)` immediately — what a proxy 403 looks like to the browser WebSocket API:

- `ws_rejected_immediately` — falls back to the base transport. Times out at 60s on `main`.
- `ws_rejected_immediately_but_ws_connectivity_available` — does *not* switch transport when websockets work on the network.
- `ws_rejected_immediately_with_no_base_transport` — goes straight to `disconnected`.

All nine tests in the file pass on Chromium, Firefox and WebKit. The red CI legs are the pre-existing flaky suites (`presence`, `rest/message-operations`, `channel_backoff_*`), each verified to fail on `main` too.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved real-time connection reliability when WebSocket connections fail during setup.
  * The app now more accurately distinguishes between unavailable connectivity and rejected connection hosts.
  * Added smoother fallback to alternative transport methods when WebSocket connections cannot be established.
  * Improved disconnected-state handling when no usable connection method is available.

### DIFF
--- a/src/common/lib/transport/connectionmanager.ts
+++ b/src/common/lib/transport/connectionmanager.ts
@@ -163,6 +163,7 @@ class ConnectionManager extends EventEmitter {
   } = { isProcessing: false, queue: [] };
   webSocketSlowTimer: NodeJS.Timeout | null;
   wsCheckResult: boolean | null;
+  wsCheckPromise: Promise<boolean> | null;
   webSocketGiveUpTimer: NodeJS.Timeout | null;
   abandonedWebSocket: boolean;
   connectCounter: number;
@@ -257,6 +258,7 @@ class ConnectionManager extends EventEmitter {
     this.forceFallbackHost = false;
     this.connectCounter = 0;
     this.wsCheckResult = null;
+    this.wsCheckPromise = null;
     this.webSocketSlowTimer = null;
     this.webSocketGiveUpTimer = null;
     this.abandonedWebSocket = false;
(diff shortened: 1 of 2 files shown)


src/common/lib/transport/connectionmanager.ts
```diff
@@ -1456,6 +1459,87 @@ class ConnectionManager extends EventEmitter {
...
+   * the webSocketGiveUpTimer does.
+   */
+  abandonWebSocketForBaseTransport(transportParams: TransportParams): void {
+    Logger.logAction(
+      this.logger,
+      Logger.LOG_MINOR,
+      'ConnectionManager.abandonWebSocketForBaseTransport()',
+      'websocket connectivity appears to be unavailable; trying base transport',
+    );
+    this.abandonedWebSocket = true;
```
owenpearson, line 1467: nit: we should remove the duplication between webSocketGiveUpTimer and this

owenpearson (approved):

src/common/lib/transport/connectionmanager.ts
VeskeR: done in https://github.com/ably/ably-js/compare/26e9bce785e03f526e8ce38cde34e95f2b109c25..2f34fdb4df08e7adfae2a5bda3decfcd3a480ae9

PR #2272, opened 2026-07-22
Release 2.26.0

ttypic:
Release 2.26.0.

### What's Changed

- Add `push.updateToken()` to update the device's push token on React Native after the underlying push platform rotates it, keeping the device registration in sync with Ably #2267
- Add a `useObject` React hook for subscribing to LiveObjects state from the `ably/react` entrypoint #2259

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added support for `push.updateToken()` on React Native to keep push token registration in sync after rotation.
  * Introduced the `useObject` React hook for subscribing to LiveObjects state via the React entrypoint.
* **Changes**
  * Updated `PathObject.instance()` behavior to return `undefined` only when nothing exists at the path; resolved values are wrapped as `LiveMap`, `LiveCounter`, or primitives (primitive-backed instances are read-only).
* **Release**
  * Bumped the library version to **2.26.0** and refreshed the changelog accordingly.

PR #2286, opened 2026-08-10
Release 2.26.1

VeskeR:
Bumps the version to 2.26.1 in `package.json`, `package-lock.json` and `src/platform/react-hooks/src/AblyReactHooks.ts`, and adds the `2.26.1` entry to `CHANGELOG.md`.

The headline change is #2285: a browser client behind a proxy that rejects the WebSocket upgrade outright never connected at all, rotating through fallback hosts on `web_socket` indefinitely instead of falling back to `xhr_polling`.

### PRs included since 2.26.0

**User-facing (in changelog)**

- #2285 (WebSocket fallback fix — headline)
- #2273 (server-side resumability)
- #2276
- #2261

**Not user-facing (not in changelog)**

- #2274 — CI
- #2278 — test-only
- #2282 — docs
- #2245 — docs
- #2244 — docs
- #2243 — docs
- #2242 — docs

#2219 merged before the 2.26.0 tag and is already in that release; #2272 is the 2.26.0 release PR itself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 This contains only the most important and/or user-facing changes; for a full changelog, see the commit history.
 
+## [2.26.1](https://github.com/ably/ably-js/tree/2.26.1) (2026-08-10)
+
+[Full Changelog](https://github.com/ably/ably-js/compare/2.26.0...2.26.1)
+
+### What's Changed
+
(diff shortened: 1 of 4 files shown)


VeskeR:
Superseded by #2287, cut as 2.27.0 rather than 2.26.1: #2273 changes the resumption semantics of existing public members (id/key retained through SUSPENDED, so createRecoveryKey() no longer returns null there) without changing the API surface, which a patch label would misdescribe.

PR #2288, opened 2026-08-11
LiveObjects: suppress no-op diff updates (RTLC14c/RTLM22c) and skip the sync wait for an empty synthetic list (RTO20d4)

sacOO7:
> [!NOTE]
> Stacked on #2284 (`liveobjects/rto23c1-get-sync-wait-failure`) — this branch bases there because both touch the `publishAndApply` sync-wait region; review after that PR. Spec companion: https://github.com/ably/specification/pull/515 (stacked on https://github.com/ably/specification/pull/514).

## Problem

Three related behaviours diverged from the Objects spec's no-op update clauses, each delivering a spurious or wasteful result:

1. **Spurious update events on re-sync of unchanged data** ([RTLC14c](https://sdk.ably.com/builds/ably/specification/main/features/#RTLC14c) / [RTLM22c](https://sdk.ably.com/builds/ably/specification/main/features/#RTLM22c)). When an `OBJECT_SYNC` re-delivers state identical to what an object already holds (reconnect, re-attach with `HAS_OBJECTS`), `_updateFromDataDiff` produced `{ update: { amount: 0 } }` / `{ update: {} }` and `_applySync` delivered them to subscribers — change events for no change. The spec requires such diffs to be no-op updates (`noop: true`), which `notifyUpdated` already suppresses.
2. **Spurious empty update on reset of an already-empty root** ([RTO4b2a](https://sdk.ably.com/builds/ably/specification/main/features/#RTO4b2a)). The `ATTACHED`-without-`HAS_OBJECTS` reset routes through the same diff (`resetToInitialPool` → `clearObjectsData` → `notifyUpdated`), so an already-empty root emitted `{ update: {} }`. The clarified RTO4b2a specifies suppression (ably-cocoa and ably-java already suppress).
3. **`publishAndApply` parked on a wait with nothing to apply** ([RTO20d4](https://sdk.ably.com/builds/ably/specification/main/features/#RTO20d4)). When every serial in the `PublishResult` is `null` (each skipped per RTO20d1), the synthetic message list is empty, yet the code still entered the RTO20e sync wait — an unnecessary park that would surface a spurious 92008 if the channel dropped meanwhile. (Defensive: all-null serials are not an expected server behaviour today.)

## Solution

- **The no-op collapse lives in `_updateFromDataDiff` itself** (both `LiveCounter` and `LiveMap`), so every caller — the `overrideWithObjectState` re-sync path, the RTO4b reset, and any future call site — is covered by one change. The `overrideWithObjectState` sites gain a no-op passthrough (mirroring their existing RTLC6e/RTLM6e terminal-noop returns) so `objectMessage` is only stamped on real updates.
- **The tombstone path is the one mandated exception** (the RTLC14c/RTLM22c advisory: the exception "must not be applied when the diff is computed for a tombstone per RTLO4e5"). A tombstone update must never be no-op-marked — `notifyUpdated` suppresses no-ops *before* its tombstone branch, so the RTLO4b4c3c listener teardown would silently be skipped for an already-zero counter / already-empty map. `tombstone()` therefore synthesizes the typed no-change update (`{ update: { amount: 0 } }` / `{ update: {} }`) via a new `protected abstract _createNoChangeUpdate()` hook when the diff is a no-op, then stamps `tombstone = true`. The hook follows the class's existing template-method design (it joins seven existing abstract members) and matches ably-java's synthesis semantics (`BaseRealtimeLiveObject.tombstone()`); a base-class type-switch was rejected because `LiveObject` instances carry no type discriminator and a concrete literal is not assignable to the generic `TUpdate` without casts.
- **RTO20d4 guard**: `publishAndApply` returns successfully before the RTO20e wait when the synthetic list is empty.

## Tests

Seven UTS unit test ports (spec cases added in ably/specification#514 + #515):

| Test ID | Asserts |
|---|---|
| `RTLC14c/zero-delta-diff-is-noop-0` | identical-state override → `noop: true`, no event |
| `RTLM22c/empty-diff-is-noop-0` | identical non-tombstoned entries (timeserial ignored per RTLM22b3) → `noop: true` |
| `RTO20d4/empty-synthetic-list-skips-sync-wait-0` | all-null ACK serials while SYNCING → operation resolves without the sync completing |
| `RTLO5/tombstone-zero-value-counter-emits-update-0` | tombstoning a zero counter → non-no-op update, `tombstone: true`, `amount: 0` |
| `RTLO5/tombstone-empty-map-emits-update-0` | tombstoning an all-tombstoned map → non-no-op, `tombstone: true`, empty payload |
| `RTLO4b4c3c/tombstone-zero-value-counter-tears-down-0` | subscriber receives the zero-amount tombstone update and listeners are deregistered |
| `RTO4b2a/reset-of-empty-root-emits-no-update-0` | reset of an empty root emits nothing (second-channel liveness control) |

## Verification

- Full `test/uts/objects/unit` tier: **332 passing, 0 failing, 0 pending** (325 baseline + 7).
- The RTLC14c/RTLM22c tests fail without the production change (`noop` is `undefined` pre-fix); the RTO20d4 test times out parked in the RTO20e wait pre-fix.
- Existing tombstone, RTO4b/RTO27, RTLC14/RTLM22 and RTO20d1/RTO20e neighbours all green (the tombstone suite passing proves the teardown exception holds).
- `tsc --noEmit` and `eslint` clean on the touched sources.

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid live object operations and state validation errors now return HTTP 400 responses.
  * Corrected zero-value counter and empty-map updates during deletion and state replacement.
  * Tombstone events now reach listeners reliably, include deletion details, and properly stop future notifications.
  * Resetting an already-empty root no longer sends unnecessary subscriber updates.
  * Publish operations with no applicable acknowledgements now complete without waiting for synchronization.
* **Tests**
  * Added coverage for zero-delta updates, tombstones, listener behavior, root resets, and acknowledgement handling.


ttypic (approved):
LGTM

PR #2293, opened 2026-08-27
Split Pub/Sub SDK into separate device and server packages v2

ttypic:
* New public packages: Create `@ably/pubsub-device` and `@ably/pubsub-server`, allowing Pub/Sub connections to explicitly declare whether they are device- or server-side for realtime classification and MAU billing.
* Shared core: Move common Pub/Sub functionality into `@ably/pubsub-core`, which is for internal/shared usage and not intended as a public API.
* Package structure: The two public packages wrap and re-export the shared core functionality while maintaining lockstep versioning and avoiding duplicate core instances.


umair-ably:
Ran a review pass over this PR (umair, with Claude Code assistance; findings verified by building the branch and executing against the built packages) and pushed fixes as individual commits:

- **32bf23ed** — the Push and modular missing-plugin remediations dictated default imports, but those subpaths only have named exports; following them verbatim in Node ESM throws `does not provide an export named 'default'`. Added the braces.
- **4bd7ac24** — `optionsWithSideAgent` spread `undefined` into `{}`, so a no-argument factory call constructed a client that failed later with `No authentication options provided` instead of the core's instructive initialization error. Now passes `undefined`/`null` through so the core guard fires. Regression tests added (`pubsub_factories.test.js`).
- **d31564d8** — the modular `createClient` converted bare key/token strings into options, bypassing the core's deliberate rejection (`you must provide a client options object with a 'plugins' property`) and replacing it with an unactionable missing-plugin error. Strings now pass through so the core's targeted rejection surfaces.
- **6d11cf04** — `index-pubsub-side.ts` was bundled but never executed (global name nothing invokes, no page, not served, no playwright scenario), so its assertions — including the cross-package `ErrorInfo instanceof` double-core check — were dead code. Wired it up like the other entry points; verified with a full `test:package` run (new scenario executes and passes).
- **9ea9a29a** — the device README's REST guidance told readers to depend on `@ably/pubsub-core` directly, contradicting the package's no-direct-core-dependency design and inviting the exact-version-peer double-core failure. Now points at the package's own `Rest` re-export.

Also on the branch from earlier today: **e1eb68ca** adds wire-level assertions on the side-declaring `Ably-Agent` entry each factory stamps (the per-side billing contract from the rollout plan), and **9c906536** makes `test:grep` build the wrapper packages that test now requires.

All gates green locally: unit suite, `check:packages`, `test:package`, lint, prettier.

umair-ably:
@ttypic I'll merge this once CI is green... only outstanding question I had that we can pick up when you're back is whether the server package should also have a modular entry point? seems intentionally left out right now but curious if this decision thought about serverless runtimes that might also prefer modular variants?

AndyTWF:
@umair-ably I have a concern with the agent headers used for the new packages. They do not indicate the ecosystem/language anywhere, which will make them indistinguishable from other SDKs if we ever drop the "ably-js" agent which I believe is on the cards for the future. Either way, it means we can't distinguish the SDK if we ever see the pubsub-server version in isolation.

Right now the PubSub server agent is `ably-pubsub-server/2.28.0`... I would recommend we make it into `pubsub-server-js/<version>` which matches how we do Spaces and Chat. Lets catch up on this :)

PR #2296, opened 2026-09-01
Lockstep release tooling for the three packages

umair-ably:
Stacked on #2294 (base: `uts-per-side`) — implements the rollout plan's release-tooling step: *"One command releases core + device + server at the same version. Partial release is impossible or fails reversibly."*

## What

`scripts/release.js`, one command per release step, replacing the manual checklist's three hand-edited version sites and three separate publishes:

- **`npm run check:versions`** — verifies all seven version sites agree (root + three workspace `package.json`s, both wrappers' exact `@ably/pubsub-core` peer pins, and the react-hooks agent version constant). Now runs in the lint CI job, so a drifted site cannot merge.
- **`npm run release:bump <version>`** — moves every site at once and refreshes the lockfile. It rewrites the `package.json` files directly rather than using `npm version --workspaces --include-workspace-root`, which was observed (npm 11) to rewrite the files and *still exit non-zero*, then fail on re-run with `Version not changed` — exactly the partial state this tooling exists to prevent. `bump` is deterministic and re-runnable at any version.
- **`npm run release:publish [-- --dry-run] [-- --otp <code>]`** — the lockstep publish:
  - refuses to start unless every version site agrees and every build artifact the packages' `files` globs reference exists on disk (`npm pack` silently omits missing files, so a stale checkout would otherwise ship broken tarballs);
  - dry-run packs everything before publishing anything;
  - publishes in dependency order (core before the wrappers that pin it exactly), so no wrapper is ever visible on the registry before its core;
  - consults the registry first and skips already-published versions, so a run that fails partway is completed by simply re-running it;
  - prints rollback guidance (`npm unpublish` within 72h) if a release is abandoned instead.

Also fixes a first-release landmine found while building this: **none of the three scoped packages set `publishConfig.access`**, and npm publishes scoped packages as `restricted` by default on first publish — so the first release would have shipped them private (or failed, depending on org plan). All three now declare `access: public`.

CONTRIBUTING's release process shrinks accordingly (steps 4–6 → one bump command; 10–11 → one publish command).

## Verification

- `check` passes at 2.28.0 and correctly reports a deliberately drifted site with the full table.
- `bump` round-tripped 2.28.0 → 2.99.0 → 2.28.0 with all seven sites moving together.
- `publish --dry-run` runs the full preflight against the real registry (all three names correctly detected as not yet published), dry-run packs all three, and reports the publish order.
- The missing-artifact guard verified by deleting `packages/server/dist/index.mjs`: publish refuses with the exact path.

## Out of scope, flagged

First-time npm provisioning: trusted publishing can only be configured per package *after* its first publish, so release one of the new packages is necessarily an authenticated manual publish (OTP) — this tooling supports exactly that; the trusted-publishing binding (and its later re-binding for the repo rename) is registry-side configuration outside this repo.

## Summary by CodeRabbit

* **New Features**
  * Added automated version consistency checks for all related packages and release metadata.
  * Added commands to update versions, validate release readiness, and publish packages in the correct order.
  * Added support for dry-run publishing, OTP authentication, resumable releases, and skipping already-published packages.
  * Configured packages for public npm publication.

* **Documentation**
  * Updated release instructions to describe the streamlined lockstep release process.


ttypic (approved):
LGTM

PR #2297, opened 2026-09-02
Agent identifiers for the package split: versionless side flags + ably-pubsub-js

umair-ably:
Stacked on #2296 (base: `lockstep-release`) — implements both halves of the agent-identifier convention proposed on PDR-091b2 and reviewed on [ably/ably-common#361](https://github.com/ably/ably-common/pull/361). Target wire shape (live capture from the branch):

```
ably-pubsub-js/2.28.0 nodejs/24.13.0 ably-pubsub-device
```

## Commit 1 — versionless side flags

The `ably-pubsub-device` / `ably-pubsub-server` identifiers are **cross-SDK** (ably-java#1233 and ably-ruby#453 stamp the same strings), so a version on them cannot say what it versions — and under lockstep releasing it only ever duplicated the SDK entry beside it. Per the registry review, they become pure flags, like `browser`:

- `getAgentString` emits a bare identifier for a nullish `agents` value (previously that would have produced the invalid token `name/undefined`; the `Record<string, string | undefined>` type always sanctioned versionless entries, but no code path emitted them correctly). Versioned entries are emitted exactly as before, so layered-SDK attribution (`chat-js/x.y.z`) is unchanged.
- `optionsWithSideAgent` loses its `version` parameter; the factories stop importing their package.json version.
- Wire tests assert the bare token and fail on any versioned form (including the `name/undefined` shape).

## Commit 2 — SDK family identifier renames to `ably-pubsub-js`

Because the identifier flips exactly at the split and the maintenance branch is never touched, the identifier alone partitions the fleet: `ably-js/*` = legacy `ably`-package traffic, `ably-pubsub-js/*` = new-package traffic — migration tracking and legacy EOL enforcement become a string match. The identifier deliberately still names the *family*, not any one published package, exactly as `ably-js` always did (npm package was `ably`; platform axis factored into separate `browser`/`nodejs`/`react-native` entries — the side axis is factored out the same way).

One production line (`Defaults.agent`) plus the six test assertions on the identifier (UTS RSC7d/RTC6, the agent wire tests, and the side_modes null-stamp check, which now tests for the flag tokens themselves since the SDK identifier shares their prefix).

## Ordering dependencies

- Merges only after ably/ably-common#361 settles the versionless modeling.
- **Requires an `ably-pubsub-js` registry entry in ably-common before any release from this branch** (the convention: registry first, so the identifier never ships unregistered). Follow-up to #361.
- ably-java#1233 / ably-ruby#453 need the mirror changes (versionless flag + their own family identifier) before they merge.

## Verification

Unit suite 29 passing; UTS unit tier 1395 passing in **all three** side modes; `check:packages` clean; prettier clean; wire header verified by live local capture (shown above).


ttypic (approved):
LGTM

PR #2301, opened 2026-09-13
Omit null message fields from wire JSON (RSL1e, RTL6i3)

### LINKED ISSUE #2199: Null message fields included in wire JSON instead of omitted (RSL1e, RTL6i3)
## Spec points

- **RSL1e**: "If any of the values are null, then key is not sent to Ably."
- **RTL6i3**: Same requirement — a payload with null \`data\` should be sent as \`{ "name": "click" }\`, omitting the \`data\` key entirely.

## Observed behaviour

When \`data\` is \`null\`/\`undefined\`, ably-js includes it as \`"data": null\` in the JSON wire format instead of omitting the key. Same for \`name\`.

Message serialization in \`src/common/lib/types/message.ts\` does not strip null/undefined values before \`JSON.stringify\`.

## Failing tests (2)

| Test | Expected | Actual |
|------|----------|--------|
| \`RSL1e - null name omitted from body\` | Body has no \`name\` key | Body has \`"name": null\` |
| \`RSL1e - null data omitted from body\` | Body has no \`data\` key | Body has \`"data": null\` |

## Reproduction

\`\`\`bash
RUN_DEVIATIONS=1 npx mocha --grep "RSL1e - null" test/uts/rest/channel/publish.test.ts
\`\`\`

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-5713) by [Unito](https://www.unito.io)

cpruijsen:
Message fields whose value is null are sent to Ably rather than omitted, against RSL1e and RTL6i3.

`wireToJSON` in `src/common/lib/types/basemessage.ts` built the wire object with `Object.assign`,
which copies every own property including the null ones. This deletes keys whose value is null or
undefined before returning, so an unset field is absent from the payload rather than present and
null.

`encrypt` needed a matching guard. It called `encryptData(msg.data, ...)` unconditionally, so a
message with no payload had its null data encrypted into a non-null value and then travelled as an
encrypted empty payload. It now returns the message untouched when there is nothing to encrypt, which
is what leaves the field absent for the JSON step above.

The deviations file drops the entries for the tests this makes pass.

Fixes #2199

## Summary by CodeRabbit

- **Bug Fixes**
  - Messages with null or missing payloads are no longer encrypted unnecessarily.
  - Serialized messages now omit null-valued fields, so name-only and data-only messages are transmitted without empty fields.
  - Realtime and REST message publishing now consistently handles null names and data as expected.

- **Tests**
  - Automated coverage now verifies null-field handling without requiring special test settings.


src/common/lib/types/basemessage.ts
```diff
@@ -302,7 +306,14 @@ export function wireToJSON(this: BaseMessage, ...args: any[]): any {
   const format = args.length > 0 ? Utils.Format.json : Utils.Format.msgpack;
   const { data, encoding } = encodeDataForWire(this.data, this.encoding, format);
 
-  return Object.assign({}, this, { encoding, data });
+  const result: Record<string, any> = Object.assign({}, this, { encoding, data });
+  // RSL1e, RTL6i3: fields whose value is null are not sent to Ably
+  for (const key in result) {
```
ttypic, line 311: Spec only specify omitting `null` values for `data` and `name` fields. I don't think we should do this for all fields

ttypic (changes requested):

src/common/lib/types/basemessage.ts
cpruijsen: Good catch, narrowed to `data` and `name` in 838c658. The loop was dropping every null field, which goes beyond what RSL1e and RTL6i3 say. Also rebased onto main, since it had fallen behind.

PR #2302, opened 2026-09-13
Expose PresenceMessage.memberKey (TP3h)

### LINKED ISSUE #2202: PresenceMessage.memberKey not exposed (TP3h)
## Spec point

- **TP3h**: \`memberKey\` is a "string function that combines the \`connectionId\` and \`clientId\` ensuring multiple connected clients with the same clientId are uniquely identifiable." It should be a method on \`PresenceMessage\`.

## Observed behaviour

\`memberKey\` is not a method on \`PresenceMessage\`. It is computed internally as a lambda \`(item) => item.clientId + ':' + item.connectionId\` passed to \`PresenceMap\`, but not accessible to callers.

## Failing test

| Test | Expected | Actual |
|------|----------|--------|
| \`TP3h - memberKey format\` | \`typeof msg.memberKey === 'string'\` | \`typeof msg.memberKey === 'undefined'\` |

## Reproduction

\`\`\`bash
RUN_DEVIATIONS=1 npx mocha --grep "TP3h" test/uts/rest/types/presence_message_types.test.ts
\`\`\`

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-5716) by [Unito](https://www.unito.io)

cpruijsen:
`PresenceMessage.memberKey` is not exposed, against TP3h. The value is what distinguishes two
connections sharing a `clientId`, so a consumer that needs it has to rebuild it by hand and get the
field order right.

Three changes. `ably.d.ts` declares `memberKey` as a readonly string on `PresenceMessage`.
`realtimepresence.ts` builds its `PresenceMap` key from `item.memberKey` rather than reassembling
`clientId + ':' + connectionId` inline, so the library reads its own property instead of keeping a
second copy of the rule. `defaultpresencemessage.ts` takes `memberKey` out of the `fromValues` input
type, since it is derived rather than supplied.

Fixes #2202

## Summary by CodeRabbit

* **New Features**
  * Presence messages now expose a read-only `memberKey` identifier for presence members.
  * Presence member tracking uses this identifier for consistent member lookup.
  * Presence message creation APIs now distinguish derived member identifiers from user-provided values.

* **Tests**
  * Member key format validation now runs by default.

* **Documentation**
  * Updated test deviation documentation to reflect the available `memberKey` property.


ttypic (approved):
Thanks for contribution, lgtm

PR #2299, opened 2026-09-12
Reject annotation publish/delete without a type (RSAN1a3)

cpruijsen:
`constructValidateAnnotation()` now rejects a missing, empty, or non-string `type` with `ErrorInfo` code 40003 / HTTP 400, matching RSAN1a3 and the existing serial / second-argument checks in the same helper.

Fixes #2194.

## Summary

REST `annotations.publish` / `delete` and realtime `annotations.publish` / `delete` all go through that helper. Unskipped the existing UTS tests (`RSAN1a3 - type required`, `RTAN1a - publish validates type is required`) and removed the RSAN1a3 entry from `test/uts/deviations.md`.

RSAN1a3 as quoted on #2194: the SDK must validate that the user supplied a `type`; all other fields are optional; throw 40003. Public `OutboundAnnotation` already required `type`; this is the runtime check.

## Decision

Validate in the shared helper, treating empty string like a missing serial (`!value || typeof !== 'string'`). The alternatives were to only reject a missing or non-string `type` (allow `type: ''`), or to validate only on the REST path. One helper is already the publish/delete gate for both clients, and the serial check next to it already uses that truthiness test.

Can drop the truthiness check if empty string should be left to the server.

## Test plan

- [x] `RSAN1a3 - type required` (REST UTS, mocked HTTP) throws 40003
- [x] `RTAN1a - publish validates type is required` (realtime UTS, mocked WebSocket) throws 40003
- [x] Same tests fail without the source change (publish succeeds; `error.code` is undefined)
- [x] Rest of `test/uts/rest/unit/channel/annotations.test.ts` (9 passing, 1 pending RSAN1c4) and `test/uts/realtime/unit/channels/channel_annotations.test.ts` (15 passing) still pass
- [ ] Type definitions unchanged (`OutboundAnnotation.type` was already required)

## Summary by CodeRabbit

* **Bug Fixes**
  * Publishing annotations now validates that the `type` field is a non-empty string.
  * Invalid or missing annotation types return a consistent error with code `40003` and HTTP status `400`.
  * Validation is applied consistently for both REST and realtime annotation publishing.

* **Tests**
  * Added coverage confirming the required validation behavior across REST and realtime publishing.

PR #2303, opened 2026-09-14
Fix REST token renewal retry sending stale Authorization header (RSA4b, RSC10)

### LINKED ISSUE #2193: Token renewal sends old authorization header on retry, no retry limit (RSA4b, RSC10)
## Spec points

- **RSA4b**: When a REST request fails with a token error (40140–40149), the library should obtain a new token and retry the request with the new token. Renewal should retry at most once per request.
- **RSC10**: "Transparent retry after token renewal."

## Observed behaviour

Two compounding bugs in `Resource.do()`:

### 1. Old authorization header sent on retry

After a token error, the retry calls:

```javascript
await client.auth.authorize(null, null);
return withAuthDetails(client, headers, params, doRequest);
```

The `headers` parameter is the **merged** headers from the first `withAuthDetails` call, which already contains `authorization: 'Bearer <old-token>'`. Then `withAuthDetails` does `Utils.mixin(newAuthHeaders, oldMergedHeaders)`, which overwrites the new token's authorization header with the old one.

### 2. No retry limit

The retry loop in `Resource.do()` is unbounded — on each token error, it calls `authorize()` and retries recursively with no counter. Combined with bug #1, this causes an infinite loop and eventual OOM when the server persistently returns token errors.

## Failing tests (3)

| Test | Issue |
|------|-------|
| `RSA4b - renewal on 40142 error` | Retried request has old token in authorization header |
| `RSC10 - transparent retry after renewal` | Same: retried request carries old token |
| `RSA4b - renewal limit` | Unbounded retries (test caps authCallback at 3 to prevent OOM) |

## Reproduction

```bash
RUN_DEVIATIONS=1 npx mocha --grep "RSA4b|RSC10" test/uts/rest/auth/token_renewal.test.ts
```

From PR #2191, branch `uts-rest`.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-5707) by [Unito](https://www.unito.io)

cpruijsen:
Token renewal retries with the old Authorization header, and the retry has no limit. Both against
RSA4b and RSC10.

`src/common/lib/client/resource.ts` handled the token error inside `doRequest`, then recursed through
`withAuthDetails` from there. The headers passed to that recursion are the ones already built, so the
retry carries the expired token and fails the same way, and because the recursion is unbounded a
persistently rejected token loops rather than surfacing.

The retry moves outside the request, after the result comes back. The renewed credentials are derived
again from the supplied headers and params, so the second attempt carries the new token, and it runs
exactly once: RSA4b is specific that there is one renewal, meaning two auth callbacks and two requests
in total.

The one-shot bound is also what makes the failure legible. A regression that stops the renewed token
reaching the retry now fails on the second attempt rather than spinning.

Tests assert the callback and request counts rather than only the outcome, so a retry that renews
twice or not at all is distinguishable from one that renews correctly. The deviations file drops the
entries this makes pass.

Fixes #2193

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of expired authentication tokens by renewing authorization details and retrying the request once.

* **Tests**
  * Token-renewal integration and unit tests now run consistently without environment-based skips.
  * Updated test coverage and comments to reflect current retry behavior.

* **Documentation**
  * Removed outdated documentation describing previously observed token-renewal deviations.


ttypic (approved):
thanks for contribution! lgtm

ttypic (changes requested):
Could you please resolve conflict?

cpruijsen:
Rebased onto main, conflict resolved. The clash was in `test/uts/deviations.md`, where #2299 removed the section next to the one this PR removes.

ttypic (approved):

